PR #8, opened 2019-07-12
RFC for Removing and replacing SpanData

bg451:


0004-remove-spandata.md
```diff
@@ -0,0 +1,43 @@
...
+
+I'd like to propose getting rid of SpanData and `tracer.recordSpanData()` and replacing it by allowing `tracer.startSpan()` and `span.finish()` to accept start and end timestamp options. This reduces the API surface, consolidating on a single span type. Options would meet the requirements for out of band reporting.
+
+## Internal details
+
+`startSpan()` would change so you can include options such as `withStartTimestamp()`, `withTags()`, `withResource()`, `withEvents()`, `withLogs()`, etc. and for `finish()` you could have `withEndTimestamp()`, `withEvents()`, `withLogs()`, etc. The exact implementation would be language specific, so some would use an options pattern with function overloading or variadic parameters, or a span builder pattern.
```
reyang: Adding events/links/etc. has been supported by span APIs already (e.g. [`span.addEvent`](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/tracing-api.md#addevent)), what's the motivation to duplicate such functionality in `startSpan()`?

0004-remove-spandata.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Explanation
+
+SpanData's primary use case comes from the need to construct and report out of band spans, meaning that you're creating "custom" spans for an operation you don't own. A good example of this is a program that takes in structured logs that contain correlation IDs and a duration (e.g. from splunk) and [converts them](https://github.com/lightstep/splunktospan/blob/master/splunktospan/span.py#L43) to spans for your tracing system. [Another example](https://github.com/lightstep/haproxy_log2span/blob/master/lib/lib.go#L292) is running a sidecar on an HAProxy machine, tailing the request logs, and creating spans. SpanData supports the out of band reporting case, whereas you can’t with the current Span API as you cannot set the start and end timestamp, or add any tags at span creation that the sampler might need.
+
+I'd like to propose getting rid of SpanData and `tracer.recordSpanData()` and replacing it by allowing `tracer.startSpan()` and `span.finish()` to accept start and end timestamp options. This reduces the API surface, consolidating on a single span type. Options would meet the requirements for out of band reporting.
```
reyang: Maybe it was my misunderstanding, I thought `SpanData` was targeting consumers (e.g. exporters) where immutable is desired.
`Span` was targeting producers, where different components can grab the "current active span" from the context/tracer and append extra information.

0004-remove-spandata.md
bg451: There's this hidden dependency on sampling that I've heard come up, where you might want to look at the tags or start+end timestamps of SpanData. If sampling occurs in `startSpan()`, you'd want to be able to provide information in bulk upfront for the sampler. I should probably clarify that the RFC specifically wants to add start and end timestamp options, the rest is optional and should be discussed.

reyang: I see, now it makes sense to me. Thank you!

0004-remove-spandata.md
```diff
@@ -0,0 +1,43 @@
+# Remove SpanData
+
+Remove and replace SpanData by adding span start and finish options.
+
+## Motivation
+
+SpanData represents an immutable span object, creating a fairly large API for all of the fields (12 to be exact). It exposes what feels like an SDK concern and implementation detail to the API surface. As a user, this is another API I need to learn how to use. As an implementer, it’s a new data type that needs to be supported. The primary motivation for removing SpanData revolves around the desire to reduce the size of the tracing API.
```
bogdandrutu: One other motivation can be that the tracing ids generation logic must be exposed to the user in order for them to generate the right ids.

0004-remove-spandata.md
bogdandrutu: SpanData in the API has the motivation explained by the RFC. In the SDK may have a different usage, but if we need something in the SDK then we can define it there.

bogdandrutu: I think what you are proposing is not enough:
* You may need a way to pass the "spanId" if in the logs you have the SpanId generated (I may be wrong but this is what I think you may also need) - we can deffer this but worth documented.
* You may need a way to pass events with a timestamp as well.

0004-remove-spandata.md
bogdandrutu: I think the withEvents (logs) for start and end are out of scope for this RFC. I would really appreciate if we do a different RFC for sampling where the need for these APIs will be explained and discussed.

0004-remove-spandata.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Trade-offs and mitigations
+
+From https://github.com/open-telemetry/opentelemetry-specification/issues/71: If the underlying SDK automatically adds tags to spans such as thread-id, stacktrace, and cpu-usage when a span is started, they would be incorrect for out of band spans as the tracer would not know the difference between in and out of band spans. This can be mitigated by indicating that the span is out of band to prevent attaching possibly incorrect information.
+
+https://github.com/open-telemetry/opentelemetry-specification/issues/96 discusses the possibility of allowing SpanData to support lazy fields, preventing allocations until it is read by the exporter. With span options, I don’t believe we would be able to get a fully lazy span, but we can effectively satisify the requirements, especially by adding lazily evaluated events and logs. OpenTracing added bulk logging options when finishing a span for this reason, allowing users to attach logs should, say, an error occur and they explicitly sample the span. Some discussion should be had about the laziness questions in the Open Questions section.
```
bogdandrutu: SpanData is already lazy in some ways, don't really see a need to support that.

0004-remove-spandata.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Internal details
+
+`startSpan()` would change so you can include options such as `withStartTimestamp()`, `withTags()`, `withResource()`, `withEvents()`, `withLogs()`, etc. and for `finish()` you could have `withEndTimestamp()`, `withEvents()`, `withLogs()`, etc. The exact implementation would be language specific, so some would use an options pattern with function overloading or variadic parameters, or a span builder pattern.
+
+## Trade-offs and mitigations
+
+From https://github.com/open-telemetry/opentelemetry-specification/issues/71: If the underlying SDK automatically adds tags to spans such as thread-id, stacktrace, and cpu-usage when a span is started, they would be incorrect for out of band spans as the tracer would not know the difference between in and out of band spans. This can be mitigated by indicating that the span is out of band to prevent attaching possibly incorrect information.
```
bogdandrutu: I think this implies some changes in the start options to specify if the span is reported out of band?

0004-remove-spandata.md
```diff
@@ -0,0 +1,43 @@
+# Remove SpanData
```
bogdandrutu: Please include the use-cases for this as well. See as an inspiration https://github.com/open-telemetry/opentelemetry-specification/issues/71#issuecomment-506083966

0004-remove-spandata.md
```diff
@@ -0,0 +1,43 @@
...
+
+The OpenTracing specification for `tracer.startSpan()` includes an optional start timestamp and zero or more tags. It also calls out an optional end timestamp for `span.finish()`.
+
+## Open questions
+
+What does lazy mean for SpanData? When we talk about lazy logs, we normally mean lazy evaluation, where expensive formatting is deferred until they are actually needed. Is laziness a desired property of SpanData? If so, what are the other requirements for SpanData? 
+
+There also seems to be some hidden dependency between SpanData and the sampler API. For example, given a complete SpanData object with a start and finish timestamp, I imagine there's a use case where the sampler can look at the that and decide "this took a long time" and sample it. Is this a real use case? Is there a requirement to be able to provide complete span objects to the sampler?
+
+## Related Issues
```
bogdandrutu: Is https://github.com/open-telemetry/opentelemetry-specification/issues/71 not closing as well?

0004-remove-spandata.md
jmacd: There has been additional confusion because the exporters declare a `SpanData` type, but it's not the same thing that's been discussed as a way to handle "out of band" spans. This RFC is about removing `SpanData` from the API, not the observer.

tsloughter:
I reread this and now I'm back to reading it as saying there is no way to add to a span during its life and only when it starts or finishes?

If that is the case one worry I have is exception handling. If you can only add events/logs/status when starting or finishing a span it removes the possibility to add those attributes during the life of the span and finishing the span in the handling of a thrown exception.

bg451:
> as saying there is no way to add to a span during its life and only when it starts or finishes?

@tsloughter This wouldn't be the case, just talking about the concept of bulk logging when a span finishes, which is something opentracing allows.

0004-remove-spandata.md
bg451: 👍 I'll move this to the Future Work.

tsloughter:
@bg451 ah, ok. The phrasing, "getting rid of SpanData and `tracer.recordSpanData()` and replacing it by allowing `tracer.startSpan()` and `span.finish()` to accept start and end timestamp options", threw me off.

Is this also around being able to create spans after the fact from log data? This is important to me for support of an Erlang/Elixir library that outputs events like database query times after the db call that we'd like to generate spans from and attach as children to the process that makes the db request.

bg451:
> Is this also around being able to create spans after the fact from log data?

If I'm understanding the question correctly, yes this is one of the supported use cases. If you can extract a trace context/identifier from your log, you'd be able to make it a child span in that trace. For example (with opentracing start options), this is a sidecar program that would tail HAProxy logs, look for a span context to become a child of, then create/finish the span with the timestamp and tags coming from the parsed log line: https://github.com/lightstep/haproxy_log2span/blob/master/lib/lib.go#L290

tsloughter:
@bg451 yup, exactly that, thanks.

bogdandrutu:
This is currently blocked on the author side to respond to all the comments. Waiting for that before another review.

0004-remove-spandata.md
bg451: I struggle with the idea of allowing users to set an explicit span ID but I think I understand the use case. For example, you have a sidecar that's accepting serialized spans that already have a span and trace ID, say from a legacy tracing system, that you want to convert to opentelem spans. I think if you're doing that you might just want to take a direct dependency on the sdk's span type and translate to that? If you only have a trace ID, you can set the it by passing in a span context with that trace id. I'll add a note that you may want to include a `withSpanID`.

I mention `withEvents()` in the internal details, does that cover the second bullet point? Alternatively `span.addEvent` could accept an optional timestamp.

0004-remove-spandata.md
bg451: 👍 I'll remove this from the RFC

0004-remove-spandata.md
```diff
@@ -0,0 +1,51 @@
...
+
+I'd like to propose getting rid of SpanData and `tracer.recordSpanData()` and replacing it by allowing `tracer.startSpan()` to accept a start timestamp option and `span.finish()` to accept end timestamp option. This reduces the API surface, consolidating on a single span type. Options would meet the requirements for out of band reporting.
+
+## Internal details
+
+`startSpan()` would change so you can include options such as `withStartTimestamp()` and `withResource()`. When you have a span sink, out of band spans may have different resources than the tracer they are being reported to, so you want to pass an explicit resource. For `span.finish()` you would have `withEndTimestamp()`. The exact implementation would be language specific, so some would use an options pattern with function overloading or variadic parameters, or add these options to the span builder.
```
bogdandrutu: `startSpan()` would change to optionally include a `start_timestamp` and a `resource`. Let's not make this go specific (where the withX option pattern comes from, at least that is my feeling).

tylerbenson: I don't like this aspect of the OpenTracing API since the clock resolution of an external timestamp is not as good as the cpu "nano" time.  I suggest we expose a method to get the "nano" time from the same clock used by the tracer (which perhaps returns a "typed" value instead of just a primitive), and accept the return value of that as the start/stop time argument in the builder.

The description in this document is pretty generic and likely to result in significant variation in implementation.  Please clarify

bogdandrutu: I think this comment applies when we write the specs for this RFC. We should keep this RFC a bit more high level.

Happy to hear what others think.

tylerbenson: Fair point.  Consider this feedback for the future spec. Carry on.

0004-remove-spandata.md
```diff
@@ -0,0 +1,51 @@
...
+
+The second use case comes from the need to construct and report out of band spans, meaning that you're creating "custom" spans for an operation you don't own. One good example of this is a span sink that takes in structured logs that contain correlation IDs and a duration (e.g. from splunk) and [converts them](https://github.com/lightstep/splunktospan/blob/master/splunktospan/span.py#L43) to spans for your tracing system. [Another example](https://github.com/lightstep/haproxy_log2span/blob/master/lib/lib.go#L292) is running a sidecar on an HAProxy machine, tailing the request logs, and creating spans. SpanData allows you to report the out of band reporting case, whereas you can’t with the current Span API as you cannot set the start and end timestamp.
```
songy23: Nit: use a specific commit instead of master in the Github links.

0004-remove-spandata.md
```diff
@@ -0,0 +1,51 @@
...
+
+From https://github.com/open-telemetry/opentelemetry-specification/issues/71: If the underlying SDK automatically adds tags to spans such as thread-id, stacktrace, and cpu-usage when a span is started, they would be incorrect for out of band spans as the tracer would not know the difference between in and out of band spans. This can be mitigated by indicating that the span is out of band to prevent attaching incorrect information, possibly with an `isOutOfBand()` option.
+
+## Prior art and alternatives
+
+The OpenTracing specification for `tracer.startSpan()` includes an optional start timestamp and zero or more tags. It also calls out an optional end timestamp and bulk logging for `span.finish()`.
+
+## Open questions
+
+Imagine a scenario where you have a sidecar that accepts serialized spans from a legacy tracing system that you want to convert to OpenTelem spans. Those legacy spans have a trace and span ID you want to reuse. Should `startSpan()` include an option to explicitly set the span ID? Would it make more sense to take a direct dependency on the SDK's span type and translate to that?
```
songy23:
```suggestion
Imagine a scenario where you have a sidecar that accepts serialized spans from a legacy tracing system that you want to convert to OpenTelemetry spans. Those legacy spans have a trace and span ID you want to reuse. Should `startSpan()` include an option to explicitly set the span ID? Would it make more sense to take a direct dependency on the SDK's span type and translate to that?
```

0004-remove-spandata.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Motivation
+
+SpanData represents an immutable span object, creating a fairly large API for all of the fields (12 to be exact). It exposes what feels like an SDK concern and implementation detail to the API surface. As a user, this is another API I need to learn how to use, and ID generation might also need to be exposed. As an implementer, it is a new data type that needs to be supported. The primary motivation for removing SpanData revolves around the desire to reduce the size of the tracing API.
+
+## Explanation
+
+SpanData has a couple of use cases.
+
+The first use case revolves around creating a span syncronously but needing to change the start time to a more accurate timestamp. For example, in an HTTP server, you might record the time the first byte was received, parse the headers, determine the span name, and then create the span. The moment the span was created isn't representative of when the request actually began, so the time the first byte was received would become the span's start time. Since the current API doesn't allow start timestamps, you'd need to create a SpanData object. The big downside is that you don't end up with an active span object.
```
reyang: syncronously -> synchronously

0004-remove-spandata.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Prior art and alternatives
+
+The OpenTracing specification for `tracer.startSpan()` includes an optional start timestamp and zero or more tags. It also calls out an optional end timestamp and bulk logging for `span.finish()`.
+
+## Open questions
+
+Imagine a scenario where you have a sidecar that accepts serialized spans from a legacy tracing system that you want to convert to OpenTelemetry spans. Those legacy spans have a trace and span ID you want to reuse. Should `startSpan()` include an option to explicitly set the span ID? Would it make more sense to take a direct dependency on the SDK's span type and translate to that?
+
+There also seems to be some hidden dependency between SpanData and the sampler API. For example, given a complete SpanData object with a start and finish timestamp, I imagine there's a use case where the sampler can look at the that and decide "this took a long time" and sample it. Is this a real use case? Is there a requirement to be able to provide complete span objects to the sampler?
```
reyang: It is a very valid use case to have "deferred sampling decision".

0004-remove-spandata.md
```diff
@@ -0,0 +1,51 @@
...
+
+From https://github.com/open-telemetry/opentelemetry-specification/issues/71: If the underlying SDK automatically adds tags to spans such as thread-id, stacktrace, and cpu-usage when a span is started, they would be incorrect for out of band spans as the tracer would not know the difference between in and out of band spans. This can be mitigated by indicating that the span is out of band to prevent attaching incorrect information, possibly with an `isOutOfBand()` option.
+
+## Prior art and alternatives
+
+The OpenTracing specification for `tracer.startSpan()` includes an optional start timestamp and zero or more tags. It also calls out an optional end timestamp and bulk logging for `span.finish()`.
+
+## Open questions
+
+Imagine a scenario where you have a sidecar that accepts serialized spans from a legacy tracing system that you want to convert to OpenTelemetry spans. Those legacy spans have a trace and span ID you want to reuse. Should `startSpan()` include an option to explicitly set the span ID? Would it make more sense to take a direct dependency on the SDK's span type and translate to that?
```
draffensperger: This is not just for legacy span cases. In the web tracing case we have a use for this as well. The browser does not send trace headers for the initial page navigation, so in order to relate the server and client spans for the initial load, we need the server to send back its parent trace and span ID, and have the browser client span (generated from browser performance timings) set its span ID to the provided parent span ID of the server. Here's [code in OpenCensus Web for this](https://github.com/census-instrumentation/opencensus-web/blob/4ab81112ce16216929a507b8e494a95d6d83298d/packages/opencensus-web-instrumentation-perf/src/initial-load-root-span.ts#L75).

It would be really helpful for `startSpan` to take an optional span ID. I'd rather not use `SpanData` because that's feels like extra complexity and in the case of the web browser, extra JS code that needs to be shipped and parsed and so would (slightly) slow down end user code the embeds the library.

draffensperger: @bg451 would you be willing to update the language here to talk about the web use case in terms of receiving a span ID when creating a new span?

0004-remove-spandata.md
draffensperger: Would it make sense to consolidate the logic for deferred sampling into the [opentelemetry-service](https://github.com/open-telemetry/opentelemetry-service), which can handle it in the general distributed trace case?

bogdandrutu:
Ping. Can you please respond to the last round of comments?

bg451:
@bogdandrutu done, please take a look

0004-remove-spandata.md
```diff
@@ -18,7 +18,7 @@ I'd like to propose getting rid of SpanData and `tracer.recordSpanData()` and re
 
 ## Internal details
 
-`startSpan()` would change so you can include an optional start timestamp and resource. When you have a span sink, out of band spans may have different resources than the tracer they are being reported to, so you want to pass an explicit resource. For `span.finish()` you would have an optional end timestamp. The exact implementation would be language specific, so some would use an options pattern with function overloading or variadic parameters, or add these options to the span builder.
+`startSpan()` would change so you can include an optional start timestamp, span ID, and resource. When you have a span sink, out of band spans may have different resources than the tracer they are being reported to, so you want to pass an explicit resource. For `span.finish()` you would have an optional end timestamp. The exact implementation would be language specific, so some would use an options pattern with function overloading or variadic parameters, or add these options to the span builder.
```
draffensperger: Awesome!!! :+1:

bhs:
@bg451 what's the status on this? Are you/we waiting for approval from someone? Is there some other meaningful blocker to merging this PR (aside from git conflicts)?

carlosalberto:
Hey hey @bogdandrutu 

I will merge this PR later today unless you have something to say ;)

And of course, you can always provide feedback/PRs later on.

bogdandrutu:
@carlosalberto one small issue that can be resolved in the specification is time correlation (ensuring monotonic time is used). We do want to ensure inside a span or a sub-trace (spans generated by the same process) that the timestamp is correctly (monotonic) calculated.

I would suggest just a small item in the future work for the moment.

0004-remove-spandata.md
```diff
@@ -0,0 +1,49 @@
...
+
+The second use case comes from the need to construct and report out of band spans, meaning that you're creating "custom" spans for an operation you don't own. One good example of this is a span sink that takes in structured logs that contain correlation IDs and a duration (e.g. from splunk) and [converts them](https://github.com/lightstep/splunktospan/blob/58ef9bca81933a47605a51047b12742e2d13aa8f/splunktospan/span.py#L43) to spans for your tracing system. [Another example](https://github.com/lightstep/haproxy_log2span/blob/761da5bf3e4b6cce56039d65439ae7db57f48603/lib/lib.go#L292) is running a sidecar on an HAProxy machine, tailing the request logs, and creating spans. SpanData allows you to report the out of band reporting case, whereas you can’t with the current Span API as you cannot set the start and end timestamp.
+
+I'd like to propose getting rid of SpanData and `tracer.recordSpanData()` and replacing it by allowing `tracer.startSpan()` to accept a start timestamp option and `span.finish()` to accept end timestamp option. This reduces the API surface, consolidating on a single span type. Options would meet the requirements for out of band reporting.
```
pavolloffay: `s/span.finish()/s.end()/g` - the finish is defined in various places

pavolloffay:
There are benefits of defining it in the SDK. For instance sdk-processors spec can reference it.

bogdandrutu:
@bg451 please fix the last two comments and I will merge this.

bg451:
> There are benefits of defining it in the SDK.

Just to clarify, it being spandata here right? If so I agree, there still needs to be some sort of concrete type that SDK exporters can read from.

0004-remove-spandata.md
bogdandrutu: done. I changed the file.

text/0002-remove-spandata.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Internal details
+
+`startSpan()` would change so you can include an optional start timestamp, span ID, and resource. When you have a span sink, out of band spans may have different resources than the tracer they are being reported to, so you want to pass an explicit resource. For `span.end()` you would have an optional end timestamp. The exact implementation would be language specific, so some would use an options pattern with function overloading or variadic parameters, or add these options to the span builder.
+
+## Trade-offs and mitigations
+
+From https://github.com/open-telemetry/opentelemetry-specification/issues/71: If the underlying SDK automatically adds tags to spans such as thread-id, stacktrace, and cpu-usage when a span is started, they would be incorrect for out of band spans as the tracer would not know the difference between in and out of band spans. This can be mitigated by indicating that the span is out of band to prevent attaching incorrect information, possibly with an `isOutOfBand()` option.
```
bogdandrutu: Mention that the `isOutOfBand()` option is part of the startSpan.

bogdandrutu:
@bg451 please also respond to my comments in this post https://github.com/open-telemetry/rfcs/pull/8#issuecomment-518709601

bg451:
Oh huh I thought I pushed my changes yesterday morn, looks like I didn't.

@bogdandrutu https://github.com/open-telemetry/rfcs/pull/8/commits/375c3977021502808ba452f2fb5a6f6b0ded7ce5#diff-f62a07a10e742d00e2ef6471e1d80d13R39 re your comment

bogdandrutu:
@tedsuo @carlosalberto @AloisReitbauer @yurishkuro @SergeyKanzhelev @c24t any more comments on this? I would like to merge it but I am looking to have at least one or ideally two more approvers.

PR #9, opened 2019-07-23
Rename telemetry-without-manual-instrumentation 0002 -> 0001.

bogdandrutu:
This is the first RFC merged.


bogdandrutu:
@iredelmeier I think I am wrong, based on https://github.com/open-telemetry/rfcs#submitting-a-new-rfc what should be the name and path for this?

PR #11, opened 2019-07-31
Global initializer requirements proposal

jmacd:
This proposal is supported by demonstration code here:
https://github.com/jmacd/opentelemetry-go/pull/3

This is heavily influenced by: https://github.com/open-telemetry/opentelemetry-ruby/issues/19

This is attempting to resolve: https://github.com/open-telemetry/opentelemetry-go/issues/52


bhs:
@jmacd what's the current status on this PR? What are we blocked on, if anything? Thanks in advance...

0010-global-init.md
```diff
@@ -0,0 +1,110 @@
...
+Testing should be performed without depending on the global SDK.
+
+### Synchronization
+
+Since the global Tracer and Meter objects are required to begin
+working once the SDK is installed, there is some implied
+synchronization overhead at startup, overhead we expect to fall after
+the SDK is installed.  We recommend explicitly installing a No-op SDK
+to fully disable instrumentation, as this approach will have a lower
+overhead than leaving the OpenTracing library uninitialized.
```
carlosalberto: *OpenTelemetry

carlosalberto: Btw, in OpenTracing/OpenTelemetry Java a default no-op SDK is installed, so no explicit installation of a no-op SDK is needed at all for this case, just FYI ;)

0010-global-init.md
```diff
@@ -0,0 +1,110 @@
...
+As an example that does not qualify as "commonly accepted", see [Go
+issue 52](https://github.com/open-telemetry/opentelemetry-go/issues/52)
+which demonstrates using the Go `plugin` package to load a
+configurable SDK prior to first use.
+
+## Open questions
+
+What other options should be passed to the explicit global initializer?
+
+Is there a public test for "is the SDK installed; is it a no-op"?
```
carlosalberto: In OpenTracing we have `GlobalTracer.isRegistered()` (to tell whether **any** SDK has been installed) but that works mostly as a hint ;)

0010-global-init.md
```diff
@@ -0,0 +1,110 @@
...
+## Requirements: Explicit initializer
+
+OpenTelemetry specifies that the default implementation is
+non-operational (i.e., a "no-op"), requiring that API method calls
+result in effectively zero instrumentation overhead.  We expect third
+party libraries to use the global SDK before it is installed, which is
+addressed in a requirement stated below.
+
+The explicit initializer method should take independent `Tracer` and
+`Meter` objects (e.g., `opentelemetry.Init(Tracer, Meter)`).  The SDK
```
reyang: I suggest that we have separate initialization for Tracer and Meter, even in the examples here.

0010-global-init.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Meter
+
+There may be loss of metrics at startup.
+
+Metric SubMeasure objects (i.e., metrics w/ predefined labels)
+initialized before the SDK is installed will redirect to the global
+SDK after it is installed.
+
+### Concrete types
```
reyang: I wonder what shall we do with propagators (e.g. should propagators be part of tracing API). We can explore this in a separate PR though.

bogdandrutu:
@jmacd please respond to the comments.

0010-global-init.md
jmacd: I was trying to emphasize a minor point, which is that there may be a difference in performance between the default/implicit noop implementation and an explicit noop implementation. The explicit noop implementation can do nothing. The implicit noop implementation, if it is to "start working after the SDK is first registered", must do more than the explicit noop implementation. This is a super minor point, but if I wanted the lowest overhead possible, I should explicitly install a noop. 

In Java, the presence of a Service Provider Interface sidesteps this concern. In Java, the elected SDK can be installed before it's first use.

0010-global-init.md
jmacd: :+1: 
I'd lean toward having the Tracer and Meter interfaces support a "Am I not a Noop?" method.

0010-global-init.md
jmacd: I believe it's possible with this API to independently construct and configure the Tracer and Meter that will be used, all that this requires is that they be provided in a single initializer call. I believe that we should encourage a single initializer call so that when a combined implementation (i.e., an SDK that presents both APIs) is registered, it can be sure that its registration is all-or-none.

In any case, it should be possible to provide Tracer and Meter implementations that are completely independent, or to pass a Noop implementation of either in the initializer.

0010-global-init.md
jmacd: I see a separate issue here, yes.

jmacd:
I've responded to the comments. I don't see this proposal as preventing the use of independent Tracer and Meter implementations.

text/0005-global-init.md
```diff
@@ -0,0 +1,110 @@
...
+Testing should be performed without depending on the global SDK.
+
+### Synchronization
+
+Since the global Tracer and Meter objects are required to begin
+working once the SDK is installed, there is some implied
+synchronization overhead at startup, overhead we expect to fall after
+the SDK is installed.  We recommend explicitly installing a No-op SDK
+to fully disable instrumentation, as this approach will have a lower
+overhead than leaving the OpenTracing library uninitialized.
```
bogdandrutu: s/OpenTracing/OpenTelemetry

bogdandrutu:
@jmacd ping, can you please fix the last comment.

jmacd:
Done.

jmacd:
After https://github.com/open-telemetry/rfcs/pull/16, I believe we'll change the global initialization logic to install factories rather than concrete Tracer and Meter interfaces.

PR #12, opened 2019-08-01
Propose consolidating into a single implementation

iredelmeier:


text/0000-support-single-implementation.md
```diff
@@ -0,0 +1,131 @@
...
+
+Given that OpenTelemetry is still very young and not yet production-ready, migration costs should be relatively small for both end users and implementors - particularly to the extent that the ambiguity between the API and SDK has slowed down implementation.
+
+## Prior art and alternatives
+
+### Prior art
+
+The three main prior art examples are OpenTracing, OpenCensus, and the existing OpenTelemetry architecture.
+
+**OpenTracing** handles export functionality by having each vendor (or other party) reimplement the whole API. While this maximizes the flexibility for vendors, it also means that everyone must redo a lot of the same logic - including any logic that may be considered vendor-agnostic.
```
bhs: This is not a *requirement* of OpenTracing – it's feasible to create a high-quality, general-purpose implementation that vendors may choose (or not-choose) to extend. We did not have enough time/staffing to do that in OpenTracing out of the gate, but it was (and remains) an option.

text/0000-support-single-implementation.md
```diff
@@ -0,0 +1,131 @@
...
+
+**OpenCensus** handles it by having a single implementation which then sends data to tracing and stats exporters. _(**Note:** While architecturally this is very similar to the RFC's proposal, key differences include OpenCensus itself being quite opinionated and sometimes stateful, depending on the implementation.)_
+
+The **existing OpenTelemetry architecture** essentially combines the OpenTracing and OpenCensus approaches. As with OpenTracing, the core API provides an interface that can then have multiple implementations. The equivalent to the OpenCensus "opinionated implementation" is then provided as a *particular* implementation (the SDK), which in turn supports exporters.
```
bhs: If I understand correctly, the delta between this RFC and the OpenCensus approach is – essentially – that the scope of the "common implementation" proposed here is smaller than the one it OpenCensus?

text/0000-support-single-implementation.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Alternatives
+
+What are some prior and/or alternative approaches? For instance, is there a corresponding feature in OpenTracing or OpenCensus? What are some ideas that you have rejected?
+
+## Open questions
+
+What are some questions that you know aren't resolved yet by the RFC? These may be questions that could be answered through further discussion, implementation experiments, or anything else that the future may bring.
```
bhs: An alternative – and, frankly, one I would prefer – would be to maintain the separation between API and implementation but commit to a "standard" or "recommended" implementation that those of us on the ground floor of OpenTelemetry would all (most likely) use, at least "for now."

As small as a reified Span object sounds, in practice it will be the place that certain optimizations and quirks live, so I there is a meaningful "unknown unknown" here about forwards-compatibility, both with optimizations and features we haven't thought of yet, and also with vendors who are not participating in these early days but will participate later on in 2020+.

yurishkuro:
@iredelmeier could you please provide a better explanation/justification for this proposal?

Separation of interface from _any_ implementation was a pre-condition for merging OpenTracing and OpenCensus. It's been debated ad nauseam, with many vendors saying that they cannot do some of their internal house-keeping if they are forced to extend base classes, however basic they may be. And there have been use cases for using the OpenTracing API for different purposes altogether, sometimes not even related to telemetry (meaning that "basic" implementation with ID generation and timestamps would introduce unnecessary work).

I looked through the example issues in the footnote, and did not find why they would require going back on the most fundamental decision about OpenTelemetry. The only explanation in the RFC is this one sentence:

> However, the solution comes at a significant cost: it seems to be both confusing to users as well as cumbersome for anyone implementing or otherwise working on OpenTelemetry, in addition to whatever technical downsides (such as duplication of code across implementations) there may be.

I did not see any references to complaints from _end users_ in the proposal. People implementing OpenTelemetry are not end users, their numbers are nothing compared to the number of app developers who would be interacting with the API and can be completely agnostic to the implementation details. The code duplication argument doesn't hold, as Ben also commented, having an abstract API does not preclude us from having a base implementation that people can extend.

discostu105:
From our point of view, we would really like to keep the API layer. As APM vendors, we can choose to directly implement the API, or use the official OTel SDK. It gives us maximum flexibility in how we implement OpenTelemetry, in case there are features that we just have a different take on (sampling decisions, capturing additional information (cpu time, thread profiling, etc...).

Concerning confusion for end-users: Can we somehow tackle this problem differently? E.g. a better/clearer naming policy for the two packages that is also consistent across techs? Or better documentation?

Oberon00:
@iredelmeier you have not reacted to this PR since opening it. Does that mean it is abandoned? Can it be closed?

tedsuo:
There's a lot of support for leaving the API/SDK separation in place. 

But discussing this proposal in various forums showed me a couple of things.

* We need to explain the requirements that led to API/SDK separation. Though it's in the zeitgeist, We don't really have this written out anywhere in the OpenTelemetry project. We should fix that.
* A batched `Exporter` - like the one in the current SDK design - could be implemented on top of an API Observer (`SpanProcessor` in OTel lingo). This approach is more modular, and could lead to a cleaner SDK implementation. So, even with the API remaining separate, this work could be applied to cleaning up our SDK architecture/internals.

@iredelmeier, if the above looks reasonable to you, what do you think about closing this proposal and opening one focused on a layered SDK? I don't think it's as urgent as the API proposals we need for the beta, but it could be implemented as a backwards-compatible improvement to the SDK once we've got the time.

jmacd:
I endorse the layering suggestions, but I suggest we close this.  Let's see if we can define what a SpanProcessor is, since it's definitely a layer intended to be part of an "export pipeline". Is the SpanProcessor part of the SDK or is there a layer between the SDK/API surface and the SpanProcessor?

Related: https://github.com/open-telemetry/opentelemetry-specification/pull/205

tedsuo:
Since this RFC appears to be inactive, I am going to close the PR for now. @iredelmeier please feel free to open a new PR if you would like to propose further SDK changes.

PR #16, opened 2019-08-12
Propose "Named Tracers"

z1c0:


text/0000-named-tracers.md
```diff
@@ -0,0 +1,62 @@
...
+
+The purpose of _Named Tracers_ is the ability to allow for a more fine-grained control of the amount of tracing data being generated by a library or application. Associating a name (e.g. _"io.opentelemetry.contrib.mongodb"_) with a tracer allows for a logical coupling of this tracer and a certain technology or library (like _MongoDB_ in this example).
+
+By using *Named Tracers* a library can enable / disable tracing of a specific library, which might be necessary in case:
+* A library causes too much runtime or memory overhead.
+* A library has issues or bugs (e.g. does not call `span.end()` in every case).
+* "Double-Tracing" becomes a problem. This can be the case if a vendor (e.g. Dynatrace) might already have built-in auto-instrumentation for a certain library. In this case the vendor might want to decide (automatically or configuration-based) which of the implementations should be enabled or disabled.
```
bhs: This third issue is the only one that really moves me as a problem-worth-solving, at least at this stage of the project... bugs are bugs and should be fixed; and overhead can be controlled without introducing something this complex and general-purpose.

If we're focused on the double-instrumentation problem, I would suggest doing something simpler, though – it seems like we want more of a "registry" of the application's telemetry surface area where we can ensure single-coverage / conflict resolution / etc. I am very much in favor of coming up with a credible (though I'd also argue for "minimal" / simple) solution to that problem, but I don't think it would be "named tracers" per se.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+In a way, the `TracerFactory` replaces the global `Tracer` singleton object as a ubiquitous point to request a tracer instance.
+
+
+## Internal details
+
+By providing a TracerFactory and *Named Tracers*, a vendor or OpenTelemetry implementation gains more flexibility in providing tracers and can decide whether or not to produce spans for a specific traced component and even change this behavior at runtime.
+
+In the simplest case, an OpenTelemetry implementation can return a single instance for a requested tracer regardless of the name specified. This could be the case for implementations that do no want/need to enable or disable a tracer.
```
arminru:
```suggestion
In the simplest case, an OpenTelemetry implementation can return a single instance for a requested tracer regardless of the name specified. This could be the case for implementations that do not want/need to enable or disable a tracer.
```

text/0000-named-tracers.md
```diff
@@ -0,0 +1,62 @@
...
+
+Alternatively, an implementation can provide different tracers per specified tracer name, thus being able to associate this tracer with the component being traced. This allows for the possibility to enable / disable a tracer based on a custom configuration.
+* Automatically set the `component` ("the component being traced") on every span being produced.
+
+## Trade-offs and mitigations
+
+## Prior art and alternatives
+
+Alternatively, instead of having a `TracerFactory`, existing (global) tracers could return and additional indirection objects (called e.g. `TraceComponent`), which would be able to produce spans for specifically named traced components.
```
arminru:
```suggestion
Alternatively, instead of having a `TracerFactory`, existing (global) tracers could return additional indirection objects (called e.g. `TraceComponent`), which would be able to produce spans for specifically named traced components.
```

text/0000-named-tracers.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+Overall, this would not change a lot since the levels of indirection until producing an actual span are the same.
+
+
+## Open questions
+
+## Future possibilities
+
+By adapting this proposal, current implementations that do not honour the specified tracer name and provide a single global tracer, do not required much change. However they could change that behavior in future versions and provide more specific tracer implementations then. On the other side, if the mechanism of *Named Tracer* is not a part of the initial specification, such scenarios will be prevented and hard to retrofit in future version, should they be deemed necessary then.
```
arminru:
```suggestion
By adapting this proposal, current implementations that do not honor the specified tracer name and provide a single global tracer, would not require much change. However they could change that behavior in future versions and provide more specific tracer implementations then. On the other side, if the mechanism of *Named Tracer*s is not a part of the initial specification, such scenarios will be prevented and hard to retrofit in future version, should they be deemed necessary then.
```

text/0000-named-tracers.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Motivation
+
+Instead of having a global tracer singleton, every instrumentation library needs to create its own tracer, which also has a proper name that identifies the library / logical component that uses this tracer. If there are two different instrumentation libraries for the same technology (e.g. MongoDb), these instrumentation libraries should have distinct names.
+
+This proposal originates from an `opentelemetry-specification` proposal on [components](https://github.com/open-telemetry/opentelemetry-specification/issues/10), but the term "components" was quickly found to being too generic and possibly misleading, since it is already used frequently in differing contexts and connotations.
+
+The purpose of _Named Tracers_ is the ability to allow for a more fine-grained control of the amount of tracing data being generated by a library or application. Associating a name (e.g. _"io.opentelemetry.contrib.mongodb"_) with a tracer allows for a logical coupling of this tracer and a certain technology or library (like _MongoDB_ in this example).
+
+By using *Named Tracers* a library can enable / disable tracing of a specific library, which might be necessary in case:
```
arminru: I would phrase this more generically – for (1)-(3) it might be a user who decides to switch that instrumentation off in their environment (e.g. via a setting in the backend), only for (3) it is likely that one tracing library turns off another one.

```suggestion
*Named Tracers* allow enabling and disabling tracing of a specific library, which might be necessary in case:
```

arminru: I would phrase this more generically – for (1)-(3) it might be a user who decides to switch that instrumentation off in their environment (e.g. via a setting in the backend), only for (3) it is likely that one tracing library turns off another one.

```suggestion
*Named Tracers* allow enabling and disabling tracing of a specific library, which might be necessary in case:
```

text/0000-named-tracers.md
arminru: We've seen demand from various enterprise customers to quickly turn off specific instrumentation also for reasons (1) and (2) under certain circumstances. An example are cases when the instrumentation usually works well but misbehaves in certain use cases not considered when adding the instrumentation, e.g., when an instrumented library is integrated in a new, high-load environment.

To me, the registry you are proposing sounds like something that would be build up from the names passed to the TracerFactory proposed in this RFC. How else would you expect such a registry to be populated? Or are you thinking of a centrally maintained global registry of all known instrumentations hosted by OpenTelemetry rather than one dynamically built up for each specific application instance?

text/0000-named-tracers.md
```diff
@@ -0,0 +1,62 @@
+# Named Tracers 
+
```
arminru:
```suggestion

**Status:** `proposed`

```

text/0000-named-tracers.md
discostu105: Fully agree to @arminru 's comment.

I'll try to outline a scenario that explain cases (1) and (2):

Customer runs an app in production. It's monitored by a certain APM vendor (e.g. Dynatrace). Everything works smoothly.

Customer deploys a new version of the application, which has new OpenTelemetry instrumentation (either new library with OTel support or custom instrumentation). However, the new instrumentation is far too chatty (produces too many spans), is too heavyweight (e.g. while populating attributes, it queries expensive API's) or is buggy (span.end not properly called). In test and pre-prod this is maybe not really a problem, because only the No-Op Tracer is tested.

As soon as that deployment hits production, a different Tracer is used and suddenly the problems surface. In that case a customer would turn to the APM vendor for support and in enterprise scenarios a quick solution is needed. Customers are often unable to quickly deploy new application versions, so they would end up being forced to disable APM monitoring until then.

In such a case, we would like to be able to identify various active "instrumentations" and disable the offending one. That way, we could troubleshoot "live", without any application re-deployment.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,69 @@
...
+## Open questions
+
+## Future possibilities
+
+By adapting this proposal, current implementations that do not honor the specified tracer name and provide a single global tracer, would not require much change. However they could change that behavior in future versions and provide more specific tracer implementations then. On the other side, if the mechanism of *Named Tracers* is not a part of the initial specification, such scenarios will be prevented and hard to retrofit in future version, should they be deemed necessary then.
+
+## Examples (for Tracer names)
+
+Since tracer names describe the libraries which use the tracers, those names should be defined in a way that makes them as unique as possible.
```
discostu105: I would clarify this: The name of the tracer should represent the identity of the _library, class or package that provides the instrumentation_. Examples:

Some more examples:
 * "io.opentracing.contrib.spring.rabbitmq"
 * "io.opentracing.contrib.jdbc"
 * "io.opentracing.thrift"
 * "io.opentracing.contrib.asynchttpclient"
 * "io.opencensus.contrib.http.servlet"
 * "io.opencensus.contrib.spring.sleuth.v1x"
 * "io.opencensus.contrib.http.jaxrs"
 * "github.com/opentracing-contrib/go-amqp" (Go)
 * "github.com/opentracing-contrib/go-grpc" (Go)
 * "OpenTracing.Contrib.NetCore.AspNetCore" (.NET)
 * "OpenTracing.Contrib.NetCore.EntityFrameworkCore" (.NET)

text/0000-named-tracers.md
bhs: @arminru @discostu105 thanks for your replies above. Much of this makes sense to me.

It sounds like you want to create a name – and a "single throat to choke" – for some set of instrumentation (that likely pertains to a single library / component / whatever).

After sleeping on this, I think the naming here is confusing me... I think of a "Tracer" as basically a "sink" for telemetry data; where here the term "Tracer" is actually being used to describe a "source" for telemetry data.

I stand by my desire for a (simple, basic) registry API, and I would also suggest calling these something other than a "Tracer" since this new concept is used to describe where telemetry comes *from*, not where it goes *to*.

Thoughts??

discostu105: @bhs Thanks for your reply. We're totally open for a different way of achieving this goal. Do you have a suggestion for such a registry API?

In this proposal, the `TracerFactory` basically acts as such a registry. IMHO it is a very simple approach and it's borrowed from Logging (log4j), where this you'd also create a Logger for each "source". E.g.

```
/* Get the logger for the actual class name to be printed on */
private final Logger log = LoggerFactory.getLogger(App.class);
```

jmacd: I like the `log4j` model as well. This suggests that the SDK concept we've been discussing is really a factory for named Tracer and Meter interfaces, and the calling convention will be to statically allocate one per module of code, using the global SDK I presume:

```
import io.opentelemetry.GlobalSDK;

  private final Tracer tracer = GlobalSDK.getTracer(MyClass.class);
  private final Meter meter = GlobalSDK.getMeter(MyClass.class);
```

bhs: TL;DR I like @jmacd's proposal ^^^

text/0000-named-tracers.md
```diff
@@ -0,0 +1,69 @@
...
+
+*Named Tracers* allow enabling and disabling tracing of a specific library, which might be necessary in case:
+* A library causes too much runtime or memory overhead.
+* A library has issues or bugs (e.g. does not call `span.end()` in every case).
+* "Double-Tracing" becomes a problem. This can be the case if a vendor (e.g. Dynatrace) might already have built-in auto-instrumentation for a certain library. In this case the vendor might want to decide (automatically or configuration-based) which of the implementations should be enabled or disabled.
+
+
+## Explanation
+
+From a user perspective, working with *Named Tracers* and *TracerFactories* is very similar to how logging frameworks like [log4j](https://www.slf4j.org/apidocs/org/slf4j/LoggerFactory.html) work. In analogy to requesting Logger objects through LoggerFactories, a tracing library would create specific 'Tracer' objects through a 'TracerFactory'.
```
bogdandrutu: Not only log4j but the entire Java logging API works like that, see https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html#getLogger(java.lang.String).

text/0000-named-tracers.md
bogdandrutu: +1 for the factory like method.

@jmacd I think the factory should be part of the main trace/metrics/etc package and the global instance should have a `getTracerFactory().getTracer("my_component");`. The reason is that if people want to manually pass a `Tracer` to a plugin for example they should pass the `TracerFactory` not the initialized `Tracer` with component name.

Maybe in future we would like to have a "telemetry level" similar with logs.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,69 @@
...
+
+```java
+Tracer tracer = OpenTelemetry.Tracing.getTracerFactory().getTracer("io.opentelemetry.contrib.mongodb");
```
bogdandrutu: why not just:

```
Tracer tracer = OpenTelemetry.getTracerFactory().getTracer("io.opentelemetry.contrib.mongodb");
```

text/0000-named-tracers.md
```diff
@@ -0,0 +1,69 @@
...
+* [Global Instance discussions](https://github.com/open-telemetry/opentelemetry-specification/labels/global%20instance)
+
+## Motivation
+
+Instead of having a global tracer singleton, every instrumentation library needs to create its own tracer, which also has a proper name that identifies the library / logical component that uses this tracer. If there are two different instrumentation libraries for the same technology (e.g. MongoDb), these instrumentation libraries should have distinct names.
+
+This proposal originates from an `opentelemetry-specification` proposal on [components](https://github.com/open-telemetry/opentelemetry-specification/issues/10), but the term "components" was quickly found to being too generic and possibly misleading, since it is already used frequently in differing contexts and connotations.
+
+The purpose of _Named Tracers_ is the ability to allow for a more fine-grained control of the amount of tracing data being generated by a library or application. Associating a name (e.g. _"io.opentelemetry.contrib.mongodb"_) with a tracer allows for a logical coupling of this tracer and a certain technology or library (like _MongoDB_ in this example).
```
bogdandrutu: In a default implementation (SDK), should the name of the tracer be incorporated in the Span name or should be a property on the Span? Same question for Metrics as well. I think incorporating this in the name of the metrics makes a lot of sense because then the metric name must be unique inside the component.

text/0000-named-tracers.md
bogdandrutu: Does getTracer for the same name require to return the same Tracer instance?

text/0000-named-tracers.md
reyang: Same here, like @jmacd's proposal.

text/0000-named-tracers.md
reyang: Entire Python logging API also has very similar idea.

text/0000-named-tracers.md
reyang: My vote for "yes".

text/0000-named-tracers.md
```diff
@@ -0,0 +1,69 @@
...
+In a way, the `TracerFactory` replaces the global `Tracer` singleton object as a ubiquitous point to request a tracer instance.
+
+
+## Internal details
+
+By providing a TracerFactory and *Named Tracers*, a vendor or OpenTelemetry implementation gains more flexibility in providing tracers and can decide whether or not to produce spans for a specific traced component and even change this behavior at runtime.
+
+In the simplest case, an OpenTelemetry implementation can return a single instance for a requested tracer regardless of the name specified. This could be the case for implementations that do not want/need to enable or disable a tracer.
+
+Alternatively, an implementation can provide different tracer instances per specified tracer name, thus being able to associate this tracer with the component being traced. This allows for the possibility to enable / disable a tracer based on a custom configuration.
```
reyang: Do we envision multiple SDKs (or tracer implementations) to co-exist in the same process?

bogdandrutu:
Please everyone review this RFC, this is very important for us to meet our goal to release in September an alpha version.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Future possibilities
+
+By adapting this proposal, current implementations that do not honor the specified tracer name and provide a single global tracer, would not require much change. However they could change that behavior in future versions and provide more specific tracer implementations then. On the other side, if the mechanism of *Named Tracers* is not a part of the initial specification, such scenarios will be prevented and hard to retrofit in future version, should they be deemed necessary then.
+
+## Examples (for Tracer names)
+
+Since tracer names describe the libraries which use the tracers, those names should be defined in a way that makes them as unique as possible.
+
+Proposed naming is in analogy to Java package names, e.g: "io.opentelemetry.contrib.mongodb".
```
yurishkuro: I would remove the sole reference to Java and instead provide examples in different languages. For example, in Node.js and Python the module names are _not_ hierarchical, but the respective registries (npm, pypi) provide what could be considered canonical names.

text/0000-named-tracers.md
yurishkuro: This needs to prescribe the actual behavior. What does it mean to "disable" the tracer for some component that's in the middle of the stack, e.g. layer B in A->B->C? What happens when instrumentation still creates a new span? Is it not going to be exported? what happens to the child of it (layer C), what would it be its parent?

z1c0: This RFC targets the API level rather than the SDK. Hence, questions like tracer verbosity or turning tracers on / off are something the SDK or a vendor-implementation can decide based on e.g. a proprietary configuration. On an API level there would currently be no notion of a configuration.

For example, the spec does not prescribe what happens with the span name either. It (more or less) describes what it means, but it does not set any rules how implementation should process the span name. It is true that it would make sense for the SDK to have some additional config APIs to e.g. discard span operations based on the tracer name, but this is out of scope for this RFC

text/0000-named-tracers.md
z1c0: Thanks for the feedback. I understand how referencing Java and Java naming conventions can be misleading here. However, I would not get too hung up on syntactics here. Any (unique) string that can identify a a tracer for a given language is fine.

Also, see the comment by @discostu105 (https://github.com/open-telemetry/rfcs/pull/16#discussion_r313749544).
Would adding these examples to the proposal clarify that?

text/0000-named-tracers.md
yurishkuro: > the spec does not prescribe what happens with the span name either.

That's not true, it does provide guidance of what constitutes a good span name and why (at least opentracing spec did), and that guidance is based on some basic understanding of how span name will be used (e.g. to treat spans with the same name as a class, aggregate across them, etc.) The same applies to all data elements collected through the API, there's some shared understanding why they are useful and worth collecting.

If we have no idea _how_ the tracer name can be used, it doesn't make sense in the API either. The main reason I've seen so far is to "turn instrumentation for different layers on and off", but as I showed in the previous comment, a naive implementation of that doesn't work. So aside from labelling a component (which can be done via semantic convention of the tags and does not require API change), I think we need to see a concrete example of how it can be used.

In contrast, in Java logging APIs, the meaning of the layer name is pretty clear, and even more, the names are expected to be in a specific package-like format that implies a hierarchy.

text/0000-named-tracers.md
yurishkuro: Again, this goes to the purpose of collecting this namespace. If it's purely decorative, then sure, any string will do.

But even as decorative, using a Resource rather than a string could make more sense, because resources have attributes, e.g. I might want to know not just that this span came from library X, but from version N of library X.

bogdandrutu:
I really like this RFC, and we need to make progress. @z1c0 can you please focus and address all the comments/concerns? If you need more details/help let me know. Thanks

text/0000-named-tracers.md
z1c0: I would leave this open to the implementation (unless it's somehow a requirement?).

text/0000-named-tracers.md
z1c0: It seems to me that this discussion is stuck somehow. What could we do to bring this RFC forward?

Let me try and sum up the main intention of this RFC:

* Establish a basic mechanism to create multiple instances of tracers that can be treated as separate logical entities (as opposed to a single global instance).
* By giving those entities a name, we later have the possibility to:
  * have specific configuration (on/off, verbosity, ...) for a tracer by identifying it through this name.
  * Matching produced spans to a certain tracer by e.g. setting this name as their "component".

If we want to have that same naming mechanism also for metrics, we could have a similar RFC there? I see how that makes sense.

z1c0:
> I really like this RFC, and we need to make progress. @z1c0 can you please focus and address all the comments/concerns? If you need more details/help let me know. Thanks

text/0000-named-tracers.md
z1c0: The tracer name _could_ be a property on the Span, which would be a good way to e.g. address the "tracer component issue" (https://github.com/open-telemetry/opentelemetry-specification/issues/10).
However, this is deliberately left out of this RFC's scope, since it  is neither a requirement nor a consequence of it.

text/0000-named-tracers.md
z1c0: I really prefer using a string as a key here, since this is a simple and already established (see the conversation on logging frameworks) pattern.

Also, there is a proposal https://github.com/open-telemetry/oteps/pull/26/files which intends to move Resources from API into SDK, which does not make them a good candidate for a key here.

However, adding the version information you are looking for, is currently addressed in another proposal https://github.com/open-telemetry/opentelemetry-specification/issues/235 (which should probably go into oteps as well).

text/0000-named-tracers.md
z1c0: Updated PR.

text/0000-named-tracers.md
z1c0: Not sure about SDKs, but this is probably beyond the scope of this RFC. Different tracer implementations however, are entirely possible in this model.

yurishkuro:
I am going to reply here, not inline, since the comments become hidden.

Re https://github.com/open-telemetry/oteps/pull/16/files#r315490941

@z1c0 Sorry, this does not answer my questions. The way I expect an RFC document to be formulated is:
  1. this is the problem that we want to solve
  2. this is why the current Spec does not allow us to solve that problem, or makes the solution not very good
  3. this is the proposal for change
  4. this is _how the proposal solves the problem_ (unless it's obvious from the change itself, which in not the case here)

This PR has a bit of 1 (controlling verbosity and double-instrumentation), then jumps into 3, and completely ignores 4. As a result, while the proposal appears reasonable, we don't know if it actually solves the problem. "It works for logging" is not a strong argument that it would work for traces.

I would like to see a specific example how this proposal solves the problem. I would also like to see better reasoning why "name" is sufficient for identifying the layer/component, where there may be other useful attributes, like a version (since there may be multiple versions of the library in use).

@tedsuo I think we should move away from merging RFCs that have not reached consensus. I opened a ticket #32.

z1c0:
@yurishkuro maybe we are just looking at this proposal from different angles and are interpreting its main purpose differently. I'll try and explain the main intention and purpose along the outline you suggested and maybe we can reach consensus that way:

1.) The problem this RFC solves is first and foremost the fact that currently the spec only intends to have a single, global tracer. So, _this_ is what is actually addressed here. Not being able to control verbosity on a finer grain level or having double instrumentation, are problem details that directly result from the overarching fact that there is just one, single tracer. Let's call it "the global tracer problem".

2.) Since this behavior (a single, global tracer) is currently part of the spec, the above mentioned scenarios of controlling tracing on a more fine-grained level are not possible.

3.) The proposal for change is to replace the global instance with a factory mechanism, that allows for having multiple tracer instances for a given key (name).

4.) The "how" is the factory pattern.

Regarding the "name" vs. "name + version + ..." _identifier_ for tracers: This is definitely a valid point for discussion, but then again it's almost an implementation detail. Once we agree on having _some_ kind of lookup key for a tracer (whether this is string or a more complex object) to solve the "global tracer problem", the main intention of this proposal is served.

yurishkuro:
> 1.) The problem this RFC solves is first and foremost the fact that currently the spec only intends to have a single, global tracer. So, this is what is actually addressed here. Not being able to control verbosity on a finer grain level or having double instrumentation, are problem details that directly result from the overarching fact that there is just one, single tracer. Let's call it "the global tracer problem".

I accept that controlling verbosity and double instrumentation are two problems worth solving. From that it does not follow that they can be generalized into the "global tracer problem" - they can potentially be solved by different means that do not require named tracers. But that's a minor point.

You are making a claim that named tracers are a solution to the two (separate) problems. But you are not providing any evidence of that. I am looking at this from two positions, as an end user of the OpenTelemetry API/SDK, and as an implementor of an SDK, and I am not seeing how I can use named tracers to solve the two problems. If you have a solution in mind, please describe the procedure or algorithm that demonstrates the solution.

discostu105:
@yurishkuro 

The angle I look at this is from a direct implementer of the OTel API, potentially bundled with an auto-instrumentation that already traces certain components itself (not via OTel).

The TracerFactory would act as a registry that "discovers" named tracers that have registered them self. Every delivered Tracer would be attached to a configuration that would control if the tracer is "active" or "inactive". The configuration could be pre-populated: certain tracers are blacklisted, because they are known to capture the same data as an auto-injected sensor.

The list of "discovered" tracers could also be reported to a backend, where that list is presented to a user. A user could enable/disable each tracer for troubleshooting bugs or overhead problems. A changed configuration would be propagated to the OTel implementation and applied to the Tracer objects.

Of course an implementation can also chose to ignore tracer-name completely and TracerFactory would return the very same Tracer object for all requests. Without any such backend/config system in place, this is probably a valid approach (could be done so in current SDK impl).

yurishkuro:
@discostu105 what is the difference in prescribed behavior between active and inactive tracer? If you have an inactive tracer and someone calls startSpan on it, what is the expected behavior?

discostu105:
@yurishkuro it would act like a no-op tracer. So, no span is actually created and sent to any backend/exporter.

yurishkuro:
what if that instrumentation is in the middle of the call graph (as most would be)?

discostu105:
Well, as a result it should just be missing in the call-graph. That would mean child spans would have to have a different parent, so an "inactive" span must not activate itself on the current scope, but rather it's parent.

z1c0:
@bogdandrutu @yurishkuro @discostu105 I reworked that RFC and tried to focus it on the actual purpose which is **"associating a name (aka component) with a tracer and its produced spans"**.

I removed all references to topics like configuration or enabling/disabling of tracers since those are not a concern of this RFC, spec and the API surface. Please let me know what you think of it and if that covers your concerns.
Thank you.

text/0000-named-tracers.md
```diff
@@ -36,13 +31,11 @@ In a way, the `TracerFactory` replaces the global `Tracer` singleton object as a
 
 ## Internal details
 
-By providing a TracerFactory and *Named Tracers*, a vendor or OpenTelemetry implementation gains more flexibility in providing tracers and can decide whether or not to produce spans for a specific traced component and even change this behavior at runtime.
-
-In the simplest case, an OpenTelemetry implementation can return a single instance for a requested tracer regardless of the name specified. This could be the case for implementations that do not want/need to enable or disable a tracer.
+By providing a TracerFactory and *Named Tracers*, a vendor or OpenTelemetry implementation gains more flexibility in providing tracers and which attributes they set in the resulting spans that are produced.
 
-Alternatively, an implementation can provide different tracer instances per specified tracer name, thus being able to associate this tracer with the component being traced. This allows for the possibility to enable / disable a tracer based on a custom configuration.
+In the simplest case, an OpenTelemetry implementation can return a single instance for a requested tracer, regardless of the name specified as long as the name given for the tracer
```
Oberon00: This sentence seems incomplete.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,74 @@
...
+  Span span = traceComponent.spanBuilder("someMethod").startSpan();
+```
+
+Overall, this would not change a lot compared to the `TracerFactory` since the levels of indirection until producing an actual span are the same.
+
+Instead of setting the `component` property based on the given tracer names, those names could also be used as *prefixes* for produced span names (e.g. `<TracerName-SpanName>`). However, with regard to data quality and semantic conventions, a dedicated `component` set on spans is probably preferred.
+
+## Future possibilities
+
+By adapting this proposal, current implementations that do not honor the specified tracer name and provide a single global tracer, would not require much change. However they could change that behavior in future versions and provide more specific tracer implementations then. On the other side, if the mechanism of *Named Tracers* is not a part of the initial specification, such scenarios will be prevented and hard to retrofit in future version, should they be deemed necessary then.
```
Oberon00:
```suggestion
By adapting this proposal, current implementations that do not make use of the specified tracer name and provide a single global tracer, would not require much change. However they could change that behavior in future versions and provide more specific tracer implementations then. On the other side, if the mechanism of *Named Tracers* is not a part of the initial specification, such scenarios will be prevented and hard to retrofit in future version, should they be deemed necessary then. 
```

Maybe add a few line breaks after sentences so that one can point to parts of the paragraph more easily.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,74 @@
...
+
+By adapting this proposal, current implementations that do not honor the specified tracer name and provide a single global tracer, would not require much change. However they could change that behavior in future versions and provide more specific tracer implementations then. On the other side, if the mechanism of *Named Tracers* is not a part of the initial specification, such scenarios will be prevented and hard to retrofit in future version, should they be deemed necessary then. 
+
+## Examples (of Tracer names)
+
+Since tracer names describe the libraries which use the tracers, those names should be defined in a way that makes them as unique as possible. The name of the tracer should represent the identity of the library, class or package that provides the instrumentation. 
+
+Examples (based on existing contribution libraries from OpenTracing and OpenCensus):
+
+* _io.opentracing.contrib.spring.rabbitmq_
```
Oberon00:
```suggestion
* `io.opentracing.contrib.spring.rabbitmq`
```

And so on

text/0000-named-tracers.md
```diff
@@ -0,0 +1,74 @@
...
+_Creating tracers using a factory mechanism and naming those tracers in accordance with the library that provides the instrumentation for a traced component._
+
+## Suggested reading
+
+* [Proposal: Tracer Components](https://github.com/open-telemetry/opentelemetry-specification/issues/10)
+* [Global Instance discussions](https://github.com/open-telemetry/opentelemetry-specification/labels/global%20instance)
+
+## Motivation
+
+Instead of having a global tracer singleton, every instrumentation library needs to create its own tracer, which also has a proper name that identifies the library / logical component that uses this tracer. If there are two different instrumentation libraries for the same technology (e.g. MongoDb), these instrumentation libraries should have distinct names. In this sense, this proposal suggests the introduction of a "semantic namespace" concept for tracers and the resulting spans.
```
Oberon00: > [...] every instrumentation library needs to create its own tracer [...]

Having an additional need is hardly a motivation. This seems to belong to the explanation.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,74 @@
...
+
+The purpose of _Named Tracers_ is the ability to allow for associating an identifier (e.g. _"io.opentelemetry.contrib.mongodb"_) with a tracer. This further allows for logical and semantic coupling of this tracer and a certain technology or library (like _MongoDB_ in this example). Having named tracers enables support for having a dedicated property (component) depending on this name on a tracer that is different for different libraries.
```
Oberon00:
```suggestion
The purpose of _Named Tracers_ is the ability to associate an identifier (e.g. `io.opentelemetry.contrib.mongodb`) with a tracer. This further allows for logical and semantic coupling of this tracer and a certain technology or library (like _MongoDB_ in this example). Having named tracers enables support for having a dedicated property (component) depending on this name on a tracer that is different for different libraries.
```

Oberon00: I would leave out the sentence with the dedicated property. That is basically a restatement of "associating an indentifier". Also what is the difference between "logical" and "semantic" coupling? What does coupling even mean here? Is that another restatement of being able to associate an identifier?

I would rather like to see examples of what *useful* things (e.g. from the perspective of someone that deploys a traced system) can be achieved with that identifier. For example, ability of API-implementations to provide server-side disabling of certain tracers by name, adding an additional property to a Span to ease interpretation of attributes, etc.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,74 @@
...
+
+This proposal originates from an `opentelemetry-specification` proposal on [components](https://github.com/open-telemetry/opentelemetry-specification/issues/10) since having a concept of named tracers would automatically enable determining this semantic `component` property.
```
Oberon00: I wouldn't use the term "component property". The component *attribute* has a different meaning (it identifies the traced component while this new component "property" should identify the tracing component).

text/0000-named-tracers.md
```diff
@@ -0,0 +1,74 @@
...
+
+By providing a TracerFactory and *Named Tracers*, a vendor or OpenTelemetry implementation gains more flexibility in providing tracers and which attributes they set in the resulting spans that are produced.
+
+In the simplest case, an OpenTelemetry implementation can return a single instance for a requested tracer, regardless of the name specified as long as the name given for the tracer 
+
+Alternatively, an implementation can provide different tracer instances per specified tracer name, thus being able to associate this tracer with the `component` being traced.
+
+## Prior art and alternatives
+
+Alternatively, instead of having a `TracerFactory`, existing (global) tracers could return additional indirection objects (called e.g. `TraceComponent`), which would be able to produce spans for specifically named traced components.
```
Oberon00: What about making this name a mandatory parameter for span creation? (And later metric creation and so on.)

text/0000-named-tracers.md
yurishkuro: The first sentence is not a motivation, it's a solution.

The second sentence it not a motivation either, it's a requirement assuming someone is already bought into a motivation.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,74 @@
...
+
+This proposal originates from an `opentelemetry-specification` proposal on [components](https://github.com/open-telemetry/opentelemetry-specification/issues/10) since having a concept of named tracers would automatically enable determining this semantic `component` property.
+
+
+## Explanation
+
+From a user perspective, working with *Named Tracers* and *TracerFactories* is very similar to how e.g. the [Java logging API](https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html#getLogger(java.lang.String)) and logging frameworks like [log4j](https://www.slf4j.org/apidocs/org/slf4j/LoggerFactory.html) work. In analogy to requesting Logger objects through LoggerFactories, a tracing library would create specific 'Tracer' objects through a 'TracerFactory'.
+
+```java
+Tracer tracer = OpenTelemetry.getTracerFactory().getTracer("io.opentelemetry.contrib.mongodb");
```
ajacques: What do we expect users to provider as the parameter here? Is it a fully qualified class name (that includes the class), a package name, or does it represent just a name?

I like the Slf4j-style where you pass a `Class<?>` which makes it type-safe and very clear to a user what they should provide. String parameters are not always clear.
```java
getTracer(MyClass.class)
```

Oberon00: This is an OTEP not an Java proposal. Using `.class` is very Java specific and won't work in most other languages. But one could indeed provide a `Class<?>` overload when implementing this OTEP in Java.

z1c0:
This proposal got a big overhaul and we worked in the feedback we got so far.
Thoughts? @bogdandrutu  @yurishkuro @Oberon00 @discostu105

text/0000-named-tracers.md
```diff
@@ -0,0 +1,89 @@
...
+Tracer tracer = OpenTelemetry.getTracerFactory().getTracer(libraryResource);
+```
+
+In a way, the `TracerFactory` replaces the global `Tracer` singleton object as a ubiquitous point to request a Tracer instance.
+
+## Internal details
+
+By providing a `TracerFactory` and *Named Tracers*, a vendor or OpenTelemetry implementation gains more flexibility in providing Tracers and which attributes they set in the resulting spans that are produced.
+
+The span API is extended with a `getLibraryResource` function that returns the resource associated with the Tracer that created the span.
```
yurishkuro: There is no need for such method in the span API, only in the SDK.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,89 @@
...
+
+* [Proposal: Tracer Components](https://github.com/open-telemetry/opentelemetry-specification/issues/10)
+* [Global Instance discussions](https://github.com/open-telemetry/opentelemetry-specification/labels/global%20instance)
+* [Proposal: Add a version resource](https://github.com/open-telemetry/oteps/pull/38)
+
+## Motivation
+
+The mechanism of "Named Tracers" proposed here is motivated by following scenarios:
+
+* As a consumer of OpenTelemetry instrumentation libraries, there is currently no possibility of influencing the amount of the data produced by such libraries. Instrumentation libraries can easily "spam" backend systems, deliver bogus data or - in the worst case - crash or slow down applications. These problems might even occur suddenly in production environments caused by external factors such as increasing load or unexpected input data.
```
yurishkuro:
```suggestion
* For a consumer of OpenTelemetry instrumentation libraries, there is currently no mechanism of influencing the amount of the data produced by such libraries. Instrumentation libraries can easily "spam" backend systems, deliver bogus data or - in the worst case - crash or slow down applications. These problems might even occur suddenly in production environments caused by external factors such as increasing load or unexpected input data.
```

text/0000-named-tracers.md
```diff
@@ -0,0 +1,89 @@
...
+
+* If a library hasn't implemented semantic conventions correctly or those conventions change over time, it's currently hard to interpret and sanitize these data selectively. The produced Spans cannot associated with those instrumentation libraries later.
```
yurishkuro: this part is confusing - how is semver related to sanitizing data?

text/0000-named-tracers.md
```diff
@@ -0,0 +1,89 @@
...
+
+Based on such an identifier, a Sampler could be implemented that discards Spans from certain libraries. Also, by providing a custom Exporter, Span data could be sanitized before it gets processed in a back-end system. However, this is beyond the scope of this proposal, which only provides the fundamental mechanisms.
+
+## Explanation
+
+From a user perspective, working with *Named Tracers* and *TracerFactories* is conceptually similar to how e.g. the [Java logging API](https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html#getLogger(java.lang.String)) and logging frameworks like [log4j](https://www.slf4j.org/apidocs/org/slf4j/LoggerFactory.html) work. In analogy to requesting Logger objects through LoggerFactories, a tracing library would create specific 'Tracer' objects through a 'TracerFactory'.
+
+Instead of using simple strings as an argument for creating new Tracers a Resource identifying an instrumentation library is used. Such resources must have a _version_ and _name_ labels (there could be semantic convention definitions for those).
```
yurishkuro:
```suggestion
Instead of using a plain string as an argument for creating new Tracers, a Resource identifying an instrumentation library MAY be used. Such resource MUST have a `name` label, and SHOULD have a `version` label (there could be semantic convention definitions for those).
```

text/0000-named-tracers.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Motivation
+
+The mechanism of "Named Tracers" proposed here is motivated by following scenarios:
+
+* As a consumer of OpenTelemetry instrumentation libraries, there is currently no possibility of influencing the amount of the data produced by such libraries. Instrumentation libraries can easily "spam" backend systems, deliver bogus data or - in the worst case - crash or slow down applications. These problems might even occur suddenly in production environments caused by external factors such as increasing load or unexpected input data.
+
+* If a library hasn't implemented semantic conventions correctly or those conventions change over time, it's currently hard to interpret and sanitize these data selectively. The produced Spans cannot associated with those instrumentation libraries later.
+
+This proposal attempts to solve the stated problems by introducing the concept of _Named Tracers_. These associate a given identifier (e.g. _"io.opentelemetry.contrib.mongodb"_, _"1.0.0"_) with a Tracer and its produced Spans.
```
yurishkuro:
```suggestion
This proposal attempts to solve the stated problems by introducing the concepts of:
 * _Named Tracers_ identified via a Resource (e.g. _"io.opentelemetry.contrib.mongodb"_, _"1.0.0"_) which is associated with the Tracer and the Spans it produces; and
 * _Tracer Factory_ as the only means of creating a tracer.
```

text/0000-named-tracers.md
```diff
@@ -0,0 +1,89 @@
...
+// Create resource for given instrumentation library information (name + version)
+Map<String, String> libraryLabels = new HashMap<>();
+libraryLabels.put("name", "io.opentelemetry.contrib.mongodb");
+libraryLabels.put("version", "1.0.0");
+Resource libraryResource = Resource.create(libraryLabels);
+// Create tracer for given instrumentation library.
+Tracer tracer = OpenTelemetry.getTracerFactory().getTracer(libraryResource);
+```
+
+In a way, the `TracerFactory` replaces the global `Tracer` singleton object as a ubiquitous point to request a Tracer instance.
```
yurishkuro: not "in a way" though, it does replace it

text/0000-named-tracers.md
```diff
@@ -0,0 +1,89 @@
...
+
+Based on such an identifier, a Sampler could be implemented that discards Spans from certain libraries. Also, by providing a custom Exporter, Span data could be sanitized before it gets processed in a back-end system. However, this is beyond the scope of this proposal, which only provides the fundamental mechanisms.
+
+## Explanation
+
+From a user perspective, working with *Named Tracers* and *TracerFactories* is conceptually similar to how e.g. the [Java logging API](https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html#getLogger(java.lang.String)) and logging frameworks like [log4j](https://www.slf4j.org/apidocs/org/slf4j/LoggerFactory.html) work. In analogy to requesting Logger objects through LoggerFactories, a tracing library would create specific 'Tracer' objects through a 'TracerFactory'.
+
+Instead of using simple strings as an argument for creating new Tracers a Resource identifying an instrumentation library is used. Such resources must have a _version_ and _name_ labels (there could be semantic convention definitions for those).
+
```
yurishkuro: there should be some provision for what constitutes a good value for the `version` label.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,89 @@
...
+
+```java
+// Create resource for given instrumentation library information (name + version)
+Map<String, String> libraryLabels = new HashMap<>();
+libraryLabels.put("name", "io.opentelemetry.contrib.mongodb");
+libraryLabels.put("version", "1.0.0");
+Resource libraryResource = Resource.create(libraryLabels);
+// Create tracer for given instrumentation library.
+Tracer tracer = OpenTelemetry.getTracerFactory().getTracer(libraryResource);
```
yurishkuro: how should `getTracer` behave if the resource does not contain the `name` label?

text/0000-named-tracers.md
z1c0: This is refers to the "semantic conventions" part of the specification. I'll a link to it (https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md) here.

text/0000-named-tracers.md
z1c0: I'll put in a link to "Proposal: Add a version resource" https://github.com/open-telemetry/oteps/pull/38).

text/0000-named-tracers.md
z1c0: Following the suggestions in "error handling proposal" https://github.com/open-telemetry/opentelemetry-specification/pull/153 a _smart default_ could be to:
* return a default tracer implementation (that has an empty resource definition) 
* warn the user (e.g. in a logging callback meachnism, if we decide to have such a mechanism).

Would you agree with that?

yurishkuro: +1 to return default "unnamed" tracer. We just need to be specific in the docs that it is the expected behavior (as opposed to throwing an exception).

I am not sure warning the user is particularly useful because the problem may be in the instrumentation library they are forced to use and can't do much about.

text/0000-named-tracers.md
bogdandrutu: I like @Oberon00 suggestion here, we should not hardcode `_` as prefix/suffix, this is how I read it.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,91 @@
+# Named Tracers
```
bogdandrutu: I think this logic applies to Meter as well. Consider to make this change to apply to all the "signal producers" factories: metrics, traces, in the future logs.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,91 @@
...
+
+Based on such an identifier, a Sampler could be implemented that discards Spans from certain libraries. Also, by providing a custom Exporter, Span data could be sanitized before it gets processed in a back-end system. However, this is beyond the scope of this proposal, which only provides the fundamental mechanisms.
+
+## Explanation
+
+From a user perspective, working with *Named Tracers* and *TracerFactories* is conceptually similar to how e.g. the [Java logging API](https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html#getLogger(java.lang.String)) and logging frameworks like [log4j](https://www.slf4j.org/apidocs/org/slf4j/LoggerFactory.html) work. In analogy to requesting Logger objects through LoggerFactories, a tracing library would create specific 'Tracer' objects through a 'TracerFactory'.
+
+Instead of using plain strings as an argument for creating new Tracers a Resource identifying an instrumentation library is used. Such resources must have a _version_ and _name_ labels (there could be semantic convention definitions for those labels). Version values will follow the conventions proposed [here](https://github.com/open-telemetry/oteps/pull/38).
```
bogdandrutu: I am not sure I understand why a `Resource` is better here. My vote would be to use two strings, for the following reasons:
1. Simpler API to get a Tracer/Meter.
1. Limits what can be set and makes required fields more obviously.
1. Simpler API surface for OpenTelemetry - Resource does not need to be something that an instrumentation/library developer needs to learn/understand.
1. Component name can be used by some implementation as a prefix for Span name or Metric name. These implementations may not support a notion of Resource. This is specifically important for systems that do not support sub-target resources (components) and only target resources.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,91 @@
...
+
+
+```java
+// Create resource for given instrumentation library information (name + version)
+Map<String, String> libraryLabels = new HashMap<>();
+libraryLabels.put("name", "io.opentelemetry.contrib.mongodb");
+libraryLabels.put("version", "1.0.0");
```
bogdandrutu: Why `version` for component? If you have a version for the system as an application owner you can determine all the components versions (I hope everyone is using reproducible builds).

Based on the rationals that I see in this OTEP, I do not understand the need for a `version` (but I do clearly understand the need for a `name`).

yurishkuro: Some examples:
* the details of how to collect less data from certain named traces is not really described here, but it's easy to imagine a scenario where some specific range of versions of a given middleware needs to be suppressed, while other versions allowed (perhaps because of a bug in the instrumentation). Suppression may need to happen online, which makes resolving version from the build number again very complicated.
* go modules allows multiple versions of the same middleware in a single build, so you cannot resolve version just from the build

text/0000-named-tracers.md
z1c0: The _ is not a prefix but merely markdown formatting.

text/0000-named-tracers.md
z1c0: I also see the usefulness of having this version information for the reasons above. We could however consider making this part of the "name" (Resource, ...) explicitly optional?

text/0000-named-tracers.md
bogdandrutu: As @Oberon00 suggested use ` for the markdown formatting

text/0000-named-tracers.md
z1c0: I would very much prefer having dedicated RFCs for metrics and logs. As we have seen it's easy enough to get hung up on details and move forward painfully slow even with the narrow focus we have here. Widening this focus and trying to covers metrics as well with this RFC (or even logs) will not add any benefit. This RFC targets a specific part of the specification, namely tracers. Other specification topics deserve dedicated and thoroughly discussed RFCs as well.

text/0000-named-tracers.md
z1c0: Honestly, I don't have a strong opinion on this topic. Two strings would work equally well in my opinion.
To me this should come down to the decision of whether `Resource` should be part of the API or not:
* If Tracer creation is the only occurrence of this type in the API I vote in favor of using plain strings instead.
* If however Resources are a common API type, I see no reason why those should not also be used in this context.

text/0000-named-tracers.md
z1c0: Removed formatting here

text/0000-named-tracers.md
bogdandrutu: I would vote to not use Resource here to not bring back the Resource just for this, and use a "required" field name and an "optional" version (if version is useful).

text/0000-named-tracers.md
bogdandrutu: Please document the reasons if you want to keep the version.

@yurishkuro based on my understanding you can have multiple versions only if they are different major versions which implies different package names (which probably implies different names).

@yurishkuro @z1c0 I see the advantage from the first bullet point that @yurishkuro posted. So I am happy to keep the version, but make sure it is documented in the requirements so we understand why we need version.

text/0000-named-tracers.md
z1c0: The main proposal uses plain strings again and resources as an alternative are mentioned in the "Prior art and alternatives" section (also referring to https://github.com/open-telemetry/opentelemetry-specification/pull/254)

text/0000-named-tracers.md
```diff
@@ -0,0 +1,104 @@
...
+
+* [Proposal: Tracer Components](https://github.com/open-telemetry/opentelemetry-specification/issues/10)
+* [Global Instance discussions](https://github.com/open-telemetry/opentelemetry-specification/labels/global%20instance)
+* [Proposal: Add a version resource](https://github.com/open-telemetry/oteps/pull/38)
+
+## Motivation
+
+The mechanism of "Named Tracers" proposed here is motivated by following scenarios:
+
+* For a consumer of OpenTelemetry instrumentation libraries, there is currently no possibility of influencing the amount of the data produced by such libraries. Instrumentation libraries can easily "spam" backend systems, deliver bogus data or - in the worst case - crash or slow down applications. These problems might even occur suddenly in production environments caused by external factors such as increasing load or unexpected input data.
```
tigrannajaryan: I am confused a bit by the usage of " instrumentation library". Does it refer to a 3rd party library that is instrumented using OpenTelemetry API?
If so I would change the wording to be "instrumented library". "Instrumentation library" IMO is a better label for the OpenTelemetry library itself and is confusing in this context.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,104 @@
...
+Instead of using plain strings as an argument for creating new Tracers a Resource identifying an instrumentation library is used. Such resources must have a _version_ and _name_ labels (there could be semantic convention definitions for those labels). Version values will follow the conventions proposed [here](https://github.com/open-telemetry/oteps/pull/38).
+
+```java
+// Create a tracer for given instrumentation library in a specific version.
+Tracer tracer = OpenTelemetry.getTracerFactory().getTracer("io.opentelemetry.contrib.mongodb", "semver:1.0.0");
+```
+
+This `TracerFactory` replaces the global `Tracer` singleton object as a ubiquitous point to request a Tracer instance.
+
+If no tracer name (null or empty string) is specified, following the suggestions in ["error handling proposal"](open-telemetry/opentelemetry-specification#153), a "smart default" will be applied and a default tracer implementation is returned.
```
tigrannajaryan: Link to "error handling proposal" is broken.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,104 @@
...
+* [Global Instance discussions](https://github.com/open-telemetry/opentelemetry-specification/labels/global%20instance)
+* [Proposal: Add a version resource](https://github.com/open-telemetry/oteps/pull/38)
+
+## Motivation
+
+The mechanism of "Named Tracers" proposed here is motivated by following scenarios:
+
+* For a consumer of OpenTelemetry instrumentation libraries, there is currently no possibility of influencing the amount of the data produced by such libraries. Instrumentation libraries can easily "spam" backend systems, deliver bogus data or - in the worst case - crash or slow down applications. These problems might even occur suddenly in production environments caused by external factors such as increasing load or unexpected input data.
+
+* If a library hasn't implemented [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md) correctly or those conventions change over time, it's currently hard to interpret and sanitize these data selectively. The produced Spans cannot associated with those instrumentation libraries later.
```
tigrannajaryan:
```suggestion
* If a library hasn't implemented [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md) correctly or those conventions change over time, it's currently hard to interpret and sanitize these data selectively. The produced Spans cannot be associated with those instrumentation libraries later.
```

text/0000-named-tracers.md
```diff
@@ -0,0 +1,104 @@
...
+This `TracerFactory` replaces the global `Tracer` singleton object as a ubiquitous point to request a Tracer instance.
+
+If no tracer name (null or empty string) is specified, following the suggestions in ["error handling proposal"](open-telemetry/opentelemetry-specification#153), a "smart default" will be applied and a default tracer implementation is returned.
+
+
+## Internal details
+
+By providing a `TracerFactory` and *Named Tracers*, a vendor or OpenTelemetry implementation gains more flexibility in providing Tracers and which attributes they set in the resulting spans that are produced.
+
+The SpanData class is extended with a `getLibraryResource` function that returns the resource associated with the Tracer that created the span.
```
tigrannajaryan: Is `LibraryResouce` a part of SpanData that is supposed to be serialized and propagated when the data leaves the local process and thus it is available for intermediary nodes (such as Agents and Collectors) and final destination/backend to inspect or it is ephemeral and only available in the local process to the exporters and samplers?

If the intent is that this is part of serialized span data then I would expect this proposal to define where exactly in the span data this fits. Is it going to be encoded in Span attributes? Is it going to be a separate `libraryresource` field? If it is a new field how this will be encoded / translated for existing trace protocols that do not have the equivalent concept?

text/0000-named-tracers.md
Oberon00: We explicitly don't mean the instrumented library here, but the library that does the instrumentation (I've seen them called integration), whether it is a 3rd party or developed in an OpenTracing repository, or even some auto-injected code from an Agent (see Zero Source Code Modification Instrumentation OTEP). See also the examples for identifiers at the end.

tigrannajaryan: OK, just to double check an "instrumentation library"  in this doc refers to a library that adds/injects telemetry calls to another library or otherwise helps that library to make telemetry calls.

Is that distinction important for the purposes of this document? I would imagine "Named Tracers" could equally well be directly used by the instrumented library. I think the fact that the instrumented library and the library that instruments it are separate is an orthogonal concept and can be left out from this discussion.

I may be missing something so please correct me.

Oberon00: If a library (or even application) has instrumentation built-in, it is of course both the instrumenting and instrumented library and can pass its own name here. In all other cases (and to distinguish them from that case), the distinction between instrumenting and instrumented library is very important. For example, if a HTTP library com.example.http is instrumented by either io.opentelemetry.contrib.examplehttp or com.example.company.examplehttpintegration, then it is important that the tracer is *not* named com.example.http but either io.opentelemetry.contrib.examplehttp or com.example.company.examplehttpintegration.

text/0000-named-tracers.md
z1c0: Since this `LibraryResource` should be handled the same way as all kinds of ressources, questions about serialization and encoding really need to be answered in that part of the specification https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-resources.md.
The specification on resources probably needs attention anyway since it mentions the API a couple of times and there's a proposal about removing those from the API (https://github.com/open-telemetry/opentelemetry-specification/pull/254) currently. Also serialization and encoding are not addressed at all at the moment.

text/0000-named-tracers.md
z1c0: Fixed

text/0000-named-tracers.md
z1c0: Fixed

z1c0:
@bogdandrutu I'm not quite sure how we left off after yesterday's discussion in the spec meeting. Are you OK with the approach that we propose a dedicated RFC for metrics (and also logs at some later point in time) to have a better separation of concerns? 
Metrics and logs get a mention in "Future possibilities" (i.e. "very near future" for metrics).

text/0000-named-tracers.md
bogdandrutu: I think the question of serialization can be addressed later.

bogdandrutu:
@z1c0 I would prefer to not postpone the metrics:
1. This is a major API change that we need for Meter as well before alpha.
2. All reasons that we need this for Trace apply to Metrics as well.
3. We want to be consistent between Meter/Tracer in general if possible.

I suggest that we just re-write a bit the first few paragraphs to generalize the motivation for metrics as well.

Everything else stays the same:
```java
// Create a tracer/meter for given instrumentation library in a specific version.
Tracer tracer = OpenTelemetry.getTracerFactory().getTracer("io.opentelemetry.contrib.mongodb", "semver:1.0.0");
Meter meter = OpenTelemetry.getMeterFactory().getMeter("io.opentelemetry.contrib.mongodb", "semver:1.0.0");
```

Also do we need a way to access unnamed Tracer/Meter?

text/0000-named-tracers.md
tigrannajaryan: OK, makes sense, it was not apparent from the document. Perhaps worth adding a few words explaining what the "name" is intended to be.

z1c0: I updated the "Explanation" section and tried to clarify the "name" part.

z1c0:
@bogdandrutu metrics are now included in this proposal.
Unnamed Tracers/Meters are mentioned in the last "Explanation" paragraph.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,108 @@
...
+
+```java
+// Create a tracer/meter for a given instrumentation library in a specific version.
+Tracer tracer = OpenTelemetry.getTracerFactory().getTracer("io.opentelemetry.contrib.mongodb", "semver:1.0.0");
+Meter meter = OpenTelemetry.getMeterFactory().getMeter("io.opentelemetry.contrib.mongodb", "semver:1.0.0");
+```
+
+These factories (`TracerFactory` and `MeterFactory`) replace the global `Tracer` / `Meter` singleton objects as ubiquitous points to request Tracer and Meter instances.
+
+ The *name* used to create a Tracer or Meter explicitly refers to the *instrumentation* libraries (also referred to as *integrations*) and not the instrumented libraries. These instrumentation libraries could be libraries developed in an OpenTelemetry repository, a 3rd party implementation or even auto-injected code (see [Open Telemetry Without Manual Instrumentation OTEP](https://github.com/open-telemetry/oteps/blob/master/text/0001-telemetry-without-manual-instrumentation.md)). See also the examples for identifiers at the end.
```
Oberon00:
```suggestion
 The *name* used to create a Tracer or Meter must identify the *instrumentation* libraries (also referred to as *integrations*) and not the instrumented libraries. These instrumentation libraries could be libraries developed in an OpenTelemetry repository, a 3rd party implementation or even auto-injected code (see [Open Telemetry Without Manual Instrumentation OTEP](https://github.com/open-telemetry/oteps/blob/master/text/0001-telemetry-without-manual-instrumentation.md)). See also the examples for identifiers at the end.
```

BTW, a few hard line breaks here and there (e.g. after sentence) would make reviewing a bit easier, not suggesting to change that but for the future it would be nice.

Oberon00:
```suggestion
 The *name* used to create a Tracer or Meter explicitly refers to the *instrumentation* libraries (also referred to as *integrations*) and not the instrumented libraries. These instrumentation libraries could be libraries developed in an OpenTelemetry repository, a third party implementation or even auto-injected code (see [Open Telemetry Without Manual Instrumentation OTEP](https://github.com/open-telemetry/oteps/blob/master/text/0001-telemetry-without-manual-instrumentation.md)). See also the examples for identifiers at the end.
```

text/0000-named-tracers.md
```diff
@@ -0,0 +1,108 @@
...
+If a library (or application) has instrumentation built-in, it is of course both the instrumenting and instrumented library and can pass its own name here. In all other cases (and to distinguish them from that case), the distinction between instrumenting and instrumented library is very important. For example, if an HTTP library `com.example.http` is instrumented by either `io.opentelemetry.contrib.examplehttp` or `com.example.company.examplehttpintegration`, then it is important that the Tracer is not named `com.example.http` but after the actual instrumentation library.
```
Oberon00:
```suggestion
If a library (or application) has instrumentation built-in, it is both the instrumenting and instrumented library and should pass its own name here. In all other cases (and to distinguish them from that case), the distinction between instrumenting and instrumented library is very important. For example, if an HTTP library `com.example.http` is instrumented by either `io.opentelemetry.contrib.examplehttp` or `com.example.company.examplehttpintegration`, then it is important that the Tracer is not named `com.example.http` but after the actual instrumentation library.
```

text/0000-named-tracers.md
```diff
@@ -0,0 +1,108 @@
...
+If no name (null or empty string) is specified, following the suggestions in ["error handling proposal"](https://github.com/open-telemetry/opentelemetry-specification/pull/153), a "smart default" will be applied and a default Tracer / Meter implementation is returned.
+
+
+## Internal details
+
+By providing a `TracerFactory` / `MeterFactory` and *Named Tracers / Meters*, a vendor or OpenTelemetry implementation gains more flexibility in providing Tracers and Meters and which attributes they set in the resulting Spans and Metrics that are produced.
+
+On an SDK level, the SpanData class and its Metrics counterpart are extended with a `getLibraryResource` function that returns the resource associated with the Tracer / Meter that created it.
+
+If there are two different instrumentation libraries for the same technology (e.g. MongoDb), these instrumentation libraries should have distinct names.
```
Oberon00: This paragraph is now already covered in the explanation, I think.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Prior art and alternatives
+
+This proposal originates from an `opentelemetry-specification` proposal on [components](https://github.com/open-telemetry/opentelemetry-specification/issues/10) since having a concept of named Tracers would automatically enable determining this semantic `component` property.
+
+Alternatively, instead of having a `TracerFactory`, existing (global) Tracers could return additional indirection objects (called e.g. `TraceComponent`), which would be able to produce spans for specifically named traced components.
+
+```java
+TraceComponent traceComponent = OpenTelemetry.Tracing.getTracer().componentBuilder(libraryResource);
```
Oberon00: This example should be updated to the new two-argument form instead of libraryResource.

text/0000-named-tracers.md
```diff
@@ -0,0 +1,108 @@
...
+
+Those given alternatives could be applied to Meters and Metrics in the same way.
+
+## Future possibilities
+
+Based on the Resource information identifying a Tracer or Meter these could be configured (enabled / disabled) programmatically or via external configuration sources (e.g. environment).
+
+Based on this proposal, future "signal producers" (i.e. logs) can use the same or a similar creation approach.
+
+## Examples (of Tracer and Meter names)
```
Oberon00: Maybe this section should be moved to a subsection of Explanation or Internal Details, so that we don't have to point to "see examples at the end" there.

text/0000-named-tracers.md
tigrannajaryan: LGTM now, thanks.

z1c0: Could you approve the proposal, please?

Oberon00:
Seems like this is ready to merge then?

bogdandrutu:
@z1c0 please in parallel start two PRs:
1. To mark this approved.
1. A PR to change specs accordingly.

PR #17, opened 2019-08-12
Add status to RFC template

iredelmeier:


0000-template.md
```diff
@@ -1,5 +1,7 @@
 # Replace this with your awesome RFC title
 
+**Status:** `proposed`
```
bogdandrutu: Can we enumerate all the possible values? Maybe: `draft`, `in-progress`, `approved`?

iredelmeier: Definitely.

My thinking had been that `proposed` made sense for the template since that's the starting state.

Also, the possible states are enumerates here: https://github.com/open-telemetry/rfcs#the-rfc-life-cycle

jmacd: Maybe add an italicized comment w/ the URL spelled out, for reviewing the lifecycle?

bogdandrutu: Sorry I did not see all the definitions there. Looks good.

PR #18, opened 2019-08-13
Proposal: A layered approach to data formats

tedsuo:
This RFC describes the high level requirements and structure for defining OpenTelemetry as a data format. 

If accepted, this document will serve as a rough draft for the data formats overview section of the specification. The final choices for each layer will then be added to the spec, and checked against the requirements listed here.

This layered structure is unlikely to be controversial at this point. However, please review the requirements and examples, and correct or add to them before approving this RFC.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -12,4 +12,4 @@
 #  https://help.github.com/en/articles/about-code-owners
 #
 
-*  @bogdandrutu @carlosalberto @SergeyKanzhelev @yurishkuro @tedsuo @iredelmeier @aloisReitbauer @songy23 @c24t @reyang
+*  @bogdandrutu @carlosalberto @SergeyKanzhelev @yurishkuro @tedsuo @iredelmeier @aloisReitbauer @songy23 @c24t @reyang @tigrannajaryan


text/0000-data-formats.md
```diff
@@ -0,0 +1,151 @@
...
+
+# Encoding Formats
+
+While there is a single set of OpenTelemetry data structures, they can be encoded in a variety of serialization formats.
+
+## Requirements
+
+### Prefer pre-existing formats
+
+Codecs must be commonly available in all major languages. It is outside the bounds of the OpenTelemetry project to define new serialization formats.
```
yurishkuro: These two sentences seem somewhat mutually contradictory. Or do you mean "new formats aside from JSON/Proto"?

text/0000-data-formats.md
```diff
@@ -0,0 +1,151 @@
...
+OpenTelemetry is an observability system, which generates observations, and then sends that data to remote systems for analysis. Describing how  this observability data is represented and transmitted is critical part of the OpenTelemetry specification.
+
+**Proposed Layers**
+*   Metadata and Semantic Conventions
+*   Logical Structure
+*   Encodings
+*   Exchange Protocols
+
+
+# Metadata and Semantic Conventions
```
yurishkuro: Does this need to mention "metadata"?

text/0000-data-formats.md
```diff
@@ -0,0 +1,151 @@
...
+
+# Logical Structure
+
+OpenTelemetry observability model, defined as data structures instead of API calls. A logical layer describes the default data structures expected to be produced by OpenTelemetry APIs for transport.
+
+## Requirements
+
+### Support Metadata
+
+The logical structure must be able to support all of the requirements for the metadata layer.
```
yurishkuro: again, can we avoid using "metadata"?

text/0000-data-formats.md
```diff
@@ -0,0 +1,151 @@
...
+### Prefer pre-existing formats
+
+Codecs must be commonly available in all major languages. It is outside the bounds of the OpenTelemetry project to define new serialization formats.
+
+**Examples:**
+*   Protobuf
+*   JSON
+
+
+# Exchange Protocols
```
yurishkuro: This largely overlaps with @tigrannajaryan 's RFC for the protocol goals

tedsuo: Yes, I intentionally lifted them from there, since there was no separate RFC I could point to, only a spec PR. I included them here for completeness.

text/0000-data-formats.md
tedsuo: I think we could just call the top layer "Semantic Conventions" but this layer needs to mention "metadata support" as that is the facility which the logical structure must have, in order for semantic conventions to be implemented.

Is it the concept you don't like, or just the term "metadata"?

text/0000-data-formats.md
tedsuo: Probably not, I can remove that reference.

tedsuo:
Also, a general question: we've been using the term "Semantic Conventions" for a while now, especially in OpenTracing. So I'm continuing to use it, for continuity's sake. But: I can firmly say that absolutely no one has any clue what "Semantic Conventions" means when they first hear it. Is there a clearer term we can use for this concept?

text/0000-data-formats.md
tedsuo: I'm just trying to say that the OpenTelemetry project is not in the business of inventing new encodings (or transport protocols, for that matter). We are not going to invent something like BISON just to shave off a few bytes. 

The reasoning is that preexisting, popular encodings already have codecs available in all languages. So when someone wants to implement OpenTelemetry for FooLang, they aren't stuck writing a new codec on top of everything else.

tedsuo: (I agree it could be worded better)

tigrannajaryan:
>Also, a general question: we've been using the term "Semantic Conventions" for a while now, especially in OpenTracing. So I'm continuing to use it, for continuity's sake. But: I can firmly say that absolutely no one has any clue what "Semantic Conventions" means when they first hear it. Is there a clearer term we can use for this concept?

Perhaps "Attribute Name Conventions" instead? I am assuming we only use for attributes, don't know if it applies to anything else.

text/0000-data-formats.md
yurishkuro: Dapper-style tracing systems are often referred to as "based on metadata propagation", so I find the usage of "metadata" in the context of trace data confusing, so best if we can avoid id. It's also not like there is some agreed upon understanding of what "metadata" means. Since the RFC talks about layered approach, how about
  * semantic layer
  * logical layer
  * ??? layer
  * transport layer

text/0000-data-formats.md
```diff
@@ -0,0 +1,151 @@
+# RFC: Data Formats for OpenTelemetry
+
+# Overview
+OpenTelemetry is an observability system, which generates observations, and then sends that data to remote systems for analysis. Describing how  this observability data is represented and transmitted is critical part of the OpenTelemetry specification.
```
reyang:
```suggestion
OpenTelemetry is an observability system, which generates observations, and then sends that data to remote systems for analysis. Describing how this observability data is represented and transmitted is critical part of the OpenTelemetry specification.
```

text/0000-data-formats.md
```diff
@@ -0,0 +1,151 @@
...
+
+
+# Metadata and Semantic Conventions
+
+Spans represent specific operations in and between systems. Some of these operations represent calls that use well-known protocols like HTTP or database calls. Depending on the protocol and the type of operation, additional information is needed to represent and analyze a span correctly in monitoring systems. It is also important to unify how this attribution is made in different languages. This way, the operator will not need to learn specifics of a language and telemetry collected from multi-language micro-service can still be easily correlated and cross-analyzed.
+
+## Requirements
+
+
+### Handle complex data
```
reyang:
```suggestion
### Handle Complex Data
```

text/0000-data-formats.md
```diff
@@ -0,0 +1,151 @@
...
+### Easily Extensible
+
+Not all semantics conventions and common operations are known at this time. It must be possible to add more semantic definitions without creating to versioning or compatibility issues with OpenTelemetry implementations.
+
+
+## Examples of Common Operations
+
+#### RPC calls
+
+Modeling distributed transactions is a core usecase, it is common to analyze  errors and latency for RPC calls. Because multiple individual network requests may be involved in a single logical RPC operation, it must be possible to describe RPC calls as logical operations, independent from the descriptions of transport and application protocols which they contain.
```
reyang:
```suggestion
Modeling distributed transactions is a core usecase, it is common to analyze errors and latency for RPC calls. Because multiple individual network requests may be involved in a single logical RPC operation, it must be possible to describe RPC calls as logical operations, independent from the descriptions of transport and application protocols which they contain.
```

text/0000-data-formats.md
```diff
@@ -0,0 +1,151 @@
...
+
+Examples:
+*   A client sending a request to a remote server.
+*   A server receiving a request from a remote client.
+*   Logical vs physical networking: a request for “[https://www.wikipedia.org/wiki/Rabbit](https://www.wikipedia.org/wiki/Rabbit)” results in two physical HTTP requests - one 301 redirection, and another 200 okay. The  logical RPC operation represents both requests together.
+
+#### Network protocols
+
+The internet suite of network protocols is thoroughly standardized, and the details of network interactions calls are critical to analysing the behavior of distributed systems.
```
reyang: First time I've seen `analysing` versus `analyzing`. And GitHub would complain about `analysing` (probably due to the fact that it is a U.S. company?).

text/0000-data-formats.md
reyang: How about "schema layer"...

text/0000-data-formats.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Requirements
+
+### Prefer pre-existing formats
+
+Codecs must be commonly available in all major languages. It is outside the bounds of the OpenTelemetry project to define new serialization formats.
+
+**Examples:**
+*   Protobuf
+*   JSON
```
reyang: I suggest that we expand a little bit here about the schematization:
1. Protocols which are super efficient, highly schematized. Encoder and decoder need to have the schema knowledge (usually precompiled into the encoder/decoder), which is not embedded into the payload. For example, a byte array [0x41, 0x42, 0x3, 0x4] which represents ['A', 'B', 3, 4].
2. Protocols which are very efficient, highly schematized. Schema is not embedded into the payload. Examples are Protobuf, which provides more flexibility over option 1.
3. Protocols which are less efficient, but very self-contained. Schema is embedded into the payload. Examples are JSON, BJSON and MessagePack.

The decision could significantly affect the performance and usability of OpenTelemetry.

text/0000-data-formats.md
```diff
@@ -0,0 +1,151 @@
...
+
+
+# Exchange Protocols
+
+Exchange protocols describe how to transmit serialized OpenTelemetry data. Beyond just defining which application and transport protocol should be used, the exchange protocol defines transmission qualities such as streaming, retry, acks, backpressure, and throttling.
```
reyang: @tedsuo now that you've mentioned `retry`, I wonder if it'll be interesting to put `de-dupe` - they are like twins.

text/0000-data-formats.md
tedsuo: War of the spellchecks :)

text/0000-data-formats.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Requirements
+
+
+### Handle Complex Data
+
+Semantic information is often complex and multi-valued. How we represent semantic data is related to how attributes are defined at the logical layer.
+
+*   At minimum, nested lists of key/value pairs are necessary to represent them.
+*   Is it advantageous to also have strongly typed data structures?
```
bogdandrutu: Not sure if this comment refers to "strongly typed language support" or recorded data to be "strongly typed" (do not allow in java to record Object).

If the data that we support are not typed (strongly typed) there will be vendor issues (locking), an easy scenario is let's assume vendor X supports for Http spans a specific object that describes the operation (HttpDetails). If any instrumentation uses HttpDetails then that means the instrumentation is specific to vendor X not generic. Even though someone may say that HttpDetails will be serialized as a string, this instrumentation offers competitive advantages to vendor X, then other vendors will have to implement that in order to offer the same capability. Also this becomes even worse if the vendor X controls the definition of the HttpDetails.

If you refer about languages that they do not have a strongly type system then we should do our best (using static code checks or other mechanisms) to ensure only accepted types are used.

So I suggest that we do not support random objects in order to discourage any vendor locking and to try to offer vendor neutrality.

text/0000-data-formats.md
```diff
@@ -0,0 +1,151 @@
...
+*   Are there issues with having strongly typed data structures in certain environment, such as dynamic languages?
+
+
+### Compose Multiple Semantics
+
+Objects in OpenTelemetry may describe more than one type of operation. For example, a span may simultaneously represent a MySQL transaction and a TCP/IP request. Pieces of metadata must compose well, without losing the structure integrity of each piece or risking a conflict in meaning.
```
bogdandrutu: What is an "object" in this sentence?

text/0000-data-formats.md
```diff
@@ -0,0 +1,151 @@
...
+## Examples of Common Operations
+
+#### RPC calls
+
+Modeling distributed transactions is a core usecase, it is common to analyze errors and latency for RPC calls. Because multiple individual network requests may be involved in a single logical RPC operation, it must be possible to describe RPC calls as logical operations, independent from the descriptions of transport and application protocols which they contain.
+
+Examples:
+*   A client sending a request to a remote server.
+*   A server receiving a request from a remote client.
+*   Logical vs physical networking: a request for “[https://www.wikipedia.org/wiki/Rabbit](https://www.wikipedia.org/wiki/Rabbit)” results in two physical HTTP requests - one 301 redirection, and another 200 okay. The  logical RPC operation represents both requests together.
```
bogdandrutu: Why do we not want to model this as a "rabbit" operation span with two children?

Oberon00:
Why was this closed? Was this intentional? There was no explanation given and it looks like commits were made after closing?

PR #22, opened 2019-08-16
Change 0002 Remove SpanData status to approved

bg451:
If an rfc is merged, does that mean it's approved? cc @iredelmeier

### DIFF
--- a/text/0002-remove-spandata.md
+++ b/text/0002-remove-spandata.md
@@ -1,6 +1,6 @@
 # Remove SpanData
 
-**Status:** `proposed`
+**Status:** `approved`
 
 Remove and replace SpanData by adding span start and end options.
 


carlosalberto:
> If an rfc is merged, does that mean it's approved?

It *seems* like that is not always the case, but for this specific RFC, it is ;)

bogdandrutu:
Yes based on previous discussions that is not the case, for this I think we are good.

bogdandrutu:
I am waiting for 1 more approval before merging.

Oberon00:
Could the PR title be changed to include the name of the RFC? That way it might get the right people look over that quicker (probably doesn't matter anymore for this PR now, but for the future...).

bogdandrutu:
We have 5/10 approvers, do we need majority for this to merge (>5)?

PR #24, opened 2019-08-16
Propose sampling API changes

bogdandrutu:
This RFC is a collaboration between @lizthegrey and @bogdandrutu.

If accepted, this document will serve as a rough draft for the new Sampling API in the specification.


text/0006-sampling.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Motivation
+
+Different users of OpenTelemetry, ranging from library developers, packaged infrastructure binary
+developers, application developers, operators, and telemetry system owners, have separate use cases
+for OpenTelemetry that have gotten muddled in the design of the original Sampling API. Thus, we need
+to clarify what APIs each should be able to depend upon, and how they will configure sampling and
+OpenTelemetry according to their needs.
+
+![Personas](https://i.imgur.com/w1H0CfH.png)
```
c24t: Might want to replace this with a better diagram if we intend RFCs to be a record of changes to the library instead of just an artifact of development.

text/0006-sampling.md
```diff
@@ -0,0 +1,156 @@
...
+ `Span` the moment when the builder is created and not the moment when the `build()` method is
+ called.
+
+## Motivation
+
+Different users of OpenTelemetry, ranging from library developers, packaged infrastructure binary
+developers, application developers, operators, and telemetry system owners, have separate use cases
+for OpenTelemetry that have gotten muddled in the design of the original Sampling API. Thus, we need
+to clarify what APIs each should be able to depend upon, and how they will configure sampling and
+OpenTelemetry according to their needs.
```
c24t: The personas are helpful to motivate the changes in this RFC. It would be great to have these in another doc (in specs or elsewhere) separate from the sampler changes.

bogdandrutu: I will do that and use this as a starting point.

text/0006-sampling.md
bogdandrutu: Happy if you have better ascii art skills to approve a change :)

text/0006-sampling.md
```diff
@@ -0,0 +1,156 @@
...
+ * They may have their own recommendations for sampling rates, but don't run the binaries in
+ production, only provide packaged binaries. So their sampling rate configs, and sampling strategies
+ need to a finite "built in" set from OpenTelemetry's SDK.
+ * They need to deal with upstream sampling decisions made by services calling them.
+
+**Solution:**
+ * Allow different sampling strategies by default in OTel SDK, all configurable easily via YAML or
+ future flags, etc.:
+   * Trust parent sampling decision (trusting & propagating parent SpanContext SampleBit)
+   * Always keep
```
reyang: I don't know if we plan to cover this scenario: always record the telemetry, but propagate the parent SpanContext SampleBit.

text/0006-sampling.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Explanation
+
+We outline five different use cases (who may be overlapping sets of people), and how they should
+interact with OpenTelemetry:
+
+### Library developer
+Examples: gRPC, Express, Django developers.
+
+ * They must only depend upon the OTel API and not upon the SDK.
```
reyang: Not sure if we decided to call OpenTelemetry OTel.
It seems to be bad for SEO.
Also, OTel is a company name.

text/0006-sampling.md
```diff
@@ -0,0 +1,156 @@
...
+**Solution:**
+ * Infrastructure owners receive information attached to the span, after sampling hooks have already
+ been run.
+
+## Internal details
+The interface for the Sampler class takes in:
+ * `TraceID`
+ * `SpanID`
+ * Parent `SpanContext` if any
+ * `Links`
```
reyang: Does this mean all links should be provided during span construction?

text/0006-sampling.md
```diff
@@ -0,0 +1,156 @@
...
+ and OTel wire format exporter.
+ * They may have their own recommendations for sampling rates, but don't run the binaries in
+ production, only provide packaged binaries. So their sampling rate configs, and sampling strategies
+ need to a finite "built in" set from OpenTelemetry's SDK.
+ * They need to deal with upstream sampling decisions made by services calling them.
+
+**Solution:**
+ * Allow different sampling strategies by default in OTel SDK, all configurable easily via YAML or
+ future flags, etc.:
+   * Trust parent sampling decision (trusting & propagating parent SpanContext SampleBit)
```
tsloughter: This is more an additional property of the other samplers and not a sampler itself. Root spans would have to have a separate sampler configuration if the dev had to pick "trust parent sampling decision" as an explicit sampler.

Better would be if all samplers were required to offer a setting of trusting the parent or not. Then this is simply a setting on the other samplers and not a sampler in itself.

text/0006-sampling.md
```diff
@@ -0,0 +1,156 @@
...
+ * Parent `SpanContext` if any
+ * `Links`
+ * Initial set of `Attributes` for the `Span` being constructed
+
+It produces as an output:
+* A boolean indicating whether to sample or drop the span.
+* The new set of initial span Attributes (or passes along the SpanAttributes unmodified)
+* (under discussion in separate RFC) the SamplingRate float.
+
+## Trade-offs
```
yurishkuro: These read more like "alternatives considered", rather than "trade-offs". Either way, would be useful to capture the reasoning  and pros/cons of these alternatives.

text/0006-sampling.md
bogdandrutu: fixed, sorry for this and thanks for your note.

text/0006-sampling.md
bogdandrutu: Just this specific sampler does not allow users to configure that, and it has a defined behavior for root spans. I updated with more details all the samplers.

text/0006-sampling.md
bogdandrutu: I think this is a good practice, we should start with this approach and later add support if needed to record links on the Span. I also added a section to motivate why to remove them from the Span.

text/0006-sampling.md
bogdandrutu: Please re-read and let me know if after I tried to clarify all the flags/parameters/etc. is still unclear.

text/0006-sampling.md
bogdandrutu: I think I re-wrote this section, please take another look. I will close this conversation since it is not relevant anymore but you can start new conversations.

Thanks for your comment.

bogdandrutu:
@lizthegrey  @c24t  @reyang please re-read because I made some large changes and things may have changed.

text/0006-sampling.md
tsloughter: Ok, yea, looks good now.

text/0006-sampling.md
reyang: Much better! Now I have no problem understanding it. Thanks!

text/0006-sampling.md
```diff
@@ -0,0 +1,300 @@
+# Sampling API
+
+*Status: proposed*
+
+## TL;DR
+This section tries to summarize all the changes proposed in this RFC:
+ 1. Move the `Sampler` interface from the API to SDK package. Apply some minor changes to the
+ `Sampler` API.
+ 1. Add a new `SamplerHint` concept to the API package.
+ 1. Add capability to record `Attributes` that can be used for sampling decision during the `Span`
```
jmacd: I don't understand why we would call these synthetic things "Attributes".  They're not given by the programmer, they're generated by the SDK. Can we record these key:values in a way that marks them clearly as having been injected by the implementation? They're not the user's attributes.

lizthegrey: They are the user's attributes, and can/should be supplied by the programmer before being used for sampling.

lizthegrey:
We don't need a separate RFC to resolve the question of propagating the sampleRate if @tedsuo's hypothesis from https://github.com/open-telemetry/opentelemetry-specification/pull/198#issuecomment-522189434 is true; instead, we can just incorporate that `sampleRate`, if it exists, will be read from the `tracestate`, multiplied by the current `sampleRate` of the current `Sampler`, and passed into the newly created `tracestate` using the internal SDK method.

Awaiting feedback from @iredelmeier there and here as a result

text/0006-sampling.md
jmacd: Oh, that wasn't clear to me. I thought we already had such a requirement, since the OpenTracing start method includes a list of attributes. I would state this simply as a requirement that the user be allowed to specify attributes at start time, practically unrelated to sampling. Thanks.

text/0006-sampling.md
```diff
@@ -0,0 +1,300 @@
...
+ creation time.
+ 1. Add capability to start building a `Span` with a delayed `build` method. This is useful for
+ cases where some `Attributes` that are useful for sampling are not available when start building
+ the `Span`. As an example in Java the current `Span.Builder` will use as a start time for the
+ `Span` the moment when the builder is created and not the moment when the `build()` method is
+ called.
+ 1. Remove `addLink` APIs from the `Span` interface, and allow recording links only during the span
```
jmacd: Is this required? If I wanted to record my children in the parent span, a very natural way to do that would be via adding a link. In the case of attributes, we now have an API that lets you provide them at start time or during the span's lifecycle. Could we do the same with links?  This change doesn't feel like a requirement, considering how attributes are being treated.

reyang: I think we need to revisit links. Currently it has a lot of unclear stuff:
1. What's the maximum allowed number of links?
2. Do we in general keep links to all the children in a _logical parent_ span, or we put `logical parent` as a link in the child span? (or both?)
3. Do we allow links to be added postmortem (after both source and target spans are exported)?

We should track this in a separate PR instead of blocking this one.

bogdandrutu: @jmacd I applied a simple rule when I suggested this: adding a new API is considered backwards compatible so delaying this decision seems to be fine. this is why I proposed this :)

Also if we do apply the links later is going to be a discussion if they can influence the sampling/recording decision.

text/0006-sampling.md
```diff
@@ -0,0 +1,300 @@
...
+
+## Explanation
+
+We outline five different use cases (who may be overlapping sets of people), and how they should
+interact with OpenTelemetry:
+
+### Library developer
+Examples: gRPC, Express, Django developers.
+
+ * They must only depend upon the OpenTelemetry API and not upon the SDK.
```
songy23: I think it's fine to have tests depend on SDK.

text/0006-sampling.md
```diff
@@ -0,0 +1,300 @@
...
+stored to the backend, thus some of these systems usually don't scale to a high volume of traces,
+or the cost to store all the Spans may be too high. In order to support this use-case and to 
+ensure the quality of the data we send, OpenTelemetry needs to natively support sampling with some
+requirements:
+ * Send as many complete traces as possible. Sending just a subset of the spans from a trace is 
+ less useful because in this case the interaction between the spans may miss.
+ * Allow application operator to configure the sampling frequency.
+ 
+For new modern systems that need to collect all the Spans and later may or may not do a deferred 
+sampling decision, the OpenTelemetry needs to natively support a way to configure the library to
```
songy23:
```suggestion
sampling decision, OpenTelemetry needs to natively support a way to configure the library to 
```

text/0006-sampling.md
```diff
@@ -0,0 +1,300 @@
...
+ 
+For new modern systems that need to collect all the Spans and later may or may not do a deferred 
+sampling decision, the OpenTelemetry needs to natively support a way to configure the library to 
+collect and export all the Spans. This is possible even though OpenTelemetry supports sampling by
+setting a default config to always collect all the spans.
+
+### Sampling flags
+OpenTelemetry API has two flags/properties:
+ * `RecordEvents`
+   * This property is exposed in the `Span` interface.
```
songy23: `Span` or `Span.Builder`?

text/0006-sampling.md
```diff
@@ -0,0 +1,300 @@
...
+ * Parent `SpanContext` if any
+ * `SamplerHint`
+ * `Links`
+ * Initial set of `Attributes` for the `Span` being constructed
+
+It produces as an output called `SamplingResult`:
+ * A `SamplingDecision` enum [`NOT_RECORD`, `RECORD`, `RECORD_AND_PROPAGATE`].
+ * A set of span Attributes that will also be added to the `Span`.
+   * These attributes will be added after the initial set of `Attributes`.
+ * (under discussion in separate RFC) the SamplingRate float.
```
songy23: How about span name?

text/0006-sampling.md
```diff
@@ -0,0 +1,300 @@
...
+ It may be possible to always collect all the events until the sampling decision is made but this is
+ an important optimization.
+
+There are two important use-cases to be considered:
+ * All information that may be used for sampling decisions are available at the moment when the 
+ logical `Span` operation should start. This is the most common case.
+ * Some information that may be used for sampling decision are NOT available at the moment when the
+ logical `Span` operation should start (e.g. `http.route` may be determine later).
+
+The current [span creation logic][span-creation] it facilitate very well the first use-case, but
```
songy23:
```suggestion
The current [span creation logic][span-creation] facilitates very well the first use-case, but 
```

text/0006-sampling.md
```diff
@@ -0,0 +1,300 @@
...
+
+There are two important use-cases to be considered:
+ * All information that may be used for sampling decisions are available at the moment when the 
+ logical `Span` operation should start. This is the most common case.
+ * Some information that may be used for sampling decision are NOT available at the moment when the
+ logical `Span` operation should start (e.g. `http.route` may be determine later).
+
+The current [span creation logic][span-creation] it facilitate very well the first use-case, but 
+the second use-case requires users to record the logical `start_time` and collect all the 
+information necessarily to start the `Span` in custom objects, then when all the information are
```
songy23:
```suggestion
information necessarily to start the `Span` in custom objects, then when all the information is 
```

text/0006-sampling.md
bogdandrutu: I don't think the Sampler should change the Span name.

songy23: Apologies I commented on the wrong line - I mean the `Sample` interface may need to accept span names as input parameter in order to allow sampling on particular namespaces. /cc @rghetia brought this up before.

bogdandrutu: Good catch, I forgot about that.

Done.

text/0006-sampling.md
bogdandrutu: Done.

text/0006-sampling.md
bogdandrutu: Clarify that I was referring to isRecordingEvents.

text/0006-sampling.md
bogdandrutu: Fixed by using properties.

text/0006-sampling.md
```diff
@@ -0,0 +1,302 @@
...
+     MakeSamplingDecision().
+ * We considered allowing the sampling decision to be arbitrarily delayed, but guaranteed before 
+ any child `Span` is created, or `SpanContext` is accessed, or before `Span.end()` finished.
+   * Pros:
+     * Similar and smaller API that supports both use-cases defined ahead.
+   * Cons:
+     * If `SamplingHint` needs to also be delayed recorded then an extra API on Span is required 
+     to set this.
+     * Does not allow optimization to not record tracing events, all tracing events MUST be 
+     recorded before the sampling decision is made.
```
yurishkuro: if sampling decisions are based on a set of policies, it can be an inherent attribute of a policy whether it depends on a delayed datum (such as late tag or span name) and thus requires recording events prior to sampling decision being made. If user picks a policy that does not depend on such delayed data (e.g. user picks a basic probabilistic policy), then the tracer would know that no delayed sampling is going to happen and can short-circuit event recording once the decision is No. This is the approach we're taking in Jaeger SDKs.

text/0006-sampling.md
```diff
@@ -0,0 +1,302 @@
...
+ * Allow users to control recording for individual Spans.
+ * OpenCensus has this to support z-pages, so we need to keep backwards compatibility.
+
+The flag combination `SampledFlag == true` and `RecordEvents == false` can cause gaps in the 
+distributed trace, and because of this OpenTelemetry API should NOT allow this combination.
+
+It is safe to assume that users of the API should only access the `RecordEvents` property when 
+instrumenting code and never access `SampledFlag` unless used in context propagators.
+
+### SamplingHint
```
yurishkuro: I suggest adding clarification / examples when the user code would want to provide one of the 4 values.

text/0006-sampling.md
```diff
@@ -0,0 +1,302 @@
...
+This is a new concept added in the OpenTelemetry API that allows to suggest sampling hints to the
+implementation of the API:
+ * `UNSPECIFIED`
+   * This is the default option.
+   * No suggestion is made.
```
yurishkuro: I assume the hint object is going to be an optional parameter to start-span function. If so, isn't UNSPECIFIED the same as "hint == null" (i.e. no hint from the user)? It seems like an unnecessary cognitive overload to have this constant.

text/0006-sampling.md
lizthegrey: We deliberately chose to defer the "make decision later" option until later, as it adds more complexity and we can always add it later.

text/0006-sampling.md
```diff
@@ -0,0 +1,302 @@
...
+
+**Solution:**
+ * Allow application developers to link in custom samplers or write their own when using the
+ official SDK.
+   * These might include dynamic per-field sampling to achieve a target rate
+   (e.g. https://github.com/honeycombio/dynsampler-go)
+ * Sampling decisions are made within the start Span operation, after attributes relevant to the
+ span have been added to the Span start operation but before a concrete Span object exists (so that
+ either a NoOpSpan can be made, or an actual Span instance can be produced depending upon the
+ sampler's decision).
```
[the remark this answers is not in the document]
lizthegrey: ad-hoc sampling rules seem like they're ultimately a property of the collection exporter/backend rather than of the OTel SDK... e.g. I wouldn't insist on making the OTel SDK exactly replicate what Honeycomb does post-collection...

text/0006-sampling.md
```diff
@@ -0,0 +1,302 @@
...
+ 1. Move the `Sampler` interface from the API to SDK package. Apply some minor changes to the
+ `Sampler` API.
+ 1. Add a new `SamplerHint` concept to the API package.
+ 1. Add capability to record `Attributes` that can be used for sampling decision during the `Span`
+ creation time.
+ 1. Add capability to start building a `Span` with a delayed `build` method. This is useful for
+ cases where some `Attributes` that are useful for sampling are not available when start building
+ the `Span`. As an example in Java the current `Span.Builder` will use as a start time for the
+ `Span` the moment when the builder is created and not the moment when the `build()` method is
+ called.
```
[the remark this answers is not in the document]
lizthegrey: See below -- it's necessary in order to have a clean separation between decision later, and decision now based on what we know before any new child spans are created, which requires more difficult handling.

text/0006-sampling.md
rghetia: Using http://stable.ascii-flow.appspot.com/#Draw

```




                    +----------+           +-----------+
           grpc     |  Library |           |           |
           Django   |  Devs    +---------->| OTel API  |
           Express  |          |   +------>|           |
                    +----------+   |  +--->+-----------+                  +---------+
                                   |  |                                   | OTel    |
                                   |  |                                +->| Proxy   +---+
                                   |  |                                |  |         |   |
                    +----------+   |  |    +-----------+------------+  |  +---------+   |
                    |          |   |  |    |           | OTel Wire  |  |                |
           Hbase    |  Infra   |   |  |    |           | Export     |+-+                v
           Envoy    |  Binary  +---+  |    |  OTel     |            |  |           +----v-----+
                    |  Devs    |      |    |  SDK      +------------+  |           |          |
                    +----------+---------->|           |            |  +---------->|  Backend |
                                   +------>|           | Custom     |  +---------->|          |
                                   |  |    |           | Export     |  |           +----------+
                    +----------+   |  |    |           |            |+-+
                    |          +---+  |    +-----------+------------+
                    |  App     +------+
                    |  Devs    +
                    |          |
                    +----------+                                            Telemetry
                                                                            Owner
```

bogdandrutu: Thanks for the link, very useful.

text/0006-sampling.md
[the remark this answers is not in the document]
bogdandrutu: The use-case that I have in mind is an HTTP client/server framework. For example on the server side, the route may be determine after parsing the headers and after we are able to logically start the span. We will  create the Builder when headers are parsed and we will add an `http.route` later, then just before we call into the user application code we call `build`. When calling `build` we will use as a start_time the moment when the builder was created and not the moment when `build` is called.

text/0006-sampling.md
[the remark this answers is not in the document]
bogdandrutu: @yurishkuro can you clarify if the sampling decision still happens on the client side? If that is true can you clarify when is the limit when that needs to happen. One possibility that I covered in the "when does sampling happen" is that we can delay until the moment the SpanContext is accessed: to create a child, to serialize it, or for example to do log correlation.

text/0006-sampling.md
bogdandrutu: We need to define when is the limit when the sampling decision should happen. I like the idea to have on the sampler a property that says is delayed sampling so we keep all the events until the sampling happens, but when is the limit/point that guarantees that sampling happened.

text/0006-sampling.md
bogdandrutu: I am a dev that does not like NULL as a default. The reason is that your code will look like this:

```
if (hint == null) {
  do something for default
}
switch (hint) {
  case RECORD:
    ...
...
}
```

With non-null value the code is simpler:
```
switch (hint) {
  case UNSPECIFIED {
    do something for default
    break;
  }
  case RECORD:
    ...
...
}
```

It is more readable in my opinion.

text/0006-sampling.md
bogdandrutu: Here is the link for future updates http://stable.ascii-flow.appspot.com/#Draw1014856397574566501/1404005869

text/0006-sampling.md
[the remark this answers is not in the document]
yurishkuro: This implies that the http framework has some means of passing the Builder down the call stack, whereas the default context propagation mechanism only works with a Span.

text/0006-sampling.md
yurishkuro: Whether to keep the NULL or not is an internal implementation choice, and also language specific -- you could use Option, you could catch NULL in `default:` part of the switch, or in Python we don't even have enums so any private value can be used as an UNSPECIFIED sentinel. You could also keep a separate isHintSet boolean. There are lots of options, all having to deal with the possibility of user never invoking `withSamplingHint()`, which I assume we're not suggesting to make a required arg.

But exposing UNSPECIFIED at the API level allows the user to do meaningless thing like
```
spanBuilder.withSamplingHint(SamplingHind.UNSPECIFIED)
```

I would choose clean API over any implementation concerns.

text/0006-sampling.md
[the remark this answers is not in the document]
yurishkuro: Yes, sampling decision still happens in the client.

The limit depends on the type of sampling policies provided to the client. Consider these policies:
  * A: static: sample with probability p=0.1
  * B: if span.name = "/userProfile" then sample with rate limit r=10/sec
  * C: tag based: if tag X = Y, then sample with probability p=0.9

If the tracer is configured with policies [A], then sampling decision is finalized immediately on span creation, due to the nature of the sole policy A which is independent of any later span data.

If the tracer is configured with policies [B, A], then sampling decision can happen on span creation if A says Yes or B says Yes, or after span.setName() (assumption is that the code is only going to call setName once, at least as far as sampling decision is concerned).

If tracer is configured with all 3 policies, then recording of events may never stop if none of the policies match and/or say Yes. It's more expensive but it's the user's choice.

> we can delay until the moment the SpanContext is accessed

That was how we initially implemented the handling of late span.SetName() binding. However, there's no particular reason why the sampling on the parent span cannot remain "undecided" even if child spans are being created. Yes, it will result in some spans of the trace not captured, but that's just the nature of the combination of the chosen sampling policies. We're also thinking that we will use a shared sampling state across all spans for the trace in memory, so  that some of those sampling policies might fire on a child span and trigger sampling of the parent.

Disclaimer: we just brainstormed these scenarios and will prototype them in the next week or so. There's a chance it might not work at all, but so far it looks this approach is meeting our requirements (which definitely include late sampling decision based on a specific tag).

text/0006-sampling.md
yurishkuro: see brain dump above https://github.com/open-telemetry/rfcs/pull/24#discussion_r315999705

text/0006-sampling.md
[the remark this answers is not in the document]
lizthegrey: With respect, "we will prototype 'undecided' states in the next week" feels like "add to v0.0.2 but omit for v0.0.1" given Sep 1 spec freeze. Believe me, I'm all for late-binding sampling decisions, but we need to be explicit about when there is the possibility of that happening, and allowing implementations to act accordingly. We can always add complexity in a methodical fashion, I'm worried about rushing and getting it wrong.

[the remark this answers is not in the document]
yurishkuro: The solution with late binding of span name alone has been in Jaeger in production for >2yrs, implemented as Bogdan implied - events are recorded until sampling decision is finalized, which happens either on span.SetName() or a child span being created. I don't particularly like it, but we had some Node.js frameworks where span name was only known later. The approach I described above is meant to generalize and improve it.

I don't have a problem with not supporting any form of late binding in v1, but this RFC does provide for it via "late Builder", which feels even worse to me.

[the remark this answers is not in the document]
lizthegrey: Ah, I think I understand. Hm. I often know the attributes "at span creation", but I want the time to compute the attributes to be included in the span duration, if that makes sense? That's why we settled on the builder pattern as it was easier for application programmers to work with...

Is there some middle ground here that allows very-short-delayed binding without requiring it all upfront at span creation and adjusting the times afterward?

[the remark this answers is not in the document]
yurishkuro: I guess the question is how big of a deal it would be if v1 didn't support any form of late binding and punted it on v2. In Jaeger, only Node.js client supports late span name binding that impacts sampling, nobody complained in other languages (it only matters when using sampling policies sensitive to span name).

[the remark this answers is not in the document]
bogdandrutu: I am fine to remove the delayed sampling for the moment. But I will still use as an argument in some of the decisions that we may support this, for example the reason we have a different isRecordingEvents is because we may implement something like isDelayed that you suggested so we don't want people to play with the trace flags to know if they should record or not something on the Span.

text/0006-sampling.md
bogdandrutu: I don't have a strong opinion, and definitely all the options you proposed are valid. To clarify: Is your suggestion to remove UNSPECIFIED from the options that user has in the API to set the hint? In other words user has two options: does not set a hint OR sets one of the hints that are meaningful.

text/0006-sampling.md
bogdandrutu: This is in a "alternatives" section and we actually reverted the proposed change. So I would keep it here just for the record and we will discuss in more details when we do make a decision about this.

text/0006-sampling.md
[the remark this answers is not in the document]
bogdandrutu: This was removed.

text/0006-sampling.md
bogdandrutu: Closing this conversation, please re-open if needed.

text/0006-sampling.md
yurishkuro: > In other words user has two options: does not set a hint OR sets one of the hints that are meaningful.

yes

text/0006-sampling.md
```diff
@@ -0,0 +1,335 @@
...
+ but are not expected to support extensibility/plugins beyond the default OpenTelemetry SDK, 
+ OpenTelemetry SDKTracer, and OpenTelemetry wire format exporter.
+ * They may have their own recommendations for sampling rates, but don't run the binaries in
+ production, only provide packaged binaries. So their sampling rate configs, and sampling strategies
+ need to a finite "built in" set from OpenTelemetry's SDK.
+ * They need to deal with upstream sampling decisions made by services calling them.
+
+**Solution:**
+ * Allow different sampling strategies by default in OpenTelemetry SDK, all configurable easily via
+ YAML or future flags. See [default samplers](#default-samplers).
```
tsloughter:
```suggestion
 YAML or feature flags. See [default samplers](#default-samplers).
```

tsloughter:
I just realized I don't think it is made clear. My take away is that sampling happens on every span creation. If this is the case it should be called out because it differs from opencensus which only runs on the root span or one with a remote parent.

tsloughter:
Also, at first, since `RecordEvents` doesn't get used during a child's sampling decision I wasn't concerned that it was an attribute on a span instead of the span context. But thinking about it more, if it is to be used to know whether to skip adding events/attributes/etc to a span then it would be much better for us if it was kept in the span context.

My guess is the reason it isn't in the span context is because it isn't meant to be propagated? The issue for us is checking if a span is sampled is a very efficient operation because it simply checks the context, and so isn't an issue that it may get checked a lot if a lot of events are recorded. While if it is in the span then the table of spans has to be searched each time to find the value.

bogdandrutu:
>My guess is the reason it isn't in the span context is because it isn't meant to be propagated? The issue for us is checking if a span is sampled is a very efficient operation because it simply checks the context, and so isn't an issue that it may get checked a lot if a lot of events are recorded. While if it is in the span then the table of spans has to be searched each time to find the value.

@tsloughter  I think you refer by "us" to erlang. Conceptually this property is a span level property and it does not get propagated to the child span (only the sampled-flag is propagated).

> I just realized I don't think it is made clear. My take away is that sampling happens on every span creation. If this is the case it should be called out because it differs from opencensus which only runs on the root span or one with a remote parent.

Yes I think we should allow the sampler to be more configurable, we can always call the sampler and an implementation for example the probability sampler by default applies the sampling logic only if root or remote parent.

PR #29, opened 2019-08-27
RFC 0003, 0007 updates (Metrics measure type, metrics handle specification), RFC 0004 (configurable aggregation) deleted

jmacd:
Updates to 0003 following work session 8/21/2019

Some content has been moved to a new RFC on metric handles.

This captures _most_ of the discussion from the 8/21 meeting and _mostly_ eliminates the need for RFC 0004.

### DIFF
--- a/text/0007-metric-handles.md
+++ b/text/0007-metric-handles.md
@@ -10,7 +10,9 @@ The specification currently names this concept `TimeSeries`, the object returned
 
 ## Explanation
 
-The `TimeSeries` is renamed to `Handle` as the former name suggests an implementation, not an API concept. `Handle`, we feel, is more descriptive of the intended use.  Likewise with `GetOrCreateTimeSeries` to `GetHandle` and `GetDefaultTimeSeries` to `GetDefaultHandle`, these names suggest an implementation and not the intended use.	Applications are encouraged to re-use metric handles for efficiency.
+The `TimeSeries` is renamed to `Handle` as the former name suggests an implementation, not an API concept. `Handle`, we feel, is more descriptive of the intended use.  Likewise with `GetOrCreateTimeSeries` to `GetHandle` and `GetDefaultTimeSeries` to `GetDefaultHandle`, these names suggest an implementation and not the intended use.
+
+Applications are encouraged to re-use metric handles for efficiency.
 
 Handles are useful to reduce the cost of repeatedly recording a metric instrument (cumulative, gauge, or measure) with a pre-defined set of label values.  All metric kinds support declaring a set of required label keys.  These label keys, by definition, must be specified in every metric `Handle`.  We permit "unspecified" label values in cases where a handle is requested but a value was not provided.  The default metric handle has all its required keys unspecified.  We presume that fast pre-aggregation of metrics data is only possible, in general, when the pre-aggregation keys are a subset of the required keys on the metric.
 


text/0003-measure-metric-type.md
```diff
@@ -2,47 +2,145 @@
...
 
     000x-metric-pre-defined-labels
     000x-metric-measure
     000x-eliminate-stats-record
 
-## Overview
+### Updated 8/27/2019
 
-Introduce a `Measure` type of metric object that supports a `Record` API.  Like existing `Gauge` and `Cumulative` metrics, the new `Measure` metric supports pre-defined labels.  A new measurement batch API is introduced for recording multiple metric observations simultaneously.
+A working group convened on 8/21/2019 to discuss and debate the two metrics RFCs (0003 and 0004) and several surrounding concerns.  This document has been revised with related updates that were agreed upon during this working session.  See the (meeting notes)[https://docs.google.com/document/d/1d0afxe3J6bQT-I6UbRXeIYNcTIyBQv4axfjKF4yvAPA/edit#].
```
krnowak: It should be `[meeting notes](https://…)` (parentheses are mixed up).

The access to the document seems to be limited, though (I could not see it without requesting permission).

text/0003-measure-metric-type.md
```diff
@@ -2,47 +2,145 @@
 
 **Status:** `proposed`
 
-## Forward
+# Foreward
```
tedsuo:
```suggestion
# Foreword
```

jmacd:
@tedsuo please take another look

text/0003-measure-metric-type.md
```diff
@@ -2,50 +2,169 @@
...
+
+## Metrics API concepts
+
+The `Meter` interface represents the metrics portion of the OpenTelemetry API.
+
+There are three kinds of metric instrument, `CumulativeMetric`, `GaugeMetric`, and `MeasureMetric`.
+
+Metric instruments are constructed by the API, they are not constructed by any specific SDK.
+
+| Name | A string. |
```
tigrannajaryan: The formatting of this table looks broken to me: https://github.com/open-telemetry/oteps/blob/a8b7baf3d8a0514a70b770543e010e4dff6260d7/text/0003-measure-metric-type.md

tigrannajaryan:
@jmacd thank you for well-articulated RFC.

text/0003-measure-metric-type.md
```diff
@@ -2,50 +2,169 @@
 
 **Status:** `proposed`
 
-## Forward
+# Foreward
 
 This propsal was originally split into three semi-related parts. Based on the feedback, they are now combined here into a single proposal. The original proposals were:
```
krnowak: `s/propsal/proposal`

text/0003-measure-metric-type.md
```diff
@@ -2,50 +2,169 @@
...
+
+| Name | A string. |
+| Kind | One of Cumulative, Gauge, or Measure. |
+| Required Keys | List of always-defined keys in handles for this metric. |
+| Unit | The unit of measurement being recorded. |
+| Description | Information about this metric. |
+
+See the specification for more information on these fields, including formatting and uniqueness requirements.  To define a new metric, use one of the language-specific API methods (e.g., with names like `NewCumulativeMetric`, `NewGaugeMetric`, or `NewMeasureMetric`).
+
+Metric instrument Handles are SDK-provided objects that combine a metric instrument with a set of pre-defined labels.  Handles are obtained by calling a language-specific API method (e.g., `GetHandle`) on the metric instrument with its label values.  Handles may be used to `Set()`, `Add()`, or `Record()` metrics according to their kind.  The `Set()`, `Add()`, and `Record()`
```
krnowak: The last sentence in the paragraph is cut.

text/0003-measure-metric-type.md
```diff
@@ -2,50 +2,169 @@
...
+- Is the event count a matter of primary interest?
+- Is the distribution (p50, p99, etc.) a matter of primary interest?
+
+The specification will be updated with the following guidance.
+
+### Cumulative metric
+
+Likely to be the most common kind of metric, cumulative metric events express the computation of a sum.  Choose this kind of metric when the value is a quantity, the sum is of primary interest, and the event count and distribution are not of primary interest.  To raise (or lower) a cumulative metric, call the `Add()` method.
+
+If the quantity in question is always non-negative, it implies that the sum is strictly ascending.  When this is the case, the cumulative metric also serves to define a rate.  For this reason, cumulative metrics have a `NonNegative` option to be declared as non-negative.  The API will reject negative updates to non-negative cumulative metrics, instead submitting an SDK error event, which helps ensure meaningful rate calculations.
```
krnowak: Excuse me for being a nitpicker, but if quantity is zero then it is still non-negative, but the sum will be ascending, not strictly ascending. So either the quantity is positive for the sum to be strictly ascending or the quantity is non-negative for the sum to be ascending.

text/0003-measure-metric-type.md
```diff
@@ -2,50 +2,169 @@
...
+
+Only the gauge kind of metric supports observing the metric via a gauge `Observer` callback (as an option).  Semantically, there is an important difference between explicitly setting a gauge and observing it through a callback.  In case of setting the gauge explicitly, the call happens inside of an implicit or explicit context.  The implementation is free to associate the explicit `Set()` event with a context, for example.  When observing gauge metrics via a callback, there is no context associated with the event.
+
+As a special case, to support existing metrics infrastructure, a gauge metric may be declared as a precomputed cumulative sum using the `NonDescending` option, in which case it is defined as a strictly ascending.  The API will reject descending updates to non-descending gauges, instead submitting an SDK error event.
```
krnowak: Same here about the strictness of the ascending.

Also, should it be clarified whether the `Observer` callback is called if gauge is set to the same value as before?

text/0003-measure-metric-type.md
bogdandrutu: @jmacd please fix

text/0003-measure-metric-type.md
bogdandrutu: @krnowak a rate can be calculated for any ascending sum (not necessary strictly ascending). @jmacd please fix :)

text/0003-measure-metric-type.md
```diff
@@ -2,50 +2,168 @@
...
+
+1. The sum is of interest in addition to the count (rate)
+1. Quantile information is of interest.
+
+The key property of a measure metric event is that computing quantiles and/or summarizing a distribution (e.g., via a histogram) may be expensive.  Not only will implementations have various capabilities and algorithms for this task, users may wish to control the quality and cost of aggregating measure metrics.
+
+Like cumulative metrics, non-negative measures are an important case because they support rate calculations. As an option, measure metrics may be declared as `NonNegative`.  The API will reject negative metric events for non-negative measures, instead submitting an SDK error event.
+
+Because measure metrics have such wide application, implementations are likely to provide configurable behavior.  OpenTelemetry may provide such a facility in its standard SDK, but in case no configuration is provided by the application, a low-cost policy is specified as the default behavior, whic is to export the sum, the count (rate), the minimum value, and the maximum value.
```
bogdandrutu: not sure the min/max are relevant. can you please point me to how this can be useful?

text/0003-measure-metric-type.md
```diff
@@ -2,50 +2,168 @@
...
+
+### Disable selected metrics by default
```
bogdandrutu: Not sure I understand this. Do we offer an "enable" option when a metric is created? If so, please mention that clearly.

text/0003-measure-metric-type.md
```diff
@@ -2,50 +2,168 @@
...
+
+All OpenTelemetry metrics may be disabled by default, as an option.  Use this option to indicate that the default implementation should be to do nothing for events about this metric.
+
+### RecordBatch API
+
+Applications sometimes want to act upon multiple metric handles in a single API call, either because the values are inter-related to each other, or because it lowers overhead.  We agree that recording batch measurements will be restricted to measure metrics, although this support could be extended to all kinds of metric in the future.
+
+A single measurement is defined as:
```
bogdandrutu: Not sure I understand how this will be implemented:
Does the library developer need to provide Handles OR Metrics + a list of label values that will be applied to get the Handles (if an implementation decide to do that). Some other implementations may decide to record this event as it is so GetHandle can be avoided.

The second option is what we have in OpenCensus and works well.

text/0003-measure-metric-type.md
```diff
@@ -2,50 +2,168 @@
...
+
+### Eliminate `GetDefaultHandle()`
+
+Instead of a mechanism to obtain a default handle, some languages may prefer to simply operate on the metric instrument directly in this case.  Should OpenTelemetry eliminate `GetDefaultHandle` and instead specify that cumulative, gauge, and measure metric instruments implement `Add()`, `Set()`, and `Record()` with the same interpretation?
+
+The argument against this is that metric instruments are meant to be pure API objects, they are not constructed through an SDK. Therefore, the default Meter (SDK) will have to be located from the context, meaning there is a question about whether this is as efficient as storing a re-usable handle for the default case.  For metric instruments with no required keys, this will be a real question: what is the benefit of a handle of it specifies no information other than the SDK?
+
+If we eliminate `GetDefaultHandle()`, the SDK may keep a map of metric instrument to default handle on its own.
```
bogdandrutu: DefaultHandle in my understanding is a GetHandle(List<Null> /*null for all the values*). So an argument can be made to remove this. Also for metrics without predefined labels will need to call `GetHandle(empty_list)` which may require an unnecessary allocation of the empty_list object.

I want mostly to make sure that by DefaultHandle we understand the same thing.

text/0003-measure-metric-type.md
```diff
@@ -2,50 +2,168 @@
...
+- `RecordBatch` support for cumulatives and gauges will be natural for SDKs that act as forwarders for metric events . The natural implementation for `Add()` and `Set()` methods will be `RecordBatch` with a single event.
+- Likewise, it is simple for an SDK that acts as an aggregator (not a forwarder) to redirect `Add()` and `Set()` APIs to the handle-specific `Add()` and `Set()` methods; while the SDK, as the implementation, still may (not must) treat these cumulative and gauge updates as atomic.
+
+Arguments against batch recording for all metric instruments:
 
-The same discussion can be had for the `MeasurementBatch` type described here.  It can be declared with an ordered list of metrics, then the `Record` API takes only an ordered list of numbers.  Alternatively, and less prone to misuse, the `MeasurementBatch.Record` API could be declared with a list of metric:number pairs.
+- The `Record` in `RecordBatch` suggests it is to be applied to measure metrics.  This is due to measure metrics being the most general-purpose of metric instruments.
```
bogdandrutu: Also the measure metric is the most expensive to calculate so an implementation may defer that aggregation (especially quantiles require locking), in the sum case that is just an atomic fetch_add.

songy23:
https://circleci.com/gh/open-telemetry/oteps/115?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
```bash
./text/0003-measure-metric-type.md:5:2: "Foreward" is a misspelling of "Foreword"
./text/0003-measure-metric-type.md:113:294: "whic" is a misspelling of "which"
```
Please fix.

text/0003-measure-metric-type.md
jmacd: Yes, thanks. Also, I thought some more and believe that this non-negative case should be the default case. This left me wondering about a different name that reflects the opposite property, that the cumulative can go up and down. The phrase "Up-Down Counter" exists, but I also like "Bidirectional" to describe this. How do you feel about these changes?

text/0003-measure-metric-type.md
jmacd: There's now a separate RFC about the `Observer`, see https://github.com/open-telemetry/oteps/pull/39.

But, the pattern is reversed -- the Observer will be called to get the current value when the SDK wants it. It will be up to the metric exporters to decide how to handle zero-updates in the output. My preference is to report 0-change events on gauges, it makes a difference in some products.

text/0003-measure-metric-type.md
jmacd: These are P0 and P100 of the distribution, and they're the only two quantiles that can be computed in O(1) space. I realize there are problems in aggregating these, but I've used this sort of summary to get an idea of the worst-case latency in a high-throughput section of code.

text/0003-measure-metric-type.md
jmacd: I rewrote this section to say that enabled is the default and `Disabled` is an option.

text/0003-measure-metric-type.md
jmacd: It's also part of the CloudWatch metrics data model.
https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_concepts.html#Statistic

text/0003-measure-metric-type.md
jmacd: Yes, that's my understanding.

text/0003-measure-metric-type.md
jmacd: The problem is that I can see both cases being useful. I recognize that "raw statistics" in OC had a 3-tuple of (metric, value, pre-defined labels) as the measurement. I wanted to be able to use a 2-tuple for `RecordBatch` because I thought the purpose was to provide atomicity. 

We discussed whether it was possible to have per-call-site label values in the metrics API, and you requested treating those as labels derived from the context. This was agreed--and now a related point of confusion has come up about additional label values in the handle, which I favor.

This topic appears to be whether we can avoid the cost of allocating handles through the SDK for cases that are batch measurements that will only be used once. I see the benefit, but it seems to prevent recording a batch of metrics using a handle.

I could see supporting both options.  That's somewhat unsatisfying.

For the gRPC configuration prescribed in OC, I take it that `Handles` are used for the cumulative and gauge metrics, with pre-defined labels, but the raw measurements did not use pre-defined labels. But gRPC _knows_ the pre-defined labels here. I presume this use of pre-defined labels for cumulative and gauges is because these are pre-aggregated. Therefore, measure metrics do not benefit, the argument goes, from pre-defined labels as much.

Can we avoid this by:
(1) requiring labels supplied in a `RecordBatch` to be passed via context
(2) allowing `RecordBatch` to take either a metric instrument or a handle
(3) if a handle, the additional labels are passed by context; if a metric, both required and additional labels are passed by context.

At some point, thus blurs the distinction between call-site and per-handle labels...

text/0000-metric-handles.md
```diff
@@ -0,0 +1,65 @@
...
+
+`GetHandle` specifies two calling patterns that may be supported: one with ordered label values and one without.  The facility for ordered label values is provided as a potential optimization that facilitates a simple lookup for the SDK; in this form, the API is permitted to thrown an exception or return an error when there is a mismatch in the arguments to `GetHandle`.  When label values are accepted in any order, some SDKs may perform an expensive lookup to find an existing metrics handle, but they must not throw exceptions.
+
+`GetHandle` and `GetDefaultHandle` support additional label values not required in the definition of the metric instrument.  These optional labels act the same as pre-defined labels in the low-level metrics data representation, only that they are not required.  Some SDKs may elect to use additional label values as the "attachment" data on metrics.
```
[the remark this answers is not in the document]
jmacd: We discussed two cases: 
(1) how a user sometimes wants to add additional labels at the individual call site, that we might treat these using context propagation instead
(2) how a user sometimes wants to include additional labels in the handle, as a matter of shortening the program code for metrics that are used more than once in code. I don't think we pinned this down. We agreed that what I'm now calling "required keys", which are declared in the metric, that these are always defined in a handle (including potentially unspecified). Here, "Additional Labels" are associated with the handle but cannot be used for pre-aggregation because they are not required. For a metrics data exporter, they can be ignored. For a tracing system or diagnostic-event exporter, these would be additional data.

I've posted an open question at the bottom. Can we use additional labels to capture the `Attachment` concept?

text/0003-measure-metric-type.md
bogdandrutu: The way how cloudwatch makes use of this is by computing what we called "interval stats". They require you to have a period and always send the metric that represents the statistic for this period of time (kind of already rated) so that's how they can do some aggregation :)

bogdandrutu: So I can see this as one of the aggregation "Statistic over time", then min/max are useful.

text/0000-metric-handles.md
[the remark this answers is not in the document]
bogdandrutu: Can we start without the support of these extra labels for the moment? Adding a new API is always possible and backwards compatible. Trying to start a bit small :)

text/0000-metric-handles.md
```diff
@@ -0,0 +1,76 @@
...
+
+Specify the behavior of the Metrics API `Handle` type, for efficient repeated-use of metric instruments.
+
+## Motivation
+
+The specification currently names this concept `TimeSeries`, the object returned by `GetOrCreateTimeseries`, which supports binding a metric to a pre-defined set of labels for repeated use.  This proposal renames these `Handle` and `GetHandle`, respectively, and adds further detail to the API specification for handles.  
+
+## Explanation
+
+The `TimeSeries` is renamed to `Handle` as the former name suggests an implementation, not an API concept. `Handle`, we feel, is more descriptive of the intended use.  Likewise with `GetOrCreateTimeSeries` to `GetHandle` and `GetDefaultTimeSeries` to `GetDefaultHandle`, these names suggest an implementation and not the intended use.	Applications are encouraged to re-use metric handles for efficiency.
```
reyang:
```suggestion
The `TimeSeries` is renamed to `Handle` as the former name suggests an implementation, not an API concept. `Handle`, we feel, is more descriptive of the intended use.  Likewise with `GetOrCreateTimeSeries` to `GetHandle` and `GetDefaultTimeSeries` to `GetDefaultHandle`, these names suggest an implementation and not the intended use. 
 Applications are encouraged to re-use metric handles for efficiency.
```

text/0000-metric-handles.md
```diff
@@ -0,0 +1,76 @@
+# Metric Handle API specification
+
+**Status:** `proposed`
+
+Specify the behavior of the Metrics API `Handle` type, for efficient repeated-use of metric instruments.
+
+## Motivation
+
+The specification currently names this concept `TimeSeries`, the object returned by `GetOrCreateTimeseries`, which supports binding a metric to a pre-defined set of labels for repeated use.  This proposal renames these `Handle` and `GetHandle`, respectively, and adds further detail to the API specification for handles.
```
reyang: Having `gauge.GetHandle(...)` seems to be okay.
Having a class/type named `Handle` could be confusing.

jmacd:
I will revise this RFC again tomorrow with some of the outcome of today's metrics meeting. For now, I've allocated number 0007-metric-handle, making it precede 0008-metric-observer in PR#39.

text/0000-metric-handles.md
jmacd: In the Go prototype, there is indeed a `Handle` type but it's for internal use. The types returned by the `GetHandle()` methods in Go PR#100 are like `Float64GaugeHandle`, `Int64CumulativeHandle`. Is this sufficiently clear?

text/0000-metric-handles.md
jmacd: I applied this.

jmacd:
*Update*: It's been suggested that this PR is too long to be useful, and I would like to propose merging it in the current "proposed" state. I will immediately open another PR for further comments as I begin another round of updates regarding the LabelSet discussion yesterday.

@bogdandrutu

PR #34, opened 2019-08-28
add alpha/beta deprecation OTEP

lizthegrey:
We need to be able to deprecate OpenTelemetry beta SDKs, while not breaking beta APIs too often.

### DIFF
--- a/text/0033-deprecation-policy.md
+++ b/text/0033-deprecation-policy.md
@@ -1,0 +1,52 @@
+# API and SDK Compatibility & Deprecation Policy
+
+**Status:** `proposed`
+
+OpenTelemetry needs a clear policy on API and SDK compatibility for beta and major releases.
+
+## Motivation
+
+While we are rapidly evolving the SDK, exporter interfaces, and APIs, we need to balance usability with long term maintainability.
+Our API users need to have predictability in the interfaces they will use.
+We need to provide a predictable support timeframe for SDK releases, while enabling us to move onwards from decisions made early on.
+
+## Explanation
+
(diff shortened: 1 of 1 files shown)


tedsuo:
I really like the idea of timed releases, as a way to bundle up changes into a single release.

My concern for backwards compatibility support for the API during the beta is that it can be very tricky to accomplish. In my experience from OpenTracing, when we made important changes to handle new edge cases and requirements, those changes almost by definition did not play well with the existing API. So while we managed to provide backwards compatibility, it was very constraining and slowed the work down significantly, and meant that in some cases we could not pursue an idea which we thought was better. 

I'm thinking of core modeling issues here btw, like span and scope management, and not simpler issues like "should we expose tracestate or not?" Since this project is a reboot, I'm hoping we don't have many of these types of issues! But, given that we have yet to exercise any of these APIs vigorously, I'm concerned that different SIGs may still have some surprises awaiting them.

I suppose this is my question: When deprecation support is straightforwards, I think we should do it  during the beta. But if we discover a bigger or more fundamental issue in an API, I would want to be able to change things and make a hard break. How do you suggest we handle those situations?

Post-beta, strict backwards compatibility will be a critical feature for this project (it was for OpenTracing as well). After a release candidate becomes v1.0, we will need to have full API support for  existing instrumentation on the order of years - basically forever. But that's a separate issue! :)

lizthegrey:
> Post-beta, strict backwards compatibility will be a critical feature for this project (it was for OpenTracing as well). After a release candidate becomes v1.0, we will need to have full API support for existing instrumentation on the order of years - basically forever. But that's a separate issue! :)

Please take a look; have clarified that during alpha, we should feel free to make breaking changes, during beta we should be careful about but not tie our hands, and post-beta we need to maintain a seamless path.

PR #35, opened 2019-08-29
Add OpenTelemetry Protocol (OTLP) RFC

tigrannajaryan:
OpenTelemetry Protocol (OTLP) specification describes the encoding, transport and delivery mechanism of telemetry data between telemetry sources, intermediate nodes such as collectors and telemetry backends.


text/0000-opentelemetry-protocol.md
```diff
@@ -0,0 +1,513 @@
...
+
+#### Result Code
+
+`Export` response includes a `result_code` field, which indicates whether the server was able to successfully process the received data. Possible values for `result_code` field are:
+
+- `Success` - telemetry data is successfully processed by the server. If the server receives an empty request (a request that does not carry any telemetry data) the server SHOULD respond with `Success`.
+
+- `FailedNoneRetryable` - processing of telemetry data failed. The client MUST NOT retry sending the same telemetry data. The telemetry data MUST be dropped. This for example can happen when the request contains bad data and cannot be deserialized or otherwise processed by the server. The client SHOULD maintain a counter of such dropped data.
```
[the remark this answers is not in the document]
tigrannajaryan: Typo, thanks for catching, should be FailedNotRetryable.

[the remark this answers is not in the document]
yurishkuro: also in the proto definition

text/0000-opentelemetry-protocol.md
```diff
@@ -0,0 +1,513 @@
...
+
+The ID field of the Export request and response is not used in Unary mode since the underlying transport already supports acknowledgements and one-to-one mapping of requests and responses.
+
+#### Streaming
+
+In Streaming mode the client may send multiple `Export` requests without waiting for the response to the earlier sent requests, essentially creating a pipeline of requests that are currently in flight that are not acknowledged.
+
+`Export` request includes an ID field that is unique for each request. The value of ID field is a sequential uint64 number generated by the client. The server MUST respond and acknowledge each `Export` request with a corresponding response that includes the ID of the request. Note that the order of IDs in a sequence of responses coming from the server may be different than the order of IDs in a sequence of earlier sent requests coming from the client. The client MUST be able to correctly process such out-of-order responses.
+
```
rnburn: Why do you use generated IDs and support out-of-order acking like this?

Why not use a design that's more similar to how TCP acks with the sequence number of the last byte received and have the server send acks with the number of the last span it read from the stream? That would
* Not impose artificial batching and allow for allow for clients to open multiple streams and redirect spans/metrics at a granular level if it exceeds the throughput of a single stream.
* Simplify the client code since there's no need to maintain an indexed structure of the batches sent.
* Reduce the amount of data exchange (would require fewer ack responses from the server).
Is there a use case for out-of-order acking?

tigrannajaryan: @rnburn you are right that we could do acks similar to TCP. However acks are not the only reason that we want to send responses. Responses also carry the result code and we need one result code per request sent. We either have to send a response for each request or batch responses and send at once but that both complicates the implementation and requires the client wait for a batched response. This is unneccessary and it is much simpler to send one response per request.

To address your specific concerns:
>Not impose artificial batching

I am not sure why you call batching "artificial". There are several reasons why batching can be a good idea. Batching for the purpose of reducing number of requests is only one of them (another benefit is for example better compression ratio because larger data usually compresses more).

> and allow for allow for clients to open multiple streams and redirect spans/metrics at a granular level if it exceeds the throughput of a single stream.

Nothing in this proposal prevents implementations to open multiple streams if needed. In fact my reference implementation does that during stream re-openinig transition period. Implementations can easily use multiple streams using described protocol.

>Simplify the client code since there's no need to maintain an indexed structure of the batches sent.

True, however for clients where simplicity is important there is Unary mode. Streaming mode's primary focus is not implementation simplicity.

>Reduce the amount of data exchange (would require fewer ack responses from the server).

For typical use cases this is negligible. Telemetry data size is typically much larger than the tiny response.

Another benefit of sending acks immediately is that it allows the client to earlier free request data that is kept in-memory.

>Is there a use case for out-of-order acking?

Yes, there is. For example server implementation can be done using thread pools (task pools) that consume received requests, process concurrently and send a response back. This normally means the order of request processing _completion_ is undefined. To use sequential ack approach that you suggest one needs to perform ordering of completions or corresponding responses which is additional complication.

In general what you suggest is probably another approach that could be researched, and it will likely result in different tradeoffs (more complex server, more simple client). I beleive my proposal hits the sweet spot that I wanted to hit based on my previous experience with OpenCensus.

text/0000-opentelemetry-protocol.md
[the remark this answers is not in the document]
tigrannajaryan: Looks like I made the same typo in the spec: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-exporter.md
and it also carried over to the Java implementation: https://github.com/open-telemetry/opentelemetry-java/blob/78e4e10888ad9965c334112e52e0e0b0d04f8497/sdk/src/main/java/io/opentelemetry/sdk/trace/export/SpanExporter.java#L46

I will submit patches to fix everywhere.

text/0000-opentelemetry-protocol.md
rnburn: Why do we need a concept of requests for streaming at all? 

Here's how I'd see a streaming protocol working: 
1. As spans/metrics are finished, they're serialized, framed, and added into a buffer.
2. When a reporter is ready to sent data to a collection endpoint, it uses writev with the already framed and serialized spans to write the data directly to the collection endpoint's socket.

There's no need to bundle spans into requests and the recorder is low overhead because there's no serialization cost. If you wanted to add acking to this approach, you could have the collection endpoint regularly send back ack responses, at points of its choosing, denoting how far in the stream it's processed data.

> Nothing in this proposal prevents implementations to open multiple streams if needed. In fact my reference implementation does that during stream re-openinig transition period. Implementations can easily use multiple streams using described protocol.

I don't claim that it can't use multiple streams; however, you lose the ability to granularly redirect data. 

Suppose I have multiple span streams, a buffer of 1k spans, and I write the data to a non-blocking socket. If the non-blocking call only wrote 200 spans, then I can immediately start writing the remaining full spans to other streams. If, instead, I have to write batches of spans, then I lose this ability to pick up writing where last write call finished. 

Being able to granularly load balance like this allows throughput of the streams to naturally adjust to the rate that the endpoint is processing the data.

Furthermore, if all the sockets are blocked, I can block with epoll until one of the stream's sockets is writable again, and pick up writing with not just the remaining data, but also include any additional data that's been added into to the buffer between the last write call -- there's no need to batch the new data first.

> In general what you suggest is probably another approach that could be researched, and it will likely result in different tradeoffs (more complex server, more simple client). I beleive my proposal hits the sweet spot that I wanted to hit based on my previous experience with OpenCensus.

Maybe there are some cases where batched request/response streaming makes sense. But IMO eagerly serializing (i.e. writing the span/metrics serialization as methods on it are called) combined with streaming data unbatched over a raw socket is a definite sweet spot that leads to performant, high throughput tracers, and it should be supported.

tigrannajaryan: @rnburn 
>Why do we need a concept of requests for streaming at all?

That's how gRPC conceptually works and I suggest to use gRPC as a transport.

Other transports are certainly possible. My selection criteria was ease of implementation, ubiquity of client and server libraries, proven robustness, extensibility and out-of-the-box availability of necessary functionality and gRPC ticked the boxes.

If you would like to propose a different transport and a different protocol feel free to submit an alternate RFC. FYI, these are the design goals and requirements for a general-purpose telemetry data protocol that have been previously discussed and agreed on:
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/proto-design-goals.md
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/proto-requirements.md

I know you earlier suggested https://github.com/open-telemetry/opentelemetry-cpp/issues/2 and that looks interesting to me as a specialized approach but I believe it doesn't satisfy some of the requirements for general-purpose protocol. If you would like to work further on that and propose as a protocol I'll be glad to review and comment.

text/0000-opentelemetry-protocol.md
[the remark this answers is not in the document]
tigrannajaryan: Fixed in this proposal and submitted patches for spec and Java lib.

tigrannajaryan:
Admins, please assign a number to this RFC.

text/0000-opentelemetry-protocol.md
```diff
@@ -0,0 +1,513 @@
...
+![Unary](images/otlp-unary.png)
+
+Unary mode is recommended when simplicity of implementation is desirable and when the client and the server are connected via very low-latency network, such as for example when the client is an instrumented application and the server is a OpenTelemetry Service running as a local daemon.
+
+The maximum achievable throughput in unary mode is `max_request_size / (network_latency + server_response_time)`. For example if the request can contain at most 100 spans, network roundtrip latency is 200ms and server response time is 300 ms, then the maximum achievable throughput is `100 spans / (200ms+300ms)` or 200 spans per second. It is easy to see that in high latency networks or when the server response time is high it is difficult to achieve good throughput unless requests are very big (even when client, server and network throughput potentially allow much bigger processing rate).
+
+The ID field of the Export request and response is not used in Unary mode since the underlying transport already supports acknowledgements and one-to-one mapping of requests and responses.
+
+#### Streaming
```
akehlenbeck: I'd like to ask the "why streaming?" question in a slightly different way than rnburn did. What benefits are you getting from a streaming model that you wouldn't also get from a unary model that allows more than one in-flight request?

There are a variety of drawbacks to the pattern of "implement my own request/response pairing over a single streaming RPC." Some of them you've listed in the doc already.

* You have to implement your own bookkeeping on client and server to pair up requests and responses.

* Things like metrics and tracing at the level of the RPC framework become much less useful in a streaming model. For example, in a unary RPC model, I can look at the normal request counters and error counters and latency metrics and so forth to understand what's going on. When multiple logical RPCs are multiplexed over a single stream, that information is lost and the client and server have to re-implement all of those metrics by hand (with non-standard names, etc).

* I don't really trust loadbalancing of streaming RPCs. I know you ran some tests, and you rightfully talk in the doc about the need to periodically close streams, but my experience is that over time you end up turning that knob so far down that you're more or less approximating unary RPCs anyways.

I believe that GRPC (unlike the Google systems that it was based on) has connection-level flow control, which historically was the main reason to prefer streaming RPCs over unary RPCs. So it's not entirely clear to me what benefit is coming from the approach proposed here. Thoughts?

In my experience, the two motivations where streaming RPCs are useful are:

* You need RPC-level flow control (not just connection-level flow control).

* You need to pin some kind of session state on a particular server, i.e. you have several different request messages that *must* be sent to the same server process (and you don't want to do you own bookkeeping to make that happen).

It doesn't seem like either of those applies here though.

tigrannajaryan: >I'd like to ask the "why streaming?" question in a slightly different way than rnburn did. What benefits are you getting from a streaming model that you wouldn't also get from a unary model that allows more than one in-flight request?

@akehlenbeck thank you for thoughful comments.

I did consider doing Unary with concurrent requests and I actually have a simple implementation here https://github.com/tigrannajaryan/exp-otelproto/tree/master/protoimpls/grpc_unary_async (it is part of the benchmarks).

2 reasons why I chose streaming after comparing it with concurrent Unary:

1. Ability to optimize the Resource field and avoid sending it with every request if it doesn't change during streaming session. This optimization is not specified in this document, but I intend to add it. Unfortunately it is hard to quantify the benefits of the optimization without knowing the ratio and sizes of Resources vs Spans sent in a single session however this capability was part of OpenCensus protocol and I wanted to keep this optimization possibility. For this to be possible one need to be able to send a sequence of requests to the same server behind load balancer and streaming allows to do that.

2. Streaming consumes slightly less CPU than Unary per span sent (about 3% in benchmarks). This is small but nevertheless was visible in the benchmarks of particular implementation.

I want to note that my preference for streaming is not a strong opinion. I am open to re-considering and dropping streaming in favour of concurrent unary, especially if I hear voices telling that Resource optimization is unneeded.

>I don't really trust loadbalancing of streaming RPCs.

Can you expand on this? I am interested in knowing what were the problems that you observed.

>I believe that GRPC (unlike the Google systems that it was based on) has connection-level flow control, which historically was the main reason to prefer streaming RPCs over unary RPCs. So it's not entirely clear to me what benefit is coming from the approach proposed here. Thoughts?

Ability for the server to send acknowledgements out of order is one benefit with using own flow control rather than rely on stream's built-in flow control. Compared to unary obviously there is no flow control benefits, unary already allows out of order responses.

>In my experience, the two motivations where streaming RPCs are useful are:
>...
>You need to pin some kind of session state on a particular server, i.e. you have several different request messages that must be sent to the same server process (and you don't want to do you own bookkeeping to make that happen).

This one does apply in our case, that's how we would do Resource optimization.

text/0000-opentelemetry-protocol.md
```diff
@@ -0,0 +1,513 @@
...
+
+Unary mode is simpler to implement and provides good performance in sub-millisecond latency networks, at the cost of worse throughput in high latency networks.
+
+Streaming mode requires somewhat more complicated implementation but provides significantly better throughput on high latency networks. 
+
+Clients are recommended to implemented both modes but may choose to implement only one.
+
+Servers MUST implement both modes. Clients operating in different modes can connect to the same server.
+
+#### Unary
```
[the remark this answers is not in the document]
tigrannajaryan: My reply to your other comment covers this too.

tigrannajaryan:
Admins, can you please assign an RFC number to this? Also, based on the RFC process shouldn't this be merged in `proposed` status and then we continue the discussion?

tigrannajaryan:
Reviewers, please have another look. I changed wording around Unary to remove the restriction that only Unary request may be in flight.

tigrannajaryan:
UPDATE:

Further experiments have shown that Streaming implementation
results in significantly more complex code. The performance
benefits of Streaming mode were only visible in very
small batches and are not applicable to real life scenarios
when under high load we aim to have large batches.

In addition, one of the potential benefits that we could
get with Streaming - the ability to optimize and avoid
repetition of Resource does not seem to have much real
life usage.

As a result Streaming mode was removed from the RFC and
Unary mode is the only mode that is now part of the spec.

Please have another look.

text/0035-opentelemetry-protocol.md
```diff
@@ -0,0 +1,425 @@
+# OpenTelemetry Protocol Specification
+
+_Author: Tigran Najaryan, Omnition Inc._
```
bogdandrutu: Usually we don't use _Author.

text/0035-opentelemetry-protocol.md
```diff
@@ -0,0 +1,425 @@
...
+
+Examples of a Client are instrumented applications or sending side of telemetry collectors, examples of Servers are telemetry backends or receiving side of telemetry collectors (so a Collector is typically both a Client and a Server depending on which side you look from).
+
+Both the Client and the Server are also a `Node`. This term is used in the document when referring to either one.
+
+## Acknowledgements
+
+Special thanks to Owais Lone who helped to conduct experiments with Load Balancers, to Paulo Janotti, Bogdan Drutu and Yuri Shkuro for thoughtful discussions around the protocol.
+
+## Author's Address
```
bogdandrutu: Also remove this.

tigrannajaryan:
Reviewers, all questions and comments on this PR have been answered/address. Please have another look. We need to move forward with this, the protocol planned to be part of Alpha 0.3 milestone and the earlier we have an approved spec the better chances to have an implementation ready by 0.3 as well.

akehlenbeck:
I'm not an official reviewer, but I'm glad to see that the concurrent-unary-model was sufficient and that the streaming model has been removed. LGTM. Thanks for the prototyping you did around that issue Tigran.

text/0035-opentelemetry-protocol.md
```diff
@@ -0,0 +1,425 @@
...
+
+OTLP is a request/response style protocols: the clients send requests, the server replies with corresponding responses. This document defines one requests and response type: `Export`.
+
+### Export Request and Response
+
+After establishing the underlying transport the client starts sending telemetry data using `Export` requests.  The client continuously sends a sequence of `Export` requests to the server and expects to receive a response to each request:
+
+![Request-Response](images/otlp-request-response.png)
+
+_Note: this protocol is concerned with reliability of delivery between one pair of client/server nodes and aims to ensure that no data is lost in-transit between the client and the server. Many telemetry collection systems have intermediary nodes that the data must travel across until reaching the final destination (e.g. application -> agent -> collector -> backend). End-to-end delivery guarantees in such systems is outside of the scope of OTLP. The acknowledgements described in this protocol happen between a single client/server pair and do not span intermediary nodes in multi-hop delivery paths._
```
yurishkuro: Perhaps worth clarifying that it doesn't mean that those chained acks cannot happen, eg if the "server" is just an agent/sidecar that enriches/translates messages and forwards them further, can it respond to the client based on whether the downstream collector acked the message?

text/0035-opentelemetry-protocol.md
```diff
@@ -0,0 +1,425 @@
...
+
+For gRPC transport OTLP uses Unary RPC to send export requests and receives responses.
+
+After sending the request the client MAY wait until the response is received from the server. In that case there will be at most only one request in flight that is not yet acknowledged by the server.
+
+![Unary](images/otlp-sequential.png)
+
+Sequential operation is recommended when simplicity of implementation is desirable and when the client and the server are connected via very low-latency network, such as for example when the client is an instrumented application and the server is a OpenTelemetry Service running as a local daemon.
+
+The implementations that need to achieve high throughput SHOULD support concurrent Unary calls to achieve higher throughput. The client SHOULD send new requests without waiting for the response to the earlier sent requests, essentially creating a pipeline of requests that are currently in flight that are not acknowledged.
```
yurishkuro: Does this imply multiple connections to the server?

text/0035-opentelemetry-protocol.md
yurishkuro: It's not a bad idea to recognize authors and contributors this way (e.g. @akehlenbeck whose comments changed the OTEP in a significant way from the first version)

text/0035-opentelemetry-protocol.md
SergeyKanzhelev: I'd widen this - does "ack" means delivered and saved in any sort of persistence storage or can be replied when messages saved in-memory and can be lost on node restart. Or not defined?

text/0035-opentelemetry-protocol.md
```diff
@@ -0,0 +1,425 @@
...
+
+![Streaming](images/otlp-concurrent.png)
```
SergeyKanzhelev: this picture implies ordered responses. Is it intentional or just to not over-complicate the picture?

text/0035-opentelemetry-protocol.md
SergeyKanzhelev: ... and if so - to the same server?

text/0035-opentelemetry-protocol.md
```diff
@@ -0,0 +1,425 @@
...
+
+The number of concurrent requests SHOULD be configurable.
+
+The maximum achievable throughput is `max_concurrent_requests * max_request_size / (network_latency + server_response_time)`. For example if the request can contain at most 100 spans, network roundtrip latency is 200ms and server response time is 300 ms, then the maximum achievable throughput with one concurrent request is `100 spans / (200ms+300ms)` or 200 spans per second. It is easy to see that in high latency networks or when the server response time is high to achieve good throughput the requests need to be very big or a lot concurrent requests must be done.
+
+If the client is shutting down (e.g. when the containing process wants to exit) the client will optionally wait until all pending acknowledgements are received or until an implementation specific timeout expires. This ensures reliable delivery of telemetry data. The client implementation SHOULD expose an option to turn on and off the waiting during shutdown.
+
+If the client is unable to deliver a certain request (e.g. a timer expired while waiting for acknowledgements) the client SHOULD record the fact that the data was not delivered.
```
SergeyKanzhelev: I'd suggest to be explicit here. This is retry-able error, correct?

text/0035-opentelemetry-protocol.md
```diff
@@ -0,0 +1,425 @@
...
+
+#### Result Code
+
+`Export` response includes a `result_code` field, which indicates whether the server was able to successfully process the received data. Possible values for `result_code` field are:
+
+- `Success` - telemetry data is successfully processed by the server. If the server receives an empty request (a request that does not carry any telemetry data) the server SHOULD respond with `Success`.
+
+- `FailedNotRetryable` - processing of telemetry data failed. The client MUST NOT retry sending the same telemetry data. The telemetry data MUST be dropped. This for example can happen when the request contains bad data and cannot be deserialized or otherwise processed by the server. The client SHOULD maintain a counter of such dropped data.
+
+- `FailedRetryable` - processing of telemetry data failed. The client SHOULD record the error and may retry exporting the same data. This can happen when the server is temporarily unable to process the data.
```
SergeyKanzhelev: it is a good practice to also reply the `retry-after` parameter. For back-off scenarios. Do you think it's worth adding from the beginning or you want to introduce it later?

SergeyKanzhelev: Oh, I should have read further before commenting. Perhaps the only comment here is to mention the next section in this list

SergeyKanzhelev: and perhaps tell that client in case of `FailedRetryable` can retry immediately

text/0035-opentelemetry-protocol.md
```diff
@@ -0,0 +1,425 @@
...
+// Clients should wait until `retry_delay` amount of time has passed since
+// receiving the error response before retrying.  If retrying requests also
+// fail, clients should use an exponential backoff scheme to gradually increase
+// the delay between retries based on `retry_delay`, until either a maximum
+// number of retires have been reached or a maximum retry delay cap has been
+// reached.
+```
+
+The value of `retry_delay` is determined by the server and is implementation dependant. The server SHOULD choose a `retry_delay` value that is big enough to give the server time to recover, yet is not too big to cause the client to drop data while it is throttled.
+
```
SergeyKanzhelev: Do you want to cover the topic of which data should be dropped first? Older or newer.

text/0035-opentelemetry-protocol.md
```diff
@@ -0,0 +1,425 @@
...
+### Multi-Destination Exporting
+
+When the telemetry data from one client must be sent to more than one destination server there is an additional complication that must be accounted for. When one of the servers acknowledges the data and the other server does not (yet) acknowledges the client needs to make a decision about how to move forward.
+
+In such situation the the client SHOULD implement queuing, acknowledgement handling and retrying logic per destination. This ensures that servers do not block each other. The queues SHOULD reference shared, immutable data to be sent, thus minimizing the memory overhead caused by having multiple queues.
+
+![Multi-Destination Exporting](images/otlp-multi-destination.png)
+
+This ensures that all destination servers receive the data regardless of their speed of reception (within the available limits imposed by the size of the client-side queue).
```
SergeyKanzhelev: if this server is a "proxy" server - how the behavior should look like in terms of responding back to the originator? Do you want to cover this or leave to implementation?

text/0035-opentelemetry-protocol.md
```diff
@@ -0,0 +1,425 @@
...
+    // processing of telemetry data failed. The client MUST NOT retry 
+    // sending the same telemetry data. The telemetry data MUST be dropped. 
+    // This for example can happen when the request contains bad data and 
+    // cannot be deserialized or otherwise processed by the server.
+    FailedNotRetryable = 1;
+    
+    // Processing of telemetry data failed. The client SHOULD record the 
+    // error and may retry exporting the same data after some time. This 
+    // for example can happen when the server is overloaded.
+    FailedRetryable = 2;
```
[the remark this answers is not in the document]
tigrannajaryan: Unavailable is returned as a gRPC error with additional status that can include retry duration hint. See the Go example code in Throttling section.

text/0035-opentelemetry-protocol.md
```diff
@@ -0,0 +1,425 @@
...
+
+This specification defines how OTLP is implemented over [gRPC](https://grpc.io/) and specifies corresponding [Protocol Buffers](https://developers.google.com/protocol-buffers/docs/overview) schema. Future extensions to OTLP may define implementations over other transports. For details of gRPC service definition see section [gRPC Transport](#grpc-transport).
+
+OTLP is a request/response style protocols: the clients send requests, the server replies with corresponding responses. This document defines one requests and response type: `Export`.
+
+### Export Request and Response
+
+After establishing the underlying transport the client starts sending telemetry data using `Export` requests.  The client continuously sends a sequence of `Export` requests to the server and expects to receive a response to each request:
+
+![Request-Response](images/otlp-request-response.png)
```
SergeyKanzhelev: in this picture do you assume that span is self contained or that it can have a reference to "resource" transmitted separately?

text/0035-opentelemetry-protocol.md
tigrannajaryan: @yurishkuro Technically yes, nothing prevents the intermediary node to only ack after downstream acked.
It is a choice of particular _usage_ of the protocol to define what it means that the data is acknowledged. It may mean that:
- The data is received.
- The data is received and successfully decoded.
- The data is received, decoded and persisted.
- The data is received and forwarded.
etc.
The protocol does not define this, that definition belongs to the abstraction level which uses the protocol. (@SergeyKanzhelev I believe this answers your question too).

The only reason I have this note is to make it clear that acknowledgement response travels between a single connected gRPC client/server pair and OTLP does not attempt to implement end-to-end reliable data delivery on its own (but it certainly can be used to achieve that).

text/0035-opentelemetry-protocol.md
```diff
@@ -0,0 +1,425 @@
...
+
+In such situation the the client SHOULD implement queuing, acknowledgement handling and retrying logic per destination. This ensures that servers do not block each other. The queues SHOULD reference shared, immutable data to be sent, thus minimizing the memory overhead caused by having multiple queues.
+
+![Multi-Destination Exporting](images/otlp-multi-destination.png)
+
+This ensures that all destination servers receive the data regardless of their speed of reception (within the available limits imposed by the size of the client-side queue).
+
+## Trade-offs and mitigations
+
+### Request Acknowledgements
```
SergeyKanzhelev: another thing is there is no support for partial success. Perhaps can mention as a tradeoff

text/0035-opentelemetry-protocol.md
tigrannajaryan: @yurishkuro  It may be one or may be multiple connections - gRPC abstracts this away by providing an RPC interface.
gRPC multiplexes unary calls over a single or multiple connections. This can be a good read on the topic: https://grpc.io/blog/grpc_on_http2/

@SergeyKanzhelev yes, always to the same server.

text/0035-opentelemetry-protocol.md
tigrannajaryan: I am fine either way, please feel free to make an "executive TC decision" and I will apply it :-)

text/0035-opentelemetry-protocol.md
tigrannajaryan: It is self-contained. See protobuf definition below. Each request carries a Resource and a list of Spans. The ability to transmit Resource separately and optimize it away for re-transmissions is removed because with unary calls you are not guaranteed that if you have multiple server instances behind a load balancer sequentual calls will go to any particular server instance.

text/0035-opentelemetry-protocol.md
tigrannajaryan: Just to keep it simple. The responses may arrive out of order but it has no implications on the implementation. The protocol does not rely in any way in the ordering of responses.

text/0035-opentelemetry-protocol.md
tigrannajaryan: Correct, when the server wants to ask the client to backoff it returns `Unavailable` error and includes `RetryDelay` data. This approach is borrowed from Google APIs (https://github.com/googleapis/googleapis/blob/c6e62c7e5e61e6dae7fdc3bc3de81f60e6a9445c/google/rpc/error_details.proto#L40) and semantically is very similar to HTTP 503 Service Unavailable with Retry-After response header.

I will add a note that `FailedRetryable` can retry immediately.

text/0035-opentelemetry-protocol.md
tigrannajaryan: It is retryable in the sense that the server did not tell that it is not retryable. However if the client is shutting down there is no way to retry it unless the client is revived from shutdown state (not defined by this doc) or a new client is created.
I am not sure if it is worth touching these topics in this doc.

text/0035-opentelemetry-protocol.md
tigrannajaryan: I believe that decision does not belong to OTLP or even OTLP implementation. Users of OTLP need to make that decision depending on what makes most sense for the particular use case.

text/0035-opentelemetry-protocol.md
tigrannajaryan: Good point. I think it was briefly discussed in this PR a while ago. I will find the comment and add a note.

text/0035-opentelemetry-protocol.md
tigrannajaryan: >yes, always to the same server.

Correction: this may be different severs if they are behind load balancers. gRPC load balancing typicaly works by routing unary requests to different servers, there is no stickiness for unary calls. This is expected and desirable behavior.

text/0035-opentelemetry-protocol.md
tigrannajaryan: I think this should not be a protocol concern. The client may not even drop the data (e.g. it may have a large persistent queue that allows it to keep the data for a very long time). I believe the best behavior is highly dependent on the use case and should be a end-to-end system design concern that is orthogonal to the protocol (protocol only facilitates possible designs by allow the server to communicate this value to the client).

text/0035-opentelemetry-protocol.md
tigrannajaryan: Added clarification.

text/0035-opentelemetry-protocol.md
tigrannajaryan: I added a note.

tigrannajaryan:
Please have one more look.

I believe the only remaining open question is whether to include Author and Acknowledgements sections. I am happy to go either way (my preference would be to keep Acknowledgements section since this document is a result of several people's work, not only my own, which won't be visible if we only look at the PR or commit history).

tigrannajaryan:
I changed the status to approved, so that this gets merged in approved status (when it gets merged). If I understand correctly this is the right flow for RFCs.

tigrannajaryan:
Admins, I think we have exhausted questions on this PR and have a principal agreement about the primary content - the protocol itself.

The only outstanding issue is a meta regarding including or not including an Author/Acknowledgements section that lists people who contributed to the RFC. Can you please make a call on this? I believe such a decision will be applicable to other RFCs (and I am happy with either approach).

text/0035-opentelemetry-protocol.md
```diff
@@ -0,0 +1,429 @@
...
+// the delay between retries based on `retry_delay`, until either a maximum
+// number of retires have been reached or a maximum retry delay cap has been
+// reached.
+```
+
+The value of `retry_delay` is determined by the server and is implementation dependant. The server SHOULD choose a `retry_delay` value that is big enough to give the server time to recover, yet is not too big to cause the client to drop data while it is throttled.
+
+#### gRPC Service Definition
+
+`Export` requests and responses are delivered using bidirectional gRPC streams.
```
yurishkuro: Probably "bidirectional grpc connections" ("streams" might sound confusing).

text/0035-opentelemetry-protocol.md
```diff
@@ -0,0 +1,429 @@
...
+
+The current version of OTLP is the initial version that describes mandatory capabilities only. Implementations of this specification SHOULD NOT attempt to detect the capabilities of their peers and should operate as defined in this document.
+
+## Prior Art, Alternatives and Future Possibilities
+
+We have considered using gRPC streaming instead of Unary RPC calls. This would require implementations to manually perform stream closing and opening periodically to be L7 Load Balancer friendly. Reference implementation using gRPC Streaming has shown that it results in significantly more complex and error prone code without significant benefits. Because of this Unary RPC was chosen.
+
+OTLP is an evolution of OpenCensus protocol based on the research and testing of its modifications in production at Omnition. The modifications include changes to data formats (see RFCNNNN), use of Unary PRC and backpressure signaling capability.
```
yurishkuro: What does RFCNNNN refer to?

text/0035-opentelemetry-protocol.md
tigrannajaryan: Thanks for catching, this is wrong, I will fix.

text/0035-opentelemetry-protocol.md
tigrannajaryan: This is intended to refer to https://github.com/open-telemetry/oteps/pull/59/files
I will fix the reference.

text/0035-opentelemetry-protocol.md
tigrannajaryan: Fixed.

text/0035-opentelemetry-protocol.md
tigrannajaryan: Fixed.

tigrannajaryan:
Admins, I need one more approval or more comments on this.

tigrannajaryan:
@yurishkuro thank you.

PR #36, opened 2019-08-30
Proposal to rename distributed context and add support for baggage.

rghetia:


text/0000-rename-distributed-ctx.md
```diff
@@ -0,0 +1,59 @@
...
+## Explanation
+
+We do following to
+- Rename distributed-context to TagMap.
+  - It is strictly used for observability.
+  - It is a write-only object from application and framework-integrations perspective.
+  - It is readable by SDKs.
+
+- Add Baggage support. Equivalent of what is there in [OpenTracing](https://opentracing.io/docs/overview/tags-logs-baggage/#baggage-items).
+  - It is more of a convenient functionality to provide compatibility with Open Tracing.
```
reyang:
```suggestion
  - It is more of a convenient functionality to provide compatibility with OpenTracing.
```

text/0000-rename-distributed-ctx.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Explanation
+
+We do following to
+- Rename distributed-context to TagMap.
+  - It is strictly used for observability.
+  - It is a write-only object from application and framework-integrations perspective.
+  - It is readable by SDKs.
+
+- Add Baggage support. Equivalent of what is there in [OpenTracing](https://opentracing.io/docs/overview/tags-logs-baggage/#baggage-items).
```
reyang: It is unclear when to put an item in Baggage versus TagMap.
Also, if something started as an item in TagMap, later there are needs to also put it Baggage, do we move the item or we duplicate the item?

A meta question - do we believe that OpenTelemetry should provide such general purpose `Baggage` so people can use it for things such as "authentication context"?

Oberon00: > It is unclear when to put an item in Baggage versus TagMap.

As I read this OTEP, that is quite clear: An application puts a value in the TagMap if it wants to have spans, metrics, etc tagged with that value. The baggage on the other hand is used when the value should not be attached to the telemetry data but if the application itself (or another service invoked by it) wants to read the value back, for application-specific purposes.

text/0000-rename-distributed-ctx.md
```diff
@@ -0,0 +1,59 @@
...
+1. For observability, key-value pairs SHOULD be write-only from application perspective. SDKs can read them
+   for example to apply them to measurements at the time of recording. For non-observability, 
+   key-value pairs SHOULD be writable/readable for application.
+2. For non-observability, it may require that the delivery is guaranteed and pairs are not dropped.
+3. Semantically two are different and mechanically it could be different as well.
+
+## Explanation
+
+We do following to
+- Rename distributed-context to TagMap.
```
songy23: Please also rename `DistributedContext.Entry` to `Tag`, key and value to `TagKey`, `TagValue`.

rghetia: sure. My intention is to restore TagMap as defined in [OC](https://github.com/census-instrumentation/opencensus-specs/blob/master/tags/TagMap.md)

text/0000-rename-distributed-ctx.md
rghetia: > Also, if something started as an item in TagMap, later there are needs to also put it Baggage, do we move the item or we duplicate the item?

No. Two are independent. Yes, there is duplication as mentioned in Trade-offs section but it is not worth to optimize it.

> 
> A meta question - do we believe that OpenTelemetry should provide such general purpose `Baggage` so people can use it for things such as "authentication context"?

I agree that from observability perspective, 'Baggage' doesn't belong in OpenTelemetry. However, from POV of opentracing compatibility we need to provide some solution. It could very well be in another repo/contrib package.

reyang: > No. Two are independent. Yes, there is duplication as mentioned in Trade-offs section but it is not worth to optimize it.

I see, now it is clear to me. Thanks.

> I agree that from observability perspective, 'Baggage' doesn't belong in OpenTelemetry. However, from POV of opentracing compatibility we need to provide some solution. It could very well be in another repo/contrib package.

Makes sense. If this for compatibility, do we encourage more people to adopt it, or we suggest people to move away from it?

yurishkuro:
I am very supportive of replacing "distributed context" with "baggage".

I am not supportive of revisiting the already made decision that "tags" are a bad name for telemetry-specific baggage. In OpenTracing tags had a completely different meaning.

I am also not convinced that we **need** two different sets of baggage. First, it raises the questions "what if I want key X either here or there or both places, why should this mean changing how the whole key/value pair is even named". I also don't see why telemetry dimensions cannot include application baggage - that's just a matter of configuring the filters/extractors for dimensions. With this proposed separation I need one filter type for application baggage and another type for telemetry baggage - highly confusing and unnecessary. Finally, the only rationale for the separation seems to be different guarantees of propagation, never drop app baggage, could drop telemetry baggage. This is a concern orthogonal to what what the baggage key/value pair is named. It can be addressed with an extra attribute, similar to TTL, like "criticality" or "priority", instead of two separately named concepts. The priority can be a numeric value that ranks baggage items (dups are ok), so that if something needs to be truncated, we truncate items with lower priority.

text/0000-rename-distributed-ctx.md
yurishkuro: Baggage is just context propagation. Context propagation is not a part of OpenTelemetry, it's the layer on top of which OpenTelemetry is built.

tedsuo:
Working on a context prop RFC (will be out today). In, I experimented with changing the name of `TagMap` to `Correlations`. It feels better, for a couple reasons:
* "Correlations" has clearer semantic meaning than "Tags." Since this is now a separate concept from Baggage, we can be more specific about the intent.
* The W3C propagation format will most likely be called Correlation-Context. So there is better naming alignment.
* Doesn't confuse OpenTracing users with an overloaded term.

@yurishkuro @rghetia how does that sound to you?

yurishkuro:
Correlations definitely better than TagMap. 

Only concern is how it will correspond to W3C Correlation-Context, it was not clear if it was intended as more general baggage rather than telemetry baggage.

rghetia:
Correlations sounds better than TagMap. So there will Correlations (for Telemetry) and baggage (for non-Telemetry)?

text/0000-rename-distributed-ctx.md
```diff
@@ -0,0 +1,59 @@
...
+
+Context Propagation is generalized to carry observability and non-observability contexts. This
+includes SpanContext, Distributed-context, Authentication-context, etc. Distributed-context
+carries key-value pairs for observability. The name Distributed-context is
+very confusing. Hence it should be changed to TagMap (or other better suggestion). 
+That is the first motivation.
+ 
+The second motivation is the restrict use of Distributed-context to observability. As it is
+written currently it does not specifically call out for using it only for non-observability purpose.
+However, it can be used to support 'Baggage' functionality for non-observability purpose as it is
```
toumorokoshi: how is "non-observability" defined? I would imagine here that if Baggage refers to the key-values pairs that carry with the trace, then those would contribute to observability since they help define the context of the request.

text/0000-rename-distributed-ctx.md
```diff
@@ -0,0 +1,59 @@
...
+OpenTracing.
+There are few reasons to have different mechanism to address observability and
+non-observability use cases.
+1. For observability, key-value pairs SHOULD be write-only from application perspective. SDKs can read them
```
toumorokoshi: Is write-only desirable? I figured distributedcontext entries would be something I could pipe into metrics as I see fit.

For example, if I used this to propagate my originating service, There's cases I would want metrics to also have the originating service, as I may want to split my metric by that dimension.

text/0000-rename-distributed-ctx.md
rghetia: non-observability should be defined as data that needs to propagate downstream along with the request for usage specific to application, testing, etc. It is not tied to whether the request is being traced/monitored or not.

text/0000-rename-distributed-ctx.md
toumorokoshi: I would vote for not including a non-telemetry related concept directly into opentelemetry. If Baggage is needed for opentracing having it be in contrib makes sense.

Will SaaS vendors need to use this as well? authentication or identification seems like a common concept.

text/0000-rename-distributed-ctx.md
rghetia: SDK can read it so it can be applied to metrics.

toumorokoshi:
I feel like there should be two RFCs: one to add explicit support for baggage and one for the renaming of DistributedContext. For me the renaming of DistributedContext seems like a good idea, and I'm not sure about baggage support.

(Just my two cents) I'm not seeing why "TagMap" is clearer than DistributedContext in terms of it's function to:

1. store several key-value pairs
2. have the context be propagated across service

TagMap certainly makes me think it has support for 1. The fact that it's propagated across process boundaries I would probably have to read somewhere. I'd say DistributedContext is clearer in that regard.

Also I don't have strong opinions on naming but I think we should have consistent semantic prefix or suffix to help the consumer understand that I can get key-value pairs for my telemetry from them. I have an issue out to try to rename "Resources" to "StaticContext", as an attempt to make it clear to the user that anything with "Context" contains key-value pairs that you can access: https://github.com/open-telemetry/opentelemetry-specification/issues/239

But I don't think Context does a great job of summarizing that behavior either.

text/0000-rename-distributed-ctx.md
toumorokoshi: Got it! might be good to add that definition somewhere, unless there is a definition floating in the spec somewhere.

rghetia:
> I feel like there should be two RFCs: one to add explicit support for baggage and one for the renaming of DistributedContext.

SG. If we get the agreement that telemetry and non-telemetry should be separate then I can split this into two RFCs.

text/0000-rename-distributed-ctx.md
toumorokoshi: When you say the sdk, are you referring to the exporters?
In the case of metrics, I think a lot of the dimensions will be specified and passed in during the invocation (e.g. during the creation of a TimeSeries (currently terminology) or Handler (new RFC-based terminology)). In that world, DistributedContext would need to be readable by the user:

```
# create the main counter
request_count = Counter()

# in the request, get the relevant time series and add the value.
time_series = request_count.get_time_series({"service": DistributedContextSingleton.get("service"))
time_series.add(1)
```

Although it is possible for the exporter to have access to the values, there's no easy mechanism to tell the exporter which labels matter. 

Also this behavior is also valuable because metrics will pre-aggregate, so they need values to determine the aggregation before the measurement even reaches the exporter.

rghetia:
Is it fair to say that we only need an issue to track renaming Distributed context to Correlation? 
For Baggage, we may need another RFC or simply an issue after concluding on #42.

bogdandrutu:
I am going to close this in favor of #66

PR #38, opened 2019-09-03
Proposal: Add a version resource

austinlparker:
This proposes a resource (or sub-resource) that can identify the version of another resource.

### DIFF
--- a/text/0000-version-resource.md
+++ b/text/0000-version-resource.md
@@ -1,0 +1,53 @@
+# Version Resource Type
+
+**Status:** `proposed`
+
+Add a composable 'Version' resource type.
+
+## Motivation
+
+When creating trace data or metrics, it can be extremely useful to know the specific version that
+emitted the iota of span or measurement being viewed. However, versions can mean different things
+to different systems and users, leading to a situation where a semantic `Version` type that can
+encapsulate these identifiers would be useful for analysis systems in order to allow for slicing
+data by service, component, or other version.
+
+## Explanation
+
+A `Version` is a semantic resource that can be composed with other resources, such as a `Service`, 
+`Component`, `Library`, `Device`, `Platform`, etc. A `Version` is optional, but reccomended.
+The definition of a `Version` is as follows:
+
+| Key     | Value  | Description                                  |
+|---------|--------|----------------------------------------------|
+| semver  | string | version string in semver format              |
+| git_sha | string | version string as a git sha                  |
+
+Only one field should be specified in a single `Version`. 
+
+## Internal details
+
+The exact implementation of this resource would vary based on language, but it would ultimately need to be represented in the data format so that it could be commmunicated to analysis backends. A JSON representation of a version resource follows:
+
+```
+{
+    "version": {
+        "type": "semver",
+        "value": "1.0.0",
+    }
+}
+```
+
+## Trade-offs and mitigations
+
+The largest drawback to this proposal is that there is a wide variety of things that constitute a 'version string'. By design, we do not attempt to solve for all of them in this proposal - instead, we focus on versions strings that are well-understood to have some semantic meaning attached to them if properly used.
+
+## Prior art and alternatives
+
+Tagging service resources with their version is generally suggested by analysis tools -- see [JAEGER_TAGS](https://www.jaegertracing.io/docs/1.8/client-features/) for an example -- but lacks standardization.
+
+## Open questions
+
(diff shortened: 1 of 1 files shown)


text/0000-version-resource.md
```diff
@@ -0,0 +1,53 @@
...
+Only one field should be specified in a single `Version`. 
+
+## Internal details
+
+The exact implementation of this resource would vary based on language, but it would ultimately need to be represented in the data format so that it could be commmunicated to analysis backends. A JSON representation of a version resource follows:
+
+```
+{
+    "version": {
+        "type": "semver",
```
tigrannajaryan: Why not `"semver": "1.0.0"`?

austinlparker:
The biggest rationale I had for making the `Version` a structured element was to support the notion that there can be multiple types of version strings that may have different meaning to a consumer. For example, it's quite frequent to tag releases in cloud native systems with their git SHA rather than a semver string.

That said, I think it'd also work to simply say `version` is a key on each resource and any string can be set as the value, at which point it'd be incumbent on a backend to figure out ways to differentiate between them.

toumorokoshi:
Regardless of the medium, I think version will play a big factor, especially in metrics. 
Working through the problem myself I was thinking Resource would be a great place to put it, considering it's a static context collection that's available to reference.

I feel like if it's going to be defined in the spec, we should have some clear uses cases. What will this RFC enable us to do?

yurishkuro:
@austinlparker there's a simpler way to achieve that with just the `version` attribute of the resource - by using a naming scheme, e.g. `version=semver:1.2.3` or `version=git:8ae73a`.

I am curious if the type of version actually matters, i.e. what exactly would one do differently knowing that the version is semver vs. git sha?

text/0000-version-resource.md
z1c0: Is this meant in addition to existing resource objects?
e.g. 
{
	"type": "container",
	"labels": {
           ....
	},
       "version" : {
       ....
       }
}

(example from OpenCensus: https://github.com/census-instrumentation/opencensus-specs/blob/master/resource/Resource.md#exporter-translation)

bogdandrutu:
For TC members please vote:
I am proposing to "rejected" this otep based on the comments from https://github.com/open-telemetry/oteps/pull/38#issuecomment-529155129.

@bogdandrutu @yurishkuro @carlosalberto @SergeyKanzhelev

austinlparker:
I'd rather modify it per Yuri's comments, they make sense.

bogdandrutu:
@austinlparker that will become a simple semantic convention for the version resource label or Span attribute

austinlparker:
Yes, but wouldn't changes to semantic conventions also go through the RFC process?

bogdandrutu:
I would say this is a very simple change that I would be happy to just review in the specs repo.

My 2c: Let's try to keep a balance between what is required in RFC vs a small change/improvement in specs.

austinlparker:
I originally brought this up in the spec repo and it was suggested as an RFC (see https://github.com/open-telemetry/opentelemetry-specification/issues/235). If y'all want it to go back there fine, but other people are making refs to this.

bogdandrutu:
Disclaimer: I am pretty new in the RFC model, but my understanding is that RFC are required when some important and complicated changes are proposed. Not trying to say that this is not important, but it becomes a very simple change if we follow Yuri's comment.

bogdandrutu:
In the current form I am happy to approve this, but I feel that you did a lot more work to get some minor change in the semantic conventions. I would like to hear others opinion, but don't want to block your work because I think it is good work.

carlosalberto:
Looks fine to me either way - probably we should do as @yurishkuro says and approve it to reduce friction, and try to have this kind of relatively straightforward changes as Specification Issues instead (for the next times, that is).

Probably we could also add a small note on that, as part of the RFC process docs?

text/0000-version-resource.md
```diff
@@ -0,0 +1,34 @@
+# Version Semantic Attribute
+
+**Status:** `proposed`
```
yurishkuro:
```suggestion
**Status:** `approved`
```

yurishkuro: also needs a number - 0038

austinlparker:
assigned 0038 and marked as approved per comments

bogdandrutu:
Can we please get this reviewed by one more approver?

PR #39, opened 2019-09-04
Metric observer callback specification

jmacd:


text/0000-metric-observer.md
```diff
@@ -0,0 +1,39 @@
...
+
+Why are gauges different than cumulative and measure instruments?  Measure instruments, by definition, carry information in the individual event, so the callback cannot optimize any better than the SDK can in this case.  Cumulative instruments are more commonly used to record amounts that are readily available, such as the number of bytes read or written, and while this may not always be true, recall the special case of `NonDescending` gauges.
+
+`NonDescending` gauges owe their existenence to this case, that we support non-negative cumulative metrics which, being expensive to compute, are recommended for use with `Observer` callbacks.  For example, if it requires a system call or more to compute a non-descending sum, such as the _cpu seconds_ consumed by the process, we should declare a non-descending gauge `Observer` for the instrument, instead of a cumulative.  This allows the cost of the metric to be reduced according to the desired monitoring frequency.
+
+One significant difference between gauges that are explicitly `Set()`, as compared with observer callbacks, is that `Set()` happens inside a context, whereas the observer callback does not.
+
+## Details
```
bogdandrutu: In my mind I have two possibilities:
1) The observer callback gets the Gauge and uses the normal Gauge API to `Set` values.
2) The observer callback returns the map `_label set_ to gauge value.`, then I see no point in allowing the normal `Set` API for this.

The advantage for the first option is the fact that is 1 less API. The advantage of the second option is that we know this metric is always recording the observer pattern and no context correlation is necessary (so the available labels will always be the provided in the ctor (keys) + map(values)).

I would personally prefer the second option. From the caller perspective if the values are static they can be preconstructed in the list of values, and even a map can be constructed in advance (constructing a map may require unnecessary uniqueness check, so we may consider a List<Point<List<String>, value>> (or even an interface ObserverCallbackResult { addPoint(List<String, value>) } then this can be optimized.

jmacd: OK, we'll make Observer callbacks exclusive with Settable gauges.

text/0000-metric-observer.md
```diff
@@ -0,0 +1,37 @@
...
+
+Why are gauges different than cumulative and measure instruments?  Measure instruments, by definition, carry information in the individual event, so the callback cannot optimize any better than the SDK can in this case.  Cumulative instruments are more commonly used to record amounts that are readily available, such as the number of bytes read or written, and while this may not always be true, recall the special case of `NonDescending` gauges.
+
+`NonDescending` gauges owe their existenence to this case, that we support non-negative cumulative metrics which, being expensive to compute, are recommended for use with `Observer` callbacks.  For example, if it requires a system call or more to compute a non-descending sum, such as the _cpu seconds_ consumed by the process, we should declare a non-descending gauge `Observer` for the instrument, instead of a cumulative.  This allows the cost of the metric to be reduced according to the desired monitoring frequency.
```
lzchen:
```suggestion
`NonDescending` gauges owe their existence to this case, that we support non-negative cumulative metrics which, being expensive to compute, are recommended for use with `Observer` callbacks.  For example, if it requires a system call or more to compute a non-descending sum, such as the _cpu seconds_ consumed by the process, we should declare a non-descending gauge `Observer` for the instrument, instead of a cumulative.  This allows the cost of the metric to be reduced according to the desired monitoring frequency.
```

lzchen:
```suggestion
`NonDescending` gauges owe their existence to this case
```

lzchen:
```suggestion
`NonDescending` gauges owe their existence to this case, that we support non-negative cumulative metrics which, being expensive to compute, are recommended for use with `Observer` callbacks.  For example, if it requires a system call or more to compute a non-descending sum, such as the _cpu seconds_ consumed by the process, we should declare a non-descending gauge `Observer` for the instrument, instead of a cumulative.  This allows the cost of the metric to be reduced according to the desired monitoring frequency.
```

lzchen: existenence -> existence

lzchen:
LGTM

jmacd:
Since this is close to having the approvals it needs, I've allocated RFC number 0007 here.

PR #40, opened 2019-09-04
Metrics processor API proposal

jmacd:


jmacd:
We discussed that this is not of great benefit. Most SDKs that care about efficiency will not want this code, which is quite small anyway.  See https://github.com/open-telemetry/opentelemetry-go/pull/100/files#diff-9d38a3597915be3886126a62e8566d63

Issue #2, opened 2019-06-25
Rename the repo to "otrfcs"?


bogdandrutu:

iredelmeier:
For more context, there have been similar discussions for other repos with "common" names, e.g., in [community](https://github.com/open-telemetry/community/issues/12) and in [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/issues/2).

For RFCs specifically, we could also do something like rename them from RFCs to OTEPs (for "OpenTelemetry Enhancement Proposal"), à la Python's PEPs and Kubernetes's KEPs. Thoughts?

bogdandrutu:
RFCs sounds like standards and I know there is a push from CNCF to not define standards. If we scope the rfc to OpenTelemetry should be fine that's the main reason.

ernado:
Please do, collisions with IETF RFC are unnecessary.

Sadly, Rust has already done this:
https://github.com/rust-lang/rfcs

Currently, RFC 1234 is unambiguously referring to IETF (with expection to Rust), let's not make it more complicated.

bogdandrutu:
Based on the comments I think "oteps" is better. Please +1 or -1 for this proposal.

bogdandrutu:
Ok, I see no objections. I will go ahead and change this to "oteps" ("OpenTelemetry Enhancement Proposal")

bogdandrutu:
Done.

PR #42, opened 2019-09-10
Proposal to separate context propagation from observability

tedsuo:
This proposal provides a high level description of OpenTelemetry, which shows how context propagation, observability, and baggage could be cleanly decoupled.

A couple notes on terminology:
* the terms "distributed application" and "application layer" used in this document refer to systems which leverage the context propagation layer, like observability and baggage. They do not refer to the "user applications" these systems are embedded in.
* In this doc, I refer to `TagMap` as `Correlations` and `labels`. There can be a separate discussion for what to name this, but since Tag has repeatedly come up as a contentious name for this feature, I tried to use something "similar but different" to avoid this debate. I actually feel that the term Correlation contains a more accurate meaning than Tag, so it may be an improvement.
* `TTL` is referred to as `hoplimit`, to reflect the clearer name of this feature in IPv6. It is represented as enum rather than an integer, since only two values are defined. 

The goal with this spec proposal is to find a way to describe these public APIs as simple as possible, but no simpler. This required touching on most of the public APIs for OpenTelemetry. This description is meant to be high level, but _not_ inaccurate. If a detail appears to be missing, it is intentional - presumably handled at the SDK layer. Possibly, it should be present. Please review this document with this in mind.


tedsuo:
This proposal relates to https://github.com/open-telemetry/oteps/pull/37 and https://github.com/open-telemetry/oteps/pull/36.

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,195 @@
...
+Status: `proposed`
+
+Design OpenTelemetry as a set of separate applications which operate on a shared context propagation mechanism.
+
+
+## Motivation
+
+Based on prior art, we know that fusing the observability system and the context propagation system together creates issues. Observability systems have special rules for propagating information, such as sampling, and may have different requirements from other systems which require non-local information to be sent downstream.  
+* Separation of concerns
+  *   Remove dependecy the Tracer dependency from context propagation mechanisms.
```
fbogsany: Delete `dependecy` before `the Tracer`

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Propagation API
+
+Distributed applications send data to downstream processes via propagators, functions which read and write application context into RPC requests. Each distributed application creates a set of propagators for every type of supported medium - currently HTTP and Binary.
+
+**Inject( context, request)**  
+To send the data for all distributed applications downstream to the next process, the Propagation API provides a function which takes a context and a request, and mutates the request to include the encoded context. The canonical representation of a request is as a map.
+
+**Extract( context, request) context**  
+To receive the data from all distributed applications set by prior upstream processes, the Propagation API provides a function which takes a context and a request, and mutates the request to include the encoded context. The canonical representation of a request is as a map.
```
fbogsany: Everything after the first comma is identical to the corresponding `Inject` text. I think you mean it `mutates the context`?

text/0000-separate-context-propagation.md
tedsuo: whoop!

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,195 @@
...
+Baggage values, on the other hand, are explicitly added in order to be accessed by downstream by other application code. Therefore, Baggage Context must be readable, and reliably propagated in-band in order to accomplish this goal.
+
+There may be cases where a key-value pair is propagated as TagMap for observability and as a Baggage for application specific use. AB testing is one example of such use case. There is potential duplication here at call site where a pair is created and also at propagation. 
+
+Solving this issue is not worth having semantic confusion with dual purpose. However, because all observability functions take the complete context as input, it may still be possible to use baggage values as labels.
+
+
+## What about complex propagation behavior?
+
+Some OpenTelemetry proposals have called for more complex propagation behavior. For example, having a fallback to extracting B3 headersif Trace-Context headers are not found. Chained propagators and other complex behavior can be modeled as implementation details behind the Propagator interface. Therefore, the propagation system itself does not need to provide chained propagators or other additional facilities.
```
Oberon00:
```suggestion
Some OpenTelemetry proposals have called for more complex propagation behavior. For example, falling back to extracting B3 headers if W3C Trace-Context headers are not found. Chained propagators and other complex behavior can be modeled as implementation details behind the Propagator interface. Therefore, the propagation system itself does not need to provide chained propagators or other additional facilities.
```

Oberon00: > the propagation system itself does not need to provide chained propagators

So what happens when multiple propagators are registered for the same type? I think chained propagators may be cumbersome to provide on top of an API that allows only one propagator, since it requires cooperation between everyone who wants to add a propagator to the chain.

carlosalberto: I have seen code that does this, by using a stack-alike `Propagator` that tries to inject/extract from a list of propagators (order is important) - for extraction, it returns with the first successful attempt, and for injection it simply tries to inject all formats. And yes, we should have a test case for this scenario, so we know where work fine.

carlosalberto:
One thing to keep in mind is that in some languages there's an **implicitly current** `Context` object. In Java you can fetch/copy the current `Context`, whereas in Python (both through `contextvars` and the current OpenTelemetry Python implementation) you can only either set values on it (thread-local alike) or copy the **entire thing**.

For this purpose, I'd imagine for some languages make `Context` and optional parameter, falling back to the **current** one.

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,195 @@
...
+**SetBaggage(context, key, value) context**
+To record the distributed state of an application, the Baggage API provides a function which takes a context, a key, and a value as input, and returns an updated context which contains the new value.
+
+**GetBaggage( context, key) value**
+To access the distributed state of an application, the Baggage API provides a function which takes a context and a key as input, and returns a value.
+
+**RemoveBaggage( context, key) context**
+To delete distributed state from an application, the Baggage API provides a function which takes a context, a key, and a value as input, and returns an updated context which contains the new value.
+
+**CleanBaggage( context) context**
```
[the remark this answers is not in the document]
carlosalberto: I'd rather prefer `ClearBaggage` to keep the desired effect.

text/0000-separate-context-propagation.md
tedsuo: Fixed. In general, the descriptions of the functions could be better, I'll improve them this week. Please let me know if the functions themselves capture everything.

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,195 @@
...
+To allow for this extensibility, OpenTelemetry is separated into **application layer** and a **context propagation layer**.  In this architecture, multiple distributed applications - such as the observability and baggage systems provided by OpenTelemetry - simultaneously share the same underlying context propagation system in order to execute their programs.
+
+
+# Application Layer
+
+## Observability API
+
+OpenTelemetry currently contains two observability systems - Tracing and Metrics – and may be extended over time. These separate systems are bound into a unified Observability API through sharing labels – a mechanism for correlating independent observations – and through sharing propagators.
+
+**Observe( context, labels…, observations...) context**
```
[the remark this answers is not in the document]
tedsuo: Thanks @Oberon00. I'd like to experiment with the best approach for spec syntax. I like your suggestions, as it makes it more clear that this isn't supposed to be pseudo-go. :)

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,195 @@
...
+The general form for all observability APIs is a function which takes a Context, label keys, and observations as input, and returns an updated Context.
+
+**Correlate( context, label, value, hoplimit) context**
+To set the label values used by all observations in the current transaction, the Observability API provides a function which takes a context, a label key, a value, and a hoplimit, and returns an updated context. If the hoplimit is set to NO_PROPAGATION, the label will only be available to observability functions in the same process. If the hoplimit is set to UNLIMITED_PROPAGATION, it will be available to all downstream services.
+
+**GetPropagator( type) inject, extract**
```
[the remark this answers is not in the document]
tedsuo: GetPropagator returns an inject and extract function, with definitions matching those above. I'd rather write it like this than define objects and methods. But again, the syntax could be better.

perhaps:
```
GetPropagator( type) -> (inject, extract)
```

text/0000-separate-context-propagation.md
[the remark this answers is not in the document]
tedsuo: Thanks, `ClearBaggage` sounds better.

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,195 @@
...
+# Internal details
+
+![drawing](img/context_propagation_details.png)
+
+## Context details
+OpenTelemetry currently implements three context types of context propagation.
+
+**Span Context -** Observability data used by the tracing system. The readable attributes are defined to match those found in the  W3C **traceparent** header. Span Context is used as labels for metrics and traces. This can quickly add overhead when propagated in-band. But, because this data is write-only, how this information is transmitted remains undefined. 
+
+**Correlation Context -** Transaction-level observability data, which can be applied as labels to spans and metrics.
```
lizthegrey: can we better explain the difference between correlation context and span context?

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,195 @@
...
+## Baggage API
+
+In addition to observability, OpenTelemetry provides a simple mechanism for propagating arbitrary data, called Baggage. This allows new distributed applications to be implemented without having to create new propagators.
+
+To manage the state of a distributed application, the Baggage API provides a set of functions which read, write, and remove data.
+
+**SetBaggage(context, key, value) context**
+To record the distributed state of an application, the Baggage API provides a function which takes a context, a key, and a value as input, and returns an updated context which contains the new value.
+
+**GetBaggage( context, key) value**
```
reyang:
```suggestion
**GetBaggage(context, key) value**
```

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,195 @@
...
+To access the distributed state of an application, the Baggage API provides a function which takes a context and a key as input, and returns a value.
+
+**RemoveBaggage( context, key) context**
+To delete distributed state from an application, the Baggage API provides a function which takes a context, a key, and a value as input, and returns an updated context which contains the new value.
+
+**CleanBaggage( context) context**
+To avoid sending baggage to an untrusted downstream process, the Baggage API provides a function remove all baggage from a context.
+
+**GetPropagator( type) inject, extract**
```
reyang:
```suggestion
**GetPropagator(type) inject, extract**
```

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,195 @@
...
+**Correlation Context -** Transaction-level observability data, which can be applied as labels to spans and metrics. 
+
+**Baggage Context -** Transaction-level application data, meant to be shared with downstream components.
+
+Note that when possible, OpenTelemetry APIs calls are given access to the entire context object, and not a specific context type.
+
+
+## Context Management and in-process propagation
+
+In order for Context to function, it must always remain bound to the execution of code it represents. By default, this means that the programmer must pass a Context down the call stack as a function parameter.  However, many languages provide automated context management facilities, such as thread locals.  OpenTelemetry should leverage these facilities when available, in order to provide automatic context management.
```
reyang: Minor, there are mixed use of single space and double spaces.

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,195 @@
...
+
+Solving this issue is not worth having semantic confusion with dual purpose. However, because all observability functions take the complete context as input, it may still be possible to use baggage values as labels.
+
+
+## What about complex propagation behavior?
+
+Some OpenTelemetry proposals have called for more complex propagation behavior. For example, having a fallback to extracting B3 headersif Trace-Context headers are not found. Chained propagators and other complex behavior can be modeled as implementation details behind the Propagator interface. Therefore, the propagation system itself does not need to provide chained propagators or other additional facilities.
+
+
+## Did you add a context parameter to every API call because Go has infected your brain?
```
reyang: Curious, is this just for fun?

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,195 @@
...
+
+* [W3C Trace-Context](https://www.w3.org/TR/trace-context/) candidate is not yet accepted
+* Work on [W3C Correlation-Context](https://w3c.github.io/correlation-context/) has begun, but was halted to focus on Trace-Context. 
+* No work has begun on a theoretical W3C Baggage-Context.
+
+Given that we must ship with working propagators, and the W3C specifications are not yet complete, how should we move forwards with implementing context propagation?
+
+# Future possibilities
+
+Cleanly splitting OpenTelemetry into an Application and  Context Propagation layer may allow us to move the Context Propagation layer into its own, stand-alone project. This may facilitate adoption, by allowing us to share Context Propagation with gRPC and other projects.
```
reyang:
```suggestion
Cleanly splitting OpenTelemetry into an Application and Context Propagation layer may allow us to move the Context Propagation layer into its own, stand-alone project. This may facilitate adoption, by allowing us to share Context Propagation with gRPC and other projects.
```

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,195 @@
...
+
+**GetPropagator( type) inject, extract**
+To register with the propagation system, the Baggage API provides a set of propagation functions for every propagation type.
+
+
+## Additional APIs
+
+Because the application and context propagation layers are separated, it is possible to create new distributed applications which do not depend on either the Observability or Baggage APIs.
+
+**GetPropagator(type) inject, extract**
```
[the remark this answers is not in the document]
tedsuo: So, the propagation layer provides the RegisterPropagator function. Applications provide, and applications provide the Propagators to be registered. 

Is there a better way to explain this in the GetPropagator description? Right now it says "To register with the propagation system, the [BLANK] API provides a set of propagation functions for every propagation type."

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Pre-existing Context implementations
+
+In some languages, a single, widely used Context implementation exists. In other languages, there many be too many  implementations, or none at all. In the cases where there is not an extremely clear pre-existing option available, OpenTelemetry should provide its own Context implementation.
+
+While the above explanation represents the default OpenTelemetry approach to context propagation, it is important to note that some languages may already contain a form of context propagation. For example, Go has a the context.Context object, and widespread conventions for how to pass.
+
+Span data is used as labels for metrics and traces. This can quickly add overhead when propagated in-band. But, because this data is write-only, how this information is transmitted remains undefined.
```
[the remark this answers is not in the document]
tedsuo: derp, this is just a crappy unfinished section.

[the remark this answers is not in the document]
tedsuo: Agree, it's filled in better below, under "Why separate Baggage from Correlations?" I'll delet it from here.

text/0000-separate-context-propagation.md
tedsuo: So, here is the nuance: there is basic chaining provided at the API level by RegisterPropagator. All propagators are run for every type, in the order in which they are registered. This is where the cooperation happens between independent applications.

More complex propagation behavior usually ends up being specific to each application. So, the OTel SDK can provide the kind of fallback W3C -> B3 chaining for observability, described here. That sort of behavior is presented as a single, complex propagator, from the point of view of the Propagation API, since the fallback behavior is internal to a single application. Does that make sense?

text/0000-separate-context-propagation.md
tedsuo: Yeah totally. Please let me know if the new descriptions are better.

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,195 @@
...
+
+Distributed applications access data in-process using a shared context object. Each distributed application sets a single key in the context, containing all of the data for that system.
+
+**SetValue( context, key, value) context**  
+To record the local state of an application, the Context API provides a function which takes a context, a key, and a value as input, and returns an updated context which contains the new value.
+
+**GetValue( context, key) value**  
+To access the local state of an application, the Context API provides a function which takes a context and a key as input, and returns a value.
+
+**Optional: Automated Context Management**
```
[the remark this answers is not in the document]
tedsuo: I changed this to an actual subheading. Does that read better?

text/0000-separate-context-propagation.md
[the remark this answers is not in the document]
tedsuo: DELETED!

text/0000-separate-context-propagation.md
tedsuo: Haha, well I wanted to address what I perceive to be a common question, along the lines of "why is this context parameter everywhere? Is it because this is a golang project? Is it required that every language must pass context this way?"

But if the humor gets in the way of learning, I can change it.

text/0000-separate-context-propagation.md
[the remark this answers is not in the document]
tedsuo: changed

text/0000-separate-context-propagation.md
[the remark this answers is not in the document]
tedsuo: done!

text/0000-separate-context-propagation.md
tedsuo: Thank you for all the copyediting!

tedsuo:
@carlosalberto:
> One thing to keep in mind is that in some languages there's an **implicitly current** `Context` object... I'd imagine for some languages make `Context` and optional parameter, falling back to the **current** one.

Yes, I'd like to make it clear that in some languages the context is propagated automatically, in which case it is an implicit parameter in these functions. I mentioned this in the section titled "Optional: Automated Context Management" but I think this could be expressed clearer.

Also, there is some nuance around the difference between a native context object vs a native context propagation mechanism. For example, Go has a context object, but no propagation mechanism.

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,193 @@
...
+
+
+# Application Layer
+
+## Observability API
+
+OpenTelemetry currently contains two observability systems - Tracing and Metrics – and may be extended over time. These separate systems are bound into a unified Observability API through sharing labels – a mechanism for correlating independent observations – and through sharing propagators.
+
+**Observe(context, labels…, observations...) -> context**  
+The general form for all observability APIs is a function which takes a Context, label keys, and observations as input, and returns an updated Context.
```
toumorokoshi: what is an observation in this context? A single metric measurement comes to mind.

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,193 @@
...
+To register with the propagation system, the Observability API provides a set of propagation functions for every propagation type. 
+
+
+## Baggage API
+
+In addition to observability, OpenTelemetry provides a simple mechanism for propagating arbitrary data, called Baggage. This allows new distributed applications to be implemented without having to create new propagators.
+
+To manage the state of a distributed application, the Baggage API provides a set of functions which read, write, and remove data.
+
+**SetBaggage(context, key, value) -> context**
```
toumorokoshi: for these APIs, is it just returning back the instance that was passed in? or is it actually creating a copy? 

I'm not sure if it would be efficient for a language like python to rapidly create a bunch of new, but similar, objects.

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,193 @@
...
+Distributed applications access data in-process using a shared context object. Each distributed application sets a single key in the context, containing all of the data for that system.
+
+**SetValue(context, key, value) -> context**  
+To record the local state of an application, the Context API provides a function which takes a context, a key, and a value as input, and returns an updated context which contains the new value.
+
+**GetValue(context, key) -> value**  
+To access the local state of an application, the Context API provides a function which takes a context and a key as input, and returns a value.
+
+### Optional: Automated Context Management
+When possible, context should automatically be associated with program execution . Note that some languages do not provide any facility for setting and getting a current context. In these cases, the user is responsible for managing the current context.
```
toumorokoshi: might be worth expanded on what Automated Context Management is like in practice. I would imagine A ThreadLocal is an example since context is not available outside of the thread it was set or modified in?

text/0000-separate-context-propagation.md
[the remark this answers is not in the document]
Oberon00: I think it does 👍

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,193 @@
...
+Baggage values, on the other hand, are explicitly added in order to be accessed by downstream by other application code. Therefore, Baggage Context must be readable, and reliably propagated in-band in order to accomplish this goal.
+
+There may be cases where a key-value pair is propagated as TagMap for observability and as a Baggage for application specific use. AB testing is one example of such use case. There is potential duplication here at call site where a pair is created and also at propagation. 
+
+Solving this issue is not worth having semantic confusion with dual purpose. However, because all observability functions take the complete context as input, it may still be possible to use baggage values as labels.
+
+
+## What about complex propagation behavior?
+
+Some OpenTelemetry proposals have called for more complex propagation behavior. For example, falling back to extracting B3 headers if W3C Trace-Context headers are not found. Chained propagators and other complex behavior can be modeled as implementation details behind the Propagator interface. Therefore, the propagation system itself does not need to provide chained propagators or other additional facilities.
```
toumorokoshi: It sounds like a pretty universal need across languages to have multiple propagators. Any migration from one standard to another will require it. I'd personally rather see a standard need like this make it's way into the API rather than an unspecified SDK implementation whose behavior may differ across languages.

toumorokoshi: But I think this should just be a separate RFC regardless. This matches the current spec.

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,193 @@
...
+## Context details
+OpenTelemetry currently implements three context types of context propagation.
+
+**Span Context -** The serializable portion of a span, which is injected and extracted. The readable attributes are defined to match those found in the  W3C **traceparent** header. 
+
+**Correlation Context -** Correlation Context contains a map of labels and values, to be shared between metrics and traces. This allows observability data to be indexed and dimensionalized in a variety of ways. Note that correlations can quickly add overhead when propagated in-band. But because this data is write-only, it may be possible to optimize how it is transmitted.
+
+**Baggage Context -** Transaction-level application data, meant to be shared with downstream components. This data is readable, and must be propagated in-band. Because of this, Baggage should be used sparingly, to avoid ballooning the size of RPC requests.
+
+Note that when possible, OpenTelemetry APIs calls are given access to the entire context object, and not a specific context type.
```
toumorokoshi: I like the whole context object as it allows more flexibility for the propagators to modify contexts, but there's implementation details that all propagators will have to be aware of:

1. how to get and set specific contexts from the Context object. One way to accomplish this is with an explicit interface or methods attached to context, but as I understand it, this violates the idea that Yuri was evangelizing where the Context object is a primitive and is completely unaware of what values are being loaded into it.

2. SpanContext is a little tricky where the extractor should be setting the parent SpanContext, while the injector should be injecting values from the current SpanContext. This would be great to not require the propagator author to get that right. 

I'm personally voting for an explicit interface on context to retrieve such values. solving #2 is a little trickier. 

More context in this discussion:
https://github.com/open-telemetry/oteps/pull/37
And this PR:
https://github.com/open-telemetry/opentelemetry-python/pull/89

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,193 @@
...
+
+
+## Baggage API
+
+In addition to observability, OpenTelemetry provides a simple mechanism for propagating arbitrary data, called Baggage. This allows new distributed applications to be implemented without having to create new propagators.
+
+To manage the state of a distributed application, the Baggage API provides a set of functions which read, write, and remove data.
+
+**SetBaggage(context, key, value) -> context**  
+To record the distributed state of an application, the Baggage API provides a function which takes a context, a key, and a value as input, and returns an updated context which contains the new value.
```
objectiser: Would it also be beneficial to have a `hoplimit` on baggage items? For example, once supported, we may want a certain baggage value only to be propagated to the downstream service and no further - or only available within the service (`NO_PROPAGATION`) - but it is not to be treated as a label.

text/0000-separate-context-propagation.md
Oberon00: This describes

> The general form for all observability APIs

In my understanding, it is a stand-in for e.g. starting, ending Spans, but also any metric recording.

text/0000-separate-context-propagation.md
Oberon00: I think this was just meant as a sort of mathematical notation, and whether this is implemented as a function modifying a thread-local it does not even get as argument or having a signature that actually has a context as parameter and return value is a language-specific concern.

text/0000-separate-context-propagation.md
Oberon00: 👍  In general, this OTEP is very abstract, some (non-normative) examples would be helpful.

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,193 @@
...
+**GetBaggage(context, key) -> value**  
+To access the distributed state of an application, the Baggage API provides a function which takes a context and a key as input, and returns a value.
+
+**RemoveBaggage(context, key) -> context**  
+To delete distributed state from an application, the Baggage API provides a function which takes a context, a key, and a value as input, and returns an updated context which contains the new value.
+
+**ClearBaggage(context) -> context**  
+To avoid sending baggage to an untrusted downstream process, the Baggage API provides a function remove all baggage from a context, 
+
+**GetPropagator(type) -> (inject, extract)**
```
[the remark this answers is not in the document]
Oberon00: Which "different interfaces" do you mean? Did you mean to write "different interfaces for different propagators"?

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Propagation API
+
+Distributed applications send data to downstream processes via propagators, functions which read and write application context into RPC requests. Each distributed application creates a set of propagators for every type of supported medium - currently HTTP and Binary.
+
+**Inject(context, request)**  
+To send the data for all distributed applications downstream to the next process, the Propagation API provides a function which takes a context and a request, and mutates the request to include the encoded context. The canonical representation of a request is as a map.
+
+**Extract(context, request) -> context**  
+To receive data injected by prior upstream processes, the Propagation API provides a function which takes a context and a request, and returns an updated context.
```
objectiser: What is purpose of the supplied `context`? Do we not just want to decode the context from the `request` to return a `Context`? If the context is also supplied - does it need to explicitly state how duplicate fields would be handled (i.e. existing context and request containing same key).

Oberon00:
What is a bit unclear when reading this PR is what that `type` parameter is supposed to be. Just a an ID to identify the propagator? Are the returned extract/inject functions supposed to have the same signature regardless of `type`?

text/0000-separate-context-propagation.md
toumorokoshi: I guess it's still not something that immediately clicks for me. Some clarification on that, if it's not around in some other document, would be great.

text/0000-separate-context-propagation.md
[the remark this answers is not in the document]
toumorokoshi: Sorry, interfaces isn't a great description. I was basically saying the same service may need to contain multiple different propagation configurations. Expanding on my example above, I could have two http clients: one for internal services (which should propagate headers) and one for external services (which should not, and emit a span instead).

text/0000-separate-context-propagation.md
yurishkuro: How do we want to treat types of the arguments here? Specifically, `value` would be more efficiently represented as `[]byte` rather than a string (recall J.Mace's baggage as thrift-like structs).

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,193 @@
...
+
+OpenTelemetry currently contains two observability systems - Tracing and Metrics – and may be extended over time. These separate systems are bound into a unified Observability API through sharing labels – a mechanism for correlating independent observations – and through sharing propagators.
+
+**Observe(context, labels…, observations...) -> context**  
+The general form for all observability APIs is a function which takes a Context, label keys, and observations as input, and returns an updated Context.
+
+**Correlate(context, label, value, hoplimit) -> context**  
+To set the label values used by all observations in the current transaction, the Observability API provides a function which takes a context, a label key, a value, and a hoplimit, and returns an updated context. If the hoplimit is set to NO_PROPAGATION, the label will only be available to observability functions in the same process. If the hoplimit is set to UNLIMITED_PROPAGATION, it will be available to all downstream services.
+
+**GetPropagator(type) -> (inject, extract)**
```
yurishkuro: The `GetPropagator` propagator signature is confusing. The main idea of extensible propagation is that the instrumentation code that performs injection/extraction should not need to know/care about which types of contexts are being serialized. The only way this could work is if there is a shared "propagator" object that application interacts with, and different context sub-systems register their own serializers with that propagator. There's no situation where something would be calling GetPropagator on individual sub-systems like Correlations / Baggage.

text/0000-separate-context-propagation.md
yurishkuro: I've been in general skeptical about the TTL notion, especially "single-hop" whose semantics are completely vague given that requests may pass through proxies that the caller would not know about.

There is always the underlying in-memory Context object used as a storage for all kinds of contexts. I can always add a value to that context _explicitly_, and retrieve it _explicitly_ if my application needs it in-process. I don't think there's a need to provide any additional functionality here (although it's worth calling out that such capability must exist in the Context, whereas it was missing from OpenTracing Java, for example).

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,193 @@
...
+
+**GetPropagator(type) -> (inject, extract)**  
+To register with the propagation system, the Baggage API provides a set of propagation functions for every propagation type.
+
+
+## Additional APIs
+
+Because the application and context propagation layers are separated, it is possible to create new distributed applications which do not depend on either the Observability or Baggage APIs.
+
+**GetPropagator(type) -> (inject, extract)**
```
yurishkuro: same comment as above - the actual relationship is inverted, there's no actor that would ever call GetPropagator

text/0000-separate-context-propagation.md
yurishkuro: This section is confusing two different layers from the [Context Propagation Layers design doc][1]: the Context and In-Process Propagation, hence the problem. In languages like Go the In-Process Propagation layer is not relevant, because the context is passed explicitly. This could apply to other frameworks when people don't want to rely on thread-locals. That has nothing to do with the Context type itself that simply stores values.

[1]: https://docs.google.com/document/d/1UxrEYOaQlF_E4gtiPoFmcZ4YKKe1GxohvCvQDuwvD1I/

text/0000-separate-context-propagation.md
yurishkuro: @toumorokoshi in Go you can define a private namespace for the values you store in Context ([example in go][1]). Nobody else but your cross-cutting application needs to know how to get/set those values. Therefore, the Context can remain completely agnostic to what it stores and carries.

[1]: https://github.com/opentracing/opentracing-go/blob/master/gocontext.go#L5

text/0000-separate-context-propagation.md
yurishkuro: I like it. You can also stress that it's not just a Go thing - in the old versions of Node there was no CLS (or it was extremely inefficient) and explicitly passing the context was how the propagation was achieved.

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,193 @@
...
+
+Some OpenTelemetry proposals have called for more complex propagation behavior. For example, falling back to extracting B3 headers if W3C Trace-Context headers are not found. Chained propagators and other complex behavior can be modeled as implementation details behind the Propagator interface. Therefore, the propagation system itself does not need to provide chained propagators or other additional facilities.
+
+
+## Did you add a context parameter to every API call because Go has infected your brain?
+
+No. The concept of an explicit context is fundamental to a model where independent distributed applications share the same context propagation layer. How this context appears or is expressed is language specific, but it must be present in some form.
+
+
+# Prior art and alternatives
```
yurishkuro: ahem... Context Propagation Layers https://docs.google.com/document/d/1UxrEYOaQlF_E4gtiPoFmcZ4YKKe1GxohvCvQDuwvD1I/edit

text/0000-separate-context-propagation.md
```diff
@@ -0,0 +1,193 @@
...
+  *   Separate distributed context into Baggage and Correlations
+* Extensibility
+  *   Allow users to create new applications for context propagation.
+  *   For example: A/B testing, encrypted or authenticated data, and new, experimental forms of observability.
+
+## Explanation
+
+# OpenTelemetry Layered Architecture
+
+![drawing](img/context_propagation_explanation.png)
```
yurishkuro: I find this pic very confusing, sorry. There was an ascii art in one of the earlier tickets.

text/0000-separate-context-propagation.md
Oberon00: Maybe the implementations of Inject and Extract would need it?

text/0000-separate-context-propagation.md
Oberon00: I don't think the section is confusing them. It just defines a point where they interact. The context type used here is not necessarily the same the in-process layer uses, it could translate between them.

text/0000-separate-context-propagation.md
rghetia: What does Extract update the context with? Is it SpanContext or Span?
If it is SpanContext then it would require another api to set the current SpanContext in addition to Span.
If it is Span then Extract also has to start a span which I don't think it is appropriate. 
So, it would be better for this API to return SpanContext. Unless we introduce a concept of Inactive Span which simply contains remote SpanContext.

Oberon00: You are right, I also think this is an important detail that needs to be resolved.

I have another suggestion: The context always contains just a spancontext. Any SpanContext has an optional reference to the span it refers to, which is null in case of a remote span (cf https://github.com/open-telemetry/opentelemetry-specification/pull/216#issuecomment-527379659)

yurishkuro: I made a similar suggestion when we were working on the initial Java API in the spring. I think it has the benefits not just in the context of this discussion, but also as a alternative mechanism for implementing the "span". For example, in a streaming implementation the span does not even need to exist if all write operations simply work off the SpanContext. In other cases, spans may be kept by the tracer in a private dictionary, which makes memory management easier than if user code has an explicit reference to the span.

Oberon00: I don't understand how this would reduce references to Spans. I thought the user would be able to get this reference anytime from the SpanContext. But even withot such an API, startSpan etc would of course still return references to spans.

yurishkuro: StartSpan would return a new span context. Span is just an accumulator of data, it is not actually needed in the API. We could move all Span methods into Tracer directly and make them accept span context as the first argument, that would be functionally equivalent.

Oberon00: Ah, I understand! There is one functional difference though: if you allowed something like `Tracer.setSpanAttribute(context, "x", "y")`, you implicitly allow setting attributes on spans that were not created in this process. This is prevented by the current Span/SpanContext separation and I think that's a good thing.

carlosalberto: I remember that discussion on last spring, and I think the desire was that `Span`s created in the same process could share the monotonic clock (which is an optimization, more than a requirement).

yurishkuro: @Oberon00 indeed, you can extract a parent span context and set attributes on it. I think it's actually a feature of this approach, not the downside :-) . Although it would only work in the streaming-like implementations, not the current approach where "finished span is not writeable". But then, even today you can do `span.end(); span.setTag()`, which is meaningless.

@carlosalberto there are many ways of implementing SDK when the API only exposes Span Context. As @Oberon00 said, the simple implementation is still to have the Span at the SDK level which is linked from Span Context, this way you can share clock offsets.

text/0000-separate-context-propagation.md
toumorokoshi: Ok, I think I understand what you're proposing. It's

1. a property-agnostic context object.
2. a layer of getter and setters to get and set values from that context. 

I'm fine with this approach. We actually had a discussion in the python group and that's the consensus we've reached as well.

tedsuo:
Hi all. Picking this back up again, now that I have time. Sorry for the delay! I will try to address all of the current comments in the next two days.

Based on feedback, I've changed the propagation layer:
* HTTP and Binary propagators are separate, no more `type`.
* Functions have been added for explicitly chaining injectors and extractors, 
* No more registry, global propagators are now optional.

Please have a look. I only included the HTTP propagators for now. If people like this approach, I'll add Binary propagators as well.

text/0000-separate-context-propagation.md
lizthegrey: Much better.

text/0042-separate-context-propagation.md
```diff
@@ -96,21 +103,34 @@ To access the context associated with program execution, the Context API provide
...
+**ChainHTTPInjector(injector, injector) injector** 
 In order for the application layer to function correctly, Propagation choices must be syncronized between all processes in the distributed system, and multiple applications must be able to inject and extract their context into the same request. To meet these requirements, the Propagation API provides a function which registers a set of propagators, which will all be executed in order when the future calls to inject and extract are made. A canonical propagator consists of an inject and an extract function.
 
-OpenTelemetry currently contains two types of Propagators:
+**ChainHTTPExtractor(extractor, extractor) extractor** 
+In order for the application layer to function correctly, Propagation choices must be syncronized between all processes in the distributed system, and multiple applications must be able to inject and extract their context into the same request. To meet these requirements, the Propagation API provides a function which registers a set of propagators, which will all be executed in order when the future calls to inject and extract are made. A canonical propagator consists of an inject and an extract function.
```
rghetia: Description is identical to ChainHTTPInjector. Perhaps combine them?

text/0042-separate-context-propagation.md
```diff
@@ -96,21 +103,34 @@ To access the context associated with program execution, the Context API provide
...
+It is often convenient to create a chain of propagators during program initialization, and then access these combined propagators later in the program. To facilitate this, global injectors and extractors are optionally available. However, there is no requirement to use this feature.
+
+**SetHTTPInjector(injector)**  
+To update the global injector, the Propagation API provides a function which takes an injector.
+
+**GetHTTPInjector() -> injector**  
+To access the global injector, the Propagation API provides a function which returns an injector.
+
+**SetHTTPExtractor(extractor)**  
+To update the global extractor, the Propagation API provides a function which takes an injector.
```
rghetia: nit: s/takes an injector/takes an extractor/

text/0042-separate-context-propagation.md
tedsuo: Hmmm, copypasta error, one moment please...

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,213 @@
...
+To access the global extractor, the Propagation API provides a function which returns an extractor.
+
+# Internal details
+
+![drawing](img/context_propagation_details.png)
+
+## Context details
+OpenTelemetry currently implements three context types of context propagation.
+
+**Span Context -** The serializable portion of a span, which is injected and extracted. The readable attributes are defined to match those found in the  W3C **traceparent** header.
```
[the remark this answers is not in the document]
lizthegrey: Ping @tedsuo

text/0000-separate-context-propagation.md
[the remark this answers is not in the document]
tedsuo: I've removed the Registry concept, hopefully this is clearer.

text/0000-separate-context-propagation.md
tedsuo: Good call, I will emphasize that this issue exists in multiple languages.

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,213 @@
+# Proposal: Separate Layer for Context Propagation
+
+Design OpenTelemetry as a set of separate applications which operate on a shared context propagation mechanism.
+
+
+## Motivation
+
+Based on prior art, we know that fusing the observability system and the context propagation system together creates issues. Observability systems have special rules for propagating information, such as sampling, and may have different requirements from other systems which require non-local information to be sent downstream.
```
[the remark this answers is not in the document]
tedsuo: Thanks, I prefer your wording as well! Adding it in.

text/0000-separate-context-propagation.md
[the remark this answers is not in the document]
tedsuo: @Oberon00 
Following up, now that Named Tracers have been added:
https://github.com/open-telemetry/opentelemetry-specification/pull/264

At first glance, I this approach turns out to make it a little tricky to implement an extensible inject/extract call which is used by multiple separate systems (observability, baggage, etc), as the observability extractor you want is now contextualized by the tracer handle. 

I can also see baggage wanting a similar scoping to tracers - surely users will want to disable certain endpoints for propagating baggage. I'm also noting that for any given code component, we would be naming tracers, meters, baggage, etc, essentially all with the same name and version.

So, while I like the named tracer concept and believe it is good, I'm wondering if this concept needs to be adjusted in some way? Otherwise, I'm not sure how we can cleanly split baggage and context propagation out from tracing.

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Motivation
+
+Based on prior art, we know that fusing the observability system and the context propagation system together creates issues. Observability systems have special rules for propagating information, such as sampling, and may have different requirements from other systems which require non-local information to be sent downstream. 
+* Separation of concerns
+  *   Remove the Tracer dependency from context propagation mechanisms.
+  *   Separate distributed context into Baggage and Correlations
+* Extensibility
+  *   Allow users to create new applications for context propagation.
+  *   For example: A/B testing, encrypted or authenticated data, and new, experimental forms of observability.
```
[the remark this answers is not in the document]
tedsuo: Makes sense, thanks.

text/0000-separate-context-propagation.md
tedsuo: Yeah, I agree that single hop TTL is questionable...

@objectiser I agree that if we come up with a TTL scheme beyond just `NO_PROPAGATION` and `UNLIMITED_PROPAGATION` we should add it to baggage. I only left it off because `NO_PROPAGATION` is equivalent to just using the context API directly, and that did not leave any options other than `UNLIMITED_PROPAGATION`. So it wasn't clear to me why anyone would need baggage for that purpose.

text/0042-separate-context-propagation.md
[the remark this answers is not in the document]
tedsuo: Thanks, that is good feedback. We use the term "downstream" in a number of places in OTel, maybe we should add a glossary? https://github.com/search?q=org%3Aopen-telemetry+downstream&type=Code

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,213 @@
...
+OpenTelemetry is a distributed program, which requires non-local, transaction-level context in order to execute correctly. Transaction-level context can also be used to build other distributed programs, such as security, versioning, and network switching programs.
+
+To allow for this extensibility, OpenTelemetry is separated into an **application layer** and a **context propagation layer**. In this architecture, multiple distributed applications - such as the observability and baggage systems provided by OpenTelemetry - simultaneously share the same underlying context propagation system in order to execute their programs.
+
+
+# Application Layer
+
+## Observability API
+
+OpenTelemetry currently contains two observability systems - Tracing and Metrics – and may be extended over time. These separate systems are bound into a unified Observability API through sharing labels – a mechanism for correlating independent observations – and through sharing propagators.
```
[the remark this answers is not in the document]
tedsuo: I am thinking that Observability system does not automatically have a separate propagators for tracing and for metrics, but I see your point that Trace-Context and Correlation-Context are separate, and right now it's not clear if Tracing would use Correlation-Context.

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,213 @@
...
+
+**Observe(context, labels…, observations...) -> context**
```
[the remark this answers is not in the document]
tedsuo: Yeah, I should definitely remove the "all" here. 

In this RFC, I'm trying to define the `Context` and `Propagation` APIs. I want it to be clear enough about how we intend to leverage the context and propagation APIs that we can catch any design errors in those APIs. If I enumerate all of the details of Tracing and Metrics, I've hauled in too much. If I define nothing but `GetHTTPExtractor` and `GetHTTPInjector` in the Observability API, that will be insufficient for review. 

I _am_ trying to show how `DistribtuedContext` can be cleanly split into `Baggage` and `Correlations`. So I feel like  `Correlate` should be present, along with `Observe`. But, again, this is a general form. I will try to be clearer that these two definitions - `Observe` and `Correlate` - are more abstract than the rest. I would still like feedback to the tune of "even as a general form, these functions are not correct."

I suggest that the details about how exactly these changes will affect our Tracing and Metrics APIs should be worked out in specification PRs, not in an RFC.

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,213 @@
...
+
+OpenTelemetry currently contains two observability systems - Tracing and Metrics – and may be extended over time. These separate systems are bound into a unified Observability API through sharing labels – a mechanism for correlating independent observations – and through sharing propagators.
+
+**Observe(context, labels…, observations...) -> context**  
+The general form for all observability APIs is a function which takes a Context, label keys, and observations as input, and returns an updated Context.
+
+**Correlate(context, label, value, hoplimit) -> context**  
+To set the label values used by all observations in the current transaction, the Observability API provides a function which takes a context, a label key, a value, and a hoplimit, and returns an updated context. If the hoplimit is set to NO_PROPAGATION, the label will only be available to observability functions in the same process. If the hoplimit is set to UNLIMITED_PROPAGATION, it will be available to all downstream services.
+
+**GetHTTPExtractor() -> extractor**
```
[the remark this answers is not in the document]
tedsuo: My intent is just to show that I am not proposing a "generic" propagator: HTTP, BINARY, etc, are each handled by a separate function. I didn't call this out before and it confused some people into thinking we were going back to the OpenTracing way of doing things.

We could call it `TEXTMAP` instead of `HTTP`, or `HTTPText`. Don't really have an opinion on that front. HTTP is the shortest though. :)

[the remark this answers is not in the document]
tedsuo: @Oberon00 I actually assumed that we would not have separate propagators for Metrics and Tracing as something in the API layer. So I was namespacing by splitting things into an `Observability` API and a `Baggage` API.

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,213 @@
...
+Distributed applications access data in-process using a shared context object. Each distributed application sets a single key in the context, containing all of the data for that system.
+
+**SetValue(context, key, value) -> context**  
+To record the local state of an application, the Context API provides a function which takes a context, a key, and a value as input, and returns an updated context which contains the new value.
+
+**GetValue(context, key) -> value**  
+To access the local state of an application, the Context API provides a function which takes a context and a key as input, and returns a value.
+
+### Optional: Automated Context Management
+When possible, context should automatically be associated with program execution . Note that some languages do not provide any facility for setting and getting a current context. In these cases, the user is responsible for managing the current context.
```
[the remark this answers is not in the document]
tedsuo: good catch.

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,213 @@
...
+
+**HTTPInject(context, request)**  
+To send the data for all distributed applications downstream to the next process, the Propagation API provides a function which takes a context and an HTTP request, and mutates the HTTP request to include an HTTP Header representation of the context.
+
+**HTTPExtract(context, request) -> context**  
+To receive data injected by prior upstream processes, the Propagation API provides a function which takes a context and an HTTP request, and returns context which represents the state of the upstream system.
+
+**ChainHTTPInjector(injector, injector) injector** 
+In order for the application layer to function correctly, Propagation choices must be syncronized between all processes in the distributed system, and multiple applications must be able to inject and extract their context into the same request. To meet these requirements, the Propagation API provides a function which registers a set of propagators, which will all be executed in order when the future calls to inject and extract are made. A canonical propagator consists of an inject and an extract function.
```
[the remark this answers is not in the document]
tedsuo: I really struggle what to call these. I added a definition for distributed application at the beginning, to help. It feels natural to refer to this architecture as an "application layer" and a "context propagation layer" but it is easy to see how the term "application" could be confusing. Not sure what is better though - I see Yuri using Aspect Oriented Programming terms like "Cross-cutting concern" but people who know what that is might think we are being literal about leveraging AOP.

Ideas?
* Cross-cutting concerns
* Distributed subsystems
* Some made up word, idk

text/0042-separate-context-propagation.md
tedsuo: Fixed

text/0042-separate-context-propagation.md
tedsuo: Thanks, I'm getting buried by my own boilerplate here. :)

text/0042-separate-context-propagation.md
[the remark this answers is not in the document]
tedsuo: @lizthegrey sorry for the slow update on this! I believe that sampling is intended to be propagated as part of `SpanContext`. In the Trace-Context spec, sampling would be included as a "vendor-specifc detail" (with OpenTelemetry SDK counting as a vendor in this case) inside of the `tracestate` field. https://w3c.github.io/trace-context/#tracestate-header

There is also a `sampled` flag in the `traceparent` field which should be set: https://w3c.github.io/trace-context/#trace-flags

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,213 @@
...
+## Context details
+OpenTelemetry currently implements three context types of context propagation.
+
+**Span Context -** The serializable portion of a span, which is injected and extracted. The readable attributes are defined to match those found in the  W3C **traceparent** header. 
+
+**Correlation Context -** Correlation Context contains a map of labels and values, to be shared between metrics and traces. This allows observability data to be indexed and dimensionalized in a variety of ways. Note that correlations can quickly add overhead when propagated in-band. But because this data is write-only, it may be possible to optimize how it is transmitted.
+
+**Baggage Context -** Transaction-level application data, meant to be shared with downstream components. This data is readable, and must be propagated in-band. Because of this, Baggage should be used sparingly, to avoid ballooning the size of RPC requests.
+
+Note that when possible, OpenTelemetry APIs calls are given access to the entire context object, and not a specific context type.
```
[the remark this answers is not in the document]
tedsuo: No. I updated the language to be stronger here.

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,213 @@
...
+
+**ChainHTTPInjector(injector, injector) injector** 
+In order for the application layer to function correctly, Propagation choices must be syncronized between all processes in the distributed system, and multiple applications must be able to inject and extract their context into the same request. To meet these requirements, the Propagation API provides a function which registers a set of propagators, which will all be executed in order when the future calls to inject and extract are made. A canonical propagator consists of an inject and an extract function.
+
+**ChainHTTPExtractor(extractor, extractor) extractor** 
+In order for the application layer to function correctly, Propagation choices must be syncronized between all processes in the distributed system, and multiple applications must be able to inject and extract their context into the same request. To meet these requirements, the Propagation API provides a function which registers a set of propagators, which will all be executed in order when the future calls to inject and extract are made. A canonical propagator consists of an inject and an extract function.
+
+### Optional: Global Propagators
```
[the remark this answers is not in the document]
tedsuo: I tried to make it optional, but I'm now realizing this has to rethought, given named tracers.

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,213 @@
...
+
+**Baggage Context -** Transaction-level application data, meant to be shared with downstream components. This data is readable, and must be propagated in-band. Because of this, Baggage should be used sparingly, to avoid ballooning the size of RPC requests.
+
+Note that when possible, OpenTelemetry APIs calls are given access to the entire context object, and not a specific context type.
+
+
+## Context Management and in-process propagation
+
+In order for Context to function, it must always remain bound to the execution of code it represents. By default, this means that the programmer must pass a Context down the call stack as a function parameter. However, many languages provide automated context management facilities, such as thread locals. OpenTelemetry should leverage these facilities when available, in order to provide automatic context management.
```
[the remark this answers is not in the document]
tedsuo: Not for setting the active span in the current context, but when moving work from one thread to another, or context switching in async/nonblocking runtimes, this needs to be addressed. 

In general, I'm concerned that there will be important details which present themselves better when we attempt to implement this. We've gotten a bit of a pass so far because we did so much work in java at the very beginning of this process, before RFCs, etc. FWIW, in OpenTracing, we required a working implementation before adding an committing a change to the spec.

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,213 @@
...
+
+Since Baggage Context and Correlation Context appear very similar, why have two? 
+
+First and foremost, the intended uses for Baggage and Correlations are completely different. Secondly, the propagation requirements diverge significantly.
+
+Correlation values are solely to be used as labels for metrics and traces. By making Correlation data write-only, how and when it is transmitted remains undefined. This leaves the door open to optimizations, such as propagating some data out-of-band, and situations where sampling decisions may cease the need to propagate correlation context any further.
+
+Baggage values, on the other hand, are explicitly added in order to be accessed by downstream by other application code. Therefore, Baggage Context must be readable, and reliably propagated in-band in order to accomplish this goal.
+
+There may be cases where a key-value pair is propagated as TagMap for observability and as a Baggage for application specific use. AB testing is one example of such use case. There is potential duplication here at call site where a pair is created and also at propagation.
```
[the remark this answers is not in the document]
tedsuo: Eh sorry, poor proofreading on my part. Fixed now.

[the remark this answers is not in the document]
tedsuo: Nope!

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,213 @@
...
+
+Correlation values are solely to be used as labels for metrics and traces. By making Correlation data write-only, how and when it is transmitted remains undefined. This leaves the door open to optimizations, such as propagating some data out-of-band, and situations where sampling decisions may cease the need to propagate correlation context any further.
+
+Baggage values, on the other hand, are explicitly added in order to be accessed by downstream by other application code. Therefore, Baggage Context must be readable, and reliably propagated in-band in order to accomplish this goal.
+
+There may be cases where a key-value pair is propagated as TagMap for observability and as a Baggage for application specific use. AB testing is one example of such use case. There is potential duplication here at call site where a pair is created and also at propagation. 
+
+Solving this issue is not worth having semantic confusion with dual purpose. However, because all observability functions take the complete context as input, it may still be possible to use baggage values as labels.
```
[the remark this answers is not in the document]
tedsuo: Haha, that is true, it is a contradiction. I'm pointing out that it would be technically possible for the SDK to still do this. Maybe that is not a good idea. :)

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,213 @@
...
+
+
+## What about complex propagation behavior?
+
+Some OpenTelemetry proposals have called for more complex propagation behavior. For example, falling back to extracting B3 headers if W3C Trace-Context headers are not found. Chained propagators and other complex behavior can be modeled as implementation details behind the Propagator interface. Therefore, the propagation system itself does not need to provide chained propagators or other additional facilities.
```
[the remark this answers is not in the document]
tedsuo: Sorry, this language should be updated, now that the API has simple chaining. But, at the API level, we are only chaining together propagators for different applications. Complex details within one application – such as checking for Trace-Context and falling back to B3 if it is not present – does not need to be something handled at the API level. I was getting comments and questions about that kind of behavior, so I felt compelled to add this...

But, with the addition of named tracers, how a single inject/extract call can leverage multiple independent propagators needs to be thought about a bit more, since propagation behavior now depends on which tracer instance is used - or at least, I think that is what named tracers implies.

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,213 @@
...
+No. The concept of an explicit context is fundamental to a model where independent distributed applications share the same context propagation layer. How this context appears or is expressed is language specific, but it must be present in some form.
+
+
+# Prior art and alternatives
+
+Prior art:  
+* OpenTelemetry distributed context
+* OpenCensus propagators
+* OpenTracing spans
+* gRPC context
```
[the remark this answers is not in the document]
tedsuo: will do!

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,217 @@
...
+# OpenTelemetry Layered Architecture
+
+![drawing](img/context_propagation_explanation.png)
+
+Distributed tracing is an example of a cross-cutting concern, which requires non-local, transaction-level context propagation in order to execute correctly. Transaction-level context propagation can also be useful for other cross-cutting concerns, e.g., for security, versioning, and network switching. We refer to these types of cross-cutting concerns as **distributed applications**.
+
+OpenTelemetry is separated into an **application layer** and a **context propagation layer**. In this architecture, multiple distributed applications - including the observability and baggage systems provided by OpenTelemetry - share the same underlying context propagation system.
+
+
+# Application Layer
```
hodgesds: This RFC doesn't really explain how different applications would work, rather it goes into the implementation details of metrics and tracing "observability systems". It would be nice to have a better definition of what an application is in the application layer. Are all applications required to share a common interface?

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Explanation
+
+# OpenTelemetry Layered Architecture
+
+![drawing](img/context_propagation_explanation.png)
+
+Distributed tracing is an example of a cross-cutting concern, which requires non-local, transaction-level context propagation in order to execute correctly. Transaction-level context propagation can also be useful for other cross-cutting concerns, e.g., for security, versioning, and network switching. We refer to these types of cross-cutting concerns as **distributed applications**.
+
+OpenTelemetry is separated into an **application layer** and a **context propagation layer**. In this architecture, multiple distributed applications - including the observability and baggage systems provided by OpenTelemetry - share the same underlying context propagation system.
```
yurishkuro: From https://docs.google.com/document/d/1UxrEYOaQlF_E4gtiPoFmcZ4YKKe1GxohvCvQDuwvD1I/edit#

image

text/0042-separate-context-propagation.md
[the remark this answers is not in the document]
yurishkuro: there can be no separation in the propagation API for any propagators, not just metrics/tracing, but any custom/baggage propagators. The only thing the application needs to know is "I want to inject/extract context in this format".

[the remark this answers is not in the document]
yurishkuro: The notion of extractor/injector is irrelevant to the API for the application (we didn't have them in OpenTracing). Injector/extractor are only needed when registering them.

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,217 @@
...
+**RemoveBaggage(context, key) -> context**  
+To delete distributed state from an application, the Baggage API provides a function which takes a context, a key, and a value as input, and returns an updated context which contains the new value.
+
+**ClearBaggage(context) -> context**  
+To avoid sending baggage to an untrusted downstream process, the Baggage API provides a function remove all baggage from a context, 
+
+**GetHTTPExtractor() -> extractor**  
+To deserialize the state of the system sent from the the prior upstream process, the Baggage API provides a function which returns a HTTPExtract function. 
+
+**GetHTTPInjector() -> injector**
```
yurishkuro: I think it's misleading to combine injectors/extractors with the API for manipulating the baggage. I am not even sure there should be "get" methods for those - who would call that? The inter-process propagation layer (see my diagram above) sits below specific contexts, so they can register with it, but it shouldn't need to "call up"

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,217 @@
...
+Distributed applications propagate their state by data to downstream processes via injectors, functions which read and write application context into RPC requests. Each distributed application creates a set of propagators for every type of supported medium - currently only HTTP requests.
+
+**HTTPInject(context, request)**  
+To send the data for all distributed applications downstream to the next process, the Propagation API provides a function which takes a context and an HTTP request, and mutates the HTTP request to include an HTTP Header representation of the context.
+
+**HTTPExtract(context, request) -> context**  
+To receive data injected by prior upstream processes, the Propagation API provides a function which takes a context and an HTTP request, and returns context which represents the state of the upstream system.
+
+**ChainHTTPInjector(injector, injector) -> injector**  
+To allow multiple distributed applications to inject their context into the same request, the Propagation API provides a function which takes two injectors, and returns a single injector which calls the two original injectors in order.
```
yurishkuro: this is irrelevant, the application does not need to know if context systems are chained or what not, it only needs to say "I want to inject context". This is a low-level implementation detail of the propagation later.

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,217 @@
...
+
+Since Baggage Context and Correlation Context appear very similar, why have two? 
+
+First and foremost, the intended uses for Baggage and Correlations are completely different. Secondly, the propagation requirements diverge significantly.
+
+Correlation values are solely to be used as labels for metrics and traces. By making Correlation data write-only, how and when it is transmitted remains undefined. This leaves the door open to optimizations, such as propagating some data out-of-band, and situations where sampling decisions may cease the need to propagate correlation context any further.
+
+Baggage values, on the other hand, are explicitly added in order to be accessed by downstream by other application code. Therefore, Baggage Context must be readable, and reliably propagated in-band in order to accomplish this goal.
+
+There may be cases where a key-value pair is propagated as a Correlation for observability and as a Baggage item for application-specific use. AB testing is one example of such use case. This would result in extra overhead, as the same key-value pair would be present in two separate headers.
```
yurishkuro: I am not clear why we're making this point. Can't we allow telemetry sub-systems to access baggage? The metrics exported can be configured to read AB testing labels from baggage, not just from correlations - seems preferable to transmitting the same data twice.

text/0042-separate-context-propagation.md
```diff
@@ -0,0 +1,217 @@
...
+
+Correlation values are solely to be used as labels for metrics and traces. By making Correlation data write-only, how and when it is transmitted remains undefined. This leaves the door open to optimizations, such as propagating some data out-of-band, and situations where sampling decisions may cease the need to propagate correlation context any further.
+
+Baggage values, on the other hand, are explicitly added in order to be accessed by downstream by other application code. Therefore, Baggage Context must be readable, and reliably propagated in-band in order to accomplish this goal.
+
+There may be cases where a key-value pair is propagated as a Correlation for observability and as a Baggage item for application-specific use. AB testing is one example of such use case. This would result in extra overhead, as the same key-value pair would be present in two separate headers.  
+
+Solving this issue is not worth having semantic confusion with dual purpose. However, because all observability functions take the complete context as input – and baggage is not sampled – it may still be possible to use baggage values as labels for observability.
```
yurishkuro: This is not the first reference to sampling, and I find it highly confusing. Context is never sampled. Telemetry context may be _dropped_ due to bandwidth limitations (which is not mentioned here), but that's completely different from sampling.

text/0042-separate-context-propagation.md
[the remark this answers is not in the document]
yurishkuro: on chaining, I think it doesn't belong here (see earlier comment in the API section)

on named tracers - I would propose to ignore those. Propagators, like exporters, are 1-1 with TracerFactory, not with Tracer, i.e. all differently named tracers share the propagator. And the application that ultimately invokes the propagation API does it at the generic propagation layer, not via tracers.

text/0042-separate-context-propagation.md
tedsuo: niiiiice

tsloughter:
The name`hopLimit` makes it sound like an integer. 

Why is it not just a boolean named `propagate`?

tedsuo:
@yurishkuro @Oberon00 I've created a second draft of this proposal, based on all of this great feedback. Please find it here: #66

PR #49, opened 2019-09-13
OTEP 0049: LabelSet specification (to match current Metrics spec)

jmacd:
This proposal was discussed in the Sept 10, 2019 metrics call.


text/0000-metric-label-set.md
```diff
@@ -0,0 +1,123 @@
...
+requiredKeys := metric.DefineKeys("key1", "key2", "key3")
+labelSet := requiredKeys.DefineValues(meter, 1, 2, 3)
+```
+
+## Internal details
+
+Metric instruments are specified as SDK-independent objects, therefore metric handles were required to bind the instrument to the SDK in order to operate. In this proposal, `LabelSet` becomes responsible for binding the SDK at any call site where it is used.  Other than knowing the `Meter` that defined it, `LabelSet` is an opaque interface.  The three ways to use `LabelSet` in the metrics API are:
+
+```
+instrument.GetHandle(labels).Action(value)       // Action on a handle
```
toumorokoshi: are there still advantage for exposing a handles concept? I think as a user, passing in a labelset to a metric and having it resolve to a handle under the hood is a lot simpler than explicitly instantiating and incrementing handle objects.

jmacd: This has been emphasized as a requirement to achieve prometheus-level performance in critical sections of code (e.g., gRPC). I wrote "as fast as a few machine instructions" to capture this requirement.

I will add some discussion about the difference in cost of not using handles. I believe the "under the hood" computation is just a map-lookup where the key is a pair of (instrument ID, `LabelSet` ID). 

Some SDKs will prefer not to support this, but it's trivial to supply an implementation of these handles that call through to `RecordBatch(LabelSet, Measurement{Instrument, Value})` in ~50 lines of code.

text/0000-metric-label-set.md
```diff
@@ -0,0 +1,129 @@
...
+Metric instrument APIs which presently take labels in the form `{ Key: Value, ... }` will be updated to take an explicit `LabelSet`.  The `Meter.DefineLabels` API method supports getting a `LabelSet` from the SDK, allowing the programmer to pre-define labels without being required to manage handles.  This brings the number of ways to update a metric to three, via re-using a handle from `GetHandle()`:
+
+```
+cumulative := metric.NewFloat64Cumulative("my_counter", [
+					    "required_key1",
+					    "required_key2",
+					  ])
+labels := meter.DefineLabels({ "required_key1": value1,
+			       "required_key2": value2 })
+handle := cumulative.GetHandle(labels)
```
lzchen: With the introduction of labels as key-value-pairs, we don't have to support the implicit ordering of the keys in the instrument and values in the handle and match them correct?

jmacd: I wrote that it should be optional in `Ordered LabelSet Option` below. I don't think it's important, but some do.

text/0000-metric-label-set.md
```diff
@@ -0,0 +1,129 @@
...
+	      ... ])
+```
+
+This RFC proposes the new `LabelSet` concept is returned by an API named `Meter.DefineLabels({ Key: Value, ... })` which allows the SDK to potentially sort, canonicalize, or hash the set of labels once, allowing it to be re-used in several ways.  For example, the `RecordBatch` above can be written as one call to `DefineLabels` and inidividual operations.
+
+```
+Labels := meter.DefineLabels({ Key1: Value1,
+       	  		       Key2: Value2,
+			       ... })
+Cumulative.Add(Value, Labels)
```
bogdandrutu: I think we said that we want always the operation to happen on the Handle not on the metric.

jmacd: This proposal is to support both handles and operations by `LabelSet`. There were several users, myself among them, who would prefer not to use handles, and label sets can be re-used in scenarios where handles maybe cannot, so it's still an important optimization. An example is where you want to compute a labelset once and then operate on a number of metrics.  There is no advantage to constructing handles if you'll use them once, but the label set can still be re-used here.

Then, if you've allowed that `RecordBatch` can record a number of (metric, labelset, value) tuples, it's just a syntactic convenience to be able to write metric.Record(labelset, value) to generate the equivalent tuple.

jmacd:
I've updated the optional behavior names.
Counter default is Monotonic, option is NonMonotonic.
Gauge default is NonMonotonic, option is Monotonic.
Measure default is NonNegative, option is Signed.
Let's discuss.

jmacd:
This is a retroactive update to this OTEP to match the outcome from spec PR https://github.com/open-telemetry/opentelemetry-specification/pull/299. I do not believe there are any discrepancies between this and the current spec.

tigrannajaryan:
@jmacd as far as I see this is simply bringing the OTEP text to be on par with the current Specification text. Is this correct?

jmacd:
Yes. The situation was that there were several inter-related OTEPs and it was difficult to see what the spec would look like, so I went ahead and now I would like to finalize this. I have a few more edits to bring the OTEPs in-line with the spec, and then I will modify the OTEPs to reflect the debates that were ongoing when we closed v0.2.

jmacd:
Moreoever, this OTEP was never merged, so I have to merge something here. @SergeyKanzhelev has objecting to revising OTEPs, but this one isn't a revision, it's entirely original. Please review.

jmacd:
Or just please approve, since this matches the current spec.

text/0049-metric-label-set.md
```diff
@@ -0,0 +1,105 @@
+# Metric `LabelSet` specification
+
+**Status:** `proposed`
+
+Introduce a first-class `LabelSet` API type as a handle on a pre-defined set of labels for the Metrics API.
+
+## Motivation
+
+Labels are the term for key-value pairs used in the OpenTelemetry Metrics API.  Treatment of labels in the metrics API is especially important for performance, across a variety of export strategies.
```
dyladan:
```suggestion
Labels are the term for key-value pairs used in the OpenTelemetry Metrics API.  Treatment of labels in the Metrics API is especially important for performance across a variety of export strategies.
```

text/0049-metric-label-set.md
```diff
@@ -0,0 +1,105 @@
...
+
+Whether pushing or pulling metrics, whether aggregating metric events in the process or not, it is an expensive computation to translate a set of labels into either a serialized representation or a unique lookup key.  Users of the metrics API stand to benefit by re-using label sets whenever possible, because manipulating label sets often dominates the cost of processing metric events.
+
+The Metrics API supports three calling conventions: the Handle convention, the Direct convention, and the Batch convention. Each of these conventions stands to benefit when a `LabelSet` is re-used, as it allows the SDK to process the label set once instead of once per call.  Whenever more than one handle will be created with the same labels, whenever more than one instrument will be called directly with the same labels, and whenever more than one batch of metric events will be recorded with the same labels, re-using a `LabelSet` makes it possible for the SDK to improve performance.
```
dyladan:
```suggestion
The Metrics API supports three calling conventions: the Handle convention, the Direct convention, and the Batch convention. Each of these conventions stands to benefit when a `LabelSet` is re-used, as it allows the SDK to process the label set once instead of once per call.  Whenever more than one handle will be created with the same labels, more than one instrument will be called directly with the same labels, or more than one batch of metric events will be recorded with the same labels, re-using a `LabelSet` makes it possible for the SDK to improve performance.
```

text/0049-metric-label-set.md
```diff
@@ -0,0 +1,105 @@
...
+
+**Status:** `proposed`
+
+Introduce a first-class `LabelSet` API type as a handle on a pre-defined set of labels for the Metrics API.
+
+## Motivation
+
+Labels are the term for key-value pairs used in the OpenTelemetry Metrics API.  Treatment of labels in the metrics API is especially important for performance, across a variety of export strategies.
+
+Whether pushing or pulling metrics, whether aggregating metric events in the process or not, it is an expensive computation to translate a set of labels into either a serialized representation or a unique lookup key.  Users of the metrics API stand to benefit by re-using label sets whenever possible, because manipulating label sets often dominates the cost of processing metric events.
```
dyladan: This paragraph seems unnecessarily long and complex. It can be boiled down to something along the lines of:

```
Label serialization is often one of the most expensive tasks when processing metric events. Creating a `Handle` once and re-using it many times can greatly reduce the overall cost of processing many events.
```

text/0049-metric-label-set.md
```diff
@@ -0,0 +1,105 @@
...
+
+The Metrics API supports three calling conventions: the Handle convention, the Direct convention, and the Batch convention. Each of these conventions stands to benefit when a `LabelSet` is re-used, as it allows the SDK to process the label set once instead of once per call.  Whenever more than one handle will be created with the same labels, whenever more than one instrument will be called directly with the same labels, and whenever more than one batch of metric events will be recorded with the same labels, re-using a `LabelSet` makes it possible for the SDK to improve performance.
+
+## Explanation
+
+Metric instrument APIs which presently take labels in the form `{ Key: Value, ... }` will be updated to take an explicit `LabelSet`.  The `Meter.Labels()` API method supports getting a `LabelSet` from the SDK, allowing the programmer to acquire a pre-define label set.  Here are several examples of `LabelSet` re-use.  Assume we have two instruments:
```
dyladan:
```suggestion
Metric instrument APIs which presently take labels in the form `{ Key: Value, ... }` will be updated to take an explicit `LabelSet`.  The `Meter.Labels()` API method supports getting a `LabelSet` from the SDK, allowing the programmer to acquire a pre-defined label set.  Here are several examples of `LabelSet` re-use.  Assume we have two instruments:
```

text/0049-metric-label-set.md
jmacd: Accepted

text/0049-metric-label-set.md
```diff
@@ -0,0 +1,105 @@
...
+    ghandle = gauge.GetHandle(labels)
+)
+for ... {
+   // ...
+   chandle.Add(...)
+   ghandle.Set(...)
+}
+```
+
+Use a `LabelSet` to for multiple Direct calls:
```
MrAlias:
```suggestion
Use a `LabelSet` to make multiple Direct calls:
```

text/0049-metric-label-set.md
```diff
@@ -0,0 +1,105 @@
...
+    cumulative = metric.NewFloat64Cumulative("my_counter")
+    gauge      = metric.NewFloat64Gauge("my_gauge")
+)    
+```
+
+Use a `LabelSet` to construct multiple Handles:
+
+```golang
+var (
+    labels  = meter.Labels({ "required_key1": value1, "required_key2": value2 })
```
bogdandrutu: NewLabelSet for the name?

text/0049-metric-label-set.md
```diff
@@ -0,0 +1,105 @@
...
+## Internal details
+
+Metric SDKs that do not or cannot take advantage of the LabelSet optimizations are not especially burdened by having to support these APIs.  It is trivial to supply an implementation of `LabelSet` that simply stores a list of labels.  This may not be acceptable in performance-critical applications, but this is the common case in many metrics and diagnostics APIs today.
+
+## Trade-offs and mitigations
+
+In languages where overloading is a standard convenience, the metrics API may elect to offer alternate forms that elide the call to `Meter.Labels()`, for example:
+
+```
+instrument.GetHandle(meter, { Key: Value, ... })
```
bogdandrutu: I think this needs to be updated because instrument is now generated by the meter so no need to pass that.

text/0049-metric-label-set.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+as opposed to this:
+
+```
+instrument.GetHandle(meter.Labels({ Key: Value, ... }))
```
bogdandrutu: Can I simply have a version:
```
instrument.GetHandle({ Key: Value, ... })
```

bogdandrutu:
One comment:
What is the behavior if a LabelSet is created by one `Meter` instance and the instrument was created by other `Meter`? This is in case of a named meter/tracer.

text/0049-metric-label-set.md
jmacd: I have mixed feelings. I think this is an appropriate name in some languages, but I find it's quite a lot more readable to keep this short. I think `Labels()` is appropriate in Go, at least. I'm also thinking of how an implementation could decide not to allocate anything--for example a streaming SDK could simply record the set of labels and return an offset or an event ID. In that sense it's recording a new set of labels, but it's not returning anything new.

text/0049-metric-label-set.md
jmacd: Fixed.

text/0049-metric-label-set.md
jmacd: This is already stated in "Tradeoffs and Mitigations" below and appears in the (current spec)[https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-metrics-user.md#option-convenience-method-to-bypass-meterlabels].

text/0049-metric-label-set.md
jmacd: Also the spec has a broad declaration at the top stating "As in all OpenTelemetry specifications, these names are examples. Each language committee will decide on the appropriate names based on conventions in that language."

jmacd:
Your last question about named meters kind of made my head explode. I wish we had a standard term for the thing which is shared by all named meters, is that "provider", is that "engine", is that "SDK"? I would state that labels are usable by any meter from the same provider.

I'm adding this text
```
### Interaction with "Named" Meters

LabelSet values may be used with any named Meter originating from the
same Meter provider.  That is, LabelSets acquired through a named
Meter may be used by any Meter from the same Meter provider.
```

text/0049-metric-label-set.md
bogdandrutu: Was just for example purpose (to make it clear that a new label set is created). No strong feelings about this.

text/0049-metric-label-set.md
bogdandrutu: After removing the meter correct :). I missed this.

bogdandrutu:
This PR was opened for a long time and it has 3 official approver LGTM + few others coming from people interested in metrics. Entering God mode and merge this.

PR #51, opened 2019-09-16
WIP: Discuss approval for three metrics RFCs 0003, 0008, and 0009.

jmacd:
Numbers 0003, 0008, and 0009 are all related. This PR serves to discuss these RFCs and/or approve them.

### DIFF
--- a/text/0003-measure-metric-type.md
+++ b/text/0003-measure-metric-type.md
@@ -1,6 +1,6 @@
 # Consolidate pre-aggregated and raw metrics APIs
 
-**Status:** `proposed`
+**Status:** `accepted`
 
 # Foreword
 
--- a/text/0003-measure-metric-type.md
+++ b/text/0003-measure-metric-type.md
@@ -1,6 +1,6 @@
 # Consolidate pre-aggregated and raw metrics APIs
 
-**Status:** `accepted`
+**Status:** `approved`
 
 # Foreword
 
@@ -85,7 +85,7 @@ The specification will be updated with the following guidance.
 
 Likely to be the most common kind of metric, cumulative metric events express the computation of a sum.  Choose this kind of metric when the value is a quantity, the sum is of primary interest, and the event count and distribution are not of primary interest.  To raise (or lower) a cumulative metric, call the `Add()` method.
 
-If the quantity in question is always non-negative, it implies that the sum never descends.  This is the common case, where cumulative metrics only go up, and these _unidirectional_ cumulative metric instruments serve to compute a rate.  For this reason, cumulative metrics have a `Bidirectional` option to be declared as allowing negative inputs, the uncommon case.  The API will reject negative inputs to (default) unidirectional cumulative metrics, instead submitting an SDK error event, which helps ensure meaningful rate calculations.
+If the quantity in question is always non-negative, it implies that the sum never descends.  This is the common case, where cumulative metrics only go up, and these _monotonic_ cumulative metric instruments serve to compute a rate.  For this reason, cumulative metrics have a `Bidirectional` option to be declared as allowing negative inputs, the uncommon case.  The API will reject negative inputs to (default) monotonic cumulative metrics, instead submitting an SDK error event, which helps ensure meaningful rate calculations.
 
 For cumulative metrics, the default OpenTelemetry implementation exports the sum of event values taken over an interval of time.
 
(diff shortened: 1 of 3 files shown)


paivagustavo:
Hello @jmacd,  this is my first attempt to help the OpenTelemetry Community, so take all this with a **little grain of salt**.

I'm bringing here the discussions from the open-telemetry/opentelemetry-go/pull/100 pull request.

Initially, I was confused about the need of both `Bidirectional` and `Unidirectional` options and their naming. I had suggested to change it for something like `Monotonic` as this may suit better our context and is already used in the metrics universe (e.g., [prometheus](https://prometheus.io/docs/concepts/metric_types/#counter) uses it). Since it can be a boolean there is no need to have both options. A metric *is* or *is not* `Monotonic`.

Now that I have read this RFC, I started to think that these could be a little simpler. 

Why do we need to specify that a `Cumulative Metric` can or cannot be `Monotonic`?  Shouldn't we assume that all `Cumulative Metrics`are already `Monotonic`? If a metric can both go up and down shouldn't it be a `Gauge`? Moreover, Isn't a `Unidirectional Gauge` a `Monotonically Increasing Cumulative Metric`?

If we agree on those, we may revisit on which method each metric kind should have. Maybe `Gauges` could have a  `Add()` and `Sub()` method as well.

I haven't read all the accepted RFCs neither have been present on the meetings (which I'll start to attend), so if this has been already discussed or if I am missing some point, just tell me and I'll try to catch up with the project.

Oberon00:
Definitely in favor of "Monotonic" over Uni/Bidirectional. The former is a high school-level well-defined math term that describes exactly what is meant, while the latter seems to be our own invention in this context.

text/0003-measure-metric-type.md
```diff
@@ -1,6 +1,6 @@
 # Consolidate pre-aggregated and raw metrics APIs
 
-**Status:** `proposed`
+**Status:** `accepted`
```
Oberon00: accepted or approved? One of these is a typo.

jmacd: Fixed

jmacd:
@Oberon00 Yes, it was definitely a slip not to use Monotonic from the start. I've updated Unidirectional->Monotonic.

However, monotonic behavior is default for Counters while it is an option for Gauges, so Monotonic names the optional Gauge behavior and we need an opposite to name the Counter behavior.  Non-Monotonic is potentially an answer, but I took an informal poll of some folks around the office here and "BiDirectional" was still preferred. I'm open to suggestions.

jmacd:
I'm starting to prefer `NonMonotonic` as an option for counters.

jmacd:
I've updated the optional behavior names.
Counter default is Monotonic, option is NonMonotonic.
Gauge default is NonMonotonic, option is Monotonic.
Measure default is NonNegative, option is Signed.
Let's discuss.

jmacd:
There are two changes in https://github.com/open-telemetry/opentelemetry-specification/pull/250 relative to this PR:
(1) Disabled option has been removed
(2) "Required" label keys have been reworded as "Recommended" label keys. No mention of aggregation appears in the API spec at this point.

text/0003-measure-metric-type.md
```diff
@@ -124,9 +124,17 @@ The optional properties of a metric instrument are:
 |----------|-------------|-------------|
 | Required Keys | Determines labels that are always set on metric handles | All kinds |
 | Disabled | Indicates a verbose metric that does not report by default | All kinds |
```
bogdandrutu: I think this is removed?

text/0003-measure-metric-type.md
```diff
@@ -124,9 +124,17 @@ The optional properties of a metric instrument are:
...
-| Bidirectional | Indicates a cumulative metric instrument that goes up and down | Cumulative |
-| Unidirectional | Indicate a gauge that only ascends, for rate calculation | Gauge |
-| NonNegative | Indicates a measure that is never negative, for rate calculation | Measure |
+| NonMonotonic | Indicates a cumulative metric instrument that goes up and down | Cumulative |
+| Monotonic | Indicate a gauge that expresses a monotonic cumulative value | Gauge |
+| Signed | Indicates a measure that is positive or negative | Measure |
```
bogdandrutu: Can we remove the signed as we discussed in one of the meeting?

jmacd: It's a meaningful distinction to keep IMO.
Examples were provided in the referenced thread, e.g., for recording spanAsssasin scores. For recording the change in a bank account, there are just so many examples. 

Why should the implementation care? What value is there to the implementation in not having to support negative values? I can think of one. For computing summary metrics, i.e., quantile estimates, we need an approximation algorithm. It is significantly easier to approximate a "one-sided" distribution that has one boundary.  It's significantly easier to approximate a "two-sided" distribution that has no boundaries. If you're told the measure is non-negative you might choose to use a [DDSketch](https://arxiv.org/pdf/1908.10693.pdf).  If you're told the measure is signed, you might choose to use a [T-digest ](https://github.com/tdunning/t-digest/blob/master/docs/t-digest-paper/histo.pdf). 

I don't see this as being expensive to keep in, but I see it as a lack of symmetry if we exclude it.  We shouldn't be so confident that negative measures are not useful.

PR #59, opened 2019-10-11
Add OTLP Trace Data Format specification

tigrannajaryan:
This is a continuation of OTLP RFC proposal https://github.com/open-telemetry/oteps/pull/35

This change defines the data format used by Span and Resource messages.

The data format is a result of research of prior art (primarily OpenCensus and Jaeger),
as well as experimentation and benchmarking done as part of OTLP RFC proposal.

Go benchmark source code is available at https://github.com/tigrannajaryan/exp-otelproto
(use `make benchmark-encoding` target). Benchmarking shows that depending on the payload
composition this data format is about 4x-5x faster in encoding and 2x-3x faster in
decoding equivalent data compared to OpenCensus data format (all benchmarks in Go).

Notable differences from OpenCensus:

- Attribute key/value pairs are represented as a list rather than as a map.
  This results in significant performance gains and at the same time ~changes
  the semantic of attributes because now it is~ makes possible to have multiple attributes
  with the same key. This is also in-line with Jaeger's tags representation.

- Removed unnecessary wrappers such as google.protobuf.Timestamp which resulted in
  significant performance improvements for certain payload compositions (e.g. lots of
  TimedEvents).

- Resource labels use the same data type as Span attributes which now allows
  to have labels with other data types (OpenCensus only allowed strings).

### DIFF
--- a/text/0000-otlp-trace-data-format.md
+++ b/text/0000-otlp-trace-data-format.md
@@ -1,0 +1,361 @@
+# OTLP Trace Data Format
+
+_Author: Tigran Najaryan, Splunk_
+
+**Status:** `proposed`
+
+OTLP Trace Data Format specification describes the structure of the trace data that is transported by OpenTelemetry Protocol (RFC0035).
+
+## Motivation
+
+This document is a continuation of OpenTelemetry Protocol RFC0035 and is necessary part of OTLP specification.
+
+## Explanation
+
+OTLP Trace Data Format is primarily inherited from OpenCensus protocol. Several changes are introduced with the goal of more efficient serialization. Notable differences from OpenCensus protocol are:
+
+1. Removed `Node` as a concept.
+2. Extended `Resource` to better describe the source of the telemetry data.
+3. Replaced attribute maps by lists of key/value pairs.
+4. Eliminated unnecessary additional nesting in various values.
+
+Changes 1-2 are conceptual, changes 3-4 improve performance.
+
+## Internal details
+
+This section specifies data format in Protocol Buffers.
+
+### Resource
+
+```
+// Resource information. This describes the source of telemetry data.
+message Resource {
+  // Set of labels that describe the resource. See OpenTelemetry specification
+  // semantic conventions for standardized label names:
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md
+
+  repeated AttributeKeyValue labels = 3;
+  int32 dropped_labels_count = 11;
+}
+```
+
+### Span
+
+```
+// A span represents a single operation within a trace. Spans can be
+// nested to form a trace tree. Spans may also be linked to other spans
+// from the same or different trace. And form graphs. Often, a trace
+// contains a root span that describes the end-to-end latency, and one
+// or more subspans for its sub-operations. A trace can also contain
+// multiple root spans, or none at all. Spans do not need to be
+// contiguous - there may be gaps or overlaps between spans in a trace.
+//
+// The next id is 17.
+message Span {
+  // A unique identifier for a trace. All spans from the same trace share
+  // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes
+  // is considered invalid.
+  //
+  // This field is semantically required. Receiver should generate new
+  // random trace_id if empty or invalid trace_id was received.
+  //
+  // This field is required.
+  bytes trace_id = 1;
+
+  // A unique identifier for a span within a trace, assigned when the span
+  // is created. The ID is an 8-byte array. An ID with all zeroes is considered
+  // invalid.
+  //
+  // This field is semantically required. Receiver should generate new
+  // random span_id if empty or invalid span_id was received.
+  //
+  // This field is required.
+  bytes span_id = 2;
+
+  // This field conveys information about request position in multiple distributed tracing graphs.
+  // It is a list of Tracestate.Entry with a maximum of 32 members in the list.
+  //
+  // See the https://github.com/w3c/distributed-tracing for more details about this field.
+  message Tracestate {
+    message Entry {
+      // The key must begin with a lowercase letter, and can only contain
+      // lowercase letters 'a'-'z', digits '0'-'9', underscores '_', dashes
+      // '-', asterisks '*', and forward slashes '/'.
+      string key = 1;
+
+      // The value is opaque string up to 256 characters printable ASCII
+      // RFC0020 characters (i.e., the range 0x20 to 0x7E) except ',' and '='.
+      // Note that this also excludes tabs, newlines, carriage returns, etc.
+      string value = 2;
+    }
+
+    // A list of entries that represent the Tracestate.
+    repeated Entry entries = 1;
+  }
+
+  // The Tracestate on the span.
+  Tracestate tracestate = 3;
+
+  // The `span_id` of this span's parent span. If this is a root span, then this
+  // field must be empty. The ID is an 8-byte array.
+  bytes parent_span_id = 4;
+
+  // An optional resource that is associated with this span. If not set, this span
+  // should be part of a ResourceSpan that does include the resource information, unless resource
+  // information is unknown.
+  Resource resource = 5;
+
+  // A description of the span's operation.
+  //
+  // For example, the name can be a qualified method name or a file name
+  // and a line number where the operation is called. A best practice is to use
+  // the same display name at the same call point in an application.
+  // This makes it easier to correlate spans in different traces.
+  //
+  // This field is semantically required to be set to non-empty string.
+  // When null or empty string received - receiver may use string "name"
+  // as a replacement. There might be smarted algorithms implemented by
+  // receiver to fix the empty span name.
+  //
+  // This field is required.
+  string name = 6;
+
+  // Type of span. Can be used to specify additional relationships between spans
+  // in addition to a parent/child relationship.
+  enum SpanKind {
+    // Unspecified. Do NOT use as default.
+    // Implementations MAY assume SpanKind to be INTERNAL when receiving UNSPECIFIED.
+    SPAN_KIND_UNSPECIFIED = 0;
+
+    // Indicates that the span is used internally. Default value.
+    INTERNAL = 1;
+
+    // Indicates that the span covers server-side handling of an RPC or other
+    // remote network request.
+    SERVER = 2;
+
+    // Indicates that the span covers the client-side wrapper around an RPC or
+    // other remote request.
+    CLIENT = 3;
+
+    // Indicates that the span describes producer sending a message to a broker.
+    // Unlike client and  server, there is no direct critical path latency relationship
+    // between producer and consumer spans.
+    PRODUCER = 4;
+
+    // Indicates that the span describes consumer receiving a message from a broker.
+    // Unlike client and  server, there is no direct critical path latency relationship
+    // between producer and consumer spans.
+    CONSUMER = 5;
+  }
+
+  // Distinguishes between spans generated in a particular context. For example,
+  // two spans with the same name may be distinguished using `CLIENT` (caller)
+  // and `SERVER` (callee) to identify queueing latency associated with the span.
+  SpanKind kind = 7;
+
+  // The start time of the span. On the client side, this is the time kept by
+  // the local machine where the span execution starts. On the server side, this
+  // is the time when the server's application handler starts running.
+  //
+  // This field is semantically required. When not set on receive -
+  // receiver should set it to the value of end_time field if it was
+  // set. Or to the current time if neither was set. It is important to
+  // keep end_time > start_time for consistency.
+  //
+  // This field is required.
+  int64 start_time_unixnano = 8;
+
+  // The end time of the span. On the client side, this is the time kept by
+  // the local machine where the span execution ends. On the server side, this
+  // is the time when the server application handler stops running.
+  //
+  // This field is semantically required. When not set on receive -
+  // receiver should set it to start_time value. It is important to
+  // keep end_time > start_time for consistency.
+  //
+  // This field is required.
+  int64 end_time_unixnano = 9;
+
+  // The set of attributes. The value can be a string, an integer, a double
+  // or the Boolean values `true` or `false`. Note, global attributes like
+  // server name can be set as tags using resource API. Examples of attributes:
+  //
+  //     "/http/user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36"
+  //     "/http/server_latency": 300
+  //     "abc.com/myattribute": true
+  //     "abc.com/score": 10.239
+  repeated AttributeKeyValue attributes = 10;
+
+  // The number of attributes that were discarded. Attributes can be discarded
+  // because their keys are too long or because there are too many attributes.
+  // If this value is 0, then no attributes were dropped.
+  int32 dropped_attributes_count = 11;
+
+  // A time-stamped event in the Span.
+  message TimedEvent {
+    // The time the event occurred.
+    int64 time_unixnano = 1;
+
+    // A user-supplied name describing the event.
+    string name = 2;
+
+    // A set of attributes on the event.
+    repeated AttributeKeyValue attributes = 3;
+
+    int32 dropped_attributes_count = 4;
+  }
+
+  // A collection of `TimedEvent`s. A `TimedEvent` is a time-stamped annotation
+  // on the span, consisting of either user-supplied key-value pairs, or
+  // details of a message sent/received between Spans.
+  message TimedEvents {
+    // A collection of `TimedEvent`s.
+    repeated TimedEvent timed_event = 1;
+
+    // The number of dropped timed events. If the value is 0, then no events were dropped.
+    int32 dropped_timed_events_count = 2;
+  }
+
+  // The included timed events.
+  TimedEvents timed_events = 12;
+
+  // A pointer from the current span to another span in the same trace or in a
+  // different trace. For example, this can be used in batching operations,
+  // where a single batch handler processes multiple requests from different
+  // traces or when the handler receives a request from a different project.
+  message Link {
+    // A unique identifier of a trace that this linked span is part of. The ID is a
+    // 16-byte array.
+    bytes trace_id = 1;
+
+    // A unique identifier for the linked span. The ID is an 8-byte array.
+    bytes span_id = 2;
+
+    // The Tracestate associated with the link.
+    Tracestate tracestate = 3;
+
+    // A set of attributes on the link.
+    repeated AttributeKeyValue attributes = 4;
+
+    int32 dropped_attributes_count = 5;
+  }
+
+  // A collection of links, which are references from this span to a span
+  // in the same or different trace.
+  message Links {
+    // A collection of links.
+    repeated Link link = 1;
+
+    // The number of dropped links after the maximum size was enforced. If
+    // this value is 0, then no links were dropped.
+    int32 dropped_links_count = 2;
+  }
+
+  // The included links.
+  Links links = 13;
+
+  // An optional final status for this span. Semantically when Status
+  // wasn't set it is means span ended without errors and assume
+  // Status.Ok (code = 0).
+  Status status = 14;
+
+  // An optional number of child spans that were generated while this span
+  // was active. If set, allows an implementation to detect missing child spans.
+  google.protobuf.UInt32Value child_span_count = 15;
+}
+
+// The `Status` type defines a logical error model that is suitable for different
+// programming environments, including REST APIs and RPC APIs. This proto's fields
+// are a subset of those of
+// [google.rpc.Status](https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto),
+// which is used by [gRPC](https://github.com/grpc).
+message Status {
+  // The status code. This is optional field. It is safe to assume 0 (OK)
+  // when not set.
+  int32 code = 1;
+
+  // A developer-facing error message, which should be in English.
+  string message = 2;
+}
+```
+
+### AttributeKeyValue
+
+```
+message AttributeKeyValue {
+  enum ValueType {
+    STRING  = 0;
+    BOOL    = 1;
+    INT64   = 2;
+    FLOAT64 = 3;
+    BINARY  = 4;
+  };
+
+  string key = 1;
+  // The type of the value.
+  ValueType type = 2;
+  // A string up to 256 bytes long.
+  string string_value = 3;
+  // A 64-bit signed integer.
+  int64 int_value = 4;
+  // A Boolean value represented by `true` or `false`.
+  bool bool_value = 5;
+  // A double value.
+  double double_value = 6;
+  // A binary value of bytes.
+  bytes binary_value = 7;
+}
+
+```
+
+## Trade-offs and mitigations
+
+Timestamps were changed from google.protobuf.Timestamp to a int64 representation in Unix epoch nanoseconds. This change reduces the type-safety but benchmarks show that for small spans there is 15-20% encoding/decoding CPU speed gain. This is the right trade-off to make because encoding/decoding CPU consumption tends to dominate many workloads (particularly in OpenTelemetry Service).
+
+## Prior art and alternatives
+
+OpenCensus and Jaeger protocol buffer data schemas were used as the inspiration for this specification. OpenCensus was the starting point, Jaeger provided performance improvement ideas.
+
+## Open questions
+
+A follow up RFC is required to define the data format for metrics.
+
+## Appendix A - Benchmarking
+
+The following shows [benchmarking of encoding/decoding in Go](https://github.com/tigrannajaryan/exp-otelproto/) using various schemas.
+
+Legend:
+- OpenCensus    - OpenCensus protocol schema.
+- OTLP/AttrMap  - OTLP schema using map for attributes.
+- OTLP/AttrList - OTLP schema using list of key/values for attributes and with reduced nesting for values.
+- OTLP/AttrList/TimeWrapped - Same as OTLP/AttrList, except using google.protobuf.Timestamp instead of int64 for timestamps.
+
+Suffixes:
+- Attributes - a span with 3 attributes.
+- TimedEvent - a span with 3 timed events.
+
+```
+BenchmarkEncode/OpenCensus/Attributes-8         	      10	 605614915 ns/op
+BenchmarkEncode/OpenCensus/TimedEvent-8         	      10	1025026687 ns/op
+BenchmarkEncode/OTLP/AttrAsMap/Attributes-8     	      10	 519539723 ns/op
+BenchmarkEncode/OTLP/AttrAsMap/TimedEvent-8     	      10	 841371163 ns/op
+BenchmarkEncode/OTLP/AttrAsList/Attributes-8    	      50	 128790429 ns/op
+BenchmarkEncode/OTLP/AttrAsList/TimedEvent-8    	      50	 175874878 ns/op
+BenchmarkEncode/OTLP/AttrAsList/TimeWrapped/Attributes-8         	      50	 153184772 ns/op
+BenchmarkEncode/OTLP/AttrAsList/TimeWrapped/TimedEvent-8         	      30	 232705272 ns/op
+BenchmarkDecode/OpenCensus/Attributes-8                          	      10	 644103382 ns/op
+BenchmarkDecode/OpenCensus/TimedEvent-8                          	       5	1132059855 ns/op
+BenchmarkDecode/OTLP/AttrAsMap/Attributes-8                      	      10	 529679038 ns/op
+BenchmarkDecode/OTLP/AttrAsMap/TimedEvent-8                      	      10	 867364162 ns/op
+BenchmarkDecode/OTLP/AttrAsList/Attributes-8                     	      50	 228834160 ns/op
+BenchmarkDecode/OTLP/AttrAsList/TimedEvent-8                     	      20	 321160309 ns/op
+BenchmarkDecode/OTLP/AttrAsList/TimeWrapped/Attributes-8         	      30	 277597851 ns/op
+BenchmarkDecode/OTLP/AttrAsList/TimeWrapped/TimedEvent-8         	      20	 443386880 ns/op
+```
+
+The benchmark encodes/decodes 1000 batches of 100 spans, each span containing 3 attributes or 3 timed events. The total uncompressed, encoded size of each batch is around 20KBytes.
+
+The results show OTLP/AttrList is 5-6 times faster than OpenCensus in encoding and about 3 times faster in decoding.
+
+Using google.protobuf.Timestamp instead of int64-encoded unix timestamp results in 1.18-1.32 times slower encoding and 1.21-1.38 times slower decoding (depending on what the span contains).
--- a/text/0059-otlp-trace-data-format.md
+++ b/text/0059-otlp-trace-data-format.md
@@ -309,6 +309,8 @@ OpenCensus and Jaeger protocol buffer data schemas were used as the inspiration 
 
 A follow up RFC is required to define the data format for metrics.
 
+One of the original aspiring goals for OTLP was to _"support very fast pass-through mode (when no modifications to the data are needed), fast augmenting or tagging of data and partial inspection of data"_. This particular goal was not met directly (although performance improvements over OpenCensus encoding make OTLP more suitable for these tasks). This goal remains a good direction of future research and improvement.
+
 ## Appendix A - Benchmarking
 
 The following shows [benchmarking of encoding/decoding in Go](https://github.com/tigrannajaryan/exp-otelproto/) using various schemas.
--- a/text/0059-otlp-trace-data-format.md
+++ b/text/0059-otlp-trace-data-format.md
@@ -30,29 +30,31 @@ This section specifies data format in Protocol Buffers.
 ```
 // Resource information. This describes the source of telemetry data.
 message Resource {
-  // Set of labels that describe the resource. See OpenTelemetry specification
+  // labels is a list of attributes that describe the resource. See OpenTelemetry specification
   // semantic conventions for standardized label names:
   // https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md
+  repeated AttributeKeyValue labels = 1;
 
-  repeated AttributeKeyValue labels = 3;
-  int32 dropped_labels_count = 11;
+  // dropped_labels_count is the number of dropped labels. If the value is 0, then
+  // no labels were dropped.
+  int32 dropped_labels_count = 2;
 }
 ```
 
 ### Span
 
 ```
-// A span represents a single operation within a trace. Spans can be
+// Span represents a single operation within a trace. Spans can be
 // nested to form a trace tree. Spans may also be linked to other spans
-// from the same or different trace. And form graphs. Often, a trace
+// from the same or different trace and form graphs. Often, a trace
 // contains a root span that describes the end-to-end latency, and one
 // or more subspans for its sub-operations. A trace can also contain
 // multiple root spans, or none at all. Spans do not need to be
 // contiguous - there may be gaps or overlaps between spans in a trace.
 //
-// The next id is 17.
+// The next field id is 18.
 message Span {
-  // A unique identifier for a trace. All spans from the same trace share
+  // trace_id is the unique identifier of a trace. All spans from the same trace share
   // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes
   // is considered invalid.
   //
@@ -62,7 +64,7 @@ message Span {
   // This field is required.
   bytes trace_id = 1;
 
-  // A unique identifier for a span within a trace, assigned when the span
+  // span_id is a unique identifier for a span within a trace, assigned when the span
   // is created. The ID is an 8-byte array. An ID with all zeroes is considered
   // invalid.
   //
@@ -72,40 +74,35 @@ message Span {
   // This field is required.
   bytes span_id = 2;
 
-  // This field conveys information about request position in multiple distributed tracing graphs.
+  // TraceEntry is the entry that is repeated in tracestate field (see below).
+  message TraceEntry {
+    // key must begin with a lowercase letter, and can only contain
+    // lowercase letters 'a'-'z', digits '0'-'9', underscores '_', dashes
+    // '-', asterisks '*', and forward slashes '/'.
+    string key = 1;
+
+    // value is opaque string up to 256 characters printable ASCII
+    // RFC0020 characters (i.e., the range 0x20 to 0x7E) except ',' and '='.
+    // Note that this also excludes tabs, newlines, carriage returns, etc.
+    string value = 2;
+  }
+
+  // TraceState conveys information about request position in multiple distributed tracing graphs.
   // It is a list of Tracestate.Entry with a maximum of 32 members in the list.
   //
   // See the https://github.com/w3c/distributed-tracing for more details about this field.
-  message Tracestate {
-    message Entry {
-      // The key must begin with a lowercase letter, and can only contain
-      // lowercase letters 'a'-'z', digits '0'-'9', underscores '_', dashes
-      // '-', asterisks '*', and forward slashes '/'.
-      string key = 1;
+  repeated TraceEntry tracestate = 3;
 
-      // The value is opaque string up to 256 characters printable ASCII
-      // RFC0020 characters (i.e., the range 0x20 to 0x7E) except ',' and '='.
-      // Note that this also excludes tabs, newlines, carriage returns, etc.
-      string value = 2;
-    }
-
-    // A list of entries that represent the Tracestate.
-    repeated Entry entries = 1;
-  }
-
-  // The Tracestate on the span.
-  Tracestate tracestate = 3;
-
-  // The `span_id` of this span's parent span. If this is a root span, then this
-  // field must be empty. The ID is an 8-byte array.
+  // parent_span_id is the `span_id` of this span's parent span. If this is a root span, then this
+  // field must be omitted. The ID is an 8-byte array.
   bytes parent_span_id = 4;
 
-  // An optional resource that is associated with this span. If not set, this span
-  // should be part of a ResourceSpan that does include the resource information, unless resource
-  // information is unknown.
+  // resource that is associated with this span. Optional. If not set, this span
+  // should be part of a ResourceSpans message that does include the resource information,
+  // unless resource information is unknown.
   Resource resource = 5;
 
-  // A description of the span's operation.
+  // name describes the span's operation.
   //
   // For example, the name can be a qualified method name or a file name
   // and a line number where the operation is called. A best practice is to use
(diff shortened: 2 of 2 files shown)


z1c0:
> ... at the same time changes the semantic of attributes ...
This requires a dedicated PR against the specification repo.

Oberon00:
Or an dedicated OTEP, with a title like "Change SetAttribute to record all values instead of just the last one."

text/0000-otlp-trace-data-format.md
```diff
@@ -0,0 +1,361 @@
...
+// contiguous - there may be gaps or overlaps between spans in a trace.
+//
+// The next id is 17.
+message Span {
+  // A unique identifier for a trace. All spans from the same trace share
+  // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes
+  // is considered invalid.
+  //
+  // This field is semantically required. Receiver should generate new
+  // random trace_id if empty or invalid trace_id was received.
```
yurishkuro: I think they should reject the message on invalid trace_id

text/0000-otlp-trace-data-format.md
```diff
@@ -0,0 +1,361 @@
...
+  //
+  // This field is required.
+  bytes trace_id = 1;
+
+  // A unique identifier for a span within a trace, assigned when the span
+  // is created. The ID is an 8-byte array. An ID with all zeroes is considered
+  // invalid.
+  //
+  // This field is semantically required. Receiver should generate new
+  // random span_id if empty or invalid span_id was received.
```
yurishkuro: discard message on invalid span id

text/0000-otlp-trace-data-format.md
```diff
@@ -0,0 +1,361 @@
...
+// A span represents a single operation within a trace. Spans can be
+// nested to form a trace tree. Spans may also be linked to other spans
+// from the same or different trace. And form graphs. Often, a trace
+// contains a root span that describes the end-to-end latency, and one
+// or more subspans for its sub-operations. A trace can also contain
+// multiple root spans, or none at all. Spans do not need to be
+// contiguous - there may be gaps or overlaps between spans in a trace.
+//
+// The next id is 17.
+message Span {
```
yurishkuro: general recommendation - start each description with the name of the field/type (in the style of Go comments) instead of "This field" or "A (definition"). Given the volume of comments in front of each field it's easier to read when you start with the field/type name.

text/0000-otlp-trace-data-format.md
```diff
@@ -0,0 +1,361 @@
...
+  // random span_id if empty or invalid span_id was received.
+  //
+  // This field is required.
+  bytes span_id = 2;
+
+  // This field conveys information about request position in multiple distributed tracing graphs.
+  // It is a list of Tracestate.Entry with a maximum of 32 members in the list.
+  //
+  // See the https://github.com/w3c/distributed-tracing for more details about this field.
+  message Tracestate {
```
yurishkuro: * what is the value of explicitly parsing this instead of using the full `tracestate` header as a single string?
* what is the value of storing this in the span at all?

text/0000-otlp-trace-data-format.md
```diff
@@ -0,0 +1,361 @@
...
+
+    // A list of entries that represent the Tracestate.
+    repeated Entry entries = 1;
+  }
+
+  // The Tracestate on the span.
+  Tracestate tracestate = 3;
+
+  // The `span_id` of this span's parent span. If this is a root span, then this
+  // field must be empty. The ID is an 8-byte array.
```
yurishkuro: empty means null? not sure if "empty" is semantically clear

text/0000-otlp-trace-data-format.md
```diff
@@ -0,0 +1,361 @@
...
+  // A description of the span's operation.
+  //
+  // For example, the name can be a qualified method name or a file name
+  // and a line number where the operation is called. A best practice is to use
+  // the same display name at the same call point in an application.
+  // This makes it easier to correlate spans in different traces.
+  //
+  // This field is semantically required to be set to non-empty string.
+  // When null or empty string received - receiver may use string "name"
+  // as a replacement. There might be smarted algorithms implemented by
```
yurishkuro: > receiver may use string "name" as a replacement

-1 on magic strings. Empty string is better.

> There might be smarted algorithms

smarteR. What would be an example?

text/0000-otlp-trace-data-format.md
```diff
@@ -0,0 +1,361 @@
...
+  string name = 6;
+
+  // Type of span. Can be used to specify additional relationships between spans
+  // in addition to a parent/child relationship.
+  enum SpanKind {
+    // Unspecified. Do NOT use as default.
+    // Implementations MAY assume SpanKind to be INTERNAL when receiving UNSPECIFIED.
+    SPAN_KIND_UNSPECIFIED = 0;
+
+    // Indicates that the span is used internally. Default value.
```
yurishkuro: "INTERNAL ... is used internally" is a tautology. Is there a better description?

text/0000-otlp-trace-data-format.md
```diff
@@ -0,0 +1,361 @@
...
+
+    // Indicates that the span describes consumer receiving a message from a broker.
+    // Unlike client and  server, there is no direct critical path latency relationship
+    // between producer and consumer spans.
+    CONSUMER = 5;
+  }
+
+  // Distinguishes between spans generated in a particular context. For example,
+  // two spans with the same name may be distinguished using `CLIENT` (caller)
+  // and `SERVER` (callee) to identify queueing latency associated with the span.
```
yurishkuro: queueing or network

text/0000-otlp-trace-data-format.md
```diff
@@ -0,0 +1,361 @@
...
+  SpanKind kind = 7;
+
+  // The start time of the span. On the client side, this is the time kept by
+  // the local machine where the span execution starts. On the server side, this
+  // is the time when the server's application handler starts running.
+  //
+  // This field is semantically required. When not set on receive -
+  // receiver should set it to the value of end_time field if it was
+  // set. Or to the current time if neither was set. It is important to
+  // keep end_time > start_time for consistency.
```
yurishkuro: if it's semantically required, then the message should be rejected if the field is missing.

text/0000-otlp-trace-data-format.md
```diff
@@ -0,0 +1,361 @@
...
+  message Link {
+    // A unique identifier of a trace that this linked span is part of. The ID is a
+    // 16-byte array.
+    bytes trace_id = 1;
+
+    // A unique identifier for the linked span. The ID is an 8-byte array.
+    bytes span_id = 2;
+
+    // The Tracestate associated with the link.
+    Tracestate tracestate = 3;
```
yurishkuro: alternatively this could be stored as an attribute. I am not sure of the value this provides in the backend format.

text/0000-otlp-trace-data-format.md
```diff
@@ -0,0 +1,361 @@
...
+  Links links = 13;
+
+  // An optional final status for this span. Semantically when Status
+  // wasn't set it is means span ended without errors and assume
+  // Status.Ok (code = 0).
+  Status status = 14;
+
+  // An optional number of child spans that were generated while this span
+  // was active. If set, allows an implementation to detect missing child spans.
+  google.protobuf.UInt32Value child_span_count = 15;
```
yurishkuro: I would go with `int32`, seems overkill  to bring custom type just fo this field

text/0000-otlp-trace-data-format.md
```diff
@@ -0,0 +1,361 @@
...
+}
+
+// The `Status` type defines a logical error model that is suitable for different
+// programming environments, including REST APIs and RPC APIs. This proto's fields
+// are a subset of those of
```
yurishkuro: why a subset?

text/0000-otlp-trace-data-format.md
```diff
@@ -0,0 +1,361 @@
...
+
+  // A developer-facing error message, which should be in English.
+  string message = 2;
+}
+```
+
+### AttributeKeyValue
+
+```
+message AttributeKeyValue {
```
yurishkuro: may want to clarify that semantically this is `oneof`, but not using proto's `oneof` for perf or other reasons.

text/0000-otlp-trace-data-format.md
```diff
@@ -0,0 +1,361 @@
...
+    BINARY  = 4;
+  };
+
+  string key = 1;
+  // The type of the value.
+  ValueType type = 2;
+  // A string up to 256 bytes long.
+  string string_value = 3;
+  // A 64-bit signed integer.
+  int64 int_value = 4;
```
yurishkuro: for consistency with enum, may want to call this int64_value

text/0000-otlp-trace-data-format.md
```diff
@@ -0,0 +1,361 @@
...
+  // A Boolean value represented by `true` or `false`.
+  bool bool_value = 5;
+  // A double value.
+  double double_value = 6;
```
yurishkuro: for consistency with enum, may want to call this float64_value

text/0000-otlp-trace-data-format.md
tigrannajaryan: @yurishkuro I would argue that what you suggest is against robustness principle: https://en.wikipedia.org/wiki/Robustness_principle of which I am a fan and which has its share of critics. I would like to hear your thoughts. :-)

text/0000-otlp-trace-data-format.md
tigrannajaryan: same as above, let's discuss in one place.

text/0000-otlp-trace-data-format.md
yurishkuro: Speaking of SD principles to be a fan of, here's another one: fail fast. If the first receiver consistently rejects malformed messages, the sender has a better chance to figure out what they are doing wrong than if they have to dig up that message several layers down.

An invalid trace id means that even if we store the span, we won't be able to correlate it with the rest of the trace, making the data nearly useless.

There is one reason to store even the bad data, to be able to track it back to the source (bottom-up) and file a ticket against the team that generates it. The bad data needs to be clearly labeled in this case. For example, in Jaeger spans we have a section "warnings", and when we perform any data overrides we make a note in that list (with the original value).

In the end, I don't think it's a question for the data format spec, I'd expect this behavior to be configurable in the receiver: rejects, accept and clean up, or do 50/50 (to both warn the client and keep the data for bottom-up troubleshooting). But it's fine to have a recommendation in the data spec on *how* to do it (as you did).

tigrannajaryan: I tend to agree with you. I'd say robustness principle would apply more if there was a good way to substitute bad data with good data or interpret it in a way that makes most sense.
Replacing missing trace id with a new random id is probably not the best approach because we will silently lose the chance to detect the problem early.

I would say a likely better approach is to leave the field as is (empty or invalid), record the fact (e.g. in a telemetry counter or wherever else it makes sense) and in addition to that for example:

- If the use case is to forward the span (e.g. an Agent or Collector) then forward it as is.
- If however the use case requires specific interpretation of traceid (for example head sampling based on a hash of a traceid) then this should be considered bad input data and entire Span should be dropped/recorded as bad/etc.

Because the best possible course of action can be different for each use case I think it is best to remove the current recommendation to generate a new traceid and instead recommend to:

"The code that receives such Span SHOULD record the fact of invalid or missing traceid. If valid traceid and an interpretation of it is necessary for the use case then Span SHOULD be considered to be a case of bad input data otherwise the invalid or missing traceid MAY be ignored".

Thoughts?

text/0000-otlp-trace-data-format.md
tigrannajaryan: This is a good point. The structure is inherited from OC proto and I can see that for example OC Collector / OTel Collector decode/encode/translate this field from/to HTTP header however unclear _why_ they do it.
It seems unnecessary work since the field is never examined or modified (which would justify having it in structured form).
@bogdandrutu perhaps you have more info about why having this in structured form is useful?

text/0000-otlp-trace-data-format.md
tigrannajaryan: I will remove that mention since it is not important in this contect and is of historical interest only.

text/0000-otlp-trace-data-format.md
tigrannajaryan: I think it is better to keep it as double_value since the declared protobuf type is double.

text/0000-otlp-trace-data-format.md
tigrannajaryan: I will clarify.

text/0000-otlp-trace-data-format.md
tigrannajaryan: I agree. I will keep the requirement that this field must non-empty string, but if the receiver gets an empty string it should not replace it with anything and should be ready to deal with the empty string.

text/0000-otlp-trace-data-format.md
tigrannajaryan: I will add clarification.

text/0000-otlp-trace-data-format.md
yurishkuro: I still think being too rigid about what "must happen" is overkill at this stage. I would instead propose possible behaviors, but not dictate them:
  * the receiver MAY reject the invalid data and respond with the appropriate error code to the sender
  * the receiver MAY accept the invalid data and attempt to correct it. In this case the receiver SHOULD "log" the corrections it made (record where ???)

For the last part, I think it would be useful to provide a slot in the span model for these "logs" (as in https://github.com/jaegertracing/jaeger/blob/master/model/proto/model.proto#L136)

text/0000-otlp-trace-data-format.md
yurishkuro: it just disagrees with the enum FLOAT64

text/0000-otlp-trace-data-format.md
tigrannajaryan: Should be network, fixing.

text/0000-otlp-trace-data-format.md
tigrannajaryan: Replaced "empty" by "omitted".

text/0000-otlp-trace-data-format.md
tigrannajaryan: Changed as explained above.

text/0000-otlp-trace-data-format.md
tigrannajaryan: Clarified.

text/0000-otlp-trace-data-format.md
tigrannajaryan: Fixed.

text/0000-otlp-trace-data-format.md
tigrannajaryan: Removed.

text/0000-otlp-trace-data-format.md
tigrannajaryan: Fixed.

text/0000-otlp-trace-data-format.md
tigrannajaryan: Ah, sorry, that's what should be changed. I will fix.

tigrannajaryan: Fixed now.

text/0000-otlp-trace-data-format.md
tigrannajaryan: Done.

text/0000-otlp-trace-data-format.md
tigrannajaryan: "MAY" sounds reasonable to me. I will add them as possible recommendations. I will drop the "log" part since there are many ways to do it (log locally, record a metric, log in the attributes, etc) and it is hard to recommend single best universal way.

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,349 @@
...
+  // invalid.
+  //
+  // This field is semantically required. Receiver should generate new
+  // random span_id if empty or invalid span_id was received.
+  //
+  // This field is required.
+  bytes span_id = 2;
+
+  // TraceEntry is the entry that is repeated in tracestate field (see below).
+  message TraceEntry {
```
SergeyKanzhelev: why not tracestate? curious

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,349 @@
...
+  int32 dropped_links_count = 15;
+
+  // An optional final status for this span. Semantically when Status
+  // wasn't set it is means span ended without errors and assume
+  // Status.Ok (code = 0).
+  Status status = 16;
+
+  // An optional number of child spans that were generated while this span
+  // was active. If set, allows an implementation to detect missing child spans.
+  int32 child_span_count = 17;
```
Oberon00: As I read the description, this will only count "local children". I think that it would be even more valuable for the back end to know the number of potential "remote children", i.e. the number of times the Span's SpanContext was propagated (this is also one of the motivations for #58 "Unseal SpanContext", but if that was doable with the reference SDK, that would of course be even nicer).

text/0000-otlp-trace-data-format.md
Oberon00: > since the field is never examined or modified

I guess custom propagators might have logic that involves the actual trace state values.

text/0059-otlp-trace-data-format.md
Oberon00: Because it's a single entry in the tracestate. But maybe it could be named `TraceStateEntry`.

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,349 @@
...
+  //
+  // See the https://github.com/w3c/distributed-tracing for more details about this field.
+  repeated TraceEntry tracestate = 3;
+
+  // parent_span_id is the `span_id` of this span's parent span. If this is a root span, then this
+  // field must be omitted. The ID is an 8-byte array.
+  bytes parent_span_id = 4;
+
+  // resource that is associated with this span. Optional. If not set, this span
+  // should be part of a ResourceSpans message that does include the resource information,
```
Oberon00: I do not see the definition of a `ResourceSpans` message anywhere. I Imaginge it would just be
```protobuf
message ResourceSpans {
    Resource resource = 1;
    repeated Span spans = 2;
}
```

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,349 @@
...
+
+Changes 1-2 are conceptual, changes 3-4 improve performance.
+
+## Internal details
+
+This section specifies data format in Protocol Buffers.
+
+### Resource
+
+```
```
Oberon00:
```suggestion
```protobuf
```

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,349 @@
...
+
+  // dropped_labels_count is the number of dropped labels. If the value is 0, then
+  // no labels were dropped.
+  int32 dropped_labels_count = 2;
+}
+```
+
+### Span
+
+```
```
Oberon00:
```suggestion
```protobuf
```

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,349 @@
...
+  int32 code = 1;
+
+  // A developer-facing error message, which should be in English.
+  string message = 2;
+}
+```
+
+### AttributeKeyValue
+
+```
```
Oberon00:
```suggestion
```protobuf
```

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,349 @@
...
+  };
+
+  // key part of the key-value pair.
+  string key = 1;
+
+  // The type of the value.
+  ValueType type = 2;
+
+  // Only one of the following fields is supposed to contain data (determined by `type` field value).
+  // This is deliberately not using Protobuf `oneof` for performance reasons (verified by benchmarks).
```
Oberon00: It's sad that we have to work around deficiencies in the performance of the code generated by the Protobuf Go compiler (I assume that's what you used for benchmarking) here. I assume there is no fundamental reason why the `oneof` should be slower.

text/0059-otlp-trace-data-format.md
tigrannajaryan: It is defined in https://github.com/open-telemetry/oteps/pull/35/files

text/0059-otlp-trace-data-format.md
tigrannajaryan: Yes, Go Protobuf compiler. If I remember correctly it generates additional pointer indirection for oneof.

tigrannajaryan:
> I am curious about one of the stated requirements of OTLP, namely the augmenting and partial inspection. How are those achieved with proto format?

@yurishkuro 

There is no in-principle improvement in this area compared to OpenCensus. However because encoding and decoding are significantly faster one can argue that we made progress in this area.

I had more ambitious goals when I wrote the requirement and it was the reason I explored Capnproto and other encodings but ended up not going in that direction due to reasons outlined in #35.

I think this goal still holds and we can come up with a better protocol. What we have here is a good starting point for the first release though.

yurishkuro:
If you still think those additional goals are achievable in the future, we may want to list them in the otep as still-to-be-met.

text/0059-otlp-trace-data-format.md
tigrannajaryan: I will change to `TraceStateEntry`, it is a better name.

text/0000-otlp-trace-data-format.md
tigrannajaryan: Fixed.

text/0000-otlp-trace-data-format.md
tigrannajaryan: Fixed.

text/0000-otlp-trace-data-format.md
tigrannajaryan: Done.

text/0000-otlp-trace-data-format.md
tigrannajaryan: @Oberon00 can you please expand on that? I do not quite understand what a "custom propagator" is.

text/0059-otlp-trace-data-format.md
tigrannajaryan: Does the number of times SpanContext was propagated necessarily equal to the number of children created elsewhere? You may propagate once and create miltiple children, or propagate but do not create any children at all. I do not see how a number of propagations can reflect number of children or how you would otherwise be able to know how many child spans are generated somewhere else that you don't control.

tigrannajaryan:
> If you still think those additional goals are achievable in the future, we may want to list them in the otep as still-to-be-met.

@yurishkuro I added a note.

tigrannajaryan:
I changed the status to approved, so that this gets merged in approved status (when it gets merged). If I understand correctly this is the right flow for RFCs.

text/0059-otlp-trace-data-format.md
Oberon00: Yes it's true that this is only a "potential" number of children, and not even a maximum. After reading your arguments, I guess the only value in a propagation count would be to check if it is > 0. So maybe a `bool was_propagated` would be good?

Oberon00:
Actually, in the current flow you can remove the "Status" field entirely in the OTEP and PR merged=accepted.

text/0000-otlp-trace-data-format.md
Oberon00: Maybe I should have written "formatters" instead of "propagators", as I guess that's currently the customization point. E.g. imagine instead of the W3C TraceContext I need to use a custom propagation format. That format might require a "level" HTTP header in addition to the parent ID and trace flags. That level field would be a integer that is zero for a root span 1 for its remote children, 2 for their remote children and so on. To implement that, I would store that level in the TraceState and increment it each time the state is propagated using a custom HTTPTextFormat implementation.

Oberon00: What is more, at least in Python the TraceState is represented as the equivalent of a `HashMap<String, String>` anyway.

text/0000-otlp-trace-data-format.md
tigrannajaryan: @yurishkuro I spent a bit time thinking about this and I think it should not be OTLP's business to make recommendations about how exactly to behave for certain invalid input data, like this one. The best behavior depends on the specific use case.
For example if this is received by a backend that needs to display the span visually then it may reject the data or attempt to insert a substitute "correct" value just for displaying purposes (and then indicate visually that it is "corrected").
If this is received by an intermediary node like Collector then the Collector most likely doesn't care about this particular field. The Collector can forward the Span as is and let the subsequent receivers decide what to do about it (and the Collector may record the fact that it received bad data and from whom, but will still forward it).
So, what I think is appropriate in this specification is to define what is the correct data but refrain from prescribing a behavior when incorrect data is received.

text/0059-otlp-trace-data-format.md
bogdandrutu: @Oberon00 you can use a combination of Kind + this to determine if the number is correctly calculated or not. For the Kind Server/Consumer/Internal the number is correct, for the Client/Producer the number will be 0 but without having a response header that determines if the next service created or not a Span is hard (even harder for Producer where the Consumer may be async and this Span will be sent ended).

@tigrannajaryan: in opencensus this was an Uint32 message because we believed that having the ability to use not-set value was important for cases like Client/Producer. What do you think?

Oberon00: @bogdandrutu I feel you assign more meaning to SpanKind than it reallly has. Cases where your assumption breaks:
* There is a DB client span that has a HTTP client span as local child that implements executing the request.
* The client call fails before an attempt to propagate the context is even made (e.g. connection refused for HTTP request).
* A client call is made but propagation is explicitly disabled (for example an HTTP integration may offer an option to suppress propagation for URLs that are known to go to external services).

text/0000-otlp-trace-data-format.md
yurishkuro: My recommendation is to drop this field from this PR. I don't see a strong motivation/use case for keeping it in the out-of-band data format. Can always add it later.

text/0059-otlp-trace-data-format.md
tigrannajaryan: >in opencensus this was an Uint32 message because we believed that having the ability to use not-set value was important for cases like Client/Producer. What do you think?

@bogdandrutu yes, 0 in this case semantically is different from "not set" (which would mean we don't know). However I have a hard time imagining a use case where this difference matters.
In my mind I can imagine one good use case for `child_span_count` field: to indicate that there are missing children. For this I want to know that `child_span_count>0` and perhaps also show the actual counter in the UI (just to look nice). If the number is 0 or "not set" I will likely interpret it the same way, which is "I am not going to do anything special in this case".
Perhaps there is a good use case where this difference matters, if you have examples that would be great to know.

text/0000-otlp-trace-data-format.md
tigrannajaryan: @yurishkuro I am not sure I understand what "out-of-band" means in this context, please clarify.

I think the field itself is needed to be able to properly translate (without data loss) and forward spans received from HTTP-based formats (or any other formats that use w3c tracestate). If this field does not exist in this protocol how would such translation work? We would have nowhere to place the tracestate and would have to drop it.

I am thinking about Collector use case which may receive for example Zipkin HTTP with w3c tracestate in the header and forward in OTLP.

I may be missing something, please correct me.

--
I agree that it is most likely unnecessary to keep tracestate in a structured form here. It may be good enough to simply have an opaque byte array or a string (not even a map - as @Oberon00 mentioned is done in Python). So far I see no arguments in favor of any structured form. Any parties that are interested in the internal structure of tracestate are free to decode, modify and encode it according to w3c recommendation.

yurishkuro: In-band is the data that's propagated with the request in the headers, such as W3C trace context. That's where tracestate exists.

Out-of-hand is the performance data (spans) reported asynchronously, off the critical path of the request, to the tracing backend. Is there really a need for that data to include `tracestate`? (maybe there is, but it should be documented why).

tigrannajaryan: > Out-of-hand is the performance data (spans) reported asynchronously, off the critical path of the request, to the tracing backend. Is there really a need for that data to include `tracestate`? (maybe there is, but it should be documented why).

This is a good question, which I also would like to understand and which is unclear to me by reading w3c definition of `tracestate`. I raised an issue asking for a clarification and listed 2 use cases of how I think a tracing backend could use the `tracestate`: https://github.com/w3c/trace-context/issues/346

I would like to know though if my understanding is correct (and whether it is inline with w3c understanding).

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,346 @@
...
+This section specifies data format in Protocol Buffers.
+
+### Resource
+
+```protobuf
+// Resource information. This describes the source of telemetry data.
+message Resource {
+  // labels is a collection of attributes that describe the resource. See OpenTelemetry
+  // specification semantic conventions for standardized label names:
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md
```
SergeyKanzhelev: wrong link. It is a separate document

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,346 @@
...
+  // A Boolean value represented by `true` or `false`.
+  bool bool_value = 5;
+  // A double value.
+  double double_value = 6;
+}
+```
+
+## Trade-offs and mitigations
+
+Timestamps were changed from google.protobuf.Timestamp to a int64 representation in Unix epoch nanoseconds. This change reduces the type-safety but benchmarks show that for small spans there is 15-20% encoding/decoding CPU speed gain. This is the right trade-off to make because encoding/decoding CPU consumption tends to dominate many workloads (particularly in OpenTelemetry Service).
```
SergeyKanzhelev: wow. I wonder than why anybody uses `google.protobuf.Timestamp`?

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,346 @@
...
+  // key part of the key-value pair.
+  string key = 1;
+
+  // The type of the value.
+  ValueType type = 2;
+
+  // Only one of the following fields is supposed to contain data (determined by `type` field value).
+  // This is deliberately not using Protobuf `oneof` for performance reasons (verified by benchmarks).
+
+  // A string up to 256 bytes long.
```
SergeyKanzhelev: this is a new limitation. Should it be also mentioned in specs?

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,346 @@
...
+}
+
+// The `Status` type defines a logical error model that is suitable for different
+// programming environments, including REST APIs and RPC APIs.
+message Status {
+  // The status code. This is optional field. It is safe to assume 0 (OK)
+  // when not set.
+  int32 code = 1;
+
+  // A developer-facing error message, which should be in English.
```
SergeyKanzhelev: this is new comparing to the spec. Perhaps add a note here: https://github.com/open-telemetry/opentelemetry-specification/blob/485e6a5cd6290a9ef1eda20a6e68de897548ddc4/specification/api-tracing.md#set-status

SergeyKanzhelev: I mean that it should be in English

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,346 @@
...
+  repeated TimedEvent timed_events = 12;
+
+  // dropped_timed_events_count is the number of dropped timed events. If the value is 0,
+  // then no events were dropped.
+  int32 dropped_timed_events_count = 13;
+
+  // Link is a pointer from the current span to another span in the same trace or in a
+  // different trace. For example, this can be used in batching operations,
+  // where a single batch handler processes multiple requests from different
+  // traces or when the handler receives a request from a different project.
```
SergeyKanzhelev: You can also point to the overview document describing links

text/0000-otlp-trace-data-format.md
SergeyKanzhelev: there are few considerations:

1. `tracestate` is defined as something that is needed fro tracing system to operate. Like `cluster-id` where to place the trace information to. Quick access via separate field is beneficial here.
2. `tracestate` is coming from external system. And can cause privacy and security concerns. Separate field makes it easy to strip out if needed
3. Same as above if vendor only wants to preserve it's own tracestate values on recording

text/0059-otlp-trace-data-format.md
tigrannajaryan: Fixed.

text/0059-otlp-trace-data-format.md
tigrannajaryan: 1. google.protobuf.Timestamp supports larger date range, but I think we are fine with supporting only for the next 240 years in this protocol. This is plenty of time for our successors to come up with a better protocol :-)
2. google.protobuf.Timestamp value also can be omitted, unspecified, while timestamp in nanoseconds cannot (if you omit it you will get 0 value which is a valid timestamp - so cannot differentiate a valid 0 timestamp from missing timestamp). I believe this is also a non-issue for traces since trace timestamps are mandatory.

text/0059-otlp-trace-data-format.md
tigrannajaryan: I copied this from OpenCensus proto, but I wonder if we need it in protobuf definition? If there is a need for limitation I think it can be in SDK, doesn't have to be here. I am willing to remove this and allow unlimited length strings. What do you think?

text/0059-otlp-trace-data-format.md
tigrannajaryan: I rephrased and removed reference to English, please have a look again.

text/0059-otlp-trace-data-format.md
tigrannajaryan: Done.

SergeyKanzhelev:
@open-telemetry/specs-approvers as was discussed today at SIG meeting, this is ready to be merged. Please review and approve!

text/0059-otlp-trace-data-format.md
SergeyKanzhelev: go for it

tigrannajaryan: Length limitation removed.

text/0059-otlp-trace-data-format.md
tigrannajaryan: I am resolving this since there are no further objections.

text/0000-otlp-trace-data-format.md
tigrannajaryan: To make sure we are not stuck on this single item I suggest that we merge the PR and I will address this in a separate PR.
I am resolving this comment, if you disagree please reopen.

tmc:
I think this discussion from the other repository is relevant: https://github.com/open-telemetry/opentelemetry-proto/issues/11#issuecomment-525972693

tigrannajaryan:
> I think this discussion from the other repository is relevant: [open-telemetry/opentelemetry-proto#11 (comment)](https://github.com/open-telemetry/opentelemetry-proto/issues/11#issuecomment-525972693)

@tmc adding support for nested attribute values is easy. As soon as API/Spec decides to have such semantics the `AttributeKeyValue` in this proposal can be extended in backward compatible manner.

tigrannajaryan:
@open-telemetry/specs-approvers please review. I believe there are no more outstanding issues and this is ready to be merged.

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,348 @@
...
+
+### Resource
+
+```protobuf
+// Resource information. This describes the source of telemetry data.
+message Resource {
+  // labels is a collection of attributes that describe the resource. See OpenTelemetry
+  // specification semantic conventions for standardized label names:
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-resource-semantic-conventions.md
+  repeated AttributeKeyValue labels = 1;
```
yurishkuro: Why do we need to introduce the term "labels"? It feels confusing and adding little value over "attributes".

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,348 @@
...
+// multiple root spans, or none at all. Spans do not need to be
+// contiguous - there may be gaps or overlaps between spans in a trace.
+//
+// The next field id is 18.
+message Span {
+  // trace_id is the unique identifier of a trace. All spans from the same trace share
+  // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes
+  // is considered invalid.
+  //
+  // This field is semantically required. If empty or invalid trace_id was received:
```
yurishkuro:
```suggestion
  // This field is semantically required. If empty or invalid trace_id is received:
```

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,348 @@
...
+  // - The receiver MAY reject the invalid data and respond with the appropriate error
+  //   code to the sender.
+  // - The receiver MAY accept the invalid data and attempt to correct it.
+  bytes trace_id = 1;
+
+  // span_id is a unique identifier for a span within a trace, assigned when the span
+  // is created. The ID is an 8-byte array. An ID with all zeroes is considered
+  // invalid.
+  //
+  // This field is semantically required. If empty or invalid span_id was received:
```
yurishkuro:
```suggestion
  // This field is semantically required. If empty or invalid span_id is received:
```

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,348 @@
...
+  // end_time_unixnano is the end time of the span. On the client side, this is the time
+  // kept by the local machine where the span execution ends. On the server side, this
+  // is the time when the server application handler stops running.
+  //
+  // This field is semantically required and it is expected that end_time >= start_time.
+  //
+  // This field is required.
+  int64 end_time_unixnano = 9;
+
+  // attributes is a collection of attribute key/value pairs. The value can be a string,
```
yurishkuro:
```suggestion
  // attributes is a collection of key/value pairs enriching the semantic meaning of the span. The value can be a string,
```

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,348 @@
...
+  // an integer, a double or the Boolean values `true` or `false`. Note, global attributes
+  // like server name can be set as tags using resource API. Examples of attributes:
```
yurishkuro: can we avoid extra terminology?

```suggestion
  // like server name can be set using the resource API. Examples of attributes:
```

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,348 @@
...
+  //     "/http/user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36"
+  //     "/http/server_latency": 300
+  //     "abc.com/myattribute": true
+  //     "abc.com/score": 10.239
+  repeated AttributeKeyValue attributes = 10;
+
+  // dropped_attributes_count is the number of attributes that were discarded. Attributes
+  // can be discarded because their keys are too long or because there are too many
+  // attributes. If this value is 0, then no attributes were dropped.
+  int32 dropped_attributes_count = 11;
```
yurishkuro: feels like this should be DRYed into a `struct { attributes, dropped_count }` that can be reused in other entities like Resource, TimedEvent. The naming will be odd though, `span.attributes.attributes`

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,348 @@
...
+  // dropped_links_count is the number of dropped links after the maximum size was
+  // enforced. If this value is 0, then no links were dropped.
+  int32 dropped_links_count = 15;
+
+  // status is an optional final status for this span. Semantically when status
+  // wasn't set it is means span ended without errors and assume Status.Ok (code = 0).
+  Status status = 16;
+
+  // child_span_count is an optional number of child spans that were generated while this
+  // span was active. If set, allows an implementation to detect missing child spans.
```
yurishkuro: this is unclear, per resolved thread below. Is this the number of local-only children? If so, the doc string should be explicit about it.

text/0059-otlp-trace-data-format.md
```diff
@@ -0,0 +1,348 @@
...
+  int32 child_span_count = 17;
+}
+
+// The `Status` type defines a logical error model that is suitable for different
+// programming environments, including REST APIs and RPC APIs.
+message Status {
+  // The status code. This is optional field. It is safe to assume 0 (OK)
+  // when not set.
+  int32 code = 1;
```
yurishkuro: I don't quite follow this one. If there is no prescribed enumeration of the possible values for `code`, then the field seems to be of little value. The main value of "status" to me if it was normalized and independent of the actual RPC API.

yurishkuro: NB: in OpenTracing this was really just `error=true` tag. The actual RPC error code would be recorded as another tag, like `http.status_code` (the tag would be specific to the RPC protocol).

text/0059-otlp-trace-data-format.md
tigrannajaryan: This is not new in this proposal. "labels" are the term used in the specification currently: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-resource.md

I agree it would likely make sense to make this uniform. I created an issue for that https://github.com/open-telemetry/opentelemetry-specification/issues/342

text/0059-otlp-trace-data-format.md
tigrannajaryan: I agree it would be more DRY, but unfortunately it has measurable performance impact.

bogdandrutu: Should we use fixed64?

text/0059-otlp-trace-data-format.md
tigrannajaryan: Agree. I will add an enum that mirrors the spec: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#statuscanonicalcode

text/0059-otlp-trace-data-format.md
tigrannajaryan: Added clarification to make it explictely about local children.

text/0059-otlp-trace-data-format.md
tigrannajaryan: Enum added.

text/0059-otlp-trace-data-format.md
tigrannajaryan: I tried fixed32 and don't see any measurable difference in speed (only slight increase of uncompressed wire size).

I can change to fixed if there is any argument in favour of it.

text/0059-otlp-trace-data-format.md
tigrannajaryan: Resolving, issue created to track this separately.

tigrannajaryan:
@open-telemetry/specs-approvers do I need more approvals to get this merged? There are no outstanding issues or comments in this PR anymore. I believe it is ready to be merged.

tigrannajaryan:
@bogdandrutu thank you.

Oberon00:
@tigrannajaryan What is the timeline for integrating this into opentelemetry-proto?

Oberon00:
@tigrannajaryan It seems you are currently working on applying the OTEP bit by bit to opentelemetry-ptoro. Is there an overview somewhere which parts are applied and which are missing?

tigrannajaryan:
@Oberon00 yes, I am doing it now. No, there is no overview, I don't have a formal list of changes anywhere, however the target is well defined, if there is an interest on how much progress is made you can diff OTEP and proto repo content.

PR #60, opened 2019-10-22
add track current tracer in context otp

tsloughter:
Sorry this is quick and dirty, I wanted to get it out before the Spec SIG sync this morning so it can be discussed.

### DIFF
--- a/text/0060-context-current-tracer.md
+++ b/text/0060-context-current-tracer.md
@@ -1,0 +1,31 @@
+# Track Current Tracer in Context
+
+Add the requirement that the tracer used to create a span be tracked in the current context.
+
+## Motivation
+
+Without explicitly passing a `Tracer` variable to methods and functions any spans created do not know what tracer was used to create their parent. If tracers are meant to provide separate functionality and configuration then choosing a tracer at, for example, the root of an HTTP request is possibly the tracer the user wants to be used by any library the handler calls out to, such as a database library, without requiring the database library to modify its API to accept a tracer as an argument.
+
+## Explanation
+
+The tracer to start a span with is first fetched from a registry of tracers or newly created. If no explicit tracer is fetched, the tracer that is configured as the default is returned unless there is an active span in which case the tracer used to create that span is returned.
+
+## Internal details
+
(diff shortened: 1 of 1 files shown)


tsloughter:
@yurishkuro right. I hoped to get some clarification on that in the meeting today. I was hoping it would be that named tracers could essentially name another tracer. If a vendor has a specific tracer for you to use but the database instrumentation uses its own "named tracer" there must be a way for those two to work together where you are giving a name to the vendors tracer?

yurishkuro:
All tracers come from the same factory, I don't see a problem. Named tracer is just an extra namespace, not a different implementation/vendor.

tsloughter:
I don't know what that means. Is the factory the thing that is vendor specific? Can there be only 1 global factory, which will return the vendor's tracer or the noop tracer or the sdk tracer etc, and that is why there is no need to track it in the context?

Oberon00:
I think that the OTEP would make perfect sense when replacing Tracer with TraceFactory.
But I would suggest another alternative: The TracerFactory becomes a property of the Span (related https://github.com/open-telemetry/opentelemetry-specification/issues/21).

yurishkuro:
> Is the factory the thing that is vendor specific? 

yes

> Can there be only 1 global factory

There better be. There is no requirement that tracers from different vendors must interoperate. Duplicating the factory from the same vendor is possible, but wasteful in terms of resources.

yurishkuro:
@Oberon00 I would be opposed to having tracer factory accessible from the span, just like I was opposed to exposing the tracer - it only leads to bad programming practices. Your point (https://github.com/open-telemetry/opentelemetry-specification/issues/21#issuecomment-538387355) was spot on.

tsloughter:
I don't think it has been decided if a span created with a noop tracer should mean its children are noop'ed yet, but based on this information I would then expect a NoopTracerFactory so that children will also use the noop when started.

SergeyKanzhelev:
@tsloughter I'm closing this PR as the motivation for this change was challenged and need re-work. I understand there are discussions on how to get different `Resource`s or `TracerFactory`s from the current context. This is happening as a part of context separation working group. Thank you for writing this OTEP. 

Please feel free to re-open if you ready to adjust motivational part for this OTEP. Or restart as a separate OTEP

PR #65, opened 2019-11-12
Use gRPC status code to indicate retryable and not retryable errors

tigrannajaryan:
OTLP proposal originally used a separate ResultCode enumeration for server
to tell the client whether the failed request can be retried or no.

After discussion here https://github.com/open-telemetry/opentelemetry-proto/pull/47#discussion_r345262430
it became clear that the goal can be achieved using gRPC status codes without
a need for custom enumeration.

This change removes the ResultCode and explains how to use gRPC status codes.


text/0035-opentelemetry-protocol.md
```diff
@@ -82,23 +82,76 @@ If the client is shutting down (e.g. when the containing process wants to exit)
...
+```
+
+The server MAY use other gRPC codes to indicate retryable and not-retryable errors if those other gRPC codes are more appropriate for a particular erroneous situation. The client SHOULD interpret gRPC status codes as retryable or not-retryable according to the following table:
+
+|gRPC Code|Retryable?|
+|---------|----------|
+|CANCELLED|Yes|
+|UNKNOWN|No|
+|INVALID_ARGUMENT|No|
+|DEADLINE_EXCEEDED|No|
```
bogdandrutu: This should be retryable because it means you hit a high loaded server?

tigrannajaryan: I am torn on this one. Here is what the code says:
>The deadline expired before the operation could complete. For operations that change the state of the system, this error may be returned even if the operation has completed successfully. For example, a successful response from a server could have been delayed long

So it says _"this error may be returned even if the operation has completed successfully"_. This means that if we retry we may create duplicates of data and potentially this can repeat for many times if the reason for DEADLINE_EXCEEDED is because of the specific data that we send (e.g. data is composed such that it takes a long time to process). We may end up in infinite loop retrying the same data over and over again while the data is actually successfully processed.

Thoughts?

bogdandrutu: There are multiple scenarios in which deadline exceeded can be returned:
* Request did get to the server, request got processed, and got canceled before sending the response.
* Request did get to the server, and got canceled half way through processing the request.
* Request did get to the server, and got canceled before processing the request (waiting in the processing queue).
* Request did not get to the server (waiting in client sending queue) and got canceled.

I think in at least half of these cases we want to retry with a backoff.

tigrannajaryan: Yeah, as I mentioned I am torn on this one. If you feel strong that it should be retryable then we can mark it as retryable, but we need it to be marked one way or another. We can't have both unless we require extra details with the status and I think we shouldn't since it still may be returned without status details and we still need to make a decision.
I am fine with settling on making DEADLINE_EXCEEDED retryable. In practice I don't think this is going to be a big deal and if it is we can rethink it.

jmacd: Marking this retryable seems fine to me. Maybe it's still treated as a special case. There's nothing about DEADLINE_EXCEEDED that says you shouldn't try again, it suggests it may fail again though. The server should do it's best to not to let this response through by some sort of early failure, right?

tigrannajaryan: I changed DEADLINE_EXCEEDED to be retryable.

tigrannajaryan:
@open-telemetry/specs-approvers please have a look. We need this correction to finalize the protocol.

tigrannajaryan:

bogdandrutu:
Please review @open-telemetry/specs-approvers

bogdandrutu:
Please review @open-telemetry/specs-approvers

text/0035-opentelemetry-protocol.md
```diff
@@ -82,23 +82,76 @@ If the client is shutting down (e.g. when the containing process wants to exit)
...
+|gRPC Code|Retryable?|
+|---------|----------|
+|CANCELLED|Yes|
+|UNKNOWN|No|
+|INVALID_ARGUMENT|No|
+|DEADLINE_EXCEEDED|Yes|
+|NOT_FOUND|No|
+|ALREADY_EXISTS|No|
+|PERMISSION_DENIED|Yes|
+|UNAUTHENTICATED|Yes|
```
bogdandrutu: Should these PERMISSION_DENIED and UNAUTHENTICATED be retryable? It is clear that you need to change the request otherwise this error will happen again, so probably I wouldn't say that the request is retryable.

tigrannajaryan: If we make this non-retryable then a misconfiguration (e.g. wrong auth token) will result in OTLP dropping all data that we attempt to send. Hard to tell what is the best approach here. Do we want the sender to keep accumulating the data and retrying it until someone notices the problem and fixes the auth? Or do we want the sender to begin dropping data until someone notices the problem and fixes the auth? I have a weak preferences for retrying in this case, but happy to hear other opinions.

Oberon00:
I'm against using grpc codes to encode retryability. You can argue about the retryability of each single error code. I'm fine with adding an grpc code as additional information (though I think a string with an error message would be even better) to boolean `isRetryable`.

tigrannajaryan:
> I'm against using grpc codes to encode retryability. You can argue about the retryability of each single error code. I'm fine with adding an grpc code as additional information (though I think a string with an error message would be even better) to boolean `isRetryable`.

I agree that arguing about each single error is unfortunate. The problem is that we have to do it regardless. The protocol uses gRPC and you *will* get a gRPC code in response to the call and you *have* to make a decision about what to do for each particular code even if we decide to use explicit `isRetryable` boolean.

We can't escape this discussion unless we choose to ignore error handling.

tigrannajaryan:
@open-telemetry/specs-approvers Please review. We are not going to have OTLP protocol for Beta unless this has a quick resolution.

tigrannajaryan:
Any more thoughts on this? We need to approve or reject this (unclear what to do if rejected) to implement OTLP.

text/0035-opentelemetry-protocol.md
```diff
@@ -82,23 +82,76 @@ If the client is shutting down (e.g. when the containing process wants to exit)
...
 If the client is unable to deliver a certain request (e.g. a timer expired while waiting for acknowledgements) the client SHOULD record the fact that the data was not delivered.
 
-#### Result Code
+#### Export Response
 
-`Export` response includes a `result_code` field, which indicates whether the server was able to successfully process the received data. Possible values for `result_code` field are:
+The server may respond with either a success or an error to export requests.
 
-- `Success` - telemetry data is successfully processed by the server. If the server receives an empty request (a request that does not carry any telemetry data) the server SHOULD respond with `Success`.
+The success response indicates telemetry data is successfully processed by the server. If the server receives an empty request (a request that does not carry any telemetry data) the server SHOULD respond with success.
```
arminru: Should this say that the data _was processed_ by the server or that the server successfully accepted it for processing (but did not necessarily finish yet)?

text/0035-opentelemetry-protocol.md
```diff
@@ -242,22 +295,7 @@ message MetricExportRequest {
...
-    // cannot be deserialized or otherwise processed by the server.
-    FailedNotRetryable = 1;
-    
-    // Processing of telemetry data failed. The client SHOULD record the 
-    // error and may retry exporting the same data after some time. This 
-    // for example can happen when the server is overloaded.
-    FailedRetryable = 2;
-  }
-  ResultCode result_code = 2;
+    // Response in an empty message.
```
arminru: I'm neither an expert in protobuf nor gRPC so I'm asking naively: Is this just a technical necessity so that an empty response is rendered to carry the response code or is this intended as a placeholder for future use?

text/0035-opentelemetry-protocol.md
```diff
@@ -82,23 +82,76 @@ If the client is shutting down (e.g. when the containing process wants to exit)
...
+#### Export Response
 
-`Export` response includes a `result_code` field, which indicates whether the server was able to successfully process the received data. Possible values for `result_code` field are:
+The server may respond with either a success or an error to export requests.
 
-- `Success` - telemetry data is successfully processed by the server. If the server receives an empty request (a request that does not carry any telemetry data) the server SHOULD respond with `Success`.
+The success response indicates telemetry data is successfully processed by the server. If the server receives an empty request (a request that does not carry any telemetry data) the server SHOULD respond with success.
 
-- `FailedNotRetryable` - processing of telemetry data failed. The client MUST NOT retry sending the same telemetry data. The telemetry data MUST be dropped. This for example can happen when the request contains bad data and cannot be deserialized or otherwise processed by the server. The client SHOULD maintain a counter of such dropped data.
+When using gRPC transport success response is returned via `ExportResponse` message.
```
arminru: Is using anything other than gRPC not supported/possible or just not scope of the OTLP spec?

text/0035-opentelemetry-protocol.md
```diff
@@ -82,23 +82,76 @@ If the client is shutting down (e.g. when the containing process wants to exit)
...
+|RESOURCE_EXHAUSTED|Yes|
+|FAILED_PRECONDITION|No|
+|ABORTED|Yes|
+|OUT_OF_RANGE|Yes|
+|UNIMPLEMENTED|No|
+|INTERNAL|No|
+|UNAVAILABLE|Yes|
+|DATA_LOSS|Yes|
+
+When retrying the client SHOULD implement a backoff strategy (exception is Throttling case below, which provides explicit instructions about retrying interval).
```
arminru:
```suggestion
When retrying, the client SHOULD implement a backoff strategy. An exception to this is the Throttling case explained below, which provides explicit instructions about retrying interval.
```

text/0035-opentelemetry-protocol.md
tigrannajaryan: I think that's beyond the scope of the protocol. 

Responding with success means "I am now responsible for the data, don't worry about it anymore". What exactly that means for the particular server is that particular server's business. Some servers may reply immediately after deserializing the request, some others may do it after writing to disk and fsync-ing.

The protocol doesn't care about it, the protocol only provides means for the client and server to agree on who is responsible for the data after the request finishes.

This is I believe the typical separation of concerns between protocol and application.

text/0035-opentelemetry-protocol.md
tigrannajaryan: Not part of OTLP spec today, but in the future we may add other transports.

text/0035-opentelemetry-protocol.md
tigrannajaryan: Placeholder for future use.

text/0035-opentelemetry-protocol.md
arminru: Thought so. In this case it would make sense to reword it to "received" or "accepted" (for processing) to make that clear.

```suggestion
The success response indicates telemetry data was successfully received by the server. If the server receives an empty request (a request that does not carry any telemetry data) the server SHOULD respond with success.
```

text/0035-opentelemetry-protocol.md
arminru: Thanks! This was not clear to me so I was confused about the "_when_ using gRPC".

text/0035-opentelemetry-protocol.md
carlosalberto: I suggest adding a note ("[1] Request ought to be changed to proper credentials", etc) to make clear this is a retryable error that **requires** a change.

carlosalberto:
@tigrannajaryan Looks great! A few, very minor details to fix/update, but otherwise LGTM

text/0035-opentelemetry-protocol.md
tigrannajaryan: I don't think "received" conveys the intent. The server may "receive" the data, then return non-success response.

"successfully accepted" does not seem to be a better wording either (is there an "unsuccessful acceptance"?).

text/0035-opentelemetry-protocol.md
tigrannajaryan: There is already a link to https://godoc.org/google.golang.org/grpc/codes which explains in details what each of the codes means. PERMISSION_DENIED do not necessarily require the request to be changed, it may also be a server misconfiguration. I think the link to grpc codes provides sufficient information and hits the right spot of being precise enough and not too verbose. I could copy all that text here but I think it is unnecessary.

tigrannajaryan:
@open-telemetry/specs-approvers this is one of the issues that delays implementation of OTLP protocol. Please review.

text/0035-opentelemetry-protocol.md
yurishkuro:
```suggestion
When using gRPC transport, success response is returned via `ExportResponse` message.
```

text/0035-opentelemetry-protocol.md
```diff
@@ -82,23 +82,76 @@ If the client is shutting down (e.g. when the containing process wants to exit)
...
 
-- `Success` - telemetry data is successfully processed by the server. If the server receives an empty request (a request that does not carry any telemetry data) the server SHOULD respond with `Success`.
+The success response indicates telemetry data is successfully processed by the server. If the server receives an empty request (a request that does not carry any telemetry data) the server SHOULD respond with success.
 
-- `FailedNotRetryable` - processing of telemetry data failed. The client MUST NOT retry sending the same telemetry data. The telemetry data MUST be dropped. This for example can happen when the request contains bad data and cannot be deserialized or otherwise processed by the server. The client SHOULD maintain a counter of such dropped data.
+When using gRPC transport success response is returned via `ExportResponse` message.
 
-- `FailedRetryable` - processing of telemetry data failed. The client SHOULD record the error and may retry exporting the same data immediately. This can happen when the server is temporarily unable to process the data.
+When an error is returned by the server it falls into 2 broad categories: retryable and not-retryable.
```
yurishkuro:
```suggestion
When an error is returned by the server it falls into 2 broad categories: retryable and not-retryable:
```

text/0035-opentelemetry-protocol.md
```diff
@@ -82,23 +82,76 @@ If the client is shutting down (e.g. when the containing process wants to exit)
...
+
+- Retryable errors indicate that processing of telemetry data failed and the client SHOULD record the error and may retry exporting the same data. This can happen when the server is temporarily unable to process the data.
+
+- Not-retryable errors indicate that processing of telemetry data failed and the client MUST NOT retry sending the same telemetry data. The telemetry data MUST be dropped. This for example can happen when the request contains bad data and cannot be deserialized or otherwise processed by the server. The client SHOULD maintain a counter of such dropped data.
```
yurishkuro:
```suggestion
- Not-retryable errors indicate that processing of telemetry data failed and the client MUST NOT retry sending the same telemetry data. The telemetry data MUST be dropped. This can happen, for example, when the request contains bad data and cannot be deserialized or otherwise processed by the server. The client SHOULD maintain a counter of such dropped data.
```

text/0035-opentelemetry-protocol.md
yurishkuro: When we define "Retryable" above, we say "processing of *telemetry data* failed". So if we are splitting hairs, we could say PERMISSION_DENIED is retryable because it failed due to auth-z, not "telemetry data". However, if I think of a naive implementation of the sender, such as one that's given an application token to insert into headers (as opposed to using time-varying protocol like OAUTH2 where token might get refreshed after a while), then retrying would result in absolutely the same total request that will again fail.

Similarly, the server that cannot auth-z the client could be thinking "oh, but if you restart me with a different config, then perhaps this request will succeed", which also strikes me as a stretch.

So I would classify both of these as Not retryable. My 2c.

text/0035-opentelemetry-protocol.md
```diff
@@ -82,23 +82,76 @@ If the client is shutting down (e.g. when the containing process wants to exit)
...
 
 #### Throttling
 
 OTLP allows backpressure signalling.
 
 If the server is unable to keep up with the pace of data it receives from the client then it SHOULD signal that fact to the client. The client MUST then throttle itself to avoid overwhelming the server.
 
-To signal backpressure when using gRPC transport the server SHOULD return an error with code [Unavailable](https://godoc.org/google.golang.org/grpc/codes) and MAY supply additional [details via status](https://godoc.org/google.golang.org/grpc/status#Status.WithDetails) using [RetryInfo](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto#L40). Here is a sample Go code to illustrate:
+To signal backpressure when using gRPC transport the server SHOULD return an error with code [Unavailable](https://godoc.org/google.golang.org/grpc/codes) and MAY supply additional [details via status](https://godoc.org/google.golang.org/grpc/status#Status.WithDetails) using [RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40). Here is a sample Go code to illustrate:
```
yurishkuro:
```suggestion
To signal backpressure when using gRPC transport, the server SHOULD return an error with code [Unavailable](https://godoc.org/google.golang.org/grpc/codes) and MAY supply additional [details via status](https://godoc.org/google.golang.org/grpc/status#Status.WithDetails) using [RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40). Here is a sample Go code to illustrate:
```

text/0035-opentelemetry-protocol.md
tigrannajaryan: Makes sense, I will mark as non-retryable for now. We can refine this in the future if there is any other evidence, since this is a behavioral change and is not a breaking change on the protocol wire format.

text/0035-opentelemetry-protocol.md
tigrannajaryan: Fixed.

text/0035-opentelemetry-protocol.md
tigrannajaryan: Fixed.

text/0035-opentelemetry-protocol.md
tigrannajaryan: Fixed.

text/0035-opentelemetry-protocol.md
tigrannajaryan: Fixed.

tigrannajaryan:
@open-telemetry/specs-approvers all comments are replied / resolved. PTAL and merge if OK.

tigrannajaryan:
@open-telemetry/specs-approvers what is the best way to let all language SIGs know about this change? They will need to change OTLP implementation to match this.

arminru:
@tigrannajaryan That's a general issue that applies to all spec changes and we should introduce a proper process here. Sometimes changes go unnoticed by some SIGs where sometimes this is discovered weeks/months later and most probably a few times not at all.

Currently, the safest, while most tedious way would be to manually open an issue in each SIG to which that change applies.
Let's find better alternatives on https://github.com/open-telemetry/community/issues/317.

PR #67, opened 2019-11-16
Add jmacd@ to CODEOWNERS

bogdandrutu:
jmacd@ is already a spec approver.


bogdandrutu:
This is not controversial. Merging as administrator.

PR #70, opened 2019-11-27
Rename metric instrument "Handles" to "Bound Instruments"

jmacd:
The topic of "Handle" being too general purpose has come up repeatedly.  Each time this proposed renaming is discussed, people seem to agree it's an improvement.

### DIFF
--- a/text/0070-metric-bound-instrument.md
+++ b/text/0070-metric-bound-instrument.md
@@ -1,0 +1,61 @@
+# Rename metric instrument Handles to "Bound Instruments"
+
+The OpenTelemetry metrics API specification refers to a concept known
+as ["metric handles"](0009-metric-handles.md), which is a metric
+instrument bound to a `LabelSet`.  This OTEP proposes to change that
+term to "bound instruments" to avoid the more-generic term "handle".
+
+The corresponding method to create a bound instrument will be renamed
+"Bind" as opposed to "GetHandle".
+
+## Motivation
+
+The term "Handle" is widely seen as too general for its purpose in the
+metrics API.  Rather than re-use a widely-used noun for this concept,
+we instead will re-use the metric "instrument" noun and apply an
+adjective, "bound" to convey that it has been bound to a `LabelSet`.
+
+## Explanation
+
+"Handle" has been confusing from the start. However it was preceded by
+other potentially confusing terms (e.g., "TimeSeries", "Entry").  The
+term "Bound instrument" was initially suggested
+[here](https://github.com/open-telemetry/opentelemetry-specification/pull/299#discussion_r334211154)
+and widely accepted.
+
+## Internal details
+
+This is a simple renaming.  All uses of "handle" will be replaced by
+"bound instrument" in the specification.  All uses of the `GetHandle`
+method become `Bind`.
+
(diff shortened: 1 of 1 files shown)


text/0070-metric-bound-instrument.md
```diff
@@ -0,0 +1,63 @@
...
+method become `Bind`.
+
+Note that the phrase "bound instrument" may not appear directly in the
+user-facing API, nor is it required to, whereas the method `GetHandle`
+is a specified method on metric instruments.
+
+The newly-named `Bind()` method returns a bound instrument type.  The
+name of the returned type may simply take the name of its instrument
+with the prefix `Bound`.  For example, an `Int64Counter` instrument's
+`Bind()` method should return a `BoundInt64Counter` type.
```
bogdandrutu: I don't really like the `BoundInt64Counter`. I would like to know what others think.

iredelmeier: Definitely a mouthful, but nothing better comes to mind 🤔

jmacd: I'm going to bet you would not prefer "Curried", which is another term used by Prometheus in this context.

text/0070-metric-bound-instrument.md
```diff
@@ -0,0 +1,63 @@
...
+other potentially confusing terms (e.g., "TimeSeries", "Entry").  The
+term "Bound instrument" was initially suggested
+[here](https://github.com/open-telemetry/opentelemetry-specification/pull/299#discussion_r334211154)
+and widely accepted.
+
+## Internal details
+
+This is a simple renaming.  All uses of "handle" will be replaced by
+"bound instrument" in the specification.  All uses of the `GetHandle`
+method become `Bind`.
```
cijothomas: How about `GetBoundInstrument()` as it makes it more clear that this method returns something. (like `GetHandle()` .

`Bind()` may be wrongly interpreted as an action or something one can use to manually bind instrument and labelset as in like `meter.Bind(labelset, instrument)`

yurishkuro: It is an action, it creates a new object, bound instrument.

text/0070-metric-bound-instrument.md
yurishkuro: BoundInt64Instrument is a Go artifact. I expect it to show differently in other languages, eg BoundInstrument<long>, just BoundInstrument in Python/Js.

jmacd: I took @bogdandrutu's meaning to be that he does not like the "BoundX" construct entirely. I don't have any better suggestions.

text/0070-metric-bound-instrument.md
jmacd: It is an action, but @cijothomas is pointing out that "Bind" doesn't imply that it returns something.

I could be convinced to move the `Bind()` method to the `Meter` as you have done, and to use a `GetBoundX` method on the instrument, but this method will just call `Bind()` (which will return something) and I'm not sure it's an improvement in clarity.

text/0070-metric-bound-instrument.md
bogdandrutu: I am fine to keep it as is.

PR #72, opened 2019-12-06
Metric Observer instrument specification (refinement)

jmacd:
This is a clarification of the format OTEP 0008.  This topic encountered last-minute debate in https://github.com/open-telemetry/opentelemetry-specification/pull/250 and was left out. The section on calling conventions and examples here should clarify the issue.

### DIFF
--- a/text/0072-metric-observer.md
+++ b/text/0072-metric-observer.md
@@ -122,34 +122,28 @@ re-entry.
 ### Observer calling conventions
 
 Observer callbacks are called with an `ObserverResult`, an interface
-that supports capturing both bound and direct calls, as follows.
+that supports capturing events directly in the callback, as follows.
 
-For a "direct" observation with a specific `LabelSet`, call the
+To capture an observation with a specific `LabelSet`, call the
 `ObserverResult` directly using `ObserverResult.Observe(value,
-LabelSet)` to report an observation.
+LabelSet)`.
 
-For a "bound" observation, using a bound Observer instrument, first
-`Bind()` the instrument with a `LabelSet`, then call
-`ObserverResult.ObserveBound(value, BoundInstrument)`.  Callbacks MUST
-use bound instruments corresponding the Observer instrument for which
-they are registered.  It is an error if
-`ObserverResult.ObserveBound(value, BoundInstrument)` is called for an
-`ObserverResult` corresponding to a different Observer instrument.
+There is no equivalent of a "bound" observer instrument as there is
+with Counter, Gauge, and Measure instruments.  A bound calling
+convention is not needed for Observer instruments because there is
+little if any performance benefit in doing so (as Observer instruments
+are called during collection, there is no need to maintain "active"
+records concurrent with collection).
 
-If the language supports method overloading, `ObserverResult` may
-overload the `Observe()` method for both calling conventions.
+Multiple observations are permitted in a single callback invocation.
 
-Multiple observations are possible in a single callback invocation.
-Likewise, it is permissible to mix bound and direct observations in a
-single callback invocation.
-
-The `Observer`Result passed to a callback should not be used outside the
+The `ObserverResult` passed to a callback should not be used outside the
 invocation to which it is passed.
 
 #### One callback per instrument
 
-The API _could_ support registering independent callbacks for direct
-calls and individual bound instruments, but takes the approach of
+The API _could_ support registering independent callbacks tied to
+registered ("bound") label sets, instead it takes the approach of
 supporting one callback per instrument.  There are two cases to
 consider: (a) where the source of an instrument's values provides one
 value at a time, (b) where the source of an instrument's values
@@ -163,35 +157,7 @@ multiple callbacks.
 
 ### Pseudocode
 
-An example of a bound call:
-
-```
-class YourClass {
-  private static final Meter meter = ...;
-  private static final ObserverDouble cpuTemp =
-      meter.observerDoubleBuilder("cpuTemp")
-           .withKeys("core")
-           .build();
-  private static final ObserverDouble[] cpuTempByCore = new ObserverDouble.Bound[NUM_CORES];
-
-  void init() {
-    for (int i = 0; i < NUM_CORES; i++) {
-        cpuTempByCore[i] = cpuTemp.Bind(meter.createLabelSet("core", i));
-    }
-
-    cpuTemp.setCallback(
-        new ObserverDouble.Callback<ObserverDouble.Result>() {
-          @Override
-          public void update(Result result) {
-            for (int i = 0; i < NUM_CORES; i++) {
-              result.ObserveBound(getCPUTemp(i), cpuTempByCore[i]);
-          }
-        });
-  }
-}
-```
-
-An example of a direct call:
+An example:
 
 ```
 class YourClass {
--- a/text/0072-metric-observer.md
+++ b/text/0072-metric-observer.md
@@ -20,6 +20,15 @@ dedicated `Observer` kind of instrument with the same semantics as
 Gauge instruments.  Like a Gauge instrument, Observer instruments are
 used to report the current value of a variable.
 
+We may ask, why should Observer instruments be a first-class part of
+the API, as opposed to simply registering non-instrument-specific
+callbacks to call user-level code on the metrics collection interval?
+That would permit the use of ordinary Gauge instruments as a stand-in
+for the Observer instrument proposed here.  The approach proposed here
+is more flexible because it permits the Meter implementation to
+control the collection interval on a per-instrument basis as well as
+to disable instruments.
+
 ## Explanation
 
 Gauge metric instruments are typically used to reflect properties that
@@ -29,6 +38,13 @@ to the Counter or measure kind of metric instrument, there could be
 significant computational cost in computing or reading the current
 value.  When this is the case, it is understandable that we are
 interested in providing values on demand, as an optimization.
+
+The optimization aspect of Observer instruments is critical to their
+purpose.  If the simpler alternative suggested above--registering
+non-instrument-specific callbacks--were implemented instead, callers
+would demand a way to ask whether an instrument was "recording" or
+not, similar to the [`Span.IsRecording`
+API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#isrecording).
 
 Observer instruments are semantically equivalent to gauge instruments,
 except they support callbacks instead of a `Set()` operation.
@@ -98,7 +114,7 @@ Callbacks may be called synchronously in the SDK on behalf of an
 exporter, provided it does not contradict the requirement above.
 
 Callbacks should avoid calling OpenTelemetry APIs other than the
-interface provided to `Observe().  This prevents the SDK from
+interface provided to `Observe()`.  This prevents the SDK from
 potentially deadlocking itself by being called synchronously from its
 own thread.  We recognize this may be impossible or expensive to
 enforce.  SDKs should document how they respond to such attempts at
@@ -146,7 +162,7 @@ class YourClass {
         cpuTempByCore[i] = cpuTemp.Bind(meter.createLabelSet("core", i));
     }
 
-    observer.setCallback(
+    cpuTemp.setCallback(
         new ObserverDouble.Callback<ObserverDouble.Result>() {
           @Override
           public void update(Result result) {
@@ -166,7 +182,7 @@ class YourClass {
   private static final ObserverDouble cpuLoad = ...;
 
   void init() {
-    observer.setCallback(
+    cpuLoad.setCallback(
         new ObserverDouble.Callback<ObserverDouble.Result>() {
           @Override
           public void update(Result result) {
--- a/text/0072-metric-observer.md
+++ b/text/0072-metric-observer.md
@@ -1,0 +1,184 @@
+# Metric observer specification (refinement)
+
+The metric observer gauge was described in [OTEP
+0008](0008-metric-observer.md) but left out of the current metrics
+specification because the prior OTEP did not clarify the valid calling
+conventions for observer gauge metric instruments.  This proposal
+completely replaces OTEP 0008.
+
+## Motivation
+
+An [earlier version of the metrics specification](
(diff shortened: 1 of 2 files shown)


text/0072-metric-observer.md
```diff
@@ -0,0 +1,184 @@
...
+any OpenTelemetry API.  This prevents the application from potentially
+deadlocking itself by being called synchronously from its own thread.
+Implementations that cannot provide this guarantee should prefer not
+to implement Observer instrsuments.
+
+Callbacks may be called synchronously in the SDK on behalf of an
+exporter, provided it does not contradict the requirement above.
+
+Callbacks should avoid calling OpenTelemetry APIs other than the
+interface provided to `Observe().  This prevents the SDK from
```
yurishkuro:
```suggestion
interface provided to `Observe()`.  This prevents the SDK from
```

text/0072-metric-observer.md
```diff
@@ -0,0 +1,184 @@
...
+           .withKeys("core")
+           .build();
+  private static final ObserverDouble[] cpuTempByCore = new ObserverDouble.Bound[NUM_CORES];
+
+  void init() {
+    for (int i = 0; i < NUM_CORES; i++) {
+        cpuTempByCore[i] = cpuTemp.Bind(meter.createLabelSet("core", i));
+    }
+
+    observer.setCallback(
```
yurishkuro: what is `observer` here?

jmacd: Fixed. This should have been `cpuTemp`.

text/0072-metric-observer.md
jmacd: Fixed, thanks.

jmacd:
@yurishkuro I've added two paragraphs to try to address your question. I will confess to having mixed feelings about this API. I volunteered to write this specification after I removed it from the metrics spec earlier, in an agreement with @bogdandrutu.  My original draft in OTEP 0008 didn't talk much about about calling conventions, but I had not at that time been thinking of supporting "bound" observer instruments. The addition of bound observer instruments raises some questions (e.g., what prevents you from calling them in a _different_ observer's callback?). 

Another possibility (@bogdandrutu please comment) would be to support a callback for unbound observers and individual callbacks for each bound observer.  Why support one callback per instrument, not one callback per bound instrument IOW?

jmacd:
(I've asked @bogdandrutu this in the past, and he expressed a preference for a single-callback as proposed here, not a callback per bound observer.  It's fewer allocations.)

bogdandrutu:
@jmacd it is not necessary about fewer allocation but also that a system (or external) call like (getting memory) may return data for more than one bound which means that if a callback per-bound is allocated the system call will be done multiple times. In case multiple system (or external) calls are required the callback can deal with that and update multiple bounds. As an example see https://github.com/open-telemetry/opentelemetry-java/blob/master/contrib/runtime_metrics/src/main/java/io/opentelemetry/contrib/metrics/runtime/GarbageCollector.java#L59 where I pre-allocate a list of bounds and update all of them after doing only one system/external call.

text/0072-metric-observer.md
```diff
@@ -0,0 +1,200 @@
...
+https://github.com/open-telemetry/opentelemetry-specification/blob/597718b3fcfaf10bcf45d93f99b66f94a28048cb/specification/api-metrics.md)
+described metric callbacks as an alternate means of generating metric
+events, allowing the application to generate metric events only as
+often as desired by the collection interval.  It specified this
+support for all instrument kinds.
+
+This proposal restores the ability to use callbacks only with a
+dedicated `Observer` kind of instrument with the same semantics as
+Gauge instruments.  Like a Gauge instrument, Observer instruments are
+used to report the current value of a variable.
```
bogdandrutu: +1

text/0072-metric-observer.md
```diff
@@ -0,0 +1,200 @@
...
+
+
+Observer instruments, like Gauge instruments, use a "last value"
+aggregation by default.  With this default interpretation in mind, a
+monotonic Counter can be expressed as a monotonic Observer instrument
+simply by reporting the current sum from `Observe()`, in which case
+the "last value" may be interpreted directly as a sum.  Systems with
+support for rate calculations over current sums (e.g., Prometheus)
+will be able to use these metrics directly.  Systems not designed to
+calculate rates from sums may not benefit from this treatment.
```
bogdandrutu: If really needed we can calculate rates at the client and export them `(current_value - last_value) / collection_interval`.

text/0072-metric-observer.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Differences between Gauge and Observer
+
+One significant difference between gauges that are explicitly `Set()`,
+as compared with observer callbacks, is that `Set()` happens inside a
+context (i.e., its distributed context), whereas the observer callback
+does not execute with any distributed context.  
+
+Whereas Gauge values do have context at the moment `Set()` is called,
+Observer callbacks do not.
```
bogdandrutu: In other words they do not belong to a specific request. (e.g. process cpu-usage)

text/0072-metric-observer.md
```diff
@@ -0,0 +1,200 @@
...
+    for (int i = 0; i < NUM_CORES; i++) {
+        cpuTempByCore[i] = cpuTemp.Bind(meter.createLabelSet("core", i));
+    }
+
+    cpuTemp.setCallback(
+        new ObserverDouble.Callback<ObserverDouble.Result>() {
+          @Override
+          public void update(Result result) {
+            for (int i = 0; i < NUM_CORES; i++) {
+              cpuTempByCore[i].Observe(getCPUTemp(i), result);
```
bogdandrutu: I would design this API a bit different:

`result.Observe(getCPUTemp(i), cpuTempByCore[i])` - if no overloading support in the language then `ObserverWithBound` can be used for the name of the method/function.

text/0072-metric-observer.md
jmacd: I decided it's best to remove the confusing statement ("Systems not designed...").

text/0072-metric-observer.md
jmacd: I added "Observer instruments are appropriate for reporting metrics that are not request specific."

text/0072-metric-observer.md
jmacd: Done.

jmacd:
Please see Bogdan's example (also linked above):
https://github.com/open-telemetry/opentelemetry-java/blob/master/contrib/runtime_metrics/src/main/java/io/opentelemetry/contrib/metrics/runtime/GarbageCollector.java#L59

text/0072-metric-observer.md
```diff
@@ -0,0 +1,202 @@
...
+directly using `Result.Observe(value, LabelSet)` to report an
+observation.
+
+For a "bound" observation with using a bound observer instrument,
+first `Bind()` the instrument with a LabelSet, then call the bound
+instrument passing the result: `Result.ObserveBound(value,
+BoundInstrument)`.  Callbacks MUST use bound instruments corresponding the
+Observer instrument for which they are registered.  It is an error if
+`Result.ObserveBound(value, BoundInstrument)` is called for a Result
+corresponding to a different Observer instrument.
```
c24t: As I understand it, "bound" metric instruments are an optimization to avoid repeatedly updating a metric for a particular `labelSet`. This saves us the cost of constructing the `labelSet` and looking up the aggregation to update, since the bound instrument has a reference to the aggregation.

Does this optimization still make sense in the context of callbacks? For non-observer metrics, we wouldn't expect the user to create a bound instrument for a single call.

text/0072-metric-observer.md
```diff
@@ -0,0 +1,202 @@
...
+        new ObserverDouble.Callback<ObserverDouble.Result>() {
+          @Override
+          public void update(Result result) {
+            for (int i = 0; i < NUM_CORES; i++) {
+              result.ObserveBound(getCPUTemp(i), cpuTempByCore[i]);
+          }
+        });
+  }
+}
+```
```
c24t: It's not clear from the text why the user shouldn't be able to set a callback on a bound instrument directly. This seems like a significant difference between `Set` and `Observe`.

text/0072-metric-observer.md
```diff
@@ -0,0 +1,202 @@
...
+Callbacks should avoid calling OpenTelemetry APIs other than the
+interface provided to `Observe()`.  This prevents the SDK from
+potentially deadlocking itself by being called synchronously from its
+own thread.  We recognize this may be impossible or expensive to
+enforce.  SDKs should document how they respond to such attempts at
+re-entry.
+
+### Observer calling conventions
+
+Observer callbacks accept a Result interface, which supports both
```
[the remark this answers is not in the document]
jmacd: Yes. I've tried to clarify this with more explicit language.

text/0072-metric-observer.md
jmacd: You ask a good question (also below), because if we had one callback per bound Observer/LabelSet, then there's no need to actually expose a Bound Instrument type. There would be a `Observer.Bind(LabelSet, callback)` method to bind a callback to a LabelSet instead, and a callback would be called once per bound label set and it could return a value instead of calling `ObserverResult` callback. This is "case (b)" in the updated language. Potentially this approach can evaluate callbacks in parallel, although that could also cause problems.

This confusion is what led to leaving Observers out of the v0.2 spec. I came away thinking that one callback per bound instrument makes sense, but if we intend a direct calling convention for Observer instruments, then having one callback per instrument gives a smaller surface area (and simpler SDK).  I'm on the fence here.

text/0072-metric-observer.md
jmacd: I tried to add a justification, but I feel about 50/50 on the pros/cons of the one-callback-per-instrument decision.

c24t: Much clearer (even if it's a clear expression of ambiguity about the feature). Thanks!

text/0072-metric-observer.md
bogdandrutu: I would suggest one callback per instrument, the main reason is that you can simply implement a callback that updates multiple labelsets, but if you have a callback per labelset and you need to access a system information you may end up doing a lot of unnecessary system calls. See the java example for memory.

I think @c24t is asking to completely remove bounds from the observer metrics because the benefit of bounding is unclear if the callback is called rarely (once per export interval) which I agree with and we can always add the bounds later to the Observers if we prove the performance wrong.

jmacd: It's a good point. The point of the bound instrument is to avoid a map lookup for each operation, but that map lookup isn't needed for observer instruments--they can bypass the map of current records. They observer instrument's observations are output directly, and since the Batcher (in the current arch diagram--I want to rename this) is required to support duplicates I expect no benefit from bound observer instruments.

jmacd: I have removed "bound" observer instruments based on this reasoning.

yurishkuro: I think it would be a move in the wrong direction. IMO, the callback use case is far more obscure than bound instruments (which are used all over the place in Jaeger code, for example). Bound instruments allow the code to provide measurements on its own pace, usually based on the business traffic. If callback is the only way to provide those, then user code needs to be accumulating the measurements between callbacks, a serious complication.

text/0072-metric-observer.md
```diff
@@ -0,0 +1,183 @@
...
+class YourClass {
+  private static final Meter meter = ...;
+  private static final ObserverDouble cpuLoad = ...;
+
+  void init() {
+    cpuLoad.setCallback(
+        new ObserverDouble.Callback<ObserverDouble.Result>() {
+          @Override
+          public void update(Result result) {
+              result.Observe(getCPULoad(), meter.createLabelSet("low_power", isLowPowerMode()));
```
bogdandrutu: You may want to show that LabelSet can be cached :)

jmacd: Done

text/0072-metric-observer.md
jmacd: The (now removed) case of a bound observer instrument was not meant as a stand-in for the user to accumulate measurements, but for a group of fixed label sets to be repeatedly recorded on the collection interval.

I agree completely with your statement -- Observer instruments should not be used when it involves the user accumulating measurements.  There should not be any benefit or performance advantage for the user because the SDK is designed and heavily optimized to do just this -- accumulate measurements. Bound instruments should be used when recording measurements. 

Observer instruments are meant for "computed values", which are passively collected measurements, not accumulated by the user but computed on the collection interval. The observation (which I credit to @c24t above) is that binding a label set with an observer instrument does not offer any performance benefit, because the emitted values will be directly output to the export pipeline.  They are not buffered or accumulated in the SDK, therefore binding them does no good.

I'm leaving this issue resolved, since we have removed bound observer instruments from this proposal.

jmacd:
@bogdandrutu @c24t @yurishkuro please take another look.

text/0072-metric-observer.md
```diff
@@ -0,0 +1,184 @@
...
+To capture an observation with a specific `LabelSet`, call the
+`ObserverResult` directly using `ObserverResult.Observe(value,
+LabelSet)`.
+
+There is no equivalent of a "bound" observer instrument as there is
+with Counter, Gauge, and Measure instruments.  A bound calling
+convention is not needed for Observer instruments because there is
+little if any performance benefit in doing so (as Observer instruments
+are called during collection, there is no need to maintain "active"
+records concurrent with collection).
```
yurishkuro: Bounding an instrument may involve encoding the label set into a wire-ready data structure, which in itself can be expensive if it needs to be done repeatedly on every measurement.

text/0072-metric-observer.md
```diff
@@ -0,0 +1,184 @@
...
+
+An example:
+
+```
+class YourClass {
+  private static final Meter meter = ...;
+  private static final ObserverDouble cpuLoad = ...;
+
+  void init() {
+    LabelSet labelSet = meter.createLabelSet("low_power", isLowPowerMode());
```
yurishkuro: compile error, missing `final` :-)

text/0072-metric-observer.md
jmacd: The `LabelSet` value is meant to capture any encoding of label sets into wire-ready data structures, since these can be used for more than one bound instrument. In the example snippet, the label set is re-used to emphasize this. The only optimization that I can see would be if the Observer were to output more than once for the same LabelSet per collection period, in which case binding them would support the optimization.

PR #74, opened 2019-12-20
Initialization of global variables

jmacd:
This PR replaces https://github.com/open-telemetry/oteps/pull/45, since the content has been largely rewritten.

This now refers to a prototype in Go: https://github.com/open-telemetry/opentelemetry-go/pull/392


text/0074-global-init.md
```diff
@@ -0,0 +1,176 @@
...
+OpenTelemetry: (1) through dependency injection, (2) through global
+variables initialized at most once.  The feasibility of each approach
+varies by language.  The implementation MUST select one of the
+following strategies.
+
+### Service provider mechanism
+
+Where the language provides a commonly accepted way to inject SDK
+components, it should be preferred.  The Java SPI (Service Provider
+Interface) supports loading and configuring the SDK before it is first
```
jkwatson: Note: this probably won't work even in Java in general. Application servers often host many "applications" all running in the same java process. Each would need to be able to have an independently initialized SDK. And, the underlying application server might also need one of it's own. Classloader isolation could solve some of this, but it's definitely a use case that could cause troubles, especially in a java-agent auto-instrumentation scenario.

text/0074-global-init.md
```diff
@@ -0,0 +1,176 @@
...
+pass-through W3C Traceparent and Correlation-Context propagation.
+
+The default global Propagators instance will begin forwarding to the
+user's Propagators when it is installed as the global instance.
+
+## Trade-offs and mitigations
+
+### Testing support
+
+Testing should be performed without depending on the global SDK, if
```
jkwatson: Yes! Hopefully tests can depend on specific implementations of the specific pieces of the SDK, rather than having everything require the global instance, and a complex global-reset dance.

text/0074-global-init.md
```diff
@@ -0,0 +1,176 @@
...
+they are installed, and we wish for references obtained through these
+instances to become functional once the corresponding implementation
+is initialized.  The default instances returned by the global getters
+for Tracer provider, Meter provider, and Propagators must "forward" to
+the real SDK implementation once it is installed.
+
+#### Tracer
+
+Tracers obtained through the provider will become functional when the
+user's Tracer SDK is installed as the global instance.
```
jkwatson: This is a very interesting requirement. So, if I read this correctly, every Tracer/Meter must be a shim with the capability of having it's underlying implementation swapped out at runtime?

text/0074-global-init.md
jmacd: I more-or-less intended to describe what's needed in a language without DI support. @bogdandrutu convinced me that the Java SPI would take care of the Java case, so I put this paragraph in as a stub. I'd be happy to remove this discussion or strengthen it. 

I had assumed that the DI support could handle initializing SDK(s) and injecting them the way you describe--I'm not very familiar with Java DI though.

text/0074-global-init.md
jmacd: That's the idea, yes. This OTEP was originally written before the named Tracer/Meter was introduced, so the text is a bit light on this point, but I believe that if the global instance can be used before initialization, then this type of shim is needed to properly support it. 

It's certainly a pattern in metrics APIs to allocate instruments statically, so I was mostly focused on that support. I later realized that named Tracer and named Meter would require the same thing, otherwise you'd have applications starting up with a bunch of No-op implementations in static variables.

text/0074-global-init.md
jmacd: In the prototype Go PR, I made an _internal_ method for testing the global variables, but it can only be used from its own package sub-tree, so it's not generally available.

text/0074-global-init.md
```diff
@@ -0,0 +1,176 @@
...
+
+// SetPropagators initializes the global Propagators instance.  May only
+// be called once per process lifetime.  Subsequent calls will panic.
+//
+// Prior to setting the global Meter provider, the default global
+// Propagators instance performs pass-through W3C Traceparent and
+// Correlation-Context propagation.
+func SetPropagators(propagation.Propagators) { ... }
+
+// SetPropagators initializes the global Tracer provider.  May only
```
dyladan:
```suggestion
// SetTraceProvider initializes the global Tracer provider.  May only
```

text/0074-global-init.md
jkwatson: Yes, I'm not convinced that SPI is going to be flexible enough for all the myriad Java use-cases, so I'd definitely weaken this statement. Because the Java ecosystem is so varied, we'll need something very flexible. I'd be willing to bet that for languages/communities that are more opinionated will need less flexibility that Java.

text/0074-global-init.md
jkwatson: 👍

text/0074-global-init.md
jkwatson: Yes. The more I think about the crazy Java ecosystem, the more I think we're going to need something that can be swapped out at runtime, post class-loading. Trying to anticipate all classloader initialization ordering and making sure that the SDK can pre-emptively initialize the globals feels like a fools-errand. Building the dynamism into the SDk from the beginning will be much simpler, even if it's a little more complex than it needs to be on the surface.

text/0074-global-init.md
yurishkuro: > Note: this probably won't work even in Java in general. 

I don't buy this. (a) it worked in OpenTracing. (b) SPI is a widely used pattern in Java, not at all specific to OpenTelemetry; if it was so flawed as the comment implies, everything else in Java ecosystem wouldn't work (even basic logging). And yes, app servers (a legacy notion in itself for almost 10yrs) still handle this perfectly fine with "classloader per application" (which is, incidentally, a totally self-inflicted problem).

text/0074-global-init.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Service provider mechanism
+
+Where the language provides a commonly accepted way to inject SDK
+components, it should be preferred.  The Java SPI (Service Provider
+Interface) supports loading and configuring the SDK before it is first
+used.  This kind of support is the preferred choice in languages with
+common support for automatic dependency injection.
+
+In this case, there is no use-before-configuration question to
```
yurishkuro: I am not sure this use case is so much different from the next section. Even though Java often uses SPI, it's not a requirement that I would expect to be baked-in into OpenTelemetry Java API/SDK, i.e. if someone wants to instantiate OTel providers explicitly, that should still be an option, regardless of additional support for SPI.

text/0074-global-init.md
```diff
@@ -0,0 +1,176 @@
...
+is initialized.  The default instances returned by the global getters
+for Tracer provider, Meter provider, and Propagators must "forward" to
+the real SDK implementation once it is installed.
+
+#### Tracer
+
+Tracers obtained through the provider will become functional when the
+user's Tracer SDK is installed as the global instance.
+
+Spans started prior to installing the Tracer SDK will be No-op spans.
```
yurishkuro: This requires Noop to be part of the API (in some form) so that real tracers can distinguish Noop spans from their own spans. The spec needs to explain this.

text/0074-global-init.md
jkwatson: this is fair, but I want to make sure this ends up working for the auto-instrumentation agent, also. In my experience, it's actually pretty easy for SPI-like constructs (for logging, in particular) to collide with each other in a complex java agent premain environment.

text/0074-global-init.md
jkwatson: This is really all I'm asking for. Let's make sure we don't force SPI for people who don't want to use SPI, and want to manually configure their applications.

text/0074-global-init.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Efficiency concern
+
+Since the global instances are required to begin working once the real
+implementations are installed, there is some implied synchronization
+overhead and cost.  This overhead SHOULD be minimal.
+
+We recommend to explicitly install a No-op instance to lower the cost
+of instrumentation when no SDK will be installed, as opposed to
+leaving the default global instances in place, perpetually waiting to
```
yurishkuro: I don't follow this - what is the difference between "installed Noop instance" and "default global instance"?

text/0074-global-init.md
jkwatson: (note: It's currently baked into the Java SDK, and it's going to take a bit of elbow grease to unwind it at this point)

text/0074-global-init.md
jkwatson: I think "default" here is referring to the default SDK implementation that ships with Open Telemetry.

text/0074-global-init.md
yurishkuro: Playing devil's advocate, what is an example where testing can explicitly inject the SDK but the actual instrumentation cannot do that?

text/0074-global-init.md
yurishkuro: In other words, a user needs to do _something_ to install the default SDK? Is that really wise, given that in most cases people _will_ want a real SDK? Perhaps the Noop one should be an option to install.

text/0074-global-init.md
jkwatson: I don't think there are any, but if people build instrumentation always just via accessing the globals, then unit testing gets difficult. I guess they're making their own troubles, though!

text/0074-global-init.md
jkwatson: I'm not sure I understand what "most cases" means. If I'm adding instrumentation to a library, I only have access to the API, and users of my library should definitely get the No-op version if they haven't wired up an SDK in some way.

text/0074-global-init.md
carlosalberto: (I'd like to see a prototype of Java's `io.opentelemetry.OpenTelemetry` supporting both scenarios - else, we will have to stick, at least for such global object, to **either** SPI **or** manual initialization. And observe this is in the API side, which will definitely impact the SDK).

text/0074-global-init.md
```diff
@@ -0,0 +1,176 @@
...
+```
+
+#### Requirements
+
+We anticipate third party libraries using the global instances before
+they are installed, and we wish for references obtained through these
+instances to become functional once the corresponding implementation
+is initialized.  The default instances returned by the global getters
+for Tracer provider, Meter provider, and Propagators must "forward" to
+the real SDK implementation once it is installed.
```
carlosalberto: +1

text/0074-global-init.md
carlosalberto: Btw, the (old) OT Java global `Tracer` is a singleton that can be swapped once, and simply forwards its calls (and yes, it worked just fine all this time ;) ) 

https://github.com/opentracing/opentracing-java/blob/master/opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java

jmacd: Oh! I did not know that. It's nice to see a precedent. I'll add this detail.

text/0074-global-init.md
carlosalberto: In Java we 'already' can as we have `DefaultSpan` (previously called `BlankSpan`), as well as `DefaultTracer` and related components. And IIRC other SIGs have something like this as part of the API. But we definitely need to handle them properly (to prevent exceptions all over the place).

text/0074-global-init.md
```diff
@@ -0,0 +1,176 @@
...
+
+Meters obtained through the provider will become functional when the
+users's Meter SDK is installed as the global instance.
+
+Metric events will be dropped until the Meter SDK is installed.
+
+#### Propagators
+
+The default global Propagators instance will by default perform
+pass-through W3C Traceparent and Correlation-Context propagation.
```
carlosalberto: +1

text/0074-global-init.md
jmacd: I like the expression "premain environment".  If you're talking about what happens before main(), it's probably complex no matter what the language.

text/0074-global-init.md
carlosalberto: Agreed. Testing against the global SDK should be kept at minimum.

text/0074-global-init.md
jmacd: Support for both mechanisms sounds difficult to me, but possible. I would suggest that the dependency injection mechanism be allowed to provide a default instance, in which case it is initialized before use and no forwarding mechanism enters play. If DI does not provide an implementation, then the global "forwarding" implementation is used (with the same requirements as stated in this document) and the user can explicitly initialize the SDK in its `Main()`. 

It's been my intention to restrict the use of globals as much as possible, which is where the "only once" requirement comes from.  If you permit any caller to reset the global SDK at any time, it seems to invite trouble.

FWIW, I did a a Golang prototype that used the `plugin` package to "inject" an implementation before it is used. This was met with frowns in the Go community, so I abandoned it. See https://github.com/jmacd/opentelemetry-go/pull/1

text/0074-global-init.md
jmacd: Why is it a requirement for the tracer to distinguish its own spans?  I'm not aware of any circumstances that would be needed in the present API.

In the accompanying Go PR (which deals in Metrics), I created a new interface for this sort of reason. If the global default creates a `LabelSet` and the real SDK sees it, it must test for conversion to `LabelSetDelegate` and call its `Delegate()` method to get its own `LabelSet`.

This also reminds me -- there are reasons why the **user** wants to know if they're dealing with a No-op span. I'm not sure we've settled the question of how a user can ask this question in the specification. (I generally think of this as a mechanism of last resort, I would prefer to support "lazy" values that are dispatched cheaply when there is not a real span, but the requirement to know whether a span is "real" somehow never goes away completely.)

text/0074-global-init.md
jmacd: I'm not sure this answer's Yuri's question, but I find it difficult to imagine a module of code that both (a) wants to test their instrumentation and (b) wants to use the global SDK. It will be difficult to test instrumentation against the global SDK because you'll have to install an SDK such as a mock. But the requirements here say you can only install an SDK once. You can still write tests in this way, but they have to be single-threaded.

text/0074-global-init.md
jmacd: I think I should remove this paragraph. I'm saying that it will be somewhat more expensive to use the global forwarding SDK but never install an SDK, than to use an explicitly installed No-op SDK.

text/0074-global-init.md
jmacd: Thanks.

jkwatson:
Note: a PR for Java which implements the changes I think would be required for this OTEP:  https://github.com/open-telemetry/opentelemetry-java/pull/724

jmacd:
I've updated this OTEP stating that languages have an option to implement both SPI and explicit initialization, as long as it's consistent and satisfies the rules.

jmacd:
The behavior described in this PR is implemented in OTel-Go. I suspect other languages will need the same approach to global SDK behavior, but with the work already done in OTel-Go, I see no need to keep this PR.

PR #76, opened 2020-01-08
Clarify named tracers and meters

dyladan:
Based on conversations with SIGs, especially the specification, the overall motivations and use cases of named tracers seem to be poorly understood.

This change attempts to clarify the named tracer OTEP in order to address some of that confusion. It also adds a glossary of terms at the bottom in order to aid in that disambiguation.


text/0016-named-tracers.md
```diff
@@ -12,43 +12,53 @@ _Creating Tracers and Meters using a factory mechanism and naming those Tracers
...
+
+It is the eventual goal of OpenTelemetry that library vendors implement the OpenTelemetry API, obviating the need to auto-instrument their library. An operator should be able to disable the telemetry that is built into some database driver or other library and provide their own integration if the built-in telemetry is lacking in some way. This should be possible even if the developer of that database driver has not provided a configuration to disable telemetry.
+
+## Solution
 
 This proposal attempts to solve the stated problems by introducing the concept of:
- * _Named Tracers and Meters_ identified via a **name** (e.g. _"io.opentelemetry.contrib.mongodb"_) and a **version** (e.g._"semver:1.0.0"_) which is associated with the Tracer / Meter and the Spans / Metrics it produces.
- * A `TracerFactory` / `MeterFactory` as the only means of creating a Tracer or Meter.
+ * _Named Tracers and Meters_ which are associated with the **name** (e.g. _"io.opentelemetry.contrib.mongodb"_) and **version** (e.g._"semver:1.0.0"_) of the library which acquired them.
+ * A `TracerRegistry` / `MeterRegistry` as the only means of acquiring a Tracer or Meter.
```
yurishkuro: Factory or Provider are by far much more accurate terms, because they are active, where as registry is passive.

 What is the justification for this change?

text/0016-named-tracers.md
```diff
@@ -12,43 +12,53 @@ _Creating Tracers and Meters using a factory mechanism and naming those Tracers
...
 
 ## Explanation
 
-From a user perspective, working with *Named Tracers / Meters* and `TracerFactory` / `MeterFactory` is conceptually similar to how e.g. the [Java logging API](https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html#getLogger(java.lang.String)) and logging frameworks like [log4j](https://www.slf4j.org/apidocs/org/slf4j/LoggerFactory.html) work. In analogy to requesting Logger objects through LoggerFactories, a tracing library would create specific Tracer / Meter objects through a TracerFactory / MeterFactory.
+From a user perspective, working with *Named Tracers / Meters* and `TracerRegistry` / `MeterRegistry` is conceptually similar to how e.g. the [Java logging API](https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html#getLogger(java.lang.String)) and logging frameworks like [log4j](https://www.slf4j.org/apidocs/org/slf4j/LoggerRegistry.html) work. In analogy to requesting Logger objects through LoggerFactories, a trace reporting would create specific Tracer / Meter objects through a TracerRegistry / MeterRegistry.
```
yurishkuro: Slf4j class is called LoggerFactory, invalid URL.

text/0016-named-tracers.md
```diff
@@ -12,43 +12,53 @@ _Creating Tracers and Meters using a factory mechanism and naming those Tracers
...
-If a library (or application) has instrumentation built-in, it is both the instrumenting and instrumented library and should pass its own name here. In all other cases (and to distinguish them from that case), the distinction between instrumenting and instrumented library is very important. For example, if an HTTP library `com.example.http` is instrumented by either `io.opentelemetry.contrib.examplehttp` or `com.example.company.examplehttpintegration`, then it is important that the Tracer is not named `com.example.http` but after the actual instrumentation library.
+ The *name* used to create a Tracer or Meter must identify the *instrumentation* libraries (also referred to as *integrations* or *trace/meter reporting library*) and not the library being instrumented. These instrumentation libraries could be libraries developed in an OpenTelemetry repository, a 3rd party implementation, or even auto-injected code (see [Open Telemetry Without Manual Instrumentation OTEP](https://github.com/open-telemetry/oteps/blob/master/text/0001-telemetry-without-manual-instrumentation.md)). See also the examples for identifiers at the end.
```
yurishkuro: I don't recall seeing the term "reporting library", it's quite ambiguous.

text/0016-named-tracers.md
dyladan: I agree that registry is too passive. I can change it to provider. The justification was that a lot of people felt that "Factory" was too oop or java specific.

text/0016-named-tracers.md
dyladan: https://github.com/open-telemetry/oteps/pull/76/files#diff-b6bbd95144a84aa6a1c015cb047afc8fR85

text/0016-named-tracers.md
dyladan: haha woops find/replace error

text/0016-named-tracers.md
yurishkuro: > a lot of people felt that "Factory" was too oop or java specific

Factory is a design pattern that has nothing to do with OOP or Java. Even if it was, I don't see how "this feels like X" qualifies as an argument in a debate.

dyladan: I didn't say it was my argument, just that people felt that way. This was the original issue where the whole "language agnostic working group" stemmed from.

text/0016-named-tracers.md
yurishkuro: You're just linking recursively to the same document. My point is that it's not the terminology we have used so far, because it's misleading, and if anything, relates a lot more to SDK exporters than to  instrumentation. We already have a term "instrumentation library", I suggest we stick to using it consistently rather than using different terms to refer to the same thing.

text/0016-named-tracers.md
yurishkuro: Did that group also have an issue with "API is interface and SDK is implementation"? Because that also "feels like OOP" (because it is). Sorry for getting snarky, but I find that kind of reasoning very unprofessional - if people have arguments, let them state it in writing, as arguments. I've seen this factory issue bubble up several times, always with the same explanation "people feel that...", and never anyone stepping up, creating a ticket, and laying out convincing arguments for a different terminology. For all I know, I might agree with those arguments, but I can't even begin with  "this feels like ..." justification.

jmacd: I take issue with "Factory" because it implies something is being built. Nothing about this API demands that the SDK build something new when the factory is called. I believe "Provider" is the best term we have and it is the most literally accurate given what the current specification says.

dyladan: changed to provider

text/0016-named-tracers.md
danielkhan: I think that "Reporting library" exactly describes what it is.
It's the library that reports the traces or metrics. It's the thing that acquires and utilizes a meter on its behalf. 
This includes also instrumentations that are built **into** a library - let's say mongodb.
"Instrumentation library" would refer to a library that **provides** instrumentation.

jmacd:
I would like to propose that we rename this feature something other than "Named" interfaces. 

As we have learned--a number of people mis-interpreted this feature to be about providing a namespace for the spans and metrics created through the "Named" interface. The intention of this feature is to supply the "library name" and optional version.  While the library does have a name, the library name and version are used here for identification, not the form of naming that many implementors assumed.

I would call this feature Obligatory Reporter Identification.

danielkhan:
> I would call this feature Obligatory Reporter Identification.

I'm also fine with the 'Reporter' wording.

text/0016-named-tracers.md
```diff
@@ -66,22 +76,46 @@ Examples (based on existing contribution libraries from OpenTracing and OpenCens
...
+#### Reporting library
+Also known as the instrumentation library or the trace/metrics reporter, this may be either a library/module/plugin provided by OpenTelemetry that instruments an existing library, a third party integration which instruments some library, or a library that has implemented the OpenTelemetry API in order to instrument itself. In any case, the reporting library is the library which provides tracing and metrics data to OpenTelemetry.
+
+examples:
+* `@opentelemetry/plugin-http`
+* `io.opentelemetry.redis`
+* `redis-client` (in this case, `redis-client` has instrumented itself with the OpenTelemetry API)
+
+#### Tracer name / Meter name
+When it is acquired from the tracer/meter Provider, the tracer/meter is assigned a name and a version which is the name and version of the reporting library that acquired the tracer. In cases where a library has instrumented itself using the OpenTelemetry API, they may be the same.
```
MrAlias: This first sentence seems to highlight the some of the confusion I was anticipating this PR would address. Saying that a tracer/meter is assigned a "name and version" and then referring to that assigned value as a "name" is semantically confusing.

Is a tracer/meter name a 2-tuple of a library name and version, or just the library name? The second sentence, and, to a greater extent, the following example seem to treat it as only the name.

I think [the point made by @jmacd](https://github.com/open-telemetry/oteps/pull/76#issuecomment-572252679) is going to be something needed to clarify named tracers and meters and an important thing to address in this PR. The semantic importance of calling their identifiers "names" is leading to a fair amount of confusion. Using a more generalized term like "identifier" (or "id") seems important to avoid semantic confusion, but also to help remove undesired associations (namespaces).

arminru: What is currently referred to as a tracer/meter name is intended to be the 2-tuple of library name _and_ version. I can see the confusion and also why this might be mixed up as a namespace, which is why I agree that rephrasing it from _tracer/meter name_ to something better would make sense.

arminru:
@jmacd To me _Obligatory Reporter Identification_ sounds a bit bulky but I see that it would certainly make the intent clearer than _Named Tracer/Meter_, even though the latter is already an established name for it.

Would you prefer to rename the provided parameter from _instrumentation library_ to either _reporter_ or _reporting library_, or would you keep it as it is? I agree with @danielkhan's reasoning [above](https://github.com/open-telemetry/oteps/pull/76#discussion_r364383941).

text/0016-named-tracers.md
dyladan: I removed the term reporting library in favor of instrumentation library after some internal discussion

dyladan:
Based on SIG conversation Tuesday, I've updated this PR to use the Meter name as the namespace for metrics.

text/0016-named-tracers.md
```diff
@@ -91,18 +91,14 @@ examples:
...
-
 #### Meter namespace
-This is included here because it is often confused with the meter name.
-
-A Meter namespace describes the component which is being monitored. For instance, `latency` could be collected from a number of different things (http latency, disk latency, event loop latency) and a namespace ties the `latency` metric to the entity being metered.
+Meter name is used as a namespace for all metrics created by it. This allows a telemetry library to register a metric using any name, such as `latency`, without worrying about collisions with a metric registered under the same name by a different library.
 
-example: `bytes_transmitted` may be overall network bytes transmitted, or bytes transmitted specifically over `http`. In this case, it may be desirable to disambiguate these metrics by applying the `http` or `network` namespaces to the `bytes_transmitted` metric.
+example: The libraries `mongodb` and `postgresql` may both register metrics with the name `latency`. These metrics can still be uniquely identified even  though they have the same name because they are registered under different namespaces (`mongodb` and `postgresql` respectively).
```
Oberon00: Note that in practise, the names would be more like `opentelemetry.ext.mongodb` instead of just `mongodb` (see the list of examples elsewhere in this OTEP).

dyladan: This is only true in the case of external instrumentation. Those libraries could have integrated OTel hooks.

text/0016-named-tracers.md
```diff
@@ -91,18 +91,14 @@ examples:
...
 
 example: If the `http` library is being instrumented by a library with the name `io.opentelemetry.contrib.http`, then the tracer name is also `io.opentelemetry.contrib.http`. If that same `http` library has built-in instrumentation through use of the OpenTelemetry API, then the tracer name would be `http`.
 
-### Disambiguation with related terms
-
 #### Meter namespace
-This is included here because it is often confused with the meter name.
-
-A Meter namespace describes the component which is being monitored. For instance, `latency` could be collected from a number of different things (http latency, disk latency, event loop latency) and a namespace ties the `latency` metric to the entity being metered.
+Meter name is used as a namespace for all metrics created by it. This allows a telemetry library to register a metric using any name, such as `latency`, without worrying about collisions with a metric registered under the same name by a different library.
```
Oberon00: According to the definition of "Meter name" above, the namespace should also include the version. Is that intended?

dyladan: no i'll clarify that

dyladan: https://github.com/open-telemetry/oteps/pull/76/commits/312a65f055ce48cf6b21b8084c0ea9979bda3a69 does this help?

text/0016-named-tracers.md
dyladan: @MrAlias can you take a look at https://github.com/open-telemetry/oteps/pull/76/commits/312a65f055ce48cf6b21b8084c0ea9979bda3a69 and see if that helps?

text/0016-named-tracers.md
Oberon00: Looks good now!

text/0016-named-tracers.md
MrAlias: @dyladan the changes look good.

One minor change I would suggest is to the section header:

 ```diff
- #### Tracer name / Meter name and version
+ #### Tracer / Meter name and version
```

NBD, if you don't think it worth changing (given this is only an OTEP and not the spec).

File: text/0016-named-tracers.md [hunk 7]
```diff
@@ -66,22 +76,42 @@ Examples (based on existing contribution libraries from OpenTracing and OpenCens
...
 ## Internal details
 
-By providing a `TracerFactory` / `MeterFactory` and *Named Tracers / Meters*, a vendor or OpenTelemetry implementation gains more flexibility in providing Tracers and Meters and which attributes they set in the resulting Spans and Metrics that are produced.
+By providing a `TracerProvider` / `MeterProvider` and *Named Tracers / Meters*, a vendor or OpenTelemetry implementation gains more flexibility in providing Tracers and Meters and which attributes they set in the resulting Spans and Metrics that are produced.
 
 On an SDK level, the SpanData class and its Metrics counterpart are extended with a `getLibraryResource` function that returns the resource associated with the Tracer / Meter that created it.
 
+## Glossary of Terms
+
+#### Instrumentation library
```
Oberon00: I have heard the opinion that "instrumentation library" is yet another different thing, distinct from "instrumenting library" but synonymous to "telemetry library": https://github.com/open-telemetry/opentelemetry-proto/pull/94#issuecomment-574372313. I would love for instrumentation library to mean the same as instrumenting library. Still I think we might want to consistently use "instrument**ing**" and avoid "instrument**ation**".

dyladan: Instrumenting is simply the present tense verb form of instrument. When you are instrumenting something you are performing the process of instrumentation. If you want to get technical about the difference between instrumentation and telemetry, instrumentation is the process of attaching instruments and telemetry is the process of recording data from the instrumentation. In this sense, the instrumentation library definition here is correct.

That being said, I think it is a mistake to get this pedantic about definitions in the spec as it will just be confusing to people.

Oberon00: I completely agree with you @dyladan. 😃 There just seems to be some conflation of "telemetry library" (i.e. the OpenTelemetry library itself) and "instrumenation library" right now, see also https://github.com/open-telemetry/opentelemetry-proto/pull/94#discussion_r367148777.

That being said, I'm also fine with leaving the OTEP as it is now.

text/0016-named-tracers.md
jmacd: I agree that there's nothing wrong with this text. Maybe another sentence would make it even clearer that you can have a `latency` metric in `io.opentelemetry.redis` and a `latency` metric in `redis-client` and they are different and it is likely that one of them will be disabled.

carlosalberto:
I'd also *slightly* prefer to have this feature renamed to something other than **Named Tracers**. Approving, otherwise this looks great.

dyladan:
> I'd also _slightly_ prefer to have this feature renamed to something other than **Named Tracers**. Approving, otherwise this looks great.

Some other names were floated but nothing has really stuck.

jmacd:
@open-telemetry/specs-approvers Please review and offer feedback or approve.

text/0016-named-tracers.md
```diff
@@ -12,43 +12,53 @@ _Creating Tracers and Meters using a factory mechanism and naming those Tracers
...
 
 If no name (null or empty string) is specified, following the suggestions in ["error handling proposal"](https://github.com/open-telemetry/opentelemetry-specification/pull/153), a "smart default" will be applied and a default Tracer / Meter implementation is returned.
 
 ### Examples (of Tracer and Meter names)
 
 Since Tracer and Meter names describe the libraries which use those Tracers and Meters, their names should be defined in a way that makes them as unique as possible.
-The name of the Tracer / Meter should represent the identity of the library, class or package that provides the instrumentation. 
+The name of the Tracer / Meter should represent the identity of the library, class or package that provides the instrumentation.
```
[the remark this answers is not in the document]
dyladan: Currently unspecified. In the javascript version, you get a different instance of tracer per name-version pair, but you could conceivably get a new tracer every time `getTracer` is called. Do you think that needs to be specified? Since the tracer delegates all its configuration to the `TracerProvider`, it doesn't really matter if you make a new one per name, new one per name/version pair, new one every time, or just 1 instance of each tracer type (noop or working).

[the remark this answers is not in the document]
carlosalberto: My suggestion is to work on this detail in the actual Specification, as this OTEP has been approved and is ready to merge.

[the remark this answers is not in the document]
bogdandrutu: I think it is important mainly for the metrics side. One important thing we need to ensure the name of the metrics are unique inside a "component" (instrumentation library), so if the component is identified by the name then I need to keep a list of all the registered instruments at the name level vs at the name+version level.

bogdandrutu:
Will close/reopen to re-trigger CLA check.

jmacd:
This is ready to merge, right?

carlosalberto:
Merged, thanks - as a further reminder, please address the last question by Bogdan (about name/version identity) when writing the actual specification side, as it might need further refinement.

bogdandrutu:
@jmacd @carlosalberto I was not sure if this is ready since I raised one issue and I didn't hear back any comment. Not a problem that it is merged but was waiting for that response.

carlosalberto:
@bogdandrutu I see, I thought I was agreed this would be tackled when writing the specification side,  my bad. Opened an issue to keep track of this when @dyladan (or anybody else) includes this OTEP into the Specification: https://github.com/open-telemetry/opentelemetry-specification/issues/434

PR #80, opened 2020-01-22
Remove the Metric Gauge instrument, recommend use of other instruments

jmacd:
This was discussed in the 1/16 Metrics SIG meeting. This was first filed as https://github.com/open-telemetry/opentelemetry-specification/issues/412

### DIFF
--- a/text/0080-remove-metric-gauge.md
+++ b/text/0080-remove-metric-gauge.md
@@ -12,22 +12,85 @@ values" in an aggregator.
 
 ## Motivation
 
-Observer instruments improve on our ability to define certain
-aggregations, compared with the existing Gauge instrument.  Because of
-callbacks, Observer instruments support reporting a complete snapshot
-of the "last" values of an instrument, for all active label sets, for
-a single collection period.  This is accomplished requiring new state
-to be managed by the SDK, simply because the Observer instrument is
-expected to support a current value for all active label sets.
+Observer instruments improve on our ability to compute well-defined
+sum and average-value aggregations over a set of last-value aggregated
+data, compared with the existing Gauge instrument.  Using data from an
+Observer instrument, we are easily able to pose queries about the
+current sum of all current values as well as the number of distinct
+values, which together define the average value.
+
+To do the same with synchronous Gauge instruments, the SDK would
+potentially be required to maintain state outside a single collection
+window, which complicates memory management.  The SDK is required to
+maintain state about all distinct label sets over the query evaluation
+interval.  
+
+The question is: how long should the SDK remember a gauge value?
+Observer instruments do not pose this complication, because
+observations are synchronized with collection instead of with the
+application.
+
+Unlike with Gauge instruments, Observer instruments naturally define
+the current set of all values for a single collection period, making
+sum and average-value aggregations possible without mention of the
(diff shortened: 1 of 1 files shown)


jmacd:
@yurishkuro Thank you. I rewrote large parts of this document to address some of the points you raised. I have tried to argue that one of the other three instrument kinds is always appropriate. 

If I understand your position on Gauge vs. Measure, I think it's overly reductive to say that Measure could be replaced by Gauge, but I agree that Gauge can be replaced by measure in some cases and that there's not an additional burden for migration in all but the rarest cases. These rare cases are cases where Observer instruments are truly the right choice and Counter and Measure would not apply, but that the performance impact of a callback approach is not acceptable.

In the new text, I've tried to clarify the distinction between Observer and the other kinds of instrument, that of there being a "current set of measured values" across multiple label sets without reference to an interval of time.

jmacd:
I can still be convinced not to go with this proposal, but I'm definitely leaning towards it now. Thinking about how to implement the `Gauge Set` helper introduced as a mitigation strategy, it occurs to me that a special instance of the default SDK could be configured to support the `Gauge Set` by translating  `Gauge.Set` to `Measure.Record` in the special-purpose SDK. An Observer instrument would be registered to export the contents of the special-purpose SDK via observations. I.e., When the observer callback is called, the set of last-values is collected from the underlying Measure instruments of the special-purpose SDK.

tsloughter:
> I would actually prefer to see Measure removed as more ambiguous.

Measure is absolutely needed. Measure, through attaching aggregations, can cover Counter and Gauge, but not the other way around.

I still see Counter and Gauge (Observable or not) as optimizations, which I like. OpenCensus there was only `record` and aggregation done in views. If you have something that is for sure for counting, or observing as a gauge, having these as high level concepts helps with both readability and optimizing the code path, doing the aggregation in-process.

But I'm +1 on removing the Gauge instrument. Easy to add it it back if needed.

MrAlias:
This OTEP does a good job highlighting why it is advantageous to remove gauges from the developer perspective. However, one of the things that is really important in building out an API is to consider the user of the API and the APIs lifecycle. As @bogdandrutu pointed out in https://github.com/open-telemetry/opentelemetry-specification/issues/412:

> Apply the rule of minimal API surface: because there are some unknowns with the Gauge implementation and because we don't have very good use-cases for these type of instruments it is better to avoid adding them in the API for the moment. In case of a clear use-case determined later we should re-evaluate if adding the Gauge instrument back is the right decision or we do have different other solutions.

I think this is an understated point, but one that should be central to this discussion. Including Gauges in the API to only later in the lifecycle determine their meaning needs to change (how they support complex aggregations), or they need to be removed, is a bad design decision and one that will reflect poorly on the OpenTelemetry project overall.

@yurishkuro does bring up a valid point in questioning whether the lack of Gauges would result in a loss of functionality to the final user, but I think @jmacd and @tsloughter are correct in providing reasons as to why this is not the case.

I'm in favor of removing the Gauge from the API if we cannot come up with any valid use-cases for them. @jmacd do you think adding a bit about the API design choice to the motivation a needed part of the OTEP?

bogdandrutu:
@open-telemetry/specs-approvers please review, we need couple more reviews to merge this

PR #83, opened 2020-01-31
Add minimal vscode settings for md files

bogdandrutu:
Similar with https://github.com/open-telemetry/opentelemetry-specification/blob/master/.vscode/settings.json


bogdandrutu:
@open-telemetry/specs-approvers please review :)

.vscode/settings.json
```diff
@@ -0,0 +1,12 @@
+{
+    "rewrap.wrappingColumn": 80,
+    "editor.rulers": [80],
+    "markdownlint.config": {
+        "MD004": false,
```
dyladan: Will this list of exceptions need to be maintained indefinitely?

bogdandrutu: We can update them when they are not required.

bogdandrutu: I copied the file from the specs

carlosalberto:
@bogdandrutu I suggest we merge this one - after all, this is not an actual OTEP, but changes to this repo itself, and doesn't need that many approvals.

bogdandrutu:
@carlosalberto ok

Using the admin power and merging this with only 3 approvers

PR #84, opened 2020-02-03
Add InstrumentationLibrary to describe telemetry from a named Tracer/Meter

bogdandrutu:


carlosalberto:
This looks great to me, specially given we are not modifying the API itself. My opinion on two mentioned issues:

1. I think at this time, we should *only* support name/version, as we can always add things in the future.
2. Is Component's name/version a **direct** mapping from the named components in the API? Is this expected?

bogdandrutu:
> 1. I think at this time, we should _only_ support name/version, as we can always add things in the future.

I agree.

> 2. Is Component's name/version a **direct** mapping from the named components in the API? Is this expected?

The name and version of the component are the values passed to `getTracer` and `getMeter`.

text/0083-component.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Explanation
+
+```
+                         Application
++-----------------------------------------------------------------+
+|                   TracerProvider(Resource)                      |
+|                   MeterProvider(Resource)                       |
+|                                                                 |
+|      Instrumented Library 1         Instrumented Library 2      |
```
Oberon00: Your diagram is wrong. While I agree that for "components" it would be ideal to break up the application by (instrumented) library, named tracers, by design, break it up by instrumentation library. I fear we cannot really use named tracers for what you suggest.

text/0083-component.md
```diff
@@ -0,0 +1,118 @@
...
+|   +--------------------------+   +--------------------------+   |
+|                                                                 |
++-----------------------------------------------------------------+
+```
+
+Every application has one `TracerProvider` and one `MeterProvider` which have a
+`Resource` associated with them that describes the Application.
+
+Every instrumented library has one `Tracer` and one `Meter` which have a
+`Component` associated with them that describes the instrumented library.
```
Oberon00: Nope, the instrument**ing** library.

text/0083-component.md
```diff
@@ -0,0 +1,118 @@
...
+```proto
+// Resource information.
+message Resource {
+  // Set of labels that describe the resource.
+  repeated opentelemetry.proto.common.v1.AttributeKeyValue attributes = 1;
+}
+
+// Component is a message representing the instrumenting library informations
+// such as the fully qualified name and version.
+message Component {
```
Oberon00:
```suggestion
message InstrumentationLibraryInfo {
```

text/0083-component.md
```diff
@@ -0,0 +1,118 @@
...
+named `Tracer|Meter`.
+
+## Open questions
+
+1. Should we support more than name and version for the Component now?
+
+## Future possibilities
+
+In the future the `Component` can be extended to support multiple properties
+(attributes) that apply to the specific `instance` of the instrumented library.
```
Oberon00:
```suggestion
(attributes) that apply to the specific `instance` of the instrumenting library.
```

text/0083-component.md
```diff
@@ -0,0 +1,118 @@
+# Component
+
+Introducing the notion of Component as a first class concept in OpenTelemetry
+which describes the named `Tracer|Meter` in the data model.
+
+## Motivation
+
+The main motivation for this is to better model telemetry coming from different
+instrumented or instrumenting libraries inside an Application (`Resource`). This
```
Oberon00:
```suggestion
instrumenting libraries inside an Application (`Resource`). This
```

text/0083-component.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Trade-offs and mitigations
+
+Adding a new concept into the OpenTelemetry protocol and the exporter framework
+may be overkill, but this concept maps easily to an already existing concept
+in the API, and it is easy to explain to the users.
+
+## Prior art and alternatives
+
+An alternative approach was proposed in the proto [PR comment](
```
Oberon00: Apart from renaming LibraryInfo to Component, this proposal is the same, correct?

text/0083-component.md
bogdandrutu: What is the mapping between instrumentation library and instrumented library? I thought that it is 1:1 (I do agree that naming can be different).

@dyladan to your point in case of a JDBC instrumentation the instrumented library is JDBC and the fact that connects to a mysql or postgress is a property of the instance of the JDBC library which can be captured as suggested in the extra attributes that we can add later.

jmacd:
In https://github.com/open-telemetry/opentelemetry-specification/pull/354, there's a question about how to talk about the values passed to the named tracer. "Instrumenting library"?

dyladan:
> In [open-telemetry/opentelemetry-specification#354](https://github.com/open-telemetry/opentelemetry-specification/pull/354), there's a question about how to talk about the values passed to the named tracer. "Instrumenting library"?

Do you mean the values passed to the tracer provider? If so, then the OTEP (and incoming spec PR) is using the term "instrumentation library" (but it would not be incorrect to call it the instrumenting library). This is mainly to distinguish the instrument**ing** library from the instrument**ed** library.

text/0083-component.md
Oberon00: It's an n:m mapping. Granted, most often it will be 1:1, but in the general case you need to consider:
* An instrumenting library might instrument multiple "components". E.g. an auto instrumentation may instrument all implementations of a DB driver interface, so it covers mysql, postgres, oracle, ...
* There may be multiple instrumentations for the same library. E.g.
  * an opensource library instrumenting some component gets abandoned, then forked under a new name. 
  * a library to instrument some other library is created internally, at different companies under different names. Then one of them or more are published/opensourced under different names or an opensource/commercial implementation is developed independently.
  * a library might start out being instrumented by a 3rd party vendor, but then the instrumentation is built right into the library.
  * for some component, both an auto-instrumentation and wrapper solution might exist (since auto-instrumentation is not always possible or wanted, both have their advantages and disadvantages).

Depending on what is meant by the vague term "component" there may be an additional dimension that is relevant:
* A library may be used multiple times in the same application, resulting in multiple logical components, e.g. a MariaDB client library that is used to connect to maintain connections to both MariaDB and MySQL, maybe in the same process. You might want to assign each connection instance its own "component" with info like connection string, db name, ...

Named tracers are per instrumenting library, not per component or instrumented library.

dyladan: I would add another example for 1:n mapping. Consider a package released by a community member named `otel-db-all-in-one` which claims to instrument all major database libraries. It may auto-patch mysql, postgres, sqlite, mongo, etc. In this case, the Tracer Name would be `otel-db-all-in-one`, but the `component` should be the instrumented db lib, not the all in one instrumentation lib.

text/0083-component.md
```diff
@@ -0,0 +1,118 @@
+# Component
```
jmacd: I like the idea of automatically generating an attribute for the name of the Tracer|Meter, but I'm not sure "component" carries the right meaning, especially after reading the discussion about N:M mappings. A component sounds like a single "part", but this feels more complicated. Was LibraryName not good enough?

arminru: We were also discussing the term (telemetry) reporter or reporting library on https://github.com/open-telemetry/oteps/pull/76#discussion_r364383941 as another alternative.

Oberon00:
Honestly, I'm a bit confused how people are so confused about named tracers all the time. It seems people just hear the name and immediately in their mind they have an idea what that is/should be and then nobody reads the actual definitions what that "name" is. At this point maybe it's really better to rename the feature to ORI (Obligatory Reporter Info), then people *have* to read the docs.

Or am I wrong and is this OTEP not misunderstanding but attempting to completely change the definition of the tracer name?

> This does not limit the “component” to a particular instrumenting library either since NamedTracer/NamedMeter object may be passed around freely and can be used to emit telemetry about virtually anything the developer deems fit.

You *can* technically do that but the definition of the tracer name and what it is supposed to mean is very clear and explicit.

> It appears that we want to make recommendations around how NamedTraced/NamedMeter should be used

We already have definitions how named tracers MUST be used (maybe it's not 100% as clear for meters).

text/0083-component.md
bogdandrutu: Based on what I am reading, this is what I understand so far:
* It may be possible to have multiple instrumentation library instrumenting the same library (N : 1). This is why we include the package name (namespace) in the named tracer/meter name to distinguish between them.
* For the example where one released artifact/package instruments multiple libraries (like the dbs example) why do we not have one named tracer per db? It is probably important to have the ability to control the instrumentation at the db level not at the instrumenting package level. So I am not sure I understand that this is a valid use of the named Tracer.

Maybe we need to update the named Tracer/Meter to say that it applies to only one instrumented library/framework?

text/0083-component.md
bogdandrutu: @jmacd I don't care about the name that much (was proposing one option). I care only about the fact that all telemetry generated by one named "Tracer/Meter" should be annotated with the properties passed to the named "Tracer/Meter".

text/0083-component.md
```diff
@@ -0,0 +1,118 @@
...
+## Motivation
+
+The main motivation for this is to better model telemetry coming from different
+instrumented or instrumenting libraries inside an Application (`Resource`). This
+change affects only the OpenTelemetry protocol and exporters, and it does not
+require any API changes.
+
+Currently OpenTelemetry has defined a semantic convention for the `component`
+Span attribute which already caused some concerns and questions, see [here](
+https://github.com/open-telemetry/opentelemetry-specification/issues/336)
```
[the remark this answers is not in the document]
bogdandrutu: yes.

bogdandrutu:
@tigrannajaryan I am not sure I follow the entire comment. The rational to have this as a separate entity is for:
* Performance on the wire;
* For metrics, we do guarantee uniqueness of the names at the instrumenting library name, so some backends may use the instrumenting library name as prefix to all the metric names.
* Some backends may use the instrumenting library name as prefix for Spans.

I am not sure why do we try to flatten everything when we do have this structure in the API, so it will be natural to use the same structure in the data model and not do any flattening. Also if we do use an attribute in the Span: for this exercise let's call it `component` then the user also records an attribute `component` what is the expected behavior?

text/0083-component.md
bogdandrutu: I think here I refer explicitly to the instrumented library not instrumenting library. It is true that you may also have other properties for the instrumenting library as well.

text/0083-component.md
arminru: @bogdandrutu 
ad 1: correct
ad 2: I agree that it might make sense to be able to control it at a DB level. This is, however, a different use case than the one addressed by named tracers. The use case covered by them is that a certain instrumentation version misbehaves or should be superseded and therefore disabled. The misbehavior/bug occurring in that instrumentation library version would be the same over all DBs. Same goes with superseding it; if I have a better instrumentation available I'd most probably want to have the old one replaced for all DBs.

Feel free to propose a means to define the instrument*ed* component in order to use it for further configuration but please accept that this is not what named tracers are (currently) about. The covered use case as mentioned above is clearly stated in the named tracer OTEP (https://github.com/open-telemetry/oteps/blob/master/text/0016-named-tracers.md) and the meaning of the provided instrumentation library name is clearly stated in the spec (https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#obtaining-a-tracer).

text/0083-component.md
tsloughter: I like the proposal but also agree "Component" doesn't convey the meaning of the thing well. So maybe `InstrumentorInfo`/`InstrumentationInfo`

jmacd:
What can we do to move this forward?

bogdandrutu:
I am re-writing the otep using InstrumentationInfo as a replacement for the Component. Will have someting in 1h

text/0083-component.md
bogdandrutu: I think this is now outdated.

text/0083-component.md
bogdandrutu: Done.

bogdandrutu:
@open-telemetry/specs-approvers please review this proposal again.

text/0083-component.md
```diff
@@ -0,0 +1,115 @@
...
+  // Set of labels that describe the resource.
+  repeated opentelemetry.proto.common.v1.AttributeKeyValue attributes = 1;
+}
+
+// InstrumentationLibrary is a message representing the instrumentation library
+// informations such as the fully qualified name and version.
+message InstrumentationLibrary {
+  string name = 1;
+  string version = 2;
+  // Can be extended with attributes.
```
jmacd: Would you call these "attributes" properties of the instrumentation library? Or could they be arbitrary resources?

bogdandrutu:
Ping @open-telemetry/specs-approvers :)

text/0083-component.md
```diff
@@ -0,0 +1,115 @@
...
+`InstrumentationLibrary` associated with them that describes the instrumentation
+library.
+
+In case of multi-application deployments like Spring boot (or old Java
+Application servers) every Application will have it's own `TracerProvider` and
+`MeterProvider` instances.
+
+## Internal details
+
+This change affects only the OpenTelemtry protocol, and propose a way to
```
Oberon00:
```suggestion
This proposal affects only the OpenTelemtry protocol, and proposes a way to
```

text/0083-component.md
```diff
@@ -0,0 +1,115 @@
...
+represent the telemetry data in a structural way, here is an example for
```
Oberon00:
```suggestion
represent the telemetry data in a structured way.
For example, here is the protobuf definition for metrics:
```

text/0083-component.md
```diff
@@ -0,0 +1,115 @@
...
+  // A list of metrics that originate from a Instrumentation library.
+  repeated Metric metrics = 2;
+}
+```
+
+## Trade-offs and mitigations
+
+Adding a new concept into the OpenTelemetry protocol and the exporter framework
+may be overkill, but this concept maps easily to an already existing concept
+in the API, and it is easy to explain to the users.
```
Oberon00:
```suggestion
in the API, and it is easy to explain to users.
```

text/0083-component.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Open questions
+
+1. Should we support more than name and version for the InstrumentationLibrary
+now?
+
+## Future possibilities
+
+In the future the `InstrumentationLibrary` can be extended to support multiple
+properties (attributes) that apply to the specific `instance` of the
```
Oberon00:
```suggestion
properties (attributes) that apply to the specific instance of the
```

text/0083-component.md
```diff
@@ -0,0 +1,115 @@
...
+instrumenting or instrumented library.
```
Oberon00:
```suggestion
instrumented library.
Also, information about the instrumented library could be added, but that will require additional consideration about grouping (e.g. group by the pair (instrumenting lib, instrumenting lib) instead of just by instrumenting lib.
```

PR #88, opened 2020-02-25
Metrics API instrument foundation and refinements

jmacd:
This OTEP stems from discussion immediately following the v0.3 metrics API specification.

Partly resolves:
https://github.com/open-telemetry/opentelemetry-specification/issues/467
https://github.com/open-telemetry/opentelemetry-specification/issues/465
https://github.com/open-telemetry/opentelemetry-specification/issues/464
https://github.com/open-telemetry/opentelemetry-specification/issues/462


text/0088-metric-instrument-optional-refinements.md
```diff
@@ -0,0 +1,205 @@
...
+The `Monotonic` and `Absolute` options were removed in the 0.3
+specification.  Here, we propose to regain the equivalent effects
+through instrument refinements.  Instrument refinements have the same
+calling patterns as the foundational instrument they refine, adding
+either a different standard implementation or a restriction of the
+input domain.
+
+We have done away with instrument options, in other words, in favor of
+optional metric instruments.  Here we discuss three important
+capabilities achievable using instrument refinements.
```
bogdandrutu: One open question: should we allow users to create these base instruments or do they serve as "abstract" classes for all the other instruments?

For Measure instruments we can have (names not well thought):
* Counter - monotonic Sum.
* UpDownCounter - non-monotonic Sum.
* Distribution - min/max/sum/count (or just sum/count) - only positive values.
* DistributionWithNegative - min/max/sum/count (or just sum/count) - can have negative values.

It is important to differentiate between Distribution with negative and non-negative to be compatible with OpenMetrics (this is unfortunately).

For Observer (names match a bit the Prometheus/Stackdriver terminology and may not be the best names):
* GaugeObserver - instantaneous measurement of a value (values for the same labels can go up down)
* CumulativeObserver - measurements for the same labels are going up.
* DeltaObserver - measurements for the same labels represent the change since previous observer call, and can only be positive).

The main idea for my comment is the first sentence.

jmacd: I was leaning toward allowing the base instruments to be non-abstract, i.e., `DistributionWithNegative` is the same as `Measure`.

text/0088-metric-instrument-optional-refinements.md
```diff
@@ -0,0 +1,205 @@
...
+`Counter` instrument automatically has this property, but why?
+
+We can infer that a value is cumulative in the following circumstances:
+
+- Non-negative and Sum-only: This is called `Counter`, if synchronous, and can be `DeltaObserver`, potentially, if asynchronous
+- Monotonic: This can be `MonotonicObserver`, potentially.
+
+We see that these refinements satisfy their intended purpose, which is
+to convey additional semantics that exporters use to convert data into
+their system.
```
jkwatson: Interesting section here. If I'm running a prometheus backend, I would want my counter aggregations to be cumulative. If I'm running one that expects deltas to be reported for each interval, I'd want my counter aggregations to be non-cumulative (deltas). 

Do we foresee OTel providing a built-in way for an exporter to convert between the two? If not, how do we choose the default counter aggregation in a way that won't end up making someone's life difficult by having to build their own state into their exporter code, or, maybe worse, make the operator have to provide custom aggregation for all their counters?

text/0088-metric-instrument-optional-refinements.md
bogdandrutu: The only downside is that these generic names do not express that way the semantic as the other names I was proposing. Not saying I cannot live with that but just an idea.

text/0088-metric-instrument-optional-refinements.md
jkwatson: One possible solution: have count aggregators provide both a delta and a cumulative value, so exporters can choose appropriately the value that works for their backend.

bogdandrutu: I think this should be a property of the Views (options to configure aggregations). I do agree that the default should be more like other cncf projects offer

jkwatson: I don't see how the Views API will help an exporter author deal with having to map between cumulative counters & delta counters or vice-versa. Can you describe how this would work?

bogdandrutu: I think I understand your concern, probably we should always build deltas from the Measure where we have full control and then the export like Prometheus can merge deltas. For observer where we don't have control of the values that are recorded they may be deltas or cumulative, probably we need to know that in the exporter.

jmacd: My goal was to let the "Sum-only" refinement imply that a sum was being monitored. For exporters that export cumulative sums, the Integrator should know to use a stateful aggregator. For exporters that export deltas, the Integrator should know to use a stateless aggregator.

The existing Go Integrator code has a boolean to indicate whether it is stateful or not, which appears to address this problem. For Prometheus, however, it doesn't do enough. The reason is that Summary-valued aggregators are supposed to be reported as deltas, whereas Histogram-valued aggregators are suppose to be reported as cumulative values. So we need a finer-grained way to select whether an aggregator is stateful, and it appears to depend on the exporter.

Therefore, I agree with @bogdandrutu that this is a problem best left to the Views API. Still, it's possible when there is only one Exporter to infer the correct default behavior for these Sum-only instruments, but this won't work if there are multiple exporters.

bogdandrutu: @jmacd even for Histograms we can do deltas, they are easy to merge :)

jkwatson: Should exporters have to declare what kind of sum-like aggregations they want (cumulative vs. deltas)? I'm not sure how else that information would get communicated. I'm definitely not understanding how the Views API would help here.

jkwatson: Also, I think this discussion, although important, is not really relevant to this particular OTEP. :)

text/0088-metric-instrument-optional-refinements.md
```diff
@@ -0,0 +1,205 @@
...
+Other uses for built-in type refinements involve the type for duration
+measurements.  For example, where there is built-in type for the
+difference between two clock measurements, OpenTelemetry APIs should
+offer a refinement to automatically apply the correct unit of time to
+the measurement.
+
+### Counter refinement
+
+Counter is a non-negative, sum-only refinement of the Measure
+instrument.
```
c24t: Counters can't count? 🙃

text/0088-metric-instrument-optional-refinements.md
```diff
@@ -0,0 +1,205 @@
...
+events additionally have [Context](api-context.md), describing
+properties of the associated trace and distributed correlation values.
+
+Observer instruments have a well-defined _last value_ measured by the
+instrument that can be useful in defining aggregations.  To maintain
+this property, we impose this requirement: two or more calls to
+`Observe()` in a single Observer callback invocation are treated as
+duplicates of each other, and the last call to `Observe()` wins.
+
+Measure instruments do not define a _last value_ relationship.
```
c24t: This is very helpful. Now that the last value aggregation is outlawed for measures, would you expect that the order that measurements are recorded within a collection interval shouldn't affect the aggregate result?

jkwatson: My brain keeps tripping on this line in the OTEP. Would this imply that no one could write a last-value aggregation for a measure? Or just that we won't provide one out of the box, or something else?

text/0088-metric-instrument-optional-refinements.md
jmacd: I.e., Counters don't count in the default aggregation.

text/0088-metric-instrument-optional-refinements.md
jkwatson: Discussed offline in the SIG. This basically means that since measures don't have a single point-in-time associated with them, last-value isn't a well-defined thing for measures. One could write a last-value aggregation, but it would have to just make some assumptions/assertions about what "last-value" means, which isn't something we want to define as a part of the SDK.

jkwatson: [consider this resolved from my perspective...I can't resolve it myself]

jmacd:
As posted in https://gitter.im/open-telemetry/metrics-specs?at=5e67b66ef046a30bbe55b092

In last week's Metrics SIG we had a conversation about the interpretation of the "instrument refinements" proposed in the OTEP, and we came to realize a problem with the set of refinements being proposed, which were "sum-only", "non-negative", and "monotonic". The test case was how to think about exporting a cumulative CPU-usage counter. The problem was that metric events are aggregated differently with respect to time than they are with respect to the spatial dimensions for cumulative instruments.

I've updated the document with a significant amount of new text. I've introduced the term "Rate aggregation", which is defined differently based on whether the values being reported are deltas or cumulatives. Now the document describes 4 refinements, "sum-only", "non-negative" are still there from before, and the two new ones are "precomputed-sum" and "non-negative-rate".

The "pre-computed sum" refinement says that a value is cumulative. The "non-negative-rate" refinement says that the sum is monotonic.

There are still no functional changes in this proposal, just a revision of the mental model.

jkwatson:
Based on discussion in the metrics SIG, I have started a document to capture vendor and open-source backend metric types. Please add all relevant information: https://docs.google.com/document/d/1-Tut-w5lACJhrjGG1Jn9C41uDMquZpkd1_oXLpiYFK0/edit?usp=sharing

text/0088-metric-instrument-optional-refinements.md
```diff
@@ -0,0 +1,306 @@
...
+events additionally have [Context](api-context.md), describing
+properties of the associated trace and distributed correlation values.
+
+### Last-value relationship
+
+Observer instruments have a well-defined _last value_ measured by the
+instrument that can be useful in defining aggregations.  To maintain
+this property, we impose this requirement: two or more calls to
+`Observe()` in a single Observer callback invocation are treated as
+duplicates of each other, and the last call to `Observe()` wins.
```
krnowak: Are the calls to `Observe()` considered duplicate even if the calls are with a different labelset for each?

evantorrie: I have this question too.  In the current spec, there is a sentence which says "_A single Observer instrument callback can capture multiple metric events associated with different label sets._" which seems to imply they are not duplicates if the label sets are different.

However, for a metric which has [WithRecommendedKeys](https://github.com/open-telemetry/opentelemetry-specification/blob/4ab7c2fb4964c14fe972b03b5bb6825573bafb24/specification/api-metrics.md#withrecommendedkeys-declaration-on-metric-instruments), are two different observations with nominally different labelsets at the point of observation treated as duplicates if the `WithRecommendedKeys` option collapses those labelsets into the same group?

text/0088-metric-instrument-optional-refinements.md
```diff
@@ -0,0 +1,483 @@
...
+
+A rate can be computed from values that are reported as differences,
+referred to as _delta_ reporting here, or as sums, referred to as
+_cumulative_ reporting here.  The primary goal of the instrument
+refinements introduced in this proposal is to facilitate rate
+calculations in more than one way.
+
+When delta reporting, a rate is calculated by summing individual
+measurements or observations.  For Measure instruments, these values
+fall into a range of time, as indicated by the event timestamp.  For
```
lzchen: Does this mean that for `Measure` and `Counter` instruments, the `timestamp` attribute represents the time of recording, and for `Observer` it represents the time of collection? Since observers are recorded at time of collection, is it safe to say that all `timestamp` values are timestamps for each event when recorded. Does this go against what [the specs](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-metrics.md#time) say in regards to timestamp?

jmacd: I don't see a disagreement there. In the linked specification section, there's an effort to say that timestamps are implicit and might not actually be captured. So an SDK is free to use a single timestamp for all events in an interval, if it think's that is reasonable. For Observer instruments, I was trying to state that all observations for an interval happen at logically the same time. They are logically recorded at the same time, which I think satisfies the statement you're making.
(I'm not sure I'm helping here.)

lzchen: I guess the key here "...free to..."? I can see use cases where an explicit timestamp is required for certain ingestion services (Azure for example), so if there is flexibility in how the SDK implements it that would be quite versatile.

text/0088-metric-instrument-optional-refinements.md
jmacd: I was leaning the other way, but I re-read your proposal above and am starting to like it, except...

For synchronous instruments, I'm not convinced that it matters if a `Distribution` is non-negative. If we allow that `Distribution` is unrefined (i.e., positive or negative), there's no need for a `DistributionWithNegative`. Do you have any particular reasons why we care that Distribution values are non-negative? I have some in mind, but I don't think they matter. If you want to sum the values and compute a rate, a Counter configured with an alternate aggregation would be an alternative (i.e., using a View API).

For the asynchronous instruments you listed, I might replace `GaugeObserver` with `LastValueObserve`.

Then the proposal at the end of the latest draft of this PR would propose these non-abstract instruments:

Synchronous:
```
Counter
UpDownCounter
Timing
Distribution
```

Asynchronous:
```
DeltaObserver
CumulativeObserver
LastValueObserver
```

text/0088-metric-instrument-optional-refinements.md
jmacd: There is a motion to remove `WithRecommendedKeys` and the associated "defaultkeys" Batcher (a.k.a. Integrator).  Still, the kind of aggregation that happens when combining two distinct LabelSets into a group is referred to as spatial aggregation here, and it would apply whatever the ordinary aggregation is for the instrument.

This is why "last value" is referred to as a relationship given by Observers, not an aggregation. You can configure an asynchronous instrument to aggregate the last value using by adding values (as for `DeltaObserver` in this proposal), or by combining the distribution (as for the unrefined `Observer` in this proposal). See also this remark: https://github.com/open-telemetry/oteps/pull/88#discussion_r398204203

text/0088-metric-instrument-optional-refinements.md
jmacd: I take this as a recommendation that future versions of the default SDK should support different options with regards to timestamps.

How would you treat a Counter instrument, for example? If there are many `Add()` calls throughout a collection interval, the standard behavior is to add them without computing timestamps for each. Even the `Exact` aggregator, which simply builds an array of values, does not bother to store the timestamp for each. Only the current "lastvalue" aggregator bothers with timestamps, which it uses to maintain the last value. I can imagine an `ExactWithTimestamp` aggregator that would build an array of (value, timestamp) pairs. That could be used, but I was under the impression that even the Azure system was implemented by using 1-second checkpoints with a single timestamp for the whole 1-second batch.

text/0088-metric-instrument-optional-refinements.md
bogdandrutu: 1. I would keep `Timing` separately because that can be implemented as just a measurement 
 helper class or can be an instrument, here is an example for measurement helper class:
    ```
    t := Timer(distribution, labels);
    ...  // your code
    t.Record();
    ```
    OR as a standalone instrument.

1. For the negative distribution as I mentioned it is unfortunate that Prometheus (when exported as a histogram) uses different ways to export the count/sum uses `_count/_sum` or `_gcount/_gsum` for histogram with negative values https://github.com/prometheus/client_python/blob/master/prometheus_client/openmetrics/parser.py#L392 so we kind of have to be able to support that. Maybe we can start initially with `Distribution` only support positive values and later after we understand better relax that condition or add a `NonNegativeDistribution`.

1. Does DeltaObserver and CumulativeObserver imply that the raw values used when do the pre-aggregation are positive only (that would be great because will match perfectly with a Counter Sum aggregation that we can produce and LastValueObserver will match perfectly with a Sum aggregation for a UpDownCounter)?

text/0088-metric-instrument-optional-refinements.md
```diff
@@ -0,0 +1,483 @@
...
+
+### Standard implementation of Measure and Observer
+
+OpenTelemetry specifies how the default SDK should treat metric
+events, by default, when asked to export data from an instrument.
+Measure and Observer instruments compute `Sum` and `Count`
+aggregations, by default, in the standard implementation.  This pair
+of measurements, of course, defines an average value.  There are no
+restrictions placed on the numerical value in an event for the two
+foundational instruments.
```
bogdandrutu: Just a thought: Not sure if we should ask for 2 values as default, because this implies a mutex in languages like java where it is hard to do atomically both operations. Maybe we can stick with only Sum as default for performance reasons.

text/0088-metric-instrument-optional-refinements.md
```diff
@@ -0,0 +1,483 @@
...
+| Measure  | sum-only |                 |               |                   | _UpDownCounter_ |
+| Measure  | sum-only | precomputed-sum |               |                   | _UpDownCumulativeCounter_ |
+| Measure  |          |                 | non-negative  |                   | _AbsoluteMeasure_ |
+| Measure  |          |                 |               |                   | _NonAbsoluteMeasure_ |
+| Observer | sum-only |                 | non-negative  | non-negative-rate | _DeltaObserver_ |
+| Observer | sum-only | precomputed-sum |               | non-negative-rate | _CumulativeObserver_ |
+| Observer | sum-only |                 |               |                   | _UpDownDeltaObserver_ |
+| Observer | sum-only | precomputed-sum |               |                   | _UpDownCumulativeObserver_ |
+| Observer |          |                 | non-negative  |                   | _AbsoluteObserver_ |
+| Observer |          |                 |               |                   | _NonAbsoluteObserver_ |
```
bogdandrutu: This needs to be updated based on https://github.com/open-telemetry/oteps/pull/88#discussion_r398204203

text/0088-metric-instrument-optional-refinements.md
```diff
@@ -0,0 +1,483 @@
...
+instrument](https://prometheus.io/docs/concepts/metric_types).  
+
+| System     | Metric Kind  | Operation           | Aggregation          | Notes               |
+| ---------- | ------------ | ------------------- | -------------------- | ------------------- |
+| Prometheus | Counter      | Inc()               | Sum                  | Sum of positive deltas |
+| Prometheus | Counter      | Add()               | Sum                  | Sum of positive deltas |
+| Prometheus | Gauge        | Set()               | Last Value           | Non-additive or monotonic cumulative |
+| Prometheus | Gauge        | Inc()/Dec()         | Sum                  | Sum of deltas |
+| Prometheus | Gauge        | Add()/Sub()         | Sum                  | Sum of deltas |
+| Prometheus | Histogram    | Observe()           | Histogram            | |
```
bogdandrutu: Histogram in Prometheus is restricted to only non-negative values :(

text/0088-metric-instrument-optional-refinements.md
```diff
@@ -0,0 +1,483 @@
...
+
+The Statsd system supports only synchronous reporting.  
+
+| System | Metric Event | Operation           | Aggregation          | Notes               |
+| ------ | ------------ | ------------------- | -------------------- | ------------------- |
+| Statsd | Count        | Count()             | Sum                  | Sum of deltas       |
+| Statsd | Gauge        | Gauge()             | Last Value           |     |
+| Statsd | Histogram    | Histogram()         | Histogram            |     |
+| Statsd | Distribution | Distribution()      | _Not specified_      | A distribution summary |
+| Statsd | Timing       | Timing()            | _Not specified_      | A distribution summary |
```
bogdandrutu: https://github.com/b/statsd_spec#timers they are non-negative as well.

text/0088-metric-instrument-optional-refinements.md
jmacd: We have a lock-free solution to this in Go, I figure you could do that in Java if you wanted.
See https://github.com/open-telemetry/opentelemetry-go/blob/master/sdk/internal/state_locker.go or https://github.com/open-telemetry/opentelemetry-go/blob/master/sdk/metric/aggregator/minmaxsumcount/mmsc.go

bogdandrutu: See https://github.com/open-telemetry/opentelemetry-go/pull/600 to prove my point.

jmacd: 👨‍🔬

text/0088-metric-instrument-optional-refinements.md
jmacd: Done. (Thanks for pushing this resolution.)

text/0088-metric-instrument-optional-refinements.md
jmacd: Fixed.

text/0088-metric-instrument-optional-refinements.md
jmacd: OK! :)

text/0088-metric-instrument-optional-refinements.md
jmacd: (1) Yes, but are we willing to specify this helper class?  It feels like an inconsistency to say that timing measurements should be made using a different class of thing, compared with non-timing measurements. One of the key ideas is that the units are automatically specified for Timing measurements, and in the snippet of code above this could break. How can we avoid using a Timing wrapper with a Distribution instrument that has the wrong units?

(2) Ack. I'm skeptical that `NonNegativeDistribution` is an important class of instrument. Users could configure a `Counter` instrument to be aggregated as a distribution and get the same effect (a distribution of strictly non-negative values), but that would be a non-standard configuration. I have an internal debate about this: when the former Prometheus Histogram was used **correctly**, a Distribution will **correctly** replace it.  If the former Prometheus Histogram's error behavior is taken into account, a Distribution is not a correct replacement. I would argue we have a compatible instrument, in this case. I do not believe error-compatibility should be a requirement for OTel. 

The problem here is that for new users of a Distribution instrument, we don't know whether to export as a Prometheus counter-histogram or gauge-histogram. A Prometheus exporter could be configured to export one distribution for non-negative values and one distribution for negative values. Users _ought_ to be able to combine the positive and negative distributions correctly downstream--in any case IMO this is not an error at the API level, this is an incompatibility in a legacy export path. I do not believe we should let Prometheus shape the OTel API to this extent.

(3) Yes, my intent was that `DeltaObserver` and `CumulativeObserver` accept non-negative values. The `UpDown-` variation on these would support negative values. I see there's a minor complication here, but I've waved this away already in the document: I don't think it matters whether a Cumulative instrument accepts non-negative values, what we care about is that it has a non-negative rate. I'm willing to say that `CumulativeObserver` has both non-negative and non-negative-rate refinements. I wrote:

> For the precomputed-sum instruments, we technically do not care whether the inputs are non-negative, because rate aggregation computes differences.  However, it is useful for other aggregations to assume that precomputed sums start at zero, and we will ignore the case where a precomputed sum has an initial value other than zero.

bogdandrutu:
@open-telemetry/specs-approvers please review this.

text/0088-metric-instrument-optional-refinements.md
jmacd: See https://github.com/open-telemetry/opentelemetry-specification/issues/463.

jmacd:
@open-telemetry/specs-approvers We are experiencing delays. PLEASE REVIEW.

text/0088-metric-instrument-optional-refinements.md
```diff
@@ -0,0 +1,527 @@
...
+
+To maintain this property, we impose a requirement: two or more
+`Observe()` calls with an identical LabelSet during a single Observer
+callback invocation are treated as duplicates of each other, where the
+last call to `Observe()` wins.
+
+Based on the Last Value relationship, we can ask and answer questions
+such as "what is the average last value of a metric at a point in
+time?".  Observer instruments define the Last Value relationship
+without referring to the collection interval and without ambiguity.
```
jkwatson, line 105: Since "last value" is, by it's very nature, a single value for a given reporting interval, doesn't the "average" need to be defined over some set of collection intervals? And, if this is a "temporal" average (rather than a "spatial" average), doesn't there need to be some concept of a time window over which the average is computed?

text/0088-metric-instrument-optional-refinements.md
```diff
@@ -0,0 +1,526 @@
...
+floating point should offer instrument refinements for each, leading
+to type names like `Int64Measure` and `Float64Measure`.
+
+A language with support for unsigned integer types may wish to create
+dedicated instruments to report these values, leading to type names
+like `UnsignedInt64Observer` and `UnsignedFloat64Observer`.  These
+would naturally apply a non-negative refinment.
+
+Other uses for built-in type refinements involve the type for duration
+measurements.  For example, where there is built-in type for the
```
jkwatson, line 266:
```suggestion
measurements.  For example, where there is a built-in type for the
```

text/0088-metric-instrument-optional-refinements.md
```diff
@@ -0,0 +1,526 @@
...
+## Future possibilities
+
+A future OTEP will request the introduction of two standard
+refinements for the 0.4 API specification.  This will be the
+`CumulativeObserver` instrument described above plus a synchronous
+timing instrument named `TimingMeasure` that is equivalent to
+_AbsoluteMeasure_ with the correct unit and a language-specific
+duration type for measuring time.
+
+If the above open question is decided in favor of treating the
```
jkwatson, line 523: I think the open question referenced disappeared in some edit. It's not there now.

text/0088-metric-instrument-optional-refinements.md
jmacd: I think of average being well-defined for a single. interval. If there is no aggregation, there's a count of 1 and the average equals the single value. If there is spatial aggregation within the interval, the average is still well defined.

If there is temporal aggregation over multiple intervals, I think the average is still well defined. What we're not defining, however, is the average over a specific time range--the average is defined over a range of intervals.

I like the "last value" relationship defined here, since we can write queries that apply across machines, even though their collection intervals are not synchronized. The query is something like "read the last-value of metric X at time T across a set of machines and average the results". We are using the last value at the most-recent point in time as the input to a distributed spatial aggregation.

We could try to define an average across time and space using this data, but now some interpolation may be required, since the contributing points are spread across time and the query window is probably not aligned with the collection intervals of the target machines.

 See also https://github.com/open-telemetry/oteps/pull/96/files#diff-e822e6cd6cb430ee3f62ff7fbeaed0daR130

text/0088-metric-instrument-optional-refinements.md
jmacd: Fixed.

text/0088-metric-instrument-optional-refinements.md
jmacd: The question was whether names like "Measure" and "Observer" should be reserved as explanatory, descriptive names, and whether we should find new, different names for the actual, concrete unrestricted/unrefined form of these instruments handled by users. This was discussed in the SIG call but no issue specifically was filed. 

This has been referred to as "Abstract foundation instruments" vs "Non-abstract foundation instruments". I originally proposed Non-abstract, i.e., that the unrestricted Measure instrument should be called Measure. @bogdandrutu convinced me to go the other way, which I have done in this document's sample proposal and in OTEP 93. 

However, I have changed that stance again in OTEP 96, which states that we should use "Recorder" instead of "Measure", and that the default kind of Recorder instance should be named "Recorder--this is the Non-abstract proposal.

jkwatson: oh, I know the open question...it just isn't in this doc any more that I could see.  :)

text/0088-metric-instrument-optional-refinements.md
jkwatson: I think it's good. I was just confused about the question when the answer seemed very obvious. :)

PR #89, opened 2020-03-05
Add metrics view API

c24t:
This OTEP addresses the [Future Work: Configurable Aggregations / View API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-metrics.md#future-work-configurable-aggregations--view-api) section of [api-metrics.md](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-metrics.md). It proposes a specification for a _view API_ that allows users to configure aggregations for individual metric instruments.

It covers some of the same ground as https://github.com/open-telemetry/opentelemetry-specification/pull/347, and summarizes a few weeks of conversation in [metrics SIG meetings](https://docs.google.com/document/d/1LfDVyBJlIewwm3a0JtDtEjkusZjzQE3IAix8b0Fxy3Y/edit#heading=h.9tngw7jdwd6b).

See also the [google doc draft](https://docs.google.com/document/d/1ZyPNOqByj4GFxLhXeB8bENR4uWmR3NPYiaS46yrlWTQ/edit?usp=sharing).

### DIFF
--- a/text/0089-metrics-views-api.md
+++ b/text/0089-metrics-views-api.md
@@ -1,0 +1,205 @@
+# Metrics View API
+
+This OTEP addresses the [Future Work: Configurable Aggregations / View API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-metrics.md#future-work-configurable-aggregations--view-api) section of [api-metrics.md](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-metrics.md).
+It proposes a specification for a _view API_ that allows users to configure aggregations for individual metric instruments.
+
+## TL;DR
+
+The potentially contentious changes this OTEP proposes include:
+
+- Require views to be registered: don't record measurements from API for metric instruments that don't appear in any view.
+- Drop support for LastValue aggregations, don't allow multiple Observer calls per export interval.
+- Don't allow views to specify how "dropped" keys are aggregated, since we can safely ignore aggregation order for measurement instruments other than observers.
+
+
+## Motivation
+
+As of [opentelemetry-specification#430](https://github.com/open-telemetry/opentelemetry-specification/pull/430), the spec describes three different metric instrument types, and a default aggregation for each.
+These default aggregations are consistent with the semantics of each metric instrument, but not with the various exposition formats required by APM backends.
+
+For example, consider the [Prometheus histogram exposition format](https://prometheus.io/docs/concepts/metric_types/#histogram).
+The default aggregation for the Measure metric instrument is _MinMaxSumCount_.
+This aggregation computes summary statistics about (i.e. the _min_, _max_, _sum_, and _count_ of) the set of recorded values, but doesn't capture enough information to reconstruct histogram bucket counts.
+
+If instead the Measure instrument used an [_Exact_ aggregation](https://github.com/open-telemetry/opentelemetry-specification/pull/347/files#diff-5b01bbf3430dde7fc5789b5919d03001R254-R259) -- which stores the sequence of measurements without aggregating them -- the Prometheus exporter could compute the histogram for each recording interval at export time, but at the cost of significantly increased memory usage.
+
+A better solution would be to use a custom _Histogram_ aggregation, which could track bucket counts and compute summary statistics without preserving the original measurements.
+
+As the spec is written now, all metrics are aggregated following the default aggregation for the metric instrument by which they were recorded.
+By design, there is no option to change these defaults:
+
+> [Users do not have a facility in the API to select the aggregation they want for particular instruments.](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-metrics.md#aggregations)
+
+This limitation prevents users from:
+
+- Converting aggregated data to vendor-specific exposition formats
+- Configuring aggregation options, e.g. histogram bucket boundaries
+- Specifying multiple aggregations for a single metric
+
+In addition, it's often desirable to aggregate metrics with respect to a specific set of label keys.
+
+A _view API_ would allow users to use custom aggregation types, associate individual metrics with one or more aggregations, configure aggregation options, and specify the set of label keys to use to aggregate each metric.
+
+## Explanation
+
+Metric instruments are API objects that record measurements from application code.
+These measurements are aggregated by _Aggregators_, which are SDK objects.
+Aggregated measurement data are passed to _Exporters_, which convert these data into APM-specific exposition formats.
+
+Each view describes a relationship between a metric instrument and an aggregation, and specifies the set of label keys to preserve in the aggregated data.
+
+Views are tracked in a registry, which exporters may use to get the current aggregate values for all registered aggregations at export time.
+Each registered view represents a metric that is meant to be exported.
+As a corollary, the SDK can choose to drop measurements from any metric instrument that does not appear in a registered view.
+
+## From measurement to APM backend
+
+The goal of separating metric instrument types from aggregations is that it should be possible to switch to a different APM backend without changing application code.
+
+The default SDK should include conventional aggregators, and exporters should convert aggregated data into the appropriate exposition format based on the aggregator type.
+Some exporters will require custom aggregators, either because the built-in aggregators don't capture enough information to convert aggregated data into a particular exposition format, or to improve on the efficiency of the built-in aggregators.
+
+APM vendors should typically include custom exporters and any custom aggregators they rely on in the same distribution.
+
+One of the design goals of OpenTelemetry metrics is that it should work seamlessly with existing metrics systems like [StatsD](https://github.com/statsd/statsd/wiki) and [Prometheus/OpenMetrics,](https://github.com/prometheus/prometheus) and provide API instruments that are familiar to these systems' users.
+
+To check that the view API proposed here satisfies this goal, it may be helpful to compare OpenTelemetry views -- which are comprised of a metric instrument and aggregation -- to other systems' metric types:
+
+| System     | Metric Type                                                                                        | OT Metric Instrument | OT Aggregation    |
+| ---------- | -------------------------------------------------------------------------------------------------- | -------------------- | ----------------  |
+| OpenCensus | Count                                                                                              | Counter              | Sum               |
+| OpenCensus | Mean                                                                                               | Counter              | Mean\*            |
+| OpenCensus | Sum                                                                                                | Counter              | Sum               |
+| OpenCensus | Distribution                                                                                       | Measure              | Histogram         |
+| OpenCensus | LastValue                                                                                          | Observer             | LastValue         |
+| Datadog    | [COUNT](https://docs.datadoghq.com/developers/metrics/types/?tab=count#metric-types)               | Counter              | Sum               |
+| Datadog    | [RATE](https://docs.datadoghq.com/developers/metrics/types/?tab=rate#metric-types)                 | Counter              | Rate\*            |
+| Datadog    | [GAUGE](https://docs.datadoghq.com/developers/metrics/types/?tab=gauge#metric-types)               | Observer             | LastValue         |
+| Datadog    | [HISTOGRAM](https://docs.datadoghq.com/developers/metrics/types/?tab=histogram#metric-types)       | Measure              | Histogram         |
+| Datadog    | [DISTRIBUTION](https://docs.datadoghq.com/developers/metrics/types/?tab=distribution#metric-types) | Measure              | Sketch\* or Exact |
+| Prometheus | [Counter](https://prometheus.io/docs/concepts/metric_types/#counter)                               | Counter              | Sum               |
+| Prometheus | [Gauge](https://prometheus.io/docs/concepts/metric_types/#gauge)                                   | Observer             | LastValue         |
+| Prometheus | [Histogram](https://prometheus.io/docs/concepts/metric_types/#histogram)                           | Measure              | Histogram         |
+| Prometheus | [Summary](https://prometheus.io/docs/concepts/metric_types/#summary)                               | Measure              | Sketch\* or Exact |
+
+A well-designed view API should not require extensive configuration for typical use cases.
+It should also be expressive enough to support custom aggregations and exposition formats without requiring APM vendors to write custom SDKs.
+In short, it should make easy things easy and hard things possible.
+
+
+## Internal details
+
+A view is defined by:
+
+- A metric instrument
+- An aggregation
+- An optional set of label keys to preserve
+- An optional unique name
+- An optional description
+
+The aggregation may be configured with options specific to its type.
+For example, a _Histogram_ aggregation may be configured with bucket boundaries, e.g. `{0, 1, 10, 100, 200, 1000, inf}`.
+A _Sketch_ aggregation that estimates order statistics (i.e. quantiles), may be configured with a set of predetermined quantiles, e.g. `{.5, .95, .99, 1.00}`.
+
+Each view has a unique name, which may be automatically determined from the metric instrument and aggregation.
+If the same metric instrument and aggregation appear in two separate views, at least one view must specify a name.
+
+Note that a view does not describe the type (e.g. `int`, `float`) or unit of measurement (e.g. "bytes", "milliseconds") of the metric to be exported.
+For example, a hypothetical _MinMaxSumMeanCount_ aggregation of an int-valued millisecond latency measure may be exported as five separate metrics to the APM backend:
+
+- An int-valued _min_ metric with "ms" units
+- An int-valued _max_ metric with "ms" units
+- An int-valued _sum_ metric with "ms" units
+- A **float**-valued _mean_ metric with "ms" units
+- An **int** valued, unitless (i.e. unit "1") _count_ metric
+
+
+This OTEP doesn't propose a particular API for Aggregators, just that the API is sufficient for exporters to get all this information, including:
+
+- That the _min_, _max_, and _sum_ metrics preserve the type and unit of the underlying Measure
+- That the _mean_ metric is float-valued, but preserves the underlying measure's unit
+- That the _count_ metric is int-valued with unit "1", regardless of the underlying measure's unit
+
+
+### Aggregating over time
+
+An aggregation describes how multiple measurements captured via the same metric instrument in a single collection interval are combined.
+_Aggregators_ are SDK objects, and the default SDK includes an Aggregator interface (see [opentelemetry-specification#347](https://github.com/open-telemetry/opentelemetry-specification/pull/347) for details).
+
+Aggregations are assumed to be [_mergeable_](https://www.cs.utah.edu/~jeffp/papers/mergeSumm-MASSIVE11.pdf): aggregating a sequence of measurements should produce the same result as partitioning the sequence into subsequences, aggregating each subsequence, and combining the results.
+
+Said differently: given an aggregation function `agg` and sequence of measurements `S`, there should exist a function `merge` such that:
+
+<!-- <img src="https://render.githubusercontent.com/render/math?math=\mathrm{agg}(S) = \mathrm{merge}(\mathrm{agg}(S_1), \mathrm{agg}(S_2), \ldots, \mathrm{agg}(S_N))"> -->
+
+```
+agg(S) = merge(agg(S_1), agg(S_2), ..., agg(S_N))
+```
+
+<!-- <img src="https://render.githubusercontent.com/render/math?math=\{S_1, \ldots, S_N\}"> -->
+<!-- <img src="https://render.githubusercontent.com/render/math?math=S"> -->
+
+For every partition `{S_1, ..., S_N}` of `S`.
+
+For example, the _min_ aggregation is trivially mergeable:
+
+<!-- <img src="https://render.githubusercontent.com/render/math?math=\mathrm{min}([1, 2, 3, 4, 5, 6]) = \mathrm{min}([\mathrm{min}([1]), \mathrm{min}([2, 3]), \mathrm{min}([4, 5, 6])])"> -->
+
+```
+min([1, 2, 3, 4, 5, 6]) = min([min([1]), min([2, 3]), min([4, 5, 6])])
+```
+
+but quantile aggregations, e.g. _p95_ and _p99_ are not.
+Applications that export quantile metrics should use a mergeable aggregations such as [DDSketch](https://arxiv.org/abs/1908.10693), which estimates quantile values with bounded errors, or export raw measurements without aggregation and compute exact quantiles on the backend.
+
+We require aggregations to be mergeable so that they produce the same results regardless of the collection interval, or the number of collection events per export interval.
+
+
+### Aggregating across label keys
+
+Every measurement is associated with a _LabelSet_, a set of key-value pairs that describes the environment in which the measurement was captured.
+Label keys and values may be extracted from the [correlation context](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-correlationcontext.md) at the time of measurement.
+They may also be set by the user [at the time of measurement](https://github.com/open-telemetry/opentelemetry-specification/blob/f4aeb318a5b77c9c39132a8cbc5d995e222d124f/specification/api-metrics-user.md#direct-instrument-calling-convention) or at the time at which [the metric instrument is bound](https://github.com/open-telemetry/opentelemetry-specification/blob/f4aeb318a5b77c9c39132a8cbc5d995e222d124f/specification/api-metrics-user.md#bound-instrument-calling-convention).
+
+In each collection interval, measurements with the same labelset are aggregated together.
+In effect, there is a single Aggregator for each unique labelset.
+
+The view API should allow the user to specify the set of label keys to track for a given view.
+Other label keys should be dropped before aggregation.
+By default, if the user doesn't specify a set of label keys to track, the aggregator should record all label keys.
+These options correspond with the "defaultkeys" and "ungrouped" integrators described in [opentelemetry-specification#347](https://github.com/open-telemetry/opentelemetry-specification/pull/347).
+
+Because we don't require the user to specify the set of label keys up front, and because we don't prevent users from recording measurements with missing labels in the API, some label values may be undefined.
+Aggregators should preserve undefined label values, and exporters may convert them as required by the backend.
+
+For example, consider a Sum-aggregated Counter instrument that captures four consecutive measurements:
+
+1. `{labelset: {'k1': 'v11'}, value: 1}`
+1. `{labelset: {'k1': 'v12'}, value: 10}`
+1. `{labelset: {'k1': 'v11', 'k2': 'v21'}, value: 100}`
+1. `{labelset: {'k1': 'v12', 'k2': 'v22'}, value: 1000}`
+
+And consider three different views, one that tracks label key _k1_, one that tracks label key _k2_, and one that tracks both.
+
+After each measurement, the aggregator associated with each view has a different set of aggregated values:
+
+| time        | k1                                  | k2  | value | agg([k1])                                   | agg([k2])                                                                | agg([k1, k2]) (default)                                                                                                                  |
+| ----------- | ----------------------------------- | --- | ----- | -----------------------------------         | ------------------------------------------------------------             | --------------------------------------------------------------------------------------------------------------------                     |
+| 1           | v11                                 | -   | 1     | `({k1: v11}, 1)`                            | `({k2: undefined}, 1)`                                                   | `({k1: v11, k2: undefined}, 1)`                                                                                                          |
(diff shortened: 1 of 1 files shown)


text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,205 @@
+# Metrics View API
+
+This OTEP addresses the [Future Work: Configurable Aggregations / View API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-metrics.md#future-work-configurable-aggregations--view-api) section of [api-metrics.md](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-metrics.md).
+It proposes a specification for a _view API_ that allows users to configure aggregations for individual metric instruments.
```
jkwatson: Can we clarify who "users" are in this context of this OTEP? Are we talking about instrumentation authors, or application developers/operators, or authors of exporters?

c24t: Changed to "application owners and administrators".

It's probably also worth pointing out here that integration authors might want to ship a set of views too, like https://www.javadoc.io/doc/io.opencensus/opencensus-contrib-grpc-metrics/0.17.0/io/opencensus/contrib/grpc/metrics/RpcViewConstants.html.

tsloughter:
I will do my best to make the metrics sig meeting on the 11th where this might be easier to discuss, but my first issue with this proposal is simply allowing `Counter` to be used for anything but `Sum`.

Having an explicit Counter instrument type allows for optimizations that are lost if the Counter has to be recorded the same as a Measure, where within a collection interval all values are recorded and then aggregated. If the user wants this I'd expect them to use a Measure and if they want a Sum they'd use a Counter, something that can rely on high performance counters on some runtimes.

c24t:
> recorded the same as a Measure, where within a collection interval all values are recorded and then aggregated

The values should actually be continuously aggregated, so we update the aggregate sum with each new measurement.  Other aggregations would work this way too, e.g. a histogram aggregation would update a bucket count for each new recorded value, and wouldn't keep a reference to the measurement. Only the "exact" aggregation would store measurements.

It is odd that the view API lets you register a non-sum aggregation for a counter instrument though. What do you recommend here?

tsloughter:
I would just remove the non-sum aggregation from the view API for counter instrument.

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,207 @@
+# Metrics View API
+
+This OTEP addresses the [Future Work: Configurable Aggregations / View API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-metrics.md#future-work-configurable-aggregations--view-api) section of [api-metrics.md](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-metrics.md).
+It proposes a specification for a _view API_ that allows application owners and administrators to configure aggregations for individual metric instruments.
+
+## TL;DR
+
+The potentially contentious changes this OTEP proposes include:
+
+- Require views to be registered: don't record measurements from API for metric instruments that don't appear in any view.
```
jkwatson: If views are registered by the operator/application developer, does this mean that that person will get *no metric data* if they don't add additional configuration? How can they know a priori what metrics they they need to register views for?

jkwatson:
Does this proposed Views API live at the same level as the other APIs, or is it a default-SDK-only thing? I asked, because this would be the first "API" that didn't target instrumentors, and instead targets operators/app developers. It feels like that will lead to confusion (it confuses me!) from the user base, especially the instrumentation-building user.

text/0089-metrics-views-api.md
tsloughter: The way I've done it is each library exports a function with predefined views. In the simplest case the application developer either enables them or not, while still having the ability to instead define their own, or take a subset of the default views a library exports.

So the operator/application developer has to allow metrics from each library, instead of them being automatic and having to have a way to disable what you don't want.

jkwatson: I'm now thinking of the auto-instrumentation projects, where there may be dozens of libraries being instrumented that the operator may or may not be familiar with. This is going to be a very surprising thing for "normal" auto-instrumentation agent users to have to do, as they are used to getting things OOTB without extra configuration. Would the auto-instrumentation agent then be expected to just register all the defaults provided with the instrumented libraries?

c24t: > does this mean that that person will get _no metric data_ if they don't add additional configuration?

It does!

> Would the auto-instrumentation agent then be expected to just register all the defaults provided with the instrumented libraries?

That's what I would expect, and IIRC it's what we did in OC. For example, the predefined default views for gRPC are here: https://github.com/census-instrumentation/opencensus-java/blob/8b1fd5bbf98b27d0ad27394891e0c64c1171cb2b/contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViewConstants.java.

Auto-instrumentation could enable all these views when it loads the gRPC integration.

c24t:
> Does this proposed Views API live at the same level as the other APIs, or is it a default-SDK-only thing?

Views would live in the API package. Developers (as I understand the persona) would use metric instruments in application code, operators (same caveat) would specify views in either static config or code. Aggregators are SDK objects, and a view might rely on some third party package to define a particular aggregator class.

text/0089-metrics-views-api.md
jkwatson: @trask @prydin @tylerbenson thoughts on this? It definitely seems on the surface like it would make the auto-instrumentation job significantly more difficult. *Especially* since auto-instrumentation still has to support multiple backends with potentially different aggregations/metric types.

jkwatson: @c24t This only seems viable to me if there's a way for an exporter author to say (for example) "for everything that's a Measure, create a view with a MinMaxSumCount aggregation.".  Otherwise, I feel like this is a recipe for users thinking that nothing is working, causing a support nightmare.

c24t: That sounds like a good solution to me, and since measures already have unique names I don't see any reason in principle that we couldn't do this.

@bogdandrutu might be able to shed some light on the design of OC views.

c24t: I added a note about this in 46948ac.

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+```
+
+but quantile aggregations, e.g. _p95_ and _p99_ are not.
+Applications that export quantile metrics should use a mergeable aggregations such as [DDSketch](https://arxiv.org/abs/1908.10693), which estimates quantile values with bounded errors, or export raw measurements without aggregation and compute exact quantiles on the backend.
+
+We require aggregations to be mergeable so that they produce the same results regardless of the collection interval, or the number of collection events per export interval.
+
+### Aggregating across label keys
+
+Every measurement is associated with a _LabelSet_, a set of key-value pairs that describes the environment in which the measurement was captured.
```
lzchen, line 149:
```suggestion
Every measurement is associated with _labels_, a set of key-value pairs that describes the environment in which the measurement was captured.
```

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+Other label keys should be dropped before aggregation.
+By default, if the user doesn't specify a set of label keys to track, the aggregator should record all label keys.
+
+Users have three options to configure a view's label keys:
+
+1. Record all label keys.
+    This is the default option, and the behavior of the "ungrouped" integrator in [opentelemetry-specification#347](https://github.com/open-telemetry/opentelemetry-specification/pull/347).
+2. Specify a set of label keys to track at view creation time, and drop other keys from the labelsets of recorded measurements before aggregating.
+    This is the behavior of the "defaultkeys" integrator in [opentelemetry-specification#347](https://github.com/open-telemetry/opentelemetry-specification/pull/347).
+3. Drop all label keys, and aggregate all measurements from the metric instrument together regardless of their labelsets.
```
lzchen, line 165: To clarify, this means that the exported metrics will have no labels associated with them correct?

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+
+The view API should allow the user to specify the set of label keys to track for a given view.
+Other label keys should be dropped before aggregation.
+By default, if the user doesn't specify a set of label keys to track, the aggregator should record all label keys.
+
+Users have three options to configure a view's label keys:
+
+1. Record all label keys.
+    This is the default option, and the behavior of the "ungrouped" integrator in [opentelemetry-specification#347](https://github.com/open-telemetry/opentelemetry-specification/pull/347).
+2. Specify a set of label keys to track at view creation time, and drop other keys from the labelsets of recorded measurements before aggregating.
```
lzchen, line 163: Might want to remove the "labelset" term in this doc.

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+1. `{labelset: {'k1': 'v11'}, value: 1}`
+2. `{labelset: {'k1': 'v12'}, value: 10}`
+3. `{labelset: {'k1': 'v11', 'k2': 'v21'}, value: 100}`
+4. `{labelset: {'k1': 'v12', 'k2': 'v22'}, value: 1000}`
+
+And consider three different views, one that tracks label key _k1_, one that tracks label key _k2_, and one that tracks both.
+
+After each measurement, the aggregator associated with each view has a different set of aggregated values:
+
+| time        | k1                                  | k2  | value | agg([k1])                                   | agg([k2])                                                                | agg([k1, k2]) (default)                                                                                                                  |
```
lzchen, line 182: This is example is really helpful!

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+See the motivation section for an argument against this approach.
+
+This API borrows heavily from [OpenCensus views](https://opencensus.io/stats/view/).
+
+See also [DataDog metric types](https://docs.datadoghq.com/developers/metrics/types/?tab=distribution#metric-types) and [Prometheus metric types](https://prometheus.io/docs/concepts/metric_types).
+
+## Open questions
+
+Should we ignore calls to record measurements from instruments that don't appear in any view?
+This is potentially a nice optimization, but it would mean that no metrics are exported by default from instrumented code.
```
lzchen, line 238: The alternative is to keep using default aggregation types if no view is defined, and then the views would take priority if defined.

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+| 2    | 2    | A   | 10    | `({core: 1}: 1)` <br> `({core: 2}: 10)`   | `({job: A}, 11)`                         | `({core: 1, job: A}, 1)` <br> `({core: 2, job: A}, 10)`                                                                  |
+| 3    | 1    | B   | 100   | `({core: 1}: 100)` <br> `({core: 2}: 10)` | `({job: A}, 11)` <br> `({job: B}, 100)`  | `({core: 1, job: A}, 1)` <br> `({core: 2, job: A}, 10)` <br> `({core: 1, job: B}, 100)`                                  |
+| 4    | 2    | B   | 1000  | `({core: 1}: 1)` <br> `({core: 2}: 1000)` | `({job: A}, 11)` <br> `({job: B}, 1100)` | `({core: 1, job: A}, 1)` <br> `({core: 2, job: A}, 10)` <br> `({core: 1, job: B}, 100)` <br> `({core: 2, job: B}, 1000)` |
+
+Here the aggregated value of _agg([job])_ depends on the value of the "core" label, even though it's not preserved in the aggregation: it is the sum over the last-reported value for each core.
+
+This makes configuring views for Observers significantly more complicated than for Measures.
```
jmacd, line 220: I became confused at this section. I wonder how it changes given that OTEP 88 has written "Last Value" aggregation out of existence. I believe with the current state following OTEP 88, synchronous and asynchronous views should be the same.  The remaining difference is that Observers are reported strictly once per interval.

I had trouble understanding the text above, too. It makes me wonder if we have different mental models for Observers, still.

c24t:
> Well written and easy to understand! What would be the next steps?

@lzchen and I talked offline, but for the benefit of the github conversation: the next steps are to finalize the metrics API in https://github.com/open-telemetry/oteps/pull/98, figure out the data model in https://github.com/open-telemetry/opentelemetry-proto/pull/137, and start prototyping in clients like we did for e.g. the context API.

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+| Prometheus | [Histogram](https://prometheus.io/docs/concepts/metric_types/#histogram)                           | Measure              | Histogram         |
+| Prometheus | [Summary](https://prometheus.io/docs/concepts/metric_types/#summary)                               | Measure              | Sketch\* or Exact |
+
+A well-designed view API should not require extensive configuration for typical use cases.
+It should also be expressive enough to support custom aggregations and exposition formats without requiring APM vendors to write custom SDKs.
+In short, it should make easy things easy and hard things possible.
+
+## Internal details
+
+A view is defined by:
```
cijothomas, line 90: Do we allow users to create/modify Views in the middle of the program? Or Views must be created at initialization time?
I think the View Config should be provided at the time of MeterProvider construction, before any Meters are created. So that meters know about the which are the required Views, before the 1st metric bind or recording comes.

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+
+For example, consider a Sum-aggregated Counter instrument that captures four consecutive measurements:
+
+1. `{labelset: {'k1': 'v11'}, value: 1}`
+2. `{labelset: {'k1': 'v12'}, value: 10}`
+3. `{labelset: {'k1': 'v11', 'k2': 'v21'}, value: 100}`
+4. `{labelset: {'k1': 'v12', 'k2': 'v22'}, value: 1000}`
+
+And consider three different views, one that tracks label key _k1_, one that tracks label key _k2_, and one that tracks both.
+
```
cijothomas, line 179: From this example, can I assume that the labelset reduction/drop **should** occur at the Aggregator itself. 
i.e If my only view is add(k1), I'll end up with just 2 time series - as there were 2 unique values for k1.

OR, does the aggregator still keep all dimensions (storing k1_uniquecount * k2_uniquecount timeseries), and exporter/batcher drops the dimensions?

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+> [Users do not have a facility in the API to select the aggregation they want for particular instruments.](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-metrics.md#aggregations)
+
+This limitation prevents users from:
+
+- Converting aggregated data to vendor-specific exposition formats
+- Configuring aggregation options, e.g. histogram bucket boundaries
+- Specifying multiple aggregations for a single metric
+
+In addition, it's often desirable to aggregate metrics with respect to a specific set of label keys to reduce memory use and export data size.
```
cijothomas, line 38: Would it be correct to say that Views is the only way to pick/drop dimensions? Or Views is just one way. And batcher can still drop dimensions further?

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+
+A _view API_ would allow users to use custom aggregation types, associate individual metrics with one or more aggregations, configure aggregation options, and specify the set of label keys to use to aggregate each metric.
```
cijothomas, line 40: Can you clarify if the following scenario is achievable with Views.?

A counter is tracked with 2 dimensions, url,statuscode. Lets says there are 10 possible values for url and 5 possible values for statusCode. This would potentially result in 10*5 = 50 time series stored by default.

User don't want to pay for that much. He cares about url, statuscode but not their combination. So user defines 2 views, view1(url), and view(statuscode). Now this would result in 10 + 5 = 15 time series only.

So can we say Views can be used to achieve timeseries reduction in the above way?

text/0089-metrics-views-api.md
lzchen: I do not believe we should allow users to do that. There are some [issues](https://github.com/open-telemetry/opentelemetry-python/pull/596#discussion_r411819266) with defining Views after a collection pass has been done, it gets too complicated to deal with those so we should enforce Views to be created and immutable before recording.

text/0089-metrics-views-api.md
toumorokoshi: I'd also expect what you stated to be achievable. programatically something like:

1. declare metric counter-A
2. create View sum_by_url, attach to counter-A
3. create View sum_by_statuscode, attach to counter-A

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+
+Metric instruments are API objects that record measurements from application code.
+These measurements are aggregated by _Aggregators_, which are SDK objects.
+Aggregated measurement data are passed to _Exporters_, which convert these data into APM-specific exposition formats.
+
+Each view describes a relationship between a metric instrument and an aggregation, and specifies the set of label keys to preserve in the aggregated data.
+
+Views are tracked in a registry, which exporters may use to get the current aggregate values for all registered aggregations at export time.
+Each registered view represents a metric that is meant to be exported.
+As a corollary, the SDK can choose to drop measurements from any metric instrument that does not appear in a registered view.
```
toumorokoshi, line 52: should behavior for the opentelemetry-sdk be stated here? Seems like we should make a choice one way or the other.

I would opt to drop measurements, as it could be significant memory overhead to just keep measurements that will never be exported.

tsloughter: +1 to not having the SDK be able to "choose" but that being the required way it works.

This does not break with the desire to not require users to define views to get metrics if we go with my suggestion of default view generation for each instrument -- since there would by default be a view, so measurements would not be dropped without the end user/operator explicitly disable default views or setting their own.

cijothomas: I'd like to get more clarity on "Should exporters be even aware of ViewRegistry?" Or only Meter implementation need to be aware of ViewRegistry.

Exporters will continue to simply have a method like Export(collection<metrics>). And Meter implementation takes of dropping/retaining dimensions, choosing right aggregation etc from ViewRegistry.

cijothomas:
> > Does this proposed Views API live at the same level as the other APIs, or is it a default-SDK-only thing?
> 
> Views would live in the API package. Developers (as I understand the persona) would use metric instruments in application code, operators (same caveat) would specify views in either static config or code. Aggregators are SDK objects, and a view might rely on some third party package to define a particular aggregator class.

I vague remember discussion about this in SIG meeting - Views API would be part of the SDK, not in API. It makes sense to me to keep Views API in the SDK only, as the intended users are application writers/operations, and not Library authors.

text/0089-metrics-views-api.md
lzchen: This can even be achieved with only one View.

1. Declare metric counter-A
2. create View sum_by_url_and_statuscode, attach to counter-A

text/0089-metrics-views-api.md
lzchen: I believe from last week's metrics SIG, it was decided that any measurement that was not registered with a View would be aggregated with a default aggregation.

lzchen:
> > > Does this proposed Views API live at the same level as the other APIs, or is it a default-SDK-only thing?
> > 
> > 
> > Views would live in the API package. Developers (as I understand the persona) would use metric instruments in application code, operators (same caveat) would specify views in either static config or code. Aggregators are SDK objects, and a view might rely on some third party package to define a particular aggregator class.
> 
> I vague remember discussion about this in SIG meeting - Views API would be part of the SDK, not in API. It makes sense to me to keep Views API in the SDK only, as the intended users are application writers/operations, and not Library authors.

Yes I believe the comment was made before the discussion in the SIG meeting. You are correct.

text/0089-metrics-views-api.md
cijothomas: > This can even be achieved with only one View.
> 
> 1. Declare metric counter-A
> 2. create View sum_by_url_and_statuscode, attach to counter-A

With one view, the total timeseries could be 10*5. With 2 views, its 10+5. This makes a lot of difference in terms of cost. (memory cost, and financial cost of exporting more time series to APM backends)

The goal is to reduce the # of timeseries, hence splitting into multiple views.

lzchen: Ahh I misread your question. Yes you are correct.

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Explanation
+
+Metric instruments are API objects that record measurements from application code.
+These measurements are aggregated by _Aggregators_, which are SDK objects.
+Aggregated measurement data are passed to _Exporters_, which convert these data into APM-specific exposition formats.
+
+Each view describes a relationship between a metric instrument and an aggregation, and specifies the set of label keys to preserve in the aggregated data.
+
+Views are tracked in a registry, which exporters may use to get the current aggregate values for all registered aggregations at export time.
```
cijothomas, line 50: Is Views a global thing, and hence must be registered with MeterProvider, so that all meters created the MeterProvider will get access to View.?

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+    This is the default option, and the behavior of the "ungrouped" integrator in [opentelemetry-specification#347](https://github.com/open-telemetry/opentelemetry-specification/pull/347).
+2. Specify a set of label keys to track at view creation time, and drop other keys from the labelsets of recorded measurements before aggregating.
+    This is the behavior of the "defaultkeys" integrator in [opentelemetry-specification#347](https://github.com/open-telemetry/opentelemetry-specification/pull/347).
+3. Drop all label keys, and aggregate all measurements from the metric instrument together regardless of their labelsets.
+    This is equivalent to using an empty list with option 2.
+
+Because we don't require the user to specify the set of label keys up front, and because we don't prevent users from recording measurements with missing labels in the API, some label values may be undefined.
+Aggregators should preserve undefined label values, and exporters may convert them as required by the backend.
+
+For example, consider a Sum-aggregated Counter instrument that captures four consecutive measurements:
```
cijothomas, line 171: Can you also add a note about boundinstruments and views? Boundinstruments are supposed to be the fastest way to record a value, as it avoids a lookup for the time series. With Views, bound instruments should still be the fastest, and should not be doing lookups.

This maybe implementation detail, but calling it out in spec as well.

text/0089-metrics-views-api.md
toumorokoshi: Additional thought around this: would it be valuable to enable specific aggregations to be tied to specific exporters?

I feel like this is another reason why aggregations should be tied to a MeterProvider. Then you can choose different meterprovider sets for different metrics, as necessary.

Something like:
```
MeterProvider
    - exporters
    - aggregators
    - metrics (this may be unnecessary)
```

text/0089-metrics-views-api.md
toumorokoshi: Curious how that works in practice. It would require *something* to iterate through all metrics during every export and see if they have a view already registered.

text/0089-metrics-views-api.md
toumorokoshi: I need to understand the details a bit more on the issues, but there's tremendous value to being able to modify views on runtime. In a conversation with @c24t a while back, he mentioned that there's a way to configure views across application dynamically for debugging purposes. 

For example, imagine that one of the keys you store is the service name sending you the request. When your application starts seeing a significant increase in requests, you can turn on a view that splits the metrics by service name, and see which service is causing issues. Once resolved, you can turn them off, ensuring that you don't have high cardinality data points that consume significant metric count after the fact.

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+
+- A metric instrument
+- An aggregation
+- An optional set of label keys to preserve
+- An optional unique name
+- An optional description
+
+The aggregation may be configured with options specific to its type.
+For example, a _Histogram_ aggregation may be configured with bucket boundaries, e.g. `{0, 1, 10, 100, 200, 1000, inf}`.
+A _Sketch_ aggregation that estimates order statistics (i.e. quantiles), may be configured with a set of predetermined quantiles, e.g. `{.5, .95, .99, 1.00}`.
```
toumorokoshi, line 100: Could this be called "quantile" aggregation? A sketch doesn't immediately make me think of quantiles.

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+
+Each view has a unique name, which may be automatically determined from the metric instrument and aggregation.
+Implementations should refuse to register two views with the same name.
+
+Note that a view does not describe the type (e.g. `int`, `float`) or unit of measurement (e.g. "bytes", "milliseconds") of the metric to be exported.
+The unit is determined by the metric instrument, and the aggregation and exporter may preserve or change the unit.
```
toumorokoshi, line 106: It seems like there's only two operations which are appropriate for a unit modification on a metric instrument:

1. preserve the unit
2. change the unit entirely (e.g. to count)

I feel like it may be valuable to call out that there won't be a modification such as "seconds" to "milliseconds". But maybe that's just obvious.

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+For example, a hypothetical _MinMaxSumMeanCount_ aggregation of an int-valued millisecond latency measure may be exported as five separate metrics to the APM backend:
+
+- An int-valued _min_ metric with "ms" units
+- An int-valued _max_ metric with "ms" units
+- An int-valued _sum_ metric with "ms" units
+- A **float**-valued _mean_ metric with "ms" units
+- An **int** valued, unitless (i.e. unit "1") _count_ metric
+
+This OTEP doesn't propose a particular API for Aggregators, just that the API is sufficient for exporters to get all this information, including:
```
toumorokoshi, line 115: Should the otep recommend an API? reading the examples, it seems like something that may work would be (name, value, unit). I believe it seems it satisfies the examples below.

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+
+For every partition `{S_1, ..., S_N}` of `S`.
+
+For example, the _min_ aggregation is trivially mergeable:
+
+```
+min([1, 2, 3, 4, 5, 6]) = min([min([1]), min([2, 3]), min([4, 5, 6])])
+```
+
+but quantile aggregations, e.g. _p95_ and _p99_ are not.
```
toumorokoshi, line 142: Does this mean that quantiles are not supported as an aggregation? There's a strong need for quantiles, so I assume there's some way to ensure that it is measured and exported, but not clear how.

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+Applications that export quantile metrics should use a mergeable aggregations such as [DDSketch](https://arxiv.org/abs/1908.10693), which estimates quantile values with bounded errors, or export raw measurements without aggregation and compute exact quantiles on the backend.
+
+We require aggregations to be mergeable so that they produce the same results regardless of the collection interval, or the number of collection events per export interval.
```
toumorokoshi, line 145: I guess again looking at quantiles as the edge case: what's the need for ensuring that the produced result are the same, regardless of collection interval? Is there some caveat where the collection interval cannot be guaranteed? I guess in current API design it's in the pushController side of things.

text/0089-metrics-views-api.md
```diff
@@ -0,0 +1,264 @@
...
+
+### Aggregating across label keys
+
+Every measurement is associated with a _LabelSet_, a set of key-value pairs that describes the environment in which the measurement was captured.
+Label keys and values may be extracted from the [correlation context](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-correlationcontext.md) at the time of measurement.
+They may also be set by the user [at the time of measurement](https://github.com/open-telemetry/opentelemetry-specification/blob/f4aeb318a5b77c9c39132a8cbc5d995e222d124f/specification/api-metrics-user.md#direct-instrument-calling-convention) or at the time at which [the metric instrument is bound](https://github.com/open-telemetry/opentelemetry-specification/blob/f4aeb318a5b77c9c39132a8cbc5d995e222d124f/specification/api-metrics-user.md#bound-instrument-calling-convention).
```
toumorokoshi, line 151: How specifically are values from the correlation context piped in to the labelset? I figured that the user would still be responsible for that, but as this is calling out label extraction as a separate line, it sounds like that's not the case.

jkwatson:
My opinions on this:
1) I think we absolutely should have default aggregations, and I think the metric SIG did agree to that.
2) I don't think the exporters should have to interact with the Views API unless they want to tell the SDK that they want only delta aggregations for a given instrument type (for example). Exporters shouldn't their data via views, but the SDK uses the view definition to figure out what aggregation/batching strategy should be used.
3) Things I think we should shoot for for GA: Being able to specify label reductions, delta/cumulative temporality, and the type of aggregation to use, based on the instrument descriptor. 

This OTEP seems to range far and wide outside of the scope for a SDK-only basic views API that I think we should shoot for for GA.

jmacd:
Closing this as obsolete. 
❤️ @c24t

morigs:
Excuse me, it's unclear what's the current vision on metrics aggregation. Are there other OTEPs with the same target?

PR #91, opened 2020-03-25
Add Logs to OpenTelemetry vocabulary

tigrannajaryan:
This commit proposes the initial vocabulary for logs and event to be used across
OpenTelemetry project.

This is an initial proposal and we can extend / clarify the vocabulary as we go.

### DIFF
--- a/text/0091-logs-vocabulary.md
+++ b/text/0091-logs-vocabulary.md
@@ -1,0 +1,49 @@
+# Logs: Vocabulary
+
+This documents defines the vocabulary for logs to be used across OpenTelemetry project.
+
+## Motivation
+
+We need a common language and common understanding of terms that we use to
+avoid the chaos experienced by the builders of the Tower of Babel.
+
+## Proposal
+
+OpenTelemetry specification already contains a [vocabulary](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/overview.md)
+for Traces, Metrics and other relevant concepts.
+
+This proposal is to add the following concepts to the vocabulary.
+
+### Log Record
+
+A recording of an event. Typically the record includes a timestamp indicating
+when the event happened as well as other data that describes what happened,
+where it happened, etc.
+
+### Embedded Log 
+
+A log record embedded inside a [Span](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#span) 
+object, in the [Events](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#add-events) list.
+
+### Standalone Log
+
+A log record that is not embedded inside a Span and is recorded elsewhere.
+
+### Log Attributes
+
+Key/value pairs contained in a Log Record.
+
+### Structured Logs
+
+Logs that are recorded in a format that has a well-defined structure that allows
+to differentiate between different elements of a Log Record (e.g. the Timestamp,
+the Attributes, etc). For example [Syslog, RFC5425](https://tools.ietf.org/html/rfc5424)
+protocol defines `structured-data` format.
+
+### Flat File Logs
+
+Logs recorded in text files, often one line per log record (although multiline
+records are possible too). There is no common industry agreement whether
+logs written to text files in more structured formats (e.g. JSON files)
+are considered Flat File Logs or no. Where such distinction is important it is
+recommended to call it out specifically.
--- a/text/0091-logs-vocabulary.md
+++ b/text/0091-logs-vocabulary.md
@@ -20,30 +20,39 @@ A recording of an event. Typically the record includes a timestamp indicating
 when the event happened as well as other data that describes what happened,
 where it happened, etc.
 
-### Embedded Log 
+Also known as Log Entry.
 
-A log record embedded inside a [Span](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#span) 
+### Log
+
+Sometimes used to refer to a collection of Log Records. May be ambiguous, since
+people also sometimes use `Log` to refer to a single `Log Record`, thus this
+term should be used carefully and in the context where ambiguity is possible
+additional qualifiers should be used (e.g. `Log Record`).
+
+### Embedded Log
+
+`Log Records` embedded inside a [Span](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#span) 
 object, in the [Events](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#add-events) list.
 
 ### Standalone Log
 
-A log record that is not embedded inside a Span and is recorded elsewhere.
+`Log Records` that are not embedded inside a `Span` and are recorded elsewhere.
 
 ### Log Attributes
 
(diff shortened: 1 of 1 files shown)


text/0091-logs-vocabulary.md
```diff
@@ -0,0 +1,49 @@
...
+
+A log record that is not embedded inside a Span and is recorded elsewhere.
+
+### Log Attributes
+
+Key/value pairs contained in a Log Record.
+
+### Structured Logs
+
+Logs that are recorded in a format that has a well-defined schema that allows
```
jaredcnance: Is "well-defined schema" a requirement? I would consider a JSON log record to be a structured log since it has a well-defined structure but may not have a well-defined schema.

tigrannajaryan: Good point. Reworded.

text/0091-logs-vocabulary.md
```diff
@@ -0,0 +1,49 @@
...
+avoid the chaos experienced by the builders of the Tower of Babel.
+
+## Proposal
+
+OpenTelemetry specification already contains a [vocabulary](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/overview.md)
+for Traces, Metrics and other relevant concepts.
+
+This proposal is to add the following concepts to the vocabulary.
+
+### Log Record
```
arminru: "Log Entry" would sound more natural to me. What do you think?

text/0091-logs-vocabulary.md
```diff
@@ -0,0 +1,49 @@
...
+
+A recording of an event. Typically the record includes a timestamp indicating
+when the event happened as well as other data that describes what happened,
+where it happened, etc.
+
+### Embedded Log 
+
+A log record embedded inside a [Span](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#span)
```
arminru: I'd expect an "embedded log" or "standalone log" to be the whole log (file/stream) composed of multiple records/entries rather than a single record/entry.

text/0091-logs-vocabulary.md
```diff
@@ -0,0 +1,49 @@
...
+A log record embedded inside a [Span](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#span) 
+object, in the [Events](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#add-events) list.
+
+### Standalone Log
+
+A log record that is not embedded inside a Span and is recorded elsewhere.
+
+### Log Attributes
+
+Key/value pairs contained in a Log Record.
```
arminru: I'd also expect "log attributes" to be metadata about a whole log file/stream (e.g. producer, host or service information, time span, size) rather than attributes on a single log record/entry.

text/0091-logs-vocabulary.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Structured Logs
+
+Logs that are recorded in a format that has a well-defined structure that allows
+to differentiate between different elements of a Log Record (e.g. the Timestamp,
+the Attributes, etc). For example [Syslog, RFC5425](https://tools.ietf.org/html/rfc5424)
+protocol defines `structured-data` format.
```
arminru:
```suggestion
the Attributes, etc). The _Syslog protocol_ ([RFC 5425](https://tools.ietf.org/html/rfc5424)),
for example, defines a `structured-data` format.
```

the rendered version looks a bit odd otherwise

text/0091-logs-vocabulary.md
```diff
@@ -0,0 +1,49 @@
...
+
+A log record that is not embedded inside a Span and is recorded elsewhere.
+
+### Log Attributes
+
+Key/value pairs contained in a Log Record.
+
+### Structured Logs
+
+Logs that are recorded in a format that has a well-defined structure that allows
```
arminru:
```suggestion
Logs that are recorded in a format which has a well-defined structure that allows
```

text/0091-logs-vocabulary.md
```diff
@@ -0,0 +1,49 @@
...
+to differentiate between different elements of a Log Record (e.g. the Timestamp,
+the Attributes, etc). For example [Syslog, RFC5425](https://tools.ietf.org/html/rfc5424)
+protocol defines `structured-data` format.
+
+### Flat File Logs
+
+Logs recorded in text files, often one line per log record (although multiline
+records are possible too). There is no common industry agreement whether
+logs written to text files in more structured formats (e.g. JSON files)
+are considered Flat File Logs or no. Where such distinction is important it is
```
arminru:
```suggestion
are considered Flat File Logs or not. Where such distinction is important it is
```

text/0091-logs-vocabulary.md
jaredcnance: I think there are two classes of data which can be included in structured logs:

- attributes: key/value pairs that are specific to a single log record and often have high cardinality values (e.g. timestamp)
- metadata: key/value pairs that apply to one or more log records, generally constant for a given span or may be constant for the lifetime of the process (e.g. host, containerId, traceId, etc.).

tigrannajaryan: I agree with the distinction.

The constants part of "metadata" is stored in a `Resource` in OpenTelemetry. There is already a definition of a Resource.

The non-const parts are Attributes of a log record. The fact that they can be identical for a batch of log records is a data modeling topic and does not belong to the vocabulary (at least not yet, until we have the data model). We can have further clarification of types of attributes, but for now I would refrain from adding anything to the vocabulary. It is best to discuss this in a Data Model proposal (which I believe we inevitably need to have).

(BTW, "metadata" is not a well defined term, IMO it is best to avoid using it in this discussion since people have different understanding of what metadata means).

text/0091-logs-vocabulary.md
tigrannajaryan: Fixed.

text/0091-logs-vocabulary.md
tigrannajaryan: Fixed.

text/0091-logs-vocabulary.md
tigrannajaryan: Fixed.

text/0091-logs-vocabulary.md
tigrannajaryan: Both sound good and I have seen both used interchangeably. I added Log Entry as a valid synonym.

text/0091-logs-vocabulary.md
tigrannajaryan: Good point. Modified to refer to plural "Log Records".

tigrannajaryan:
I also added `Log` with a recommendation not to use it without additional qualifiers to avoid confusion.

text/0091-logs-vocabulary.md
arminru: I agree with both of your comments.
My comment was about the term "log attributes" itself, which sounds like attributes that apply to the whole log file/stream (collection of logs) at once. These could therefore, for example, be expressed and stored in a Resource or preamble to a log file/stream rather than one single log record. I'd rather call the key/value pairs contained in a single log record "log record attributes" to avoid this ambiguity.

PR #93, opened 2020-04-08
New metric instruments from OTEP 88

jmacd:
This PR follows OTEP 88 with a concrete proposal for 4 new instruments.

### DIFF
--- a/text/0093-metric-instruments.md
+++ b/text/0093-metric-instruments.md
@@ -50,29 +50,42 @@ as well, to support built-in value types (e.g., timestamps).
 
 ## Internal details
 
-The existing specification includes three instruments.  In this
-proposal, the two foundational instruments become abstract, in the
-sense that all synchronous instruments are instances of a Measure
-instrument and all asynchronous instruments are instances of an
-Observer instrument.  New instrument names are given to the
-unrestricted, unrefined versions of rthe foundational instruments:
+The existing specification includes three instruments: Counter,
+Measure, and Observer.  In this proposal, the two foundational
+instruments, Measure and Observer (synchronous and asynchronous,
+respectively), become abstract names, in the sense that all
+synchronous instruments are instances of a Measure instrument and all
+asynchronous instruments are instances of an Observer instrument.  New
+instrument names are given to the unrestricted, unrefined versions of
+the foundational instruments.  The following table summarizes this
+proposal:
 
-1. **Distribution** is an an unrefined Measure instrument.  Distribution accepts positive or negative values and uses MinMaxSumCount aggregation by default.
-2. **LastValueObserver** is an unrefined Observer instrument.  LastValueObserver accepts positive or negative values and uses MinMaxSumCount aggregation by default.
+| Existing name | Proposed name | Sync or Async | Refinements |
+| ------------- | ------------- | ------------- | ----------- |
+| Counter       | Counter       | Sync          | Sum-only, Non-negative, Non-negative-rate |
+| Measure       | Distribution  | Sync          | _None_  |
+|               | UpDownCounter | Sync          | Sum-only |
+|               | Timing        | Sync          | Non-negative, correct Duration units  |
+| Observer      | LastValueObserver | Async     | _None_ |
+|               | DeltaObserver | Async         | Sum-only, Non-negative, Non-negative-rate |
+|               | CumulativeObserver | Async    | Sum-only, Precomputed-sum, Non-negative-rate |
 
-The existing Counter instrument is unchanged in this proposal.  It has
-Sum-only and Non-negative refinements, which imply the
-Non-negative-rate refinement.  Counter uses Sum aggregation by default
+### Synchronous instruments
 
-Two new synchronous instruments are introduced in this proposal.
+Two new synchronous instruments are introduced in this proposal, bringing the total to four.
 
-1. **UpDownCounter** is a Sum-only instrument with no other refinements.  It supports capturing positive and negative changes to a sum (deltas).  UpDownCounter uses Sum aggregation by default.
-2. **Timing** is a Non-negative instrument specialized for the native clock duration measured on the platform.  It ensures that duration values are always captured with correct units, that ensures exporters can convert duration measurements correctly.
+1. **Counter** remains unchanged.  It uses Sum aggregation by default.
+2. **Distribution** is an an unrefined Measure instrument.  Distribution accepts positive or negative values and uses MinMaxSumCount aggregation by default.
+3. **UpDownCounter** is a Sum-only instrument with no other refinements.  It supports capturing positive and negative changes to a sum (deltas).  UpDownCounter uses Sum aggregation by default.
+4. **Timing** is a Non-negative instrument specialized for the native clock duration measured on the platform.  It ensures that duration values are always captured with correct units, that ensures exporters can convert duration measurements correctly.
 
-Two new asynchronous instruments are introduced in this proposal.
+### Asynchronous instruments
 
-1. **CumulativeObserver** is a Sum-only, Precomputed-sum, Non-negative, Non-negative-rate instrument useful when reporting precomputed sums.
+Two new asynchronous instruments are introduced in this proposal, bringing the total to three.
+
+1. **LastValueObserver** is an unrefined Observer instrument.  LastValueObserver accepts positive or negative values and uses MinMaxSumCount aggregation by default.
 2. **DeltaObserver** is a Sum-only, Non-negative, Non-negative-rate instrument useful for capturing deltas that accumulate during a collection interval.
+3. **CumulativeObserver** is a Sum-only, Precomputed-sum, Non-negative, Non-negative-rate instrument useful when reporting precomputed sums.
 
 Both new asynchronous instruments are meant to be used for aggregating rate information from a callback.
 
@@ -103,7 +116,7 @@ instruments.
 
 There are potential incompatibilities related to the input range of
 existing exporters and the new instruments.  For example, a Prometheus
-Histogram is used to capture distribution with non-negative values.
+Histogram is used to capture a distribution with non-negative values.
 This proposal does not specify a standard AbsoluteDistribution
 instrument, which has the corresponding input-range restriction.
 
--- a/text/0093-metric-instruments.md
+++ b/text/0093-metric-instruments.md
@@ -18,6 +18,20 @@ The four instrument refinements discussed in OTEP 88 are:
 * Precomputed-sum: When the application has computed a cumulative sum itself
 * Non-negative-rate: When a negative rate is invalid.
 
+These refinements are not for exposing directly to users at the API
+level.  These concepts are purely explanatory, used to define the
+properties of the metric instruments presented in the API.  Following
+OTEP 88:
+
+* Users will select instruments based on their specified properties
+* Instruments are associated with a Descriptor, that includes the instrument kind (an enumeration)
+* Exported metric events include the instrument Descriptor, allowing exporters to interpret event values.
+
+In other words, these refinements serve to define the set of
+instruments.  Both users and exporters will deal in concrete kinds of
+instrument, these refinements are just for explaining their
+properties.
+
 OTEP 88 describes how we are meant to compute rate information from
 instruments having the Non-negative-rate refinement.  Temporal
 aggregation (over time) must be treated as a special case, compared
--- a/text/0093-metric-instruments.md
+++ b/text/0093-metric-instruments.md
@@ -56,6 +56,24 @@ properties of the instrument.
 OTEP 88 also also proposes to support language-specific specialization
 as well, to support built-in value types (e.g., timestamps).
 
+### Delta, Cumulative, and Instantaneous instruments
+
+Instruments can be categorized as to whether their values are "Delta",
+"Cumulative", or "Instantaneous".  This depends on which refinements
+are present and is required knowledge for exporters of the Sum-only
+instruments, since they may need to convert back and forth between
+delta-values and cumulative-values.  The categorization is determined
+as follows:
+
+Instruments with the Sum-only refinement, but without the
+Precomputed-sum refinement, produce delta instruments.
+
+Instruments with the Sum-only and Precomputed-sum refinements produce
+cumulative instruments.
+
+Instruments without the Sum-only refinement produce instantaneous
+measurements.
+
 ## Internal details
 
 The existing specification includes three instruments: Counter,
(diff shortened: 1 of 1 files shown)


text/0093-metric-instruments.md
```diff
@@ -0,0 +1,104 @@
...
+2. **LastValueObserver** is an unrefined Observer instrument.  LastValueObserver accepts positive or negative values and uses MinMaxSumCount aggregation by default.
+
+The existing Counter instrument is unchanged in this proposal.  It has
+Sum-only and Non-negative refinements, which imply the
+Non-negative-rate refinement.  Counter uses Sum aggregation by default
+
+Two new synchronous instruments are introduced in this proposal.
+
+1. **UpDownCounter** is a Sum-only instrument with no other refinements.  It supports capturing positive and negative changes to a sum (deltas).  UpDownCounter uses Sum aggregation by default.
+2. **Timing** is a Non-negative instrument specialized for the native timestamp of the platform.  It ensures that values are always reported in conventional units (milliseconds).
```
jkwatson, line 56: Is it worth calling out float vs integer on this one? 
Is the idea that OTel will say "this is the timing instrument, it will always measure times in milliseconds", or would the units be customizable by instrumentation?

bogdandrutu: Personally I would do ns because that is usually the unit for the monotonic time available in couple of languages like Java, go. If we go with millis then is definitely a float :)

text/0093-metric-instruments.md
```diff
@@ -0,0 +1,104 @@
...
+as well, to support built-in value types (e.g., timestamps).
+
+## Internal details
+
+The existing specification includes three instruments.  In this
+proposal, the two foundational instruments become abstract, in the
+sense that all synchronous instruments are instances of a Measure
+instrument and all asynchronous instruments are instances of an
+Observer instrument.  New instrument names are given to the
+unrestricted, unrefined versions of rthe foundational instruments:
```
bogdandrutu, line 44: s/rthe/the?

text/0093-metric-instruments.md
```diff
@@ -0,0 +1,104 @@
...
+
+1. **Distribution** is an an unrefined Measure instrument.  Distribution accepts positive or negative values and uses MinMaxSumCount aggregation by default.
+2. **LastValueObserver** is an unrefined Observer instrument.  LastValueObserver accepts positive or negative values and uses MinMaxSumCount aggregation by default.
```
bogdandrutu, line 47: I think this instrument tries to combine actually 2 instruments:
* "DistributionObserver" which accepts a `List<Value>` for every label combination where a MinMaxSumCount makes a lot of sense.
* "LastValueObserver" (GaugeObeserver) which accepts one value per label combination where a "Sum" makes more sense if we talk about merging labels - as an example memory-usage.

Otherwise is confusing to me how the "MinMaxSumCount" aggregation applies to one value reported.

I would suggest that we use "Sum" as default aggregation for this instrument, and add later a DistributionObserver if needed.

text/0093-metric-instruments.md
```diff
@@ -0,0 +1,104 @@
...
+refinements, and each with a carefully selected name for the
+properties of the instrument.
+
+OTEP 88 also also proposes to support language-specific specialization
+as well, to support built-in value types (e.g., timestamps).
+
+## Internal details
+
+The existing specification includes three instruments.  In this
+proposal, the two foundational instruments become abstract, in the
```
bogdandrutu, line 40: May make sense to iterate the once that exists for people to have a full picture. Maybe a table that includes all old + new?

text/0093-metric-instruments.md
jkwatson: for sure. I'm mostly wondering if we're planning on building that in to the instrument refinement. I think it's worth calling out, either way.

c24t, line 56: Not `Timer`?

c24t: It's surprising to see restrictions on the type (int/float) and unit (ns/ms) here when the refinements above don't mention either. I think it's fine for the spec to describe conventions for instruments, but this does suggest there are more refinements in practice than are described in this OTEP.

text/0093-metric-instruments.md
```diff
@@ -0,0 +1,104 @@
...
+1. **Distribution** is an an unrefined Measure instrument.  Distribution accepts positive or negative values and uses MinMaxSumCount aggregation by default.
+2. **LastValueObserver** is an unrefined Observer instrument.  LastValueObserver accepts positive or negative values and uses MinMaxSumCount aggregation by default.
+
+The existing Counter instrument is unchanged in this proposal.  It has
+Sum-only and Non-negative refinements, which imply the
+Non-negative-rate refinement.  Counter uses Sum aggregation by default
+
+Two new synchronous instruments are introduced in this proposal.
+
+1. **UpDownCounter** is a Sum-only instrument with no other refinements.  It supports capturing positive and negative changes to a sum (deltas).  UpDownCounter uses Sum aggregation by default.
```
c24t, line 55: Do we still need to call this `UpDown` if there's no `Counter` instrument to distinguish it from?

text/0093-metric-instruments.md
```diff
@@ -0,0 +1,104 @@
...
+There are known limitations caused by not standardizing all possible
+instrument refinements.  Creating too many instruments will create
+confusion of its own, so we choose to limit the set of standard
+instruments.  It is possible that SDK support for configuring
+alternate aggregations will avoid the need for more standard
+instruments.
+
+There are potential incompatibilities related to the input range of
+existing exporters and the new instruments.  For example, a Prometheus
+Histogram is used to capture distribution with non-negative values.
```
c24t, line 92:
```suggestion
Histogram is used to capture a distribution with non-negative values.
```

text/0093-metric-instruments.md
```diff
@@ -0,0 +1,104 @@
...
+## Internal details
+
+The existing specification includes three instruments.  In this
+proposal, the two foundational instruments become abstract, in the
+sense that all synchronous instruments are instances of a Measure
+instrument and all asynchronous instruments are instances of an
+Observer instrument.  New instrument names are given to the
+unrestricted, unrefined versions of rthe foundational instruments:
+
+1. **Distribution** is an an unrefined Measure instrument.  Distribution accepts positive or negative values and uses MinMaxSumCount aggregation by default.
```
c24t, line 46: As a naming nit: if the term `Measure` doesn't appear in the user-facing API I think it would be simpler to call this `Measure` and name the abstract type (which AFAICT is only used internally) to something like `AbstractMeasure`.

text/0093-metric-instruments.md
```diff
@@ -0,0 +1,104 @@
...
+This proposal uses that proposal as a starting point.
+
+## Explanation
+
+The four instrument refinements discussed in OTEP 88 are:
+
+* Sum-only: When computing only a sum is the instrument's primary purpose
+* Non-negative: When negative values are invalid
+* Precomputed-sum: When the application has computed a cumulative sum itself
+* Non-negative-rate: When a negative rate is invalid.
```
c24t, line 19: One question this OTEP doesn't answer: do we expect these refinements to show up in the API? Or is this just the set of refinements required for the extended instruments described in 88?

text/0093-metric-instruments.md
c24t, line 44: It would help readability to make the instruments explicit here:

- Measure
  - Distribution
  - UpDownCounter
  - Timing
- Observer
  - CumulativeObserver
  - DeltaObserver

text/0093-metric-instruments.md
jmacd: I've added some text to clarify that users and exporters will not refer to these refinements directly.

text/0093-metric-instruments.md
jmacd: The abstract Measure doesn't need a name, it's the concrete instrument without refinements we need to name. There was debate over this and I don't feel extremely good about either option right now. I like "Counter" because it sounds like a real kind of Instrument you'd use for counting--it's not a figurative word. "Observer" is a thing that observes, I'm happy with it too. 

"Gauge" was figurative, like "reading a gauge", except that the operation is to set something.  "Gauger" doesn't make sense, I feel it was not a good instrument name.

By these criteria, neither Distribution nor Measure is not a good name, they describe the thing the instrument produces, not the thing the instrument does. Perhaps `Recorder` is makes more sense?

This is at the very least making me want to remove "Measure" and "Observer" and refer to synchronous and asynchronous instruments everywhere.

text/0093-metric-instruments.md
jmacd: I do not believe there should be a DistributionObserver. If you want to record a list of values, please use a synchronous instrument.

MinMaxSumCount is a good aggregator for LastValue because it includes the Sum but also characterizes the distribution w/ min/max and average. If you wanted only a Sum aggregation, why not use a DeltaObserver?

MinMaxSumCount is redundant when it includes only a single value, but when pre-aggregating LastValueObserver values you'll get both a distribution and a sum. Exporters that only want to export a sum can do so, the MMSC is cheap enough not to worry about this, I think.

text/0093-metric-instruments.md
jmacd: There is still a Counter! This document was trying to describe new instrument names, and that's the only one that survived. I've made this clearer.

text/0093-metric-instruments.md
jmacd: :+1:

jmacd:
After this discussion, I'm seriously considering that `Recorder` is a good name, better than either `Measure` or `Distribution`.

text/0093-metric-instruments.md
jmacd: I think `Timer` implies that the instrument _does the timing_. This is meant to convey it's useful for recording a timing. But as noted below, I'm starting to doubt the naming scheme. Could we call this `TimingRecorder`?

text/0093-metric-instruments.md
jmacd: Fixed, I hope.

jmacd:
I've responded to all the feedback and left two conversations open.

text/0093-metric-instruments.md
jmacd: This topic is addressed in #96.

text/0093-metric-instruments.md
jmacd: This topic is addressed in #96.

text/0093-metric-instruments.md
jmacd: This topic is addressed in #96.

jmacd: I've added a note that the names in this document are not finalized.  See #96.

text/0093-metric-instruments.md
jmacd: I've added a note that the names in this document are not finalized.  See #96.

text/0093-metric-instruments.md
jmacd: I've added a note that the names in this document are not finalized.  See #96.

text/0093-metric-instruments.md
```diff
@@ -0,0 +1,139 @@
...
+|               | DeltaObserver | Async         | Sum-only, Non-negative, Non-negative-rate |
+|               | CumulativeObserver | Async    | Sum-only, Precomputed-sum, Non-negative-rate |
+
+### Synchronous instruments
+
+Two new synchronous instruments are introduced in this proposal, bringing the total to four.
+
+1. **Counter** remains unchanged.  It uses Sum aggregation by default.
+2. **Distribution** is an an unrefined Measure instrument.  Distribution accepts positive or negative values and uses MinMaxSumCount aggregation by default.
+3. **UpDownCounter** is a Sum-only instrument with no other refinements.  It supports capturing positive and negative changes to a sum (deltas).  UpDownCounter uses Sum aggregation by default.
```
lzchen, line 87: How come `UpDownCounter` cannot be treated as a refinement to a `Counter`?

jmacd: Counter has non-negative, which implies Rate support.  We think that's the proper default behavior, so `UpDownCounter` is actually less refined than `Counter`.

lzchen: `Non-negative-rate` is the refinement that defines rate support correct? Can there be a refinement like `up-down-rate` to `Counter` that would essentially be what `UpDownCounter` is capturing? Unless if I'm missing something.

jmacd: For a delta-oriented sum-only instrument, `non-negative` implies `non-negative rate`. I think of the `UpDownCounter` as lacking this `non-negative` refinement, thus lacking the `non-negative-rate` refinement, leaving it with only the `sum-only` refinement. 

Maybe we should be talking about a "property" of the instrument, which is not so much a refinement but a choice for sum-only instruments. Either they have the "delta" property or the "cumulative" property. `UpDownCounter` and `Counter` both have the "delta" property, as does `DeltaObserver`.  I think this "delta" property implies the `up-down-rate` you're asking about, but I don't see it as a refinement so much as a lack of the `non-negative` refinement.

(I'm not sure I'm helping...)

jmacd: I added a paragraph and a column to the table, hoping to explain how measurements are categorized as Delta, Cumulative, or Instantaneous based on which refinements are present. https://github.com/open-telemetry/oteps/pull/93/files#diff-998af0c6471c96e97ea74443a922ed92R59

lzchen:
Thanks Josh for this. It greatly helped me clarify my understanding of the various use cases that metrics needs to cover. How would these translate into actual work that needs to be done? If there are no API changes being made, how can these refinement concepts be reflected through what the user does?

jmacd:
OTEP 88 didn't make API changes. This PR (#93) proposes new instruments be added, but I don't want to start on that until we get it merged and into the specification. 

I think beginning to prototype a Views API is a good plan. In order to get Cumulative export from Delta instruments and Delta export from Cumulative instruments, we'll need some new structure in the SDK. Since I haven't prototyped this myself, I haven't much guidance, but this is a project I'd like to be doing now.

text/0093-metric-instruments.md
```diff
@@ -0,0 +1,157 @@
...
+|               | CumulativeObserver | Async    | Sum-only, Precomputed-sum, Non-negative-rate | Cumulative |
+
+### Synchronous instruments
+
+Two new synchronous instruments are introduced in this proposal, bringing the total to four.
+
+1. **Counter** remains unchanged.  It uses Sum aggregation by default.
+2. **Distribution** is an an unrefined Measure instrument.  Distribution accepts positive or negative values and uses MinMaxSumCount aggregation by default.
+3. **UpDownCounter** is a Sum-only instrument with no other refinements.  It supports capturing positive and negative changes to a sum (deltas).  UpDownCounter uses Sum aggregation by default.
+4. **Timing** is a Non-negative instrument specialized for the native clock duration measured on the platform.  It ensures that duration values are always captured with correct units, that ensures exporters can convert duration measurements correctly.
```
MrAlias, line 106: I 100% think we should include this as an instrument, but having a unit based refinement wasn't specified at the start of the OTEP. Can we add something there about this?

text/0093-metric-instruments.md
```diff
@@ -0,0 +1,157 @@
...
+best addressed in a separate OTEP.  See [OTEP
+96](https://github.com/open-telemetry/oteps/pull/96) and consider the
+names used in this proposal to be provisional--OTEP 96 proposes a
+final naming scheme with greater consistency.
+
+## Explanation
+
+The four instrument refinements discussed in OTEP 88 are:
+
+* Sum-only: When computing only a sum is the instrument's primary purpose
```
MrAlias, line 24: For consistency, should this explicitly say that only a sum aggregation is to be used for instruments of this kind of refinement.

text/0093-metric-instruments.md
```diff
@@ -0,0 +1,157 @@
...
+instruments in terms of their refinements.  This proposal raised
+several open questions about naming and default aggregations that were
+best addressed in a separate OTEP.  See [OTEP
+96](https://github.com/open-telemetry/oteps/pull/96) and consider the
+names used in this proposal to be provisional--OTEP 96 proposes a
+final naming scheme with greater consistency.
+
+## Explanation
+
+The four instrument refinements discussed in OTEP 88 are:
```
MrAlias, line 22: It seems like this document has identified 2 different categories of refinements:

- Restrictions on the input values (non-negative, precomputed-sum, and non-negative-rate). This is a refinement that acts as a contract between the instrumenter and the exporter. The former agrees to only send a type of data the latter agrees to handle/interpret the data as such.
   I imagine the unit refinement (i.e. the `Timer` related restriction) to fit in here as well, right?
- Restrictions on meaningful aggregations (sum-only). This is a refinement limiting what processing make sense to be done on the data (though I imagine you could define a view that would go against this?).

Are there any other that people see?

Did we want to explicitly include these categories in this OTEP?

text/0093-metric-instruments.md
```diff
@@ -0,0 +1,157 @@
+# List of new metric instruments
+
+Formalize the new metric instruments proposed in [OTEP 88](https://github.com/open-telemetry/oteps/pull/88).
+
+## Motivation
+
+OTEP 88 introduced a framework for reasoning about new metric
+instruments with various refinements and ended with a [sample
+proposal](https://github.com/open-telemetry/oteps/pull/88#sample-proposal).
```
c24t, line 9:
```suggestion
instruments with various refinements and ended with a [sample
proposal](https://github.com/open-telemetry/oteps/blob/master/text/0088-metric-instrument-optional-refinements.md#sample-proposal).

```

jmacd:
This will be replaced by #98.

PR #96, opened 2020-04-10
Rationalize naming of metric instruments and their default aggregations

jmacd:
Based on recent feedback for https://github.com/open-telemetry/oteps/pull/93 and lingering questions from https://github.com/open-telemetry/oteps/pull/88, I've proposed a new naming scheme with more consistency. This also hopes to address an open question about aggregating last-values.

### DIFF
--- a/text/0096-metric-instrument-terminology.md
+++ b/text/0096-metric-instrument-terminology.md
@@ -53,15 +53,15 @@ The following table summarizes the four synchronous instruments and
 three asynchronous instruments that will be standardized as a result
 of this set of proposals.
 
-| Existing name | OTEP 93 name       | **Final name**         | Sync or Async | Default aggregation | Rate support |
-| ------------- | ------------------ | ---------------------- | ----------- | ---------- | ---- |
-| Counter       | Counter            | **Counter**            | Sync  | Sum | Yes | 
-|               | UpDownCounter      | **UpDownCounter**      | Sync  | Sum | Yes |
-| Measure       | Distribution       | **Recorder**           | Sync  | MinMaxSumCount | No |
-|               | Timing             | **TimingRecorder**     | Sync  | MinMaxSumCount  | No |
-| Observer      | LastValueObserver  | **GaugeObserver**      | Async | MinMaxSumCount | No |
-|               | DeltaObserver      | **DeltaObserver**      | Async | Sum | Yes |
-|               | CumulativeObserver | **CumulativeObserver** | Async | Sum | Yes |
+| Existing name | OTEP 93 name       | **Final name**         | Sync or Async | Function | Default aggregation | Rate support |
+| ------------- | ------------------ | ---------------------- | ----------- | ------------- | ---------- | ---- |
+| Counter       | Counter            | **Counter**            | Sync  | Add() | Sum | Yes | 
+|               | UpDownCounter      | **UpDownCounter**      | Sync  | Add() | Sum | Yes |
+| Measure       | Distribution       | **Recorder**           | Sync  | Record | MinMaxSumCount | No |
+|               | Timing             | **TimingRecorder**     | Sync  | Record | MinMaxSumCount  | No |
+| Observer      | LastValueObserver  | **GaugeObserver**      | Async | Observe | MinMaxSumCount | No |
+|               | DeltaObserver      | **DeltaObserver**      | Async | Observe | Sum | Yes |
+|               | CumulativeObserver | **CumulativeObserver** | Async | Observe | Sum | Yes |
 
 The argument for "Recorder" instead of "Distribution" is that we
 should prefer instrument descriptives associated with the action being
@@ -77,6 +77,35 @@ value ("reads a gauge").  A "Recorder" records an arbitrary value.  A
 This proposal consolidates OTEP 88 and OTEP 93 and proposes a consistent
 pattern for naming instruments.  It will be the source of truth when
 applying OTEP 88 and OTEP 93 to the OpenTelemetry metrics specification.
+
+### Function names
+
+The function names of the standard instruments are determined as
+follows.
+
+#### Counter and UpDownCounter
+
+Counter and UpDownCounter instruments use `Add()` as the function
+name, since they capture deltas to a Sum-only instrument.  We prefer
+`Add()` as opposed to `Count()`, since floating point numbers are
+supported, to avoid some an association with "Countable" numbers, a
+mathemtical term associated with natural numbers.
+
+#### Recorder and TimingRecorder
+
+Recorder and TimingRecorder use `Record()` as the function name, as in
+the existing specification for Measure instruments.  This conveys the
+fact that these are not a sum, and that individual events are of
+importance.
+
+#### Asynchronous instruments
+
+Asynchronous instruments use `Observe()` as the function name.  This
+signifies that the instrument passively captures a measurement, is not
+an active participant, as implied by `Record()`.  _Observation_ also
+conveys the last-value relationship specified for asynchronous
+instruments.  The observer can only observe one value at a time, where
+the last-observed value wins.
 
 ### Default aggregations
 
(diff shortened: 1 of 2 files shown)


text/0096-metric-instrument-terminology.md
```diff
@@ -0,0 +1,153 @@
...
+three asynchronous instruments that will be standardized as a result
+of this set of proposals.
+
+| Existing name | OTEP 93 name       | **Final name**         | Sync or Async | Function | Default aggregation | Rate support |
+| ------------- | ------------------ | ---------------------- | ----------- | ------------- | ---------- | ---- |
+| Counter       | Counter            | **Counter**            | Sync  | Add() | Sum | Yes | 
+|               | UpDownCounter      | **UpDownCounter**      | Sync  | Add() | Sum | Yes |
+| Measure       | Distribution       | **Recorder**           | Sync  | Record() | MinMaxSumCount | No |
+|               | Timing             | **TimingRecorder**     | Sync  | Record() | MinMaxSumCount  | No |
+| Observer      | LastValueObserver  | **GaugeObserver**      | Async | Observe() | MinMaxSumCount | No |
```
bogdandrutu, line 62: I really want this to be `Sum` because of memory-usage metric. Look at the Go example for process metrics https://github.com/open-telemetry/opentelemetry-specification/issues/549#issue-595864966 I don't want to see "process/sys_heap" as a minmaxsumcount by default.

Another argument is default cost (money), if I am using a system like Stackdriver exporting "minmaxsumcount" will cost 4 times more than just sum. So I think it is important that default aggregations are less "expensive".

PS: I know I work for a company that makes money if the user sends more data :)

text/0096-metric-instrument-terminology.md
```diff
@@ -0,0 +1,153 @@
...
+The following table summarizes the four synchronous instruments and
+three asynchronous instruments that will be standardized as a result
+of this set of proposals.
+
+| Existing name | OTEP 93 name       | **Final name**         | Sync or Async | Function | Default aggregation | Rate support |
+| ------------- | ------------------ | ---------------------- | ----------- | ------------- | ---------- | ---- |
+| Counter       | Counter            | **Counter**            | Sync  | Add() | Sum | Yes | 
+|               | UpDownCounter      | **UpDownCounter**      | Sync  | Add() | Sum | Yes |
+| Measure       | Distribution       | **Recorder**           | Sync  | Record() | MinMaxSumCount | No |
+|               | Timing             | **TimingRecorder**     | Sync  | Record() | MinMaxSumCount  | No |
```
bogdandrutu, line 61: Does "timing" allow negative values?  What is the difference between Recorder and Timing. I feel that for all the other 6 types we do have them because they express different types of data or time correlation but this is just a helper for "Recorder" that has a specific "unit". I would suggest we focus on the main types first then we add extensions.

text/0096-metric-instrument-terminology.md
```diff
@@ -0,0 +1,153 @@
...
+| Observer      | LastValueObserver  | **GaugeObserver**      | Async | Observe() | MinMaxSumCount | No |
+|               | DeltaObserver      | **DeltaObserver**      | Async | Observe() | Sum | Yes |
+|               | CumulativeObserver | **CumulativeObserver** | Async | Observe() | Sum | Yes |
+
+The argument for "Recorder" instead of "Distribution" is that we
+should prefer instrument descriptives associated with the action being
+performed ("occupationally connected"), not the value being computed,
+as the latter is dependent on SDK configuration.  A "Recorder" records
+a value that is part of a distribution.  A "Counter" counts a value
```
bogdandrutu, line 70: Have we considered other names for the "Recorder", I do understand the argument against "Distribution" but I do not really like "Recorder". "Recorder" is a very generic name and even in this sentence you had to refer to what types of values you record "A "Recorder" records a value that is part of a distribution.". I am worried that "Recorder" will supersede all the others in terms of usage. I would like to hear more opinions/suggestions.

text/0096-metric-instrument-terminology.md
```diff
@@ -0,0 +1,153 @@
...
+## Explanation
+
+The following table summarizes the four synchronous instruments and
+three asynchronous instruments that will be standardized as a result
+of this set of proposals.
+
+| Existing name | OTEP 93 name       | **Final name**         | Sync or Async | Function | Default aggregation | Rate support |
+| ------------- | ------------------ | ---------------------- | ----------- | ------------- | ---------- | ---- |
+| Counter       | Counter            | **Counter**            | Sync  | Add() | Sum | Yes | 
+|               | UpDownCounter      | **UpDownCounter**      | Sync  | Add() | Sum | Yes |
```
bogdandrutu, line 59: Rate for this?

jmacd: Fixed.

text/0096-metric-instrument-terminology.md
jmacd: No, Timing wouldn't allow negatives (as stated in 0093). It is just a helper, but it helps with a mistake that's extremely easy to make. We don't want Java processes reporting milliseconds while Go processes report nanoseconds. Without a conventional instrument, that's going to happen.

text/0096-metric-instrument-terminology.md
jmacd: Firstly, I think minmaxsumcount is perfectly sensible for a heap. If you aggregate this across a cluster, you'd like to know what the min/max values are.  If you don't aggregate the count as well, the metric system will requires domain-level knowledge to construct an average using some other knowledge (i.e.,, that there's a corresponding count available, implicitly).

Moreover, I've documented (here) how there's an obvious optimization that results in sending just one value for the min/max/sum/count when there is just one value, which reduces the data-size problem for the case you care about.  Each machine will report one min/max/sum/count as a single value in the ordinary case, but when we aggregate those values, we'll get the proper min/max/sum and count out--no domain-knowledge required.

Lastly, the metric instrument you're describing is exactly the hypothetical `UpDownCumulativeObserver` detailed in OTEP 88.  If you want a Sum-only aggregation, then you should use a Sum-only instrument. That instrument does exactly what you want, should we standardize it?

text/0096-metric-instrument-terminology.md
jmacd: Any time there is a set of values, you have a distribution, so I considered "that is part of a distribution" to be almost unnecessary, except there is emphasis being placed on it producing a distribution.  I would expect Counter is the most common instrument, after that I would expect Recorder to be second-most-common.

MrAlias: Given the `Recorder` is a foundational instrument I think this generic nature is ideal. It is the base from which refined instruments are created. I think defaulting to a `Recorder` is what we want to happen if you are not sure a refinement matches your intended need.

text/0096-metric-instrument-terminology.md
```diff
@@ -0,0 +1,153 @@
...
+|               | DeltaObserver      | **DeltaObserver**      | Async | Observe() | Sum | Delta | Yes |
+|               | CumulativeObserver | **CumulativeObserver** | Async | Observe() | Sum | Cumulative | Yes |
+
+The argument for "Recorder" instead of "Distribution" is that we
+should prefer instrument descriptives associated with the action being
+performed ("occupationally connected"), not the value being computed,
+as the latter is dependent on SDK configuration.  A "Recorder" records
+a value that is part of a distribution.  A "Counter" counts a value
+that is part of a sum.  An "GaugeObserver" observes an instantaneous
+value ("reads a gauge").  A "Recorder" records an arbitrary value.  A
```
MrAlias, line 72: It is confusing that "Recorder" is included twice in this list. Once saying it records a distribution and the other saying it records an arbitrary value.  I think the latter is more accurate. If a view of a `Recorder` were to 'aggregate' with an array (i.e. no aggregation) saying a "Recorder" records a value that is part of a distribution, while not incorrect, isn't precise.

text/0096-metric-instrument-terminology.md
```diff
@@ -0,0 +1,153 @@
...
+it is permitted to observe just one measurement per interval?  This
+says that when observed values are aggregated they should be treated
+like a distribution--we are intersted in more than a sum, by
+definition.  If observing only a sum, the DeltaObserver or
+CumulativeObserver should be used instead.
+
+Clearly, when Count equals 1, the Min, Max, and Sum are equal to the
+value.  Exporters may be able take advantage of this fact when
+exporting data from these instruments.  In particular, since it is
+known that asynchronous instruments produce only one valiue per
```
MrAlias, line 142:
```suggestion
known that asynchronous instruments produce only one value per
```

jmacd:
This will be replaced by #98.

PR #98, opened 2020-04-18
Explain the Metric API instruments

jmacd:
This document replaces OTEPs https://github.com/open-telemetry/oteps/pull/93 and https://github.com/open-telemetry/oteps/pull/96.

The number of instruments is expanded to 6.

Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/467
Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/465
Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/462


text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,97 @@
...
+
+Measurements can be described in terms of their relationship with time.
+
+Delta measurements are those that measure a change to a sum.  Delta instruments are usually selected because the program does not need to compute the sum and is able to measure the change.  In these cases, it would require extra state for the user to report cumulative values.
+
+Cumulative measurements are those that report the current value of a sum.  Cumulative instruments are usually selected because the program is able to measure the sum.  In these cases, it would require extra state for the user to report delta values.
+
+Delta and Cumulative instruments are referred to, collectively, as Additive instruments.
+
+Instantaneous measurements are those that report a non-additive measurement, one where it is not natural to compute a sum.  Instantaneous instruments are usually chosen to when the distribution of values is of interest, not only the sum.
```
MrAlias, line 50:
```suggestion
Instantaneous measurements are those that report a non-additive measurement, one where it is not natural to compute a sum.  Instantaneous instruments are usually chosen when the distribution of values is of interest, not only the sum.
```

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Function names
+
+Synchronous delta instruments support an `Add()` function, signifying that they add to a sum and do not report a total count.
+
+Synchronous instantaneous instruments support a `Record` function, signifying that they capture individual events, not only a sum.
```
MrAlias, line 56: For consistency might want to include parenthesis

```suggestion
Synchronous instantaneous instruments support a `Record()` function, signifying that they capture individual events, not only a sum.
```

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,196 @@
...
+
+| Existing name | **Standard name** | Instrument kind | Function name | Default aggregation | Measurement kind | Kind of data | Rate support (Monotonic) | Notes |
+| ------------- | ----------------------- | ----- | --------- | -------------- | ------------- | --- | ------------------------------------ | --- |
+| Counter       | **Counter**             | Sync  | Add()     | Sum            | Delta         | Additive | Yes | Per-request, part of a monotonic sum |
+|               | **UpDownCounter**       | Sync  | Add()     | Sum            | Delta         | Additive | No  | Per-request, part of a non-monotonic sum |
+| Measure       | **ValueRecorder**       | Sync  | Record()  | MinMaxSumCount | Instantaneous | Event    | No  | Per-request, any non-additive measurement |
+| Observer      | **DeltaObserver**       | Async | Observe() | Sum            | Delta         | Additive | Yes | Per-interval, part of a monotonic sum |
+|               | **UpDownDeltaObserver** | Async | Observe() | Sum            | Delta         | Additive | No  | Per-interval, part of a non-monotonic sum |
```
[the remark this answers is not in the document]
jmacd: Maybe. I was searching for good examples of these instruments, but couldn't find any. I wanted to find a well-known system-level API for measurement that gives deltas, not cumulative values.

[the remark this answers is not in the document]
bogdandrutu: I am happy to remove them for the moment if that is your suggestion :). We just need to tell user that we expect cumulative data for the other 2. I feel that somehow having these 2 deltas make user aware of the other, but this is a small benefit. So if you want to remove them +1, I also don't know of any system, and I suggested them mostly for completeness which is the wrong reason :)

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Detail
+
+Here we discuss the eight proposed instruments individually and mention other names considered for each.
+
+### Counter
+
+`Counter` is the most common synchronous instrument.  This instrument supports an `Add(delta)` function for reporting a sum, and is restricted to non-negative deltas.  The default aggregation is `Sum`, as for any additive instrument, which are those instruments with Delta or Cumulative measurement kind.
+
+Example uses for `Counter`:
```
bogdandrutu, line 106: One of the classic example is the number of requests started/completed.

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,196 @@
...
+
+These example instruments would be useful for monitoring the rate of any of these quantities.  In these situations, it is usually more convenient to report a change of the associated sums, as the change happens, as opposed to maintaining and reporting the sum.
+
+Other names considered: `Adder`.
+
+### UpDownCounter
+
+`UpDownCounter` is similar to `Counter` except that `Add(delta)` supports negative deltas.  This makes `UpDownCounter` not useful for computing a rate aggregation.  It aggregates a `Sum`, only the sum is non-monotonic.  It is generally useful for counting changes in an amount of resources used, or any quantity that rises and falls, in a request context.
+
+Example uses for `UpDownCounter`:
```
bogdandrutu, line 120: Classic example number of active requests

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,196 @@
...
+
+Example uses for `SumObserver`.
+- capture process user/system CPU seconds
+- capture the number of cache misses
+
+### UpDownSumObserver
+
+...
+
+Example uses for `SumObserver`.
```
bogdandrutu, line 176: s/`SumObserver`/`UpDownSumObserver`

text/0098-metric-instruments-explained.md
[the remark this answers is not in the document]
jkwatson: Yes, if we can't think of real examples, we should definitely remove them until we can (or maybe we never can, bonus!).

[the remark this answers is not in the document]
jmacd: I've removed the asynchronous delta instruments.

I have to say I do not prefer `SumUpDownObserver` over `UpDownSumObserver`. It's open for debate, but I'm resolving this.

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,207 @@
...
+
+A `SumObserver` is a good choice in situations where a measurement is expensive to compute, such that it would be wasteful to compute on every request.  For example, a system call is needed to capture process CPU usage, therefore it should be done periodically, not on each request.  A `SumObserver` is also a good choice in situations where it would be impractical or wasteful to instrument individual deltas that comprise a sum.  For example, even though the number of cache misses is a sum of individual cache-miss events, it would be too expensive to synchronously capture each event using a `Counter`.
+
+Other names considered: `CumulativeObserver`.
+
+### UpDownSumObserver
+
+`UpDownSumObserver` is the asynchronous instrument corresponding to `UpDownCounter`, used to capture a non-monotonic count.  "Sum" appears in the name to remind users that it is a cumulative instrument.  Use a `UpDownSumObserver` to capture any value that starts at zero and rises or falls throughout the process lifetime.
+
+Example uses for `SumObserver`.
```
cijothomas, line 161:
```suggestion
Example uses for `UpDownSumObserver`.
```

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,207 @@
...
+
+The same considerations mentioned for choosing `SumObserver` over the synchronous `Counter` apply for choosing `UpDownSumObserver` over the synchronous `UpDownCounter`.  If a measurement is expensive to compute, or if the corresponding delta events happen so frequently that it would be impractical to instrument them, use a `UpDownSumObserver`.
+
+Other names considered: `UpDownCumulativeObserver`.
+
+### ValueObserver
+
+`ValueObserver` is the asynchronous instrument corresponding to `ValueRecorder`, used to capture non-additive measurements that are expensive to compute and/or are not request-oriented.  
+
+Example uses for `SumObserver`:
```
cijothomas, line 175:
```suggestion
Example uses for `ValueObserver`:
```

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,207 @@
...
+
+The `UpDown-` forms of additive instrument are not suitable for aggregating rates because the up- and down-changes in state may cancel each other. 
+
+Non-additive instruments can be used to derive a sum, meaning rate aggregation is possible when the values are non-negative. There is not a standard non-additive instrument with a non-negative refinement in the standard.
+
+### Defalt Aggregations
+
+Additive instruments use `Sum` aggregation by default, since by definition they are used when only the sum is of interest.
+
+Instantaneous instruments use `MinMaxSumCount` aggregation by default, which is an inexpensive way to summarize a distribution of values.
```
cijothomas, line 86: Was [this](https://github.com/open-telemetry/opentelemetry-specification/issues/462) spec issue about deciding whether MinMaxSumCount or just SumCount aggregation is required for measure instruments? If yes, what section of this otep #98 help resolve that?

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Rate support
+
+Rate aggregation is supported for Counter and SumObserver instruments in the default implementation.
+
+The `UpDown-` forms of additive instrument are not suitable for aggregating rates because the up- and down-changes in state may cancel each other. 
+
+Non-additive instruments can be used to derive a sum, meaning rate aggregation is possible when the values are non-negative. There is not a standard non-additive instrument with a non-negative refinement in the standard.
+
+### Defalt Aggregations
```
cijothomas, line 82:
```suggestion
### Default Aggregations
```

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,207 @@
...
+2. There is inconsistency in the hypothetical naming scheme for instruments presented in OTEP 88.  Note that "Counter" and "Observer" end in "-er", a noun suffix used in the sense of "[person occupationally connected with](https://www.merriam-webster.com/dictionary/-er)", while the term "Measure" does not fit this pattern.  This proposal proposes to replace the abstract term "Measure" by "Recorder", since the associated function name (verb) is specified as `Record()`.
+
+This proposal also repeats the current specification--and the justification--for the default aggregation of each standard instrument.
+
+## Explanation
+
+The following table summarizes the final proposed standard instruments resulting from this set of proposals.  The columns are described in more detail below.
+
+| Existing name | **Standard name** | Instrument kind | Function name | Default aggregation | Measurement kind | Rate support (Monotonic) | Notes |
```
cijothomas, line 22: Instead of the column header `Measurement kind`, would using "Temporal Quality" make better sense, as the latter is explained later.
Also https://github.com/open-telemetry/opentelemetry-proto/pull/140/files calls this `Temporality`

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,207 @@
...
+| ------------- | ----------------------- | ----- | --------- | -------------- | ------------- | --- | ------------------------------------ |
+| Counter       | **Counter**             | Sync  | Add()     | Sum            | Delta         | Yes | Per-request, part of a monotonic sum |
+|               | **UpDownCounter**       | Sync  | Add()     | Sum            | Delta         | No  | Per-request, part of a non-monotonic sum |
+| Measure       | **ValueRecorder**       | Sync  | Record()  | MinMaxSumCount | Instantaneous | No  | Per-request, any non-additive measurement |
```
bogdandrutu, line 26: I think this is Delta, because you want the min and max from in the interval (last export, now]. Also sum an more importantly count (which is a Counter) needs time association.

text/0098-metric-instruments-explained.md
jmacd: Done.

text/0098-metric-instruments-explained.md
jmacd: Done.

text/0098-metric-instruments-explained.md
jmacd: It was, but it wasn't clear. Here I've added a specific answer: https://github.com/open-telemetry/oteps/pull/98/commits/a3af52ba0f1dc4a2ff19f7629710d00291570edd

text/0098-metric-instruments-explained.md
jmacd: (I'm composing an answer to this. I still believe this is a Instantaneous case, and will work to justify that reasoning.)

bogdandrutu: Not convinced because you need to inform when you did a reset of the values. Would you expect to reset min/max/sum/count every export cycle?

jmacd: Yes, I would expect these values to "reset" every cycle, as they are instantaneous. I've written a more thorough explanation of what it means to aggregate `ValueObserver` results across both time and space dimensions.

Specifically, when there is no aggregation being applied, each ValueObserver outputs one value which is equal to the min, the max, and the sum, with a count of 1. 

See how you like this explanation: https://github.com/open-telemetry/oteps/pull/98/commits/299c442a9658a23333032d37cbd562ffb709019f

bogdandrutu: Let's make sure we talk about the same thing. I am referring to **ValueRecorder** not ** ValueObserver**. Are we talking about the same thing?

bogdandrutu: Also on the `ValueObserver` I do agree instantaneous is better.

jmacd: Oh, sorry. I misread your comment. We were talking about different instruments.

OK, so as for `ValueRecorder`, I see what you mean. I still think of the **values** as instantaneous measurements. I see how after aggregating measurements over an interval, the result of aggregation may take on a delta temporal quality, but I don't view that as automatic.

May we just say that the whenever aggregating results, an implementation must keep track of the time interval covered by the aggregation. After aggregating `ValueRecorder` measurements for several intervals, we'll have a distribution of instantaneous values covering a range of time, from which we can draw averages.

bogdandrutu: I do agree that individual values are "instantaneous measurements" but the aggregated result is a "Delta" based on my understanding. This was one of the point from @jkwatson raised, that he needs the aggregation to reset every export (delta), and other backends prefer (cumulative). So we need to know in the exporter if this is delta or cumulative so I can do the right thing and build what I need for that exporter.

bogdandrutu: Also in my mind every "measurement" that we record using sync instruments is "instantaneous" but the "Temporal" property apply to the aggregation (a.k.a the aggregated values were recorded in this interval).

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,240 @@
...
+
+Aggregating `ValueObserver` measurements across the time dimension for a given instrument and label set yields a set of measurements that were taken across a span of time, but this does not automatically lead us to consider them delta measurements.  If we aggregate 10 consecutive collection intervals for a given label set, what we have is distribution of instantaneous measurements with Count equal to 10, with the Min, Max and Sum serving to convey the average value and the range of values present in the distribution.  The result is a time-averaged distribution of instantaneous measurements.
+
+Whether aggregating across time or space, it has been argued, the result of a `ValueObserver` instrument is has the Instantaneous temporal quality.
```
MikeGoldsmith, line 207:
```suggestion
Whether aggregating across time or space, it has been argued, the result of a `ValueObserver` instrument has the Instantaneous temporal quality.
```

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,240 @@
...
+4. Aggregate the timestamped measurements from step 3.
+
+Steps 2 and 3 ensure that measurements taken less frequently have equal representation in the output, by virtue of computing the spatial aggregation first.  If we were to compute the temporal aggregation first, then aggreagate across spatial dimensions, then instruments collected at a higher frequency will contribute correspondingly more points to the aggregation.  Thus, we must aggregate across `ValueObserver` instruments across spatial dimensions before averaging across time.
+
+## Open Questions
+
+### Timing instrument
+
+One potentially important special-purpose instrument, found in some metrics APIs, is a dedicated instrument for reporting timings.  The rationale is that when reporting timings, getting the units right is important and often not easy.  Many programming languages use a different type to represent time or a difference between times.  To correctly report a timing distribution in OpenTelemetry requires using a `ValueRecorder` but also configuring it for the units output by the clock that was used.
```
MikeGoldsmith, line 224:
```suggestion
One potentially important special-purpose instrument, found in some metrics APIs, is a dedicated instrument for reporting timings.  The rationale is that when reporting timings, getting the units right is important and often not easy.  Many programming languages use a different type to represent time or a difference between times.  To correctly report a timing distribution OpenTelemetry requires using a `ValueRecorder` but also configuring it for the units output by the clock that was used.
```

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,240 @@
...
+
+In the past, a proposal to create a dedicated `TimingValueRecorder` instrument was rejected.  This instrument would be identical to a `ValueRecorder`, but its `Record()` method would be specialized for the correct type used to represent a duration, so that the units could be set correctly and automatically.  A related pattern is a `Timer` or `StopWatch` instrument, one responsible for both measuring and capturing a timing.  
+
+Should types such as these be added as helpers?  For example, should `TimingValueRecorder` be a real instrument, or should it be a helper that wraps around a `ValueRecorder`?  There is a concern that making `TimingValueRecorder` into a helper makes it less visible, less standard, and that not having it at all will encourage instrumentation mistakes.
+
+This may be revisited in the future.
+
+### Synchronous cumulative and asynchronous delta helpers
+
+A cumulative measurement can be converted into delta measurement by remember the last-reported value.  A helper instrument could offer to emulate synchronous cumulative measurements by remembering the last-reported value and reporting deltas synchronously.
```
MikeGoldsmith, line 234:
```suggestion
A cumulative measurement can be converted into delta measurement by remembering the last-reported value.  A helper instrument could offer to emulate synchronous cumulative measurements by remembering the last-reported value and reporting deltas synchronously.
```

text/0098-metric-instruments-explained.md
jmacd: I've added to try to clarify my understanding here: https://github.com/open-telemetry/oteps/pull/98/commits/d2a90fe32df36d5d90112580f0799e634a3b9588

jmacd:
Thanks @MikeGoldsmith @codeboten. I found other `.md` files in OTel that use a syntax to avoid repeating the same link a million times, and used that for otep-88 here.

text/0098-metric-instruments-explained.md
jmacd: I believe the thing you and @jkwatson are referring to will be a topic for the SDK specification.

bogdandrutu: I don't think so, I would like to understand better then what the "temporal" property means in the proto. May be different concepts.

jkwatson: Yes, I think it's important to distinguish between what kind of values the instruments *accept* vs. how they will be aggregated, and eventually exported. These are 3 different things.

bogdandrutu: The question is: Is this "Delta" refer to the same "Delta" in the proto?

jkwatson: My reading is 'no'. The "delta" here is a characteristic of the data being sent to the instrument. How it gets aggregated, and ends up the proto is a different thing.

jmacd: I just re-read https://github.com/open-telemetry/opentelemetry-proto/pull/140 and I don't see any disagreement here. The examples for Delta and Cumulative in the proto appear to agree with the usage here, in that it talks about the values being passed to the instrument.

I think I see where the confusion lies, but I don't think we want to change anything here. There are two ways in the proto that we can talk about something being cumulative. One: to say that a metric descriptor has the cumulative temporal quality, says something about the values it took in. Two: when we export a series of aggregations for a metric, continuing to use the same start time throughout the series. 

The first usage of cumulative refers to the kind of instrument: SumObserver and UpDownSumObserver are cumulative in this sense.

The second usage of cumulative is not the same: it refers to exported aggregations. Any exporter can be generate cumulative data for any kind of instrument, if it wants to (and has the memory to do so).

Example: I have a `ValueRecorder` and I'm measuring something non-additive (e.g., pedal position); these measurements are instantaneous.  A delta-oriented exporter for this instrument will output the distribution of instantaneous values for one interval (e.g., MMSC, DDSketch, T-digest, ...). A cumulative-oriented exporter for this instrument will output the distribution of instantaneous values since the beginning of time--still having the instantaneous temporal quality.

This will be taken up in the SDK specification: if an exporter wishes to convert a cumulative instrument to a cumulative aggregation, or a delta instrument to a delta aggregation, these are easy. To convert delta-to-cumulative or cumulative-to-delta requires more memory, and we can provide these configurations.

jkwatson: Agreed. I was just saying that the value in the column here is not the value that will end up in the proto...that's something that depends on aggregation, view, or something else in the SDK.

The enumeration of possible temporal qualities may indeed end up the same, though. Just the mapping is not set by this OTEP, or by the API in general.

bogdandrutu: Then this is confusing, because even the column name suggest that they are the same.

jmacd: These are the same concepts, though, used in separate places to describe inputs and outputs. It's OK with me, but if you'd like to suggest a way out (either here or in the proto), please do.

bogdandrutu: Just to make sure I understand the output of the ValueRecorder if we would make a proto every export cycle will be (MinMaxSumCount, Delta), is that a correct understanding? And for a SumObserver will be (Int64 (assuming int), Cumulative)? What will be the output for ValueObserver?

If that is the case I would suggest to make sure we document that the column in this table refers to an interpretation of the input data and not the export data.

bogdandrutu: @jmacd my understanding from the open-telemetry/opentelemetry-proto#140 is that it refers to the aggregated data, the kind of instrument is no longer important at that point, maybe this is a source of misunderstanding for me.

MrAlias: Could we rename the `Temporal quality` column to `Input Temporal quality`, or maybe `Argument Temporal quality`?

I was initially confused into thinking this `Temporal quality` was describing the output quality, not the input data.

Is the `Rate Support (Monotonic)` column also describing input? I'm guessing not (it describes output), but want to check.

MrAlias: Sorry, don't mean to add more work here. We can just make this clear in the eventual spec PR. I think it's important we all are on the same page conceptually here, but maybe the language can be a bit loose.

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,244 @@
+# Explain the metric instruments
+
+Propose and explain final names for the standard metric instruments theorized in [OTEP 88][otep-88] and address related confusion.
+
+[otep-88]: https://github.com/open-telemetry/oteps/blob/master/text/0088-metric-instrument-optional-refinements.md
```
bogdandrutu, line 5: nit: I usually put this at the end of the file so I can easily find all defined links.

text/0098-metric-instruments-explained.md
cijothomas: I think its important to call out that the Temporal Quality in this table is NOT same as the one in Proto. Or use a different name if possible to avoid confusion.

I was trying to model this in my mind with this example below:

I am using ValueRecorder instrument to capture ServerLatency.

The following recordings occurred between times T1 and T2
valueRecorderInstrument.Record(100ms)
valueRecorderInstrument.Record(200ms)
valueRecorderInstrument.Record(50ms)
valueRecorderInstrument.Record(100ms)
valueRecorderInstrument.Record(5ms)

and following occurs between T2 and T3
valueRecorderInstrument.Record(1ms)
valueRecorderInstrument.Record(2000ms)
valueRecorderInstrument.Record(50ms)

The above calls are providing inputs, and the values are "instantaneous" inputs. This matches the temporal quality "Instantaneous" specified for ValueRecorder in the table above.

Now what I export would be one of the below (depending on my SDK/Batcher/Exporter)
{
 [Min 5 Max 200 Sum 500 Count 5, T1 to T2, TemporalQuality: Delta]
 [Min 1 Max 2000 Sum 2051 Count 3, T2 to T3, TemporalQuality: Delta]
}
or
{
[Min 1 Max 2000 Sum 2551 Count 8, T1 to T3, TemporalQuality: Delta]
}
or
{
 [Min 1 Max 2000 Sum 2551 Count 8, T1 onwards, TemporalQuality: Cumulative]
}

The TemporalQuality in the proto and instrument are not same. This is confusing for me. Would love to see this clarified!

jmacd: I don't see why we would use the cumulative vs. delta distinction in the way you're describing, and I also don't read this into the current protocol. I see the usage of these two terms as the same, they're adjectives, and they apply to inputs or to outputs of the metrics processing pipeline.

I would pass through the Instantaneous/Delta/Cumulative temporal quality of the original instrument in each case. Using Prometheus and Statsd as archetypes for the two kinds of exporter:

For Prometheus at T2:
{
[Min: 5, Max: 200, Sum: 455, Count: 5, TimeRange: T1 to T2, TemporalQuality: Instantaneous]
}

For Prometheus at T3:
{
[Min: 1, Max: 2000, Sum: 2506, Count: 8, TimeRange: T1 to T3, TemporalQuality: Instantaneous]
}

For Statsd at T2:
{
[Min: 5, Max: 200, Sum: 455, Count: 5, TimeRange: T1 to T2, TemporalQuality: Instantaneous]
}

For Statsd at T3:
{
[Min: 1, Max: 2000, Sum: 2051, Count: 3, TimeRange: T2 to T3, TemporalQuality: Instantaneous]
}

jmacd: Just for emphasis, I'm using a repeating start time to indicate that the aggregate result spans multiple collection periods. The temporal quality of export should match the instrument.

jmacd:
There is an outstanding discussion here, https://github.com/open-telemetry/oteps/pull/98#discussion_r417023428, but it's not about the matter at hand. If we agree about the names of the instruments and their default aggregation, then I think this PR is ready to merge. 

The debate above is about how we document the protocol and the internals of the SDK and export pipeline. I do not think the outcome of this debate will have an impact on the specification we write for the API. In any case we shouldn't hold this up over a decision about use of "Temporal quality" in the API documentation.

text/0098-metric-instruments-explained.md
cijothomas: > For Prometheus at T3:
> {
> [Min: 1, Max: 2000, Sum: 2506, Count: 8, TimeRange: T1 to T3, TemporalQuality: Instantaneous]
> }

Per my understanding, the above should be Cumulative. Proto file says for Instantaneous, starttime is not expected to be present (https://github.com/open-telemetry/opentelemetry-proto/blob/master/opentelemetry/proto/metrics/v1/metrics.proto#L165)

> For Statsd at T3:
> {
> [Min: 1, Max: 2000, Sum: 2051, Count: 3, TimeRange: T2 to T3, TemporalQuality: Instantaneous]
> }

Per my understanding, the above should be Delta. (https://github.com/open-telemetry/opentelemetry-proto/blob/master/opentelemetry/proto/metrics/v1/metrics.proto#L191)

I fully understand this is more for a SDK/Exporter topic, not API. I am happy with your stated purpose of this Otep - define standard instruments and their default aggregation.

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Sync vs Async instruments
+
+Synchronous instruments are called in a request context, meaning they potentially have an associated tracing context and distributed correlation values.  Multiple metric events may occur for a synchronous instrument within a given collection interval.
+
+Asynchronous instruments are reported by a callback, once per collection interval, and lack request context.  They are permitted to report only one value per distinct label set per period.  If the application observes multiple values in a single callback, for one collection interval, the last value "wins".
+
+### Temporal quality
+
+Measurements can be described in terms of their relationship with time.
```
tigrannajaryan, line 58: I found comments with examples that explain temporal qualities in https://github.com/open-telemetry/opentelemetry-proto/blob/28e2774262a651c4c2e701c3a05820c3e43ead71/opentelemetry/proto/metrics/v1/metrics.proto#L156 very helpful. It may be useful to have similar examples in this OTEP.

text/0098-metric-instruments-explained.md
```diff
@@ -0,0 +1,245 @@
...
+  - If synchronous and not monotonic, use `UpDownCounter` with arbitrary values
+  - If asynchronous and a cumulative, monotonic sum is measured, use `SumObserver`
+  - If asynchronous and a cumulative, arbitrary sum is measured, use `UpDownSumObserver`
+- If the measurements are non-additive or additive with an interest in the distribution, use an instantaneous instrument:
+  - If synchronous, use `ValueRecorder` to record a value that is part of a distribution
+  - If asynchronous use `ValueObserver` to record a single measurement nearing the end of a collection interval.
+
+### Sync vs Async instruments
+
+Synchronous instruments are called in a request context, meaning they potentially have an associated tracing context and distributed correlation values.  Multiple metric events may occur for a synchronous instrument within a given collection interval.
```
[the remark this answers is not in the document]
jmacd: Yes, you've got it. I added "Note that synchronous instruments may be called outside of a request context, such as for background computation.  In these scenarios, we may simply consider the Context to be empty."

text/0098-metric-instruments-explained.md
jmacd: We've had a lot of debate about this topic in the last few days. Surprisingly much of it, and the leading conclusion is that we should avoid these terms in the specification. Without rewriting this proposal, I've added to clarify that these terms will not be used going forward:

> Note: although this term logically applies and is used throughout this OTEP, discussion in the Metrics SIG meeting (4/30/2020) leads us to exclude this term from use in documenting the Metric API.  The explanation of terms here is consistent with the [terminology used in the protocol], but we will prefer to use these adjectives to describe properties of an aggregation, not properties of an instrument (despite this document continuing to use the terms freely).  In the API specification, this distinction will be described using "additive synchronous" in contrast with "additive asynchronous".

I have also written in the forthcoming API spec for version 0.4 this on the same topic:

> The additive instruments (`Counter`, `UpDownCounter`, `SumObserver`,
`UpDownSumObserver`) use Sum aggregation, by default.  Details about
computing a Sum aggregation vary, but from the user's perspective this
means they will be able to monitor the sum of values entered.  The
distinction between synchronous and asynchronous instruments is
crucial to specifying how exporters work, a topic that is covered in
the [SDK specification](TODO).

bogdandrutu:
@open-telemetry/specs-approvers this OTEP got very popular, can we review this to make progress?

text/0098-metric-instruments-explained.md
jmacd: I filed this issue: https://github.com/open-telemetry/opentelemetry-proto/issues/147

jmacd:
@open-telemetry/specs-approvers please approve this. We have reached agreement by all, and the only outstanding discussion point belongs to the SDK: https://github.com/open-telemetry/opentelemetry-proto/issues/147

bogdandrutu:
Please fix markdown lint errors :)

jmacd:
@bogdandrutu Done. Needs one more approver 😆

jmacd:
@bogdandrutu Let's merge this.

bogdandrutu:
Yes Sir :)

PR #99, opened 2020-04-23
OTLP/HTTP: HTTP Transport Extension for OTLP

tigrannajaryan:
OTLP can be currently communicated only via one transport: gRPC. While using
gRPC has certain benefits there are also drawbacks:

- Some users have infrastructure limitations that make gRPC-based protocol
  usage impossible. For example AWS ALB does not support gRPC connections.

- gRPC is a relatively big dependency, which some clients are not willing to
  take. Plain HTTP is a smaller dependency and is built in the standard
  libraries of many programming languages.

This proposal keeps the existing specification of OTLP over gRPC transport
(OTLP/gRPC for short) and defines an additional way to use OTLP protocol over
HTTP transport (OTLP/HTTP for short). OTLP/HTTP uses the same ProtoBuf payload
that is used by OTLP/gRPC and defines how this payload is communicated over HTTP
transport.

### DIFF
--- a/text/0000-otlp-http.md
+++ b/text/0000-otlp-http.md
@@ -1,0 +1,179 @@
+# OTLP/HTTP: HTTP Transport Extension for OTLP
+
+_Author: Tigran Najaryan, Splunk._
+
+This is a proposal to add HTTP Transport extension for
+[OTLP](0035-opentelemetry-protocol.md) (OpenTelemetry Protocol).
+
+## Table of Contents
+
+* [Motivation](#motivation)
+* [OTLP/HTTP Protocol Details](#otlphttp-protocol-details)
+  * [Request](#request)
+  * [Response](#response)
+    * [Success](#success)
+    * [Failures](#failures)
+    * [Throttling](#throttling)
+    * [All Other Responses](#all-other-responses)
+  * [Connection](#connection)
+  * [Parallel Connections](#parallel-connections)
+* [Prior Art and Alternatives](#prior-art-and-alternatives)
+
+## Motivation
+
+OTLP can be currently communicated only via one transport: gRPC. While using
+gRPC has certain benefits there are also drawbacks:
+
+- Some users have infrastructure limitations that make gRPC-based protocol
+  usage impossible. For example AWS ALB does not support gRPC connections.
+
+- gRPC is a relatively big dependency, which some clients are not willing to
+  take. Plain HTTP is a smaller dependency and is built in the standard
+  libraries of many programming languages.
+
+## OTLP/HTTP Protocol Details
+
+This proposal keeps the existing specification of OTLP over gRPC transport
+(OTLP/gRPC for short) and defines an additional way to use OTLP protocol over
+HTTP transport (OTLP/HTTP for short). OTLP/HTTP uses the same ProtoBuf payload
+that is used by OTLP/gRPC and defines how this payload is communicated over HTTP
+transport.
+
+OTLP/HTTP uses HTTP POST requests to send telemetry data from clients to
+servers. Implementations MAY use HTTP/1.1 or HTTP/2 transports. Implementations
+that use HTTP/2 transport SHOULD fallback to HTTP/1.1 transport if HTTP/2
+connection cannot be established.
+
+### Request
+
+Telemetry data is sent via HTTP POST request. The request body is a
+ProtoBuf-encoded message `TelemetryRequest` as defined below:
+
+```protobuf
+enum RequestType {
+    _ = 0;
+    TraceExport = 1;
+    MetricExport = 2;
+}
+message TelemetryRequest {
+    RequestType request_type = 1;
+
+    // Only one of the following fields is supposed to contain data (determined
+    // by `request_type` field). This is deliberately not using ProtoBuf `oneof`
+    // for performance reasons (verified by benchmarks).
+    ExportTraceServiceRequest export_trace_request = 2;
+    ExportMetricsServiceRequest export_metrics_request = 3;
+}
+```
+
+`TelemetryRequest` references `ExportTraceServiceRequest` and
+`ExportMetricsServiceRequest` message types that are defined in
+[OTLP ProtoBuf schema](https://github.com/open-telemetry/opentelemetry-proto/tree/master/opentelemetry/proto)
+specification.
+
+The client MUST set "Content-Type: application/x-protobuf" request header. The
+client MAY gzip the content and in that case SHOULD include "Content-Encoding:
+gzip" request header. The client MAY include "Accept-Encoding: gzip" request
+header if it can receive gzip-encoded responses.
+
+The default request URL path is `/telemetry` (for example the full URL when
+connecting to "example.com" server will be `https://example.com/telemetry`). A
+different URL path MAY be configured on the client and server sides.
+
+### Response
+
+#### Success
+
+On success the server MUST respond with `HTTP 200 OK` and include
+ProtoBuf-encoded `TelemetryResponse` message in the response body.
+`TelemetryResponse` is defined as follows:
+
+```protobuf
+enum ResponseType {
+    _ = 0;
+    TraceExport = 1;
+    MetricExport = 2;
+}
+
+message TelemetryResponse {
+    ResponseType response_type = 1;
+
+    // Only one of the following fields is supposed to contain data (determined
+    // by `response_type` field). This is deliberately not using ProtoBuf
+    // `oneof` for performance reasons (verified by benchmarks).
+    ExportTraceServiceResponse export_trace_response = 2;
+    ExportMetricsServiceResponse export_metrics_response = 3;
+}
+```
+
+`TelemetryResponse` references `ExportTraceServiceResponse` and
+`ExportTraceServiceResponse` message types that are defined in [OTLP Protocol
+Buffers](https://github.com/open-telemetry/opentelemetry-proto/tree/master/opentelemetry/proto)
+specification.
+
+The server MUST set "Content-Type: application/x-protobuf" response header. If
+the request header "Accept-Encoding: gzip" is present in the request the server
+MAY gzip-encode the response and set "Content-Encoding: gzip" response header.
+
+The server SHOULD respond with success no sooner than after successfully
+decoding and validating the request.
+
+#### Failures
+
+If the processing of the request fails because the request contains data that
+cannot be decoded or is otherwise invalid and such failure is permanent then the
+server MUST respond with `HTTP 400 Bad Request`. The client MUST NOT retry the
+request when it receives `HTTP 400 Bad Request` response.
+
+#### Throttling
+
+If the server receives more requests than the client is allowed or the server is
+overloaded the server SHOULD respond with `HTTP 429 Too Many Requests` or
+`HTTP 503 Service Unavailable` and MAY include "Retry-After" header with a
+recommended time interval in seconds to wait before retrying. The client SHOULD
+honour the waiting interval specified in "Retry-After" header if it is present.
+If the client receives `HTTP 429` or `HTTP 503` response and "Retry-After"
+header is not present in the response then the client SHOULD implement an
+exponential backoff strategy between retries.
+
+#### All Other Responses
+
+All other HTTP responses that are not explicitly listed in this document should
+be treated according to HTTP specification.
+
+If the server disconnects without returning a response the client SHOULD retry
+and send the same request. The client SHOULD implement an exponential backoff
+strategy between retries to avoid overwhelming the server.
+
+### Connection
+
+If the client is unable to connect to the server the client SHOULD retry the
+connection using exponential backoff strategy between retries. The interval
+between retries must have a random jitter.
+
+The client SHOULD keep the connection alive between requests. The client SHOULD
+use "Connection: Keep-Alive" request header.
+
+Server implementations MAY handle OTLP/gRPC and OTLP/HTTP requests on the same
+port and multiplex the connections to the corresponding transport handler based
+on "Content-Type" request header.
+
+### Parallel Connections
+
+To achieve higher total throughput the client MAY send requests using several
+parallel HTTP connections. In that case the maximum number of parallel
+connections SHOULD be configurable.
+
+## Prior Art and Alternatives
+
+I have also considered HTTP/1.1+WebSocket transport. Experimental implementation
+of OTLP over WebSocket transport has shown that it typically has better
+performance than plain HTTP transport implementation (WebSocket uses less CPU,
+higher throughput in high latency connections). However WebSocket transport
+requires slightly more complicated implementation and WebSocket libraries are
+less ubiquitous than plain HTTP, which may make implementation in certain
+languages difficult or impossible.
+
+HTTP/1.1+WebSocket transport may be considered as a future transport for
+high-performance use cases as it exhibits better performance than OTLP/gRPC and
+OTLP/HTTP.
--- a/text/0099-otlp-http.md
+++ b/text/0099-otlp-http.md
@@ -1,7 +1,5 @@
 # OTLP/HTTP: HTTP Transport Extension for OTLP
 
-_Author: Tigran Najaryan, Splunk._
-
 This is a proposal to add HTTP Transport extension for
 [OTLP](0035-opentelemetry-protocol.md) (OpenTelemetry Protocol).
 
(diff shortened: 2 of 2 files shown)


text/0000-otlp-http.md
```diff
@@ -0,0 +1,258 @@
...
+`ExportMetricsServiceRequest` message types that are defined in
+[OTLP ProtoBuf schema](https://github.com/open-telemetry/opentelemetry-proto/tree/master/opentelemetry/proto)
+specification.
+
+The client MUST set "Content-Type: application/x-protobuf" request header. The
+client MAY gzip the content and in that case SHOULD include "Content-Encoding:
+gzip" request header. The client MAY include "Accept-Encoding: gzip" request
+header if it can receive gzip-encoded responses.
+
+The default request URL path is `/telemetry` (for example the full URL when
```
bogdandrutu, line 80: Why not using two paths instead of an enum? It is easier to configure LBs I think

text/0000-otlp-http.md
```diff
@@ -0,0 +1,258 @@
...
+The following is the benchmarking result of
+[Go implementations](https://github.com/tigrannajaryan/exp-otelproto/tree/master/cmd/benchmark),
+running on on Ubuntu 18 LTS system with i7 7500U processor, 16 GB RAM. The test
+benchmarks loopback sending and receiving of span batches. (Note that the
+benchmarking script sets "performance" CPU governor during execution and sets
+nice value of the process for more consistent results).
+
+```
+====================================================================================
+Legend:
```
bogdandrutu, line 193: It is not necessary a correct comparison. There are some features like deadline/retries/etc. that gRPC has and I don't know if they are present in the benchmarks for HTTP.

tigrannajaryan: OK, let me remove the benchmark since it does not really matter for the motivation of this proposal.

text/0000-otlp-http.md
tigrannajaryan: That's a good point. It will definitely make routing the traffic easier.

tigrannajaryan: Changed to use URL path.

text/0099-otlp-http.md
```diff
@@ -0,0 +1,146 @@
...
+that use HTTP/2 transport SHOULD fallback to HTTP/1.1 transport if HTTP/2
+connection cannot be established.
+
+### Request
+
+Telemetry data is sent via HTTP POST request.
+
+The default URL path for requests that carry trace data is `/traces` (for
+example the full URL when connecting to "example.com" server will be
+`https://example.com/telemetry`). The request body is a ProtoBuf-encoded
```
yurishkuro, line 53: `/telemetry` or `/traces`?

text/0099-otlp-http.md
```diff
@@ -0,0 +1,146 @@
...
+
+The server SHOULD respond with success no sooner than after successfully
+decoding and validating the request.
+
+#### Failures
+
+If the processing of the request fails because the request contains data that
+cannot be decoded or is otherwise invalid and such failure is permanent then the
+server MUST respond with `HTTP 400 Bad Request`. The client MUST NOT retry the
+request when it receives `HTTP 400 Bad Request` response.
```
yurishkuro, line 93: no payload explaining the issue?

text/0099-otlp-http.md
```diff
@@ -0,0 +1,146 @@
...
+I have also considered HTTP/1.1+WebSocket transport. Experimental implementation
+of OTLP over WebSocket transport has shown that it typically has better
+performance than plain HTTP transport implementation (WebSocket uses less CPU,
+higher throughput in high latency connections). However WebSocket transport
+requires slightly more complicated implementation and WebSocket libraries are
+less ubiquitous than plain HTTP, which may make implementation in certain
+languages difficult or impossible.
+
+HTTP/1.1+WebSocket transport may be considered as a future transport for
+high-performance use cases as it exhibits better performance than OTLP/gRPC and
```
yurishkuro, line 145: Can you share any numbers how much better the performance with WebSocket is?

bogdandrutu: Also I think we can share some numbers in this comment, but I would suggest that a written version in specs or collector protocol to be posted after we verify that we are not doing some trivial mistakes.

text/0099-otlp-http.md
tigrannajaryan: Fixed.

text/0099-otlp-http.md
tigrannajaryan: We don't have the equivalent in OTLP/gRPC either. You are right that it is a good idea to have one, especially if we know what we want to have in the payload (perhaps it is just an empty ProtoBuf message initially).
Let me think.

yurishkuro: in gRPC we at least had guidelines on return codes and using grpc error details for additional information. At minimum, it's customary to return some error message in the HTTP body, in addition to the status code.

text/0099-otlp-http.md
tigrannajaryan: Not to distract from this PR, just to satisfy curiosity :-) below is plain HTTP (this PR) vs WebSocket. Both using single HTTP/1.1 connection.

These are results from quick code I put together (which likely has bugs). No profiling or optimizations done, so should not be considered definitive (code here https://github.com/tigrannajaryan/exp-otelproto). 

WebSocket wins for small batches (in CPU) or high latencies (in throughput) are pretty big.

The only outlier/counterexample is very large batches (unclear why it loses in that case, I did not dig into it - can be done if we seriously consider WebSocket in the future).

```
64000 nano batches, 1 spans per batch, 10 attrs per span
OTLP/HTTP1.1/1                 64000 spans| 13.1 cpusec|  8.9 wallsec| 4885.5 batches/cpusec|  7165.3 batches/wallsec|204.7 cpuμs/span
WebSocket/Stream/Async/1       64000 spans|  3.2 cpusec|  1.4 wallsec|19937.7 batches/cpusec| 44880.6 batches/wallsec| 50.2 cpuμs/span

16000 tiny batches, 10 spans per batch, 10 attrs per span
OTLP/HTTP1.1/1                160000 spans|  7.4 cpusec|  4.9 wallsec| 2176.9 batches/cpusec|  3256.3 batches/wallsec| 45.9 cpuμs/span
WebSocket/Stream/Async/1      160000 spans|  2.8 cpusec|  1.5 wallsec| 5714.3 batches/cpusec| 10597.7 batches/wallsec| 17.5 cpuμs/span

8000 small batches, 100 spans per batch, 10 attrs per span
OTLP/HTTP1.1/1                800000 spans| 14.5 cpusec| 11.2 wallsec|  550.6 batches/cpusec|   714.0 batches/wallsec| 18.2 cpuμs/span
WebSocket/Stream/Async/1      800000 spans| 10.3 cpusec|  6.1 wallsec|  779.7 batches/cpusec|  1313.7 batches/wallsec| 12.8 cpuμs/span

800 large batches, 500 spans per batch, 10 attrs per span
OTLP/HTTP1.1/1                400000 spans|  5.2 cpusec|  4.5 wallsec|  154.4 batches/cpusec|   177.7 batches/wallsec| 12.9 cpuμs/span
WebSocket/Stream/Async/1      400000 spans|  4.7 cpusec|  2.9 wallsec|  169.9 batches/cpusec|   277.5 batches/wallsec| 11.8 cpuμs/span

100 very large batches, 5000 spans per batch, 10 attrs per span
OTLP/HTTP1.1/1                500000 spans|  5.5 cpusec|  5.3 wallsec|   18.1 batches/cpusec|    18.9 batches/wallsec| 11.1 cpuμs/span
WebSocket/Stream/Async/1      500000 spans|  8.0 cpusec|  3.6 wallsec|   12.5 batches/cpusec|    27.4 batches/wallsec| 16.0 cpuμs/span

2ms network roundtrip latency
800 large batches, 500 spans per batch, 10 attrs per span
OTLP/HTTP1.1/1                400000 spans|  5.1 cpusec|  6.3 wallsec|  155.3 batches/cpusec|   126.8 batches/wallsec| 12.9 cpuμs/span
WebSocket/Stream/Async/1      400000 spans|  4.8 cpusec|  3.0 wallsec|  167.7 batches/cpusec|   270.6 batches/wallsec| 11.9 cpuμs/span

20ms network roundtrip latency
400 large batches, 500 spans per batch, 10 attrs per span
OTLP/HTTP1.1/1                200000 spans|  2.6 cpusec| 10.8 wallsec|  151.5 batches/cpusec|    36.9 batches/wallsec| 13.2 cpuμs/span
WebSocket/Stream/Async/1      200000 spans|  2.3 cpusec|  1.6 wallsec|  170.9 batches/cpusec|   250.2 batches/wallsec| 11.7 cpuμs/span

200ms network roundtrip latency
40 large batches, 500 spans per batch, 10 attrs per span
OTLP/HTTP1.1/1                 20000 spans|  0.3 cpusec| 12.7 wallsec|  148.1 batches/cpusec|     3.2 batches/wallsec| 13.5 cpuμs/span
WebSocket/Stream/Async/1       20000 spans|  0.2 cpusec|  2.5 wallsec|  166.7 batches/cpusec|    15.9 batches/wallsec| 12.0 cpuμs/span

200ms network roundtrip latency
400 large batches, 500 spans per batch, 10 attrs per span
OTLP/HTTP1.1/1                200000 spans|  2.8 cpusec| 86.4 wallsec|  144.9 batches/cpusec|     4.6 batches/wallsec| 13.8 cpuμs/span
WebSocket/Stream/Async/1      200000 spans|  2.4 cpusec| 28.4 wallsec|  167.4 batches/cpusec|    14.1 batches/wallsec| 11.9 cpuμs/span
```

text/0099-otlp-http.md
tigrannajaryan: You are right, I forgot that it is possible to include error details in gRPC, not just the status code. Let me add here too.

yurishkuro: I wonder if there's already some standard around this, e.g. when using grpc-gateway to expose gRPC service over HTTP/JSON.

text/0099-otlp-http.md
```diff
@@ -0,0 +1,146 @@
...
+
+To achieve higher total throughput the client MAY send requests using several
+parallel HTTP connections. In that case the maximum number of parallel
+connections SHOULD be configurable.
+
+## Prior Art and Alternatives
+
+I have also considered HTTP/1.1+WebSocket transport. Experimental implementation
+of OTLP over WebSocket transport has shown that it typically has better
+performance than plain HTTP transport implementation (WebSocket uses less CPU,
```
flands, line 138: What about performance in general -- how will HTTP perform over gRPC?

text/0099-otlp-http.md
```diff
@@ -0,0 +1,146 @@
...
+  * [Response](#response)
+    * [Success](#success)
+    * [Failures](#failures)
+    * [Throttling](#throttling)
+    * [All Other Responses](#all-other-responses)
+  * [Connection](#connection)
+  * [Parallel Connections](#parallel-connections)
+* [Prior Art and Alternatives](#prior-art-and-alternatives)
+
+## Motivation
```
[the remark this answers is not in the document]
tigrannajaryan: HTTP/1.1 works fine with HTTP load balancers. I did not check HTTP/2 but I suspect it should work fine too because we send individual requests.

text/0099-otlp-http.md
```diff
@@ -0,0 +1,146 @@
...
+OTLP/HTTP uses HTTP POST requests to send telemetry data from clients to
+servers. Implementations MAY use HTTP/1.1 or HTTP/2 transports. Implementations
+that use HTTP/2 transport SHOULD fallback to HTTP/1.1 transport if HTTP/2
+connection cannot be established.
+
+### Request
+
+Telemetry data is sent via HTTP POST request.
+
+The default URL path for requests that carry trace data is `/traces` (for
```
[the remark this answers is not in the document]
owais: Specifying version makes sense. We can also use HTTP headers to specify exact version like `OTLP_VERSION: 1.2`.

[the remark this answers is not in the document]
tigrannajaryan: I'll go with version number in the path. Additional header is a lot more bytes on the wire.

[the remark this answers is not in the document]
bogdandrutu: Also easier to configure for LB rules.

text/0099-otlp-http.md
tigrannajaryan: grpc-gateway returns a JSON-encoded [Status](https://godoc.org/google.golang.org/genproto/googleapis/rpc/status#Status) message. Status contains `code` (which is superfluous, since it is already reflected in HTTP Status), `message` string and `details` (which is `Any` object).

We also already use the same `Status` message in OTLP/gRPC to report errors and use `details` to provide additional information (e.g. throttling interval).

While some of the info in `Status` ProtoBuf message is a duplicate of what is available in HTTP response (in HTTP status code and in HTTP headers) I am still inclined to use `Status` for consistency, instead of inventing yet another message format just for OTLP/HTTP error response bodies.

Thoughts?

text/0099-otlp-http.md
```diff
@@ -0,0 +1,146 @@
...
+message for metrics.
+
+The server MUST set "Content-Type: application/x-protobuf" response header. If
+the request header "Accept-Encoding: gzip" is present in the request the server
+MAY gzip-encode the response and set "Content-Encoding: gzip" response header.
+
+The server SHOULD respond with success no sooner than after successfully
+decoding and validating the request.
+
+#### Failures
```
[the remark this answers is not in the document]
tigrannajaryan: This is specified in the catch-all "All Other Responses" section, which says - just follow HTTP spec. I do not think this protocol should prescribe any specific auth behavior. HTTP has provisions for authentication (e.g. Basic auth and 401 Unauthorized), but some backends use custom secrets in header (aka API key).

Since this is a vendor-specific behavior and there are known practices for auth in HTTP I do not want to prescribe any particular auth approach in this protocol.

In practice this means that OTLP/HTTP client implementations will have to allow users to supply any HTTP headers to be included in the request. This is what we already do for OTLP and OpenCensus exporter implementations in OpenTelemetry Collector. Perhaps the ability to include user-defined headers is worth calling out as a SHOULD requirement for clients? I am not sure.

text/0099-otlp-http.md
tigrannajaryan: Performance is comparable, but I will need to do more testing to make sure the differences I see are not measurement errors, so for now I omitted the results.

text/0099-otlp-http.md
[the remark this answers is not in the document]
yurishkuro: A general section stating "use what's normally used in HTTP" wouldn't hurt, to address such questions going forward.

text/0099-otlp-http.md
```diff
@@ -0,0 +1,144 @@
...
+If the processing of the request fails because the request contains data that
+cannot be decoded or is otherwise invalid and such failure is permanent then the
+server MUST respond with `HTTP 400 Bad Request`. The client MUST NOT retry the
+request when it receives `HTTP 400 Bad Request` response.
+
+#### Throttling
+
+If the server receives more requests than the client is allowed or the server is
+overloaded the server SHOULD respond with `HTTP 429 Too Many Requests` or
+`HTTP 503 Service Unavailable` and MAY include "Retry-After" header with a
```
yurishkuro, line 97: Is `Retry-After` based on some RFC? If not, could we use returned Status for this info?

text/0099-otlp-http.md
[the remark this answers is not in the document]
tigrannajaryan: @yurishkuro I think it is already in the doc:

>All other HTTP responses that are not explicitly listed in this document should
be treated according to HTTP specification.

Does this need additional clarification?

text/0099-otlp-http.md
```diff
@@ -0,0 +1,144 @@
...
+If the client is unable to connect to the server the client SHOULD retry the
+connection using exponential backoff strategy between retries. The interval
+between retries must have a random jitter.
+
+The client SHOULD keep the connection alive between requests. The client SHOULD
+use "Connection: Keep-Alive" request header.
+
+Server implementations MAY handle OTLP/gRPC and OTLP/HTTP requests on the same
+port and multiplex the connections to the corresponding transport handler based
+on "Content-Type" request header.
```
yurishkuro, line 124: In Jaeger we used sometimes wonky `http/1.1` vs. `http/2` check to separate grpc from http traffic on the same port. Is gRPC going to send something different than `"Content-Type: application/x-protobuf"`? It may be word adding a bit more detail here about how to reuse ports correctly if we're bringing this up at all.

text/0099-otlp-http.md
tigrannajaryan: Yes, see HTTP/1.1 RFC https://tools.ietf.org/html/rfc7231#section-6.6.4:

> The 503 (Service Unavailable) status code indicates that the server
>    is currently unable to handle the request due to a temporary overload
>    or scheduled maintenance, which will likely be alleviated after some
>    delay.  The server MAY send a Retry-After header field
>    (Section 7.1.3) to suggest an appropriate amount of time for the
>    client to wait before retrying the request.

text/0099-otlp-http.md
yurishkuro: Actually, it looks like we switched to context-type as well: https://github.com/jaegertracing/jaeger/blob/46fc5ebe08a90a7fa7d3222f62d0535118a7ba5c/cmd/query/app/server.go#L116

tigrannajaryan: Yes, based "Content-Type". gRPC uses "Content-Type: application/grpc", OTLP/HTTP (this proposal) uses "Content-Type: application/x-protobuf".
I can mentioned that "Content-Type" can be used for multiplexing in the doc if you think it is worth it.

text/0099-otlp-http.md
```diff
@@ -0,0 +1,144 @@
...
+strategy between retries to avoid overwhelming the server.
+
+### Connection
+
+If the client is unable to connect to the server the client SHOULD retry the
+connection using exponential backoff strategy between retries. The interval
+between retries must have a random jitter.
+
+The client SHOULD keep the connection alive between requests. The client SHOULD
+use "Connection: Keep-Alive" request header.
```
tigrannajaryan, line 120: After re-reading the RFC I am going to remove "Connection: Keep-Alive" requirement as apparently it is not needed for HTTP/1.1 connections to keep alive: https://tools.ietf.org/html/rfc7230#section-6.3
The client can keep the connection alive without this header.

text/0099-otlp-http.md
tigrannajaryan: OK, adding.

text/0099-otlp-http.md
tigrannajaryan: Adding the link to the text.

text/0099-otlp-http.md
tigrannajaryan: @yurishkuro Added, please have a look.

text/0099-otlp-http.md
tigrannajaryan: Done.

PR #101, opened 2020-04-29
Add performance goals to log vision

PettitWesley:
Based on the recent discussion in the Collector SIG, I felt we should write down the resource utilization goals for the unified collector. 

I expect disagreement/comment on this; this is also my first otep PR, so I understand I might not have used all of the correct terminology.

At AWS, resource consumption is one of our biggest concerns/criteria for evaluating agents. The side-car/serverless use case is especially important to us/me; the resource usage goal for that segment mentioned in this change is potentially a hard requirement for us.

Signed-off-by: Wesley Pettit <wppttt@amazon.com>

### DIFF
--- a/text/0092-logs-vision.md
+++ b/text/0092-logs-vision.md
@@ -1,6 +1,6 @@
 # OpenTelemetry Logs Vision
 
-The following are high-level items that define our long-term vision for 
+The following are high-level items that define our long-term vision for
 Logs support in OpenTelemetry project, what we aspire to achieve.
 
 This a vision document that reflects our current desires. It is not a commitment
@@ -34,8 +34,8 @@ of logs. The purpose of the data model is to have a common understanding of what
 a log record is, what data needs to be recorded, transferred, stored and
 interpreted by a logging system.
 
-Existing log formats can be unambiguously mapped to this data model. Reverse 
-mapping from this data model is also possible to the extent that the target log 
+Existing log formats can be unambiguously mapped to this data model. Reverse
+mapping from this data model is also possible to the extent that the target log
 format has equivalent capabilities.
 
 We will produce mapping recommendations for commonly used log formats.
@@ -72,6 +72,15 @@ Unified collection is important for the following reasons:
 - Uniform tagging of all 3 types of telemetry data (enrichment by attributes
   of resources where the data comes from or by user-defined attributes),
   enabling correct correlation across Resource dimensions later on the backend.
+
+#### Performance
+
+The unified collector must be performant; its resource utilization should be minimal for most use cases, but must also handle high throughput when needed.
+
+We have identified three sets of use cases for the collector with differing requirements:
+- *Side-car*: There are a set of use cases where a single instance of the collector will be deployed alongside a single instance of an application- a 1:1 relationship. This set includes many serverless environments. A single instance of an application may produce telemetry events (logs, metrics, traces) measured in 10s to 1000s per second. For these use cases, the collector's resource consumption should not exceed 100 MB memory, or 20% of a CPU on a modern server. We will try to optimize the collector so that for many users in this group, the collector will consume less than 50 MB of memory and less than 10% of a CPU.
+- *Daemon*: Many users will deploy the collector as a daemon, a single instance per node or host, to collect telemetry from multiple applications. In containerized environments, [roughly 10-20 containers are generally run per node](https://www.datadoghq.com/docker-adoption/). As noted above, these applications may emit 10s to 1000s of telemetry events per second. For this segment, our goal is that the collector not be a significant consumer of resources when compared with applications deployed on that node. Its resource consumption should be a fraction of the usage of the applications that it serves; preferably <10%, ideally <5%.
+- *Service*: The collector may also be deployed as a standalone service that serves multiple applications across multiple nodes. For this use case, low resource consumption is not the primary concern; high throughput is the goal.
 
 ## Cloud Native
 


text/0092-logs-vision.md
```diff
@@ -73,6 +73,15 @@ Unified collection is important for the following reasons:
...
+- *Daemon*: Many users will deploy the collector as a daemon, a single instance per node or host, to collect telemetry from multiple applications. In containerized environments, [roughly 10-20 containers are generally run per node](https://www.datadoghq.com/docker-adoption/). As noted above, these applications may emit 10s to 1000s of telemetry events per second. For this segment, our goal is that the collector not be a significant consumer of resources when compared with applications deployed on that node. Its resource consumption should be a fraction of the usage of the applications that it serves; preferably <10%, ideally <5%.
```
PettitWesley, line 82: I'm currently working benchmarking the performance of two agents (Fluent Bit and the AWS CloudWatch Agent) when reading log files and sending them to AWS. The CloudWatch Agent is written in Go, and seems to be able to meet these requirements. Therefore, we think that the OT Collector could as well. (Fluent Bit of course also meets these requirements). I'm hoping to be able to share the results in a few days.

jeffalder:
Apologies if this was already covered, but this feels like _requirements_ rather than a _vision_. A vision, to me, answers the question "where do we see ourselves going in the short-, medium- and long-term?" Requirements are more about "what do we need to produce _today_?". 

I don't know how performance requirements fit in the OTEP process, so I can't offer a suggestion; it might be best as a separate document. 

It also sounds like these are requirements for a collector that is potentially getting traces and/or metrics, in addition to logging, so there's a scope problem: If these are overall collector requirements, those requirements belong to the planning process for the collector itself.  

Does that make sense? These are all perfectly valid constraints, I'm just wondering where the best place to actually _put_ them is.

tigrannajaryan:
> Apologies if this was already covered, but this feels like _requirements_ rather than a _vision_. A vision, to me, answers the question "where do we see ourselves going in the short-, medium- and long-term?" Requirements are more about "what do we need to produce _today_?".

I agree. I think we need to keep the vision doc short and high-level without going into too much specific details otherwise it can become too large for easy consumption. We want anyone who wants to contribute reads this doc to be aligned on what we do and it is important to be concise in such documents.

@PettitWesley Performance requirements are very important, thanks for producing them.

I think they belong to a different place. It is likely best placed as a new requirement document in Collector repo. It can also be linked from a (to be added) "Log support" line item in the existing Roadmap doc in Collector repo.

text/0092-logs-vision.md
```diff
@@ -73,6 +73,15 @@ Unified collection is important for the following reasons:
   of resources where the data comes from or by user-defined attributes),
   enabling correct correlation across Resource dimensions later on the backend.
 
+#### Performance
+
+The unified collector must be performant; its resource utilization should be minimal for most use cases, but must also handle high throughput when needed.
+
+We have identified three sets of use cases for the collector with differing requirements:
+- *Side-car*: There are a set of use cases where a single instance of the collector will be deployed alongside a single instance of an application- a 1:1 relationship. This set includes many serverless environments. A single instance of an application may produce telemetry events (logs, metrics, traces) measured in 10s to 1000s per second. For these use cases, the collector's resource consumption should not exceed 100 MB memory, or 20% of a CPU on a modern server. We will try to optimize the collector so that for many users in this group, the collector will consume less than 50 MB of memory and less than 10% of a CPU.
```
yurishkuro, line 81: The exact memory & CPU numbers seem to be rather arbitrary to be included in the spec. I think a more reasonable approach is to require that collector needs to be adaptable to certain memory and CPU constraints, using load-shedding techniques, but the exact limits should be left to the end user to define.

text/0092-logs-vision.md
```diff
@@ -73,6 +73,15 @@ Unified collection is important for the following reasons:
   of resources where the data comes from or by user-defined attributes),
   enabling correct correlation across Resource dimensions later on the backend.
 
+#### Performance
+
+The unified collector must be performant; its resource utilization should be minimal for most use cases, but must also handle high throughput when needed.
```
tigrannajaryan, line 78: I think this is not necessary since it is already covered in Collector vision doc: https://github.com/open-telemetry/opentelemetry-collector/blob/master/docs/vision.md#performant

>Highly stable and performant under varying loads. Well-behaved under extreme load, with predictable, low resource consumption.

text/0092-logs-vision.md
tigrannajaryan, line 81: Collector CPU and RAM usage normally is load-dependent. Performance requirements that do not specify the load can be of 2 types:
- Resource usage under zero load - idle mode.
- Resource usage under overload, i.e. maximum allowed CPU and RAM usage no matter how high the load (up to certain limit until it saturates the ability of the process to trim the input load).

I am not sure which of these cases the requirement in this PR attempts to cover or if it is about some other case. This needs clarification.

There is another case for performance requirements: the normal operating case, where the load is above zero but below the point where overload occurs. The requirement for this case must include the load description - the rate of incoming data and the composition of the data (e.g. 1000 spans per second, each span with 10 attributes of average (random) length of 20 bytes, etc). Given that the resource consumption may have non-linear correlation with load rate it may be necessary to specify multiple data points for this. We will also likely need different requirements for traces, metrics and logs (although we should try to be consistent across signals where we can). Furthermore, resource usage is protocol dependent, so this needs to say which protocol we want to test. Finally, performance is highly affected by processors, presumably these are numbers without any processing, in passthrough mode. This needs to be called out specifically or we need to call out what processing is part of the benchmarks.

There are other cases that may be interesting from performance perspective. I think these 3 are most important.

text/0092-logs-vision.md
```diff
@@ -73,6 +73,15 @@ Unified collection is important for the following reasons:
...
+- *Daemon*: Many users will deploy the collector as a daemon, a single instance per node or host, to collect telemetry from multiple applications. In containerized environments, [roughly 10-20 containers are generally run per node](https://www.datadoghq.com/docker-adoption/). As noted above, these applications may emit 10s to 1000s of telemetry events per second. For this segment, our goal is that the collector not be a significant consumer of resources when compared with applications deployed on that node. Its resource consumption should be a fraction of the usage of the applications that it serves; preferably <10%, ideally <5%.
+- *Service*: The collector may also be deployed as a standalone service that serves multiple applications across multiple nodes. For this use case, low resource consumption is not the primary concern; high throughput is the goal.
```
tigrannajaryan, line 83: >low resource consumption is not the primary concern

I know people for whom resource consumption in standalone mode is a significant concern, to the point that they are willing to ditch the Collector and write a custom service to proxy telemetry data faster than the Collector can do today.

I would delete the first part of the statement and perhaps only mention the "throughout" part, which is a valid concern. Additionally, throughput needs to be quantified. It is highly dependent on the protocol and on roundtrip latency.

tedsuo:
@PettitWesley I'm closing this as we now have a log implementation in the Collector.

PR #102, opened 2020-04-29
Add list of log sources to log vision

PettitWesley:
I created this list by looking at the list of log inputs for Fluentd, Fluent Bit, and Microsoft's recently announced [Container Log Monitor](https://techcommunity.microsoft.com/t5/containers/windows-containers-log-monitor-opensource-release/ba-p/973947#).

### DIFF
--- a/text/0092-logs-vision.md
+++ b/text/0092-logs-vision.md
@@ -1,6 +1,6 @@
 # OpenTelemetry Logs Vision
 
-The following are high-level items that define our long-term vision for 
+The following are high-level items that define our long-term vision for
 Logs support in OpenTelemetry project, what we aspire to achieve.
 
 This a vision document that reflects our current desires. It is not a commitment
@@ -34,8 +34,8 @@ of logs. The purpose of the data model is to have a common understanding of what
 a log record is, what data needs to be recorded, transferred, stored and
 interpreted by a logging system.
 
-Existing log formats can be unambiguously mapped to this data model. Reverse 
-mapping from this data model is also possible to the extent that the target log 
+Existing log formats can be unambiguously mapped to this data model. Reverse
+mapping from this data model is also possible to the extent that the target log
 format has equivalent capabilities.
 
 We will produce mapping recommendations for commonly used log formats.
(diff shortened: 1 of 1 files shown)


text/0092-logs-vision.md
```diff
@@ -96,6 +96,16 @@ This is important because we believe software that was created before
 OpenTelemetry should not be disregarded and should benefit from OpenTelemetry
 efforts where possible.
 
+Specifically, we should be able to collect the following sources:
+- Log files, especially those emitted by container runtimes
+- Syslog
+- Systemd/Journald
+- Windows Event Log
+- Event Tracing for Windows (ETW)
+- JSON over TCP or UDP
```
PettitWesley, line 105: We use raw JSON over TCP at AWS with some libraries: https://github.com/awslabs/aws-embedded-metrics-python

Anecdotally, having worked with Fluentd and Fluent Bit for 2 years, these inputs are rare/niche compared to others, but definitely in use by a good number of people.

text/0092-logs-vision.md
```diff
@@ -96,6 +96,16 @@ This is important because we believe software that was created before
 OpenTelemetry should not be disregarded and should benefit from OpenTelemetry
 efforts where possible.
 
+Specifically, we should be able to collect the following sources:
+- Log files, especially those emitted by container runtimes
+- Syslog
+- Systemd/Journald
+- Windows Event Log
```
PettitWesley, line 103: I think these four are obvious and won't lead to any disagreement, the others were added for completeness, but I'm happy to remove them if thats desired

tigrannajaryan: I agree, these look like basic things that we do want to support. We may need to discuss and decide the priorities.

These are requirements for the Collector, right? We do not intend to collect these using e.g. OpenTelemetry SDK. If these are for the Collector then best to make these Collector roadmap items.

text/0092-logs-vision.md
tigrannajaryan: These are probably good fit for the Contrib. I do not think they should be part of the vision or core roadmap. Collector currently encourages to actively contribute components that support specific formats to Contrib repo and we already see this happening for traces and metrics. I believe we need to continue the practice for logs. If anyone is interested in these particular formats they should be able to add these as components.

PR #103, opened 2020-05-01
Enforce lint, link checks, fix errors. Remove "approved" status.

bogdandrutu:


text/0001-telemetry-without-manual-instrumentation.md
```diff
@@ -56,23 +53,24 @@ There is also a school of thought that we should only be focusing on the bits an
...
+
 * First and foremost, **portable OpenTelemetry instrumentation can be installed without manual source code modification**
 * There’s one “clear winner” when it comes to portable, automatic instrumentation; just like with OpenTracing and OpenCensus, this is a situation where choice is not necessarily a good thing. End-users who wish to contribute instrumentation plugins should not have their enthusiasm and generosity diluted across competing projects.
 * As much as such a thing is possible, consistency across languages
 * Broad coverage / “plugin support”
 * Broad vendor support for OpenTelemetry
 * All other things being equal, get all of these ^^ benefits ASAP!
 
-### What's the basic proposal?
+### What's the basic proposal
```
Oberon00, line 67: This is a content change. I don't think this belongs here.

bogdandrutu: There is a rule in the markdown lint to not use punctuation signs in headers.

bogdandrutu: And I would argue that this change makes any significant "content change" :)

text/0016-named-tracers.md
```diff
@@ -29,16 +27,18 @@ It is the eventual goal of OpenTelemetry that library vendors implement the Open
...
 
 ## Explanation
 
 From a user perspective, working with *Named Tracers / Meters* and `TracerProvider` / `MeterProvider` is conceptually similar to how e.g. the [Java logging API](https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html#getLogger(java.lang.String)) and logging frameworks like [log4j](https://www.slf4j.org/apidocs/org/slf4j/LoggerFactory.html) work. In analogy to requesting Logger objects through LoggerFactories, an instrumentation library would create specific Tracer / Meter objects through a TracerProvider / MeterProvider.
 
-New Tracers or Meters can be created by providing the name and version of an instrumentation library. The version (following the convention proposed in https://github.com/open-telemetry/oteps/pull/38) is basically optional but *should* be supplied since only this information enables following scenarios:
+New Tracers or Meters can be created by providing the name and version of an instrumentation library. The version (following the convention proposed in <https://github.com/open-telemetry/oteps/pull/38)> is basically optional but *should* be supplied since only this information enables following scenarios:
```
arminru, line 40:
```suggestion
New Tracers or Meters can be created by providing the name and version of an instrumentation library. The version (following the convention proposed in <https://github.com/open-telemetry/oteps/pull/38>) is basically optional but *should* be supplied since only this information enables following scenarios:
```

text/0092-logs-vision.md
```diff
@@ -9,38 +9,39 @@ document is to ensure all contributors work in alignment. As our vision changes
 over time maintainers reserve the right to add, modify, and remove items from
 this document.
 
-This document uses vocabulary introduced in https://github.com/open-telemetry/oteps/pull/91.
+This document uses vocabulary introduced in <https://github.com/open-telemetry/oteps/pull/91.>
```
arminru, line 12:
```suggestion
This document uses vocabulary introduced in <https://github.com/open-telemetry/oteps/pull/91>.
```

text/0001-telemetry-without-manual-instrumentation.md
arminru: Since this is used rather commonly throughout a repo and in my opinion shouldn't be taken as a mistake we could consider disabling this rule.

bogdandrutu: Change config to allow `?` in the headers. I think not allowing `.` is still valuable.

PR #107, opened 2020-05-18
Proposal for sampling.priority

Kanwaldeep:


Kanwaldeep:
I signed it

Kanwaldeep:
I signed it

Kanwaldeep:
We could consider changing the name to make it clear also I was pointed out  by @reyang  that it is not following the ABNF format for the key . As the trace passes service boundaries that are using different algorithms for sampling how do we ensure that the trace that was sampled in by the upstream component is honored. So as the trace is initiated the sampling priority is calculated based on any algorithm and then the other components in the call path use this value to decide whether the trace should be recorded based on it's sampling rate.

Oberon00:
To try clearing up some confusion: Apart from the 0-1000 instead of 0..1 range, when you talk about "priority", did you mean to use the word "probability" instead, @Kanwaldeep?

text/trace/0107-sampling-priority.md
```diff
@@ -0,0 +1,79 @@
...
+class.
+
+This recommendation suggests OpenTelemetry will allow to return an updated
+`tracestate` that will be used for a newly created `Span`. This will allow
+samplers from a single vendor coordinate sampling decisions across the
+different components of the trace.
+
+### Probability Sampler default behavior
+
+The default behavior of probability sampler is to calculate a stateless hash
```
Oberon00, line 46: Note that this is not yet agreed upon, although there are some PRs open namely https://github.com/open-telemetry/opentelemetry-specification/pull/331 which is several months old and just yesterday we got https://github.com/open-telemetry/opentelemetry-specification/pull/611

text/trace/0107-sampling-priority.md
```diff
@@ -0,0 +1,79 @@
...
+function of `trace-id` to make a sampling decision.
+
+It is suggested to extend the sampler behavior to expose a `tracestate` field
+called `sampling.priority` with the integer value `[0-1000]` that will indicate
```
Oberon00, line 50: indicate how? What does a value of 408 mean? 40.8% probability?

text/trace/0107-sampling-priority.md
```diff
@@ -0,0 +1,79 @@
...
+sampling priority of the current span.
+
+This will allow to align sampling algorithms between various components.
+Especially for the transition scenarios from the different SDKs.
```
Oberon00, line 54: Which "transition" scenarios? Switching from an old to a new version? Please clarify.

SergeyKanzhelev:
@Oberon00 name is coming from OpenTracing and is already used in collector: 

https://github.com/open-telemetry/opentelemetry-collector/blob/60b03d0d2d503351501291b30836d2126487a741/processor/samplingprocessor/probabilisticsamplerprocessor/testdata/config.yaml#L10

text/trace/0107-sampling-priority.md
SergeyKanzhelev: yes. The max value need to be agreed on. 0-1000 would work for most services (sampling down to 1 out of 1000). For extremely loaded services may need more, but likely they may opt out of propagating this flag to optimize performance.

Alternative approach is to go 1..100 and allow non-integers.

text/trace/0107-sampling-priority.md
SergeyKanzhelev: yes. If there are different hashing algos used for historical reasons - you can make compoents respect sampling.priority first and then change algo underneath.

SergeyKanzhelev: it may be expanded in a real spec. @Oberon00 does this note clarify things or you think more descriptive tesxt is needed here. If it's very unclear - @Kanwaldeep worth expanding a bit.

Oberon00:
Please write about this in the prior art section. I can't find the connection of the word "priority" as defined in https://www.merriam-webster.com/dictionary/priority to the meaning I assume it has in this OTEP. But following the link behind your link to https://github.com/opentracing/specification/blob/11dd7f8b16739f02761a9541f04e39f27738f525/semantic_conventions.md#span-tags-table the definition of priority seems to be effectively a simple boolean hasPriority:

> If greater than 0, a hint to the Tracer to do its best to capture the trace. If 0, a hint to the trace to not-capture the trace. If absent, the Tracer should use its default sampling mechanism.

(OK, actually a tri-bool, with the three states absent or less than zero: default; zero: don't sample; greater than zero: please sample).

Kanwaldeep:
> (OK, actually a tri-bool, with the three states absent or less than zero: default; zero: don't sample; greater than zero: please sample).

I can add the link to previous art as mentioned. Also I don't think the tri-bool would work as the proposal is to use the sampling.priority to make the sampling decision if passed. If the sampling priority is within the adjusted sampling rate configured for the component. @SergeyKanzhelev let me know if this is what you are thinking as well. If so I can work on updating the proposal.

SergeyKanzhelev:
@Oberon00 most implementation just set a value `>0` to indicate "sampled" decision. This proposal extends it to also propagate the probability of sampling. In other words, even so the decision was to DO sample, as component really wants to sample, it can use this probability instead of a hash of traceID.

Do you know of other prior art and names that can be used? Are you concerned with the previous art as some might have used the attribute with the same name differently before and didn't set probabilistic value?

Oberon00:
I'm mostly concerned that you never define what your "priority" is in this PR. If you clearly define what it is, maybe we can then find a better name as a next step, or maybe we find that priority does fit the definition.

Kanwaldeep:
> I'm mostly concerned that you never define what your "priority" is in this PR. If you clearly define what it is, maybe we can then find a better name as a next step, or maybe we find that priority does fit the definition.

I've updated the proposal with the priority calculation.

jmacd:
I still do not understand why the word "priority" is being used. It seems misleading, and I also do not see why we would create a standard that encourages loss of information due to rounding.

I can see two useful and non-lossy conventions we might adopt:

1. `sampling.probability` with a floating point value in the range 0-1
2. `sampling.adjusted_count` with a floating point value in the range 0-infinity

One is the inverse of the other. The reason why "adjusted_count" (i.e., the inverse probability)( is my preference is that it has a natural interpretation. The adjusted count of a sampled item is the number of items it is representative for in the whole population.  Items generated with a 40.8% probability sample have adjusted counts of 1/0.408 = 2.45098039216.  When you see one of these items, you would count it as 2.45098039216 items. 

It is important not to lose precision and accuracy here because we would like to transform sampled span data into metric data. When we generate a metric count from a 40.8% sampled span, we should count precisely 2.45098039216 items.

MikeGoldsmith:
I like _probability_ over _adjusted_count_ with the bound range between 0 and 1. However, I generally agree these terms are more accurate to what the intention is versus _priority_.

Kanwaldeep:
I agree with sampling.probability instead of priority with a floating point value in the range 0-1. My only concern is if the name would be confusing with this open PR - https://github.com/open-telemetry/opentelemetry-specification/pull/570

jmacd:
It looks like open-telemetry/opentelemetry-specification#570 and this proposal are the same, other than "sampling" vs "sample" as the label key prefix. Maybe we don't need this OTEP, just need to merge 570.

Oberon00:
Except for the in-band propagation via tracestate, which seems pretty important.

jmacd:
OK. I read through this and feel a bit befuddled by the discussion about coordinating sampling. Can we agree that the in-band propagation of sampling probability is independent and orthogonal to the in-band propagation of the sampling policy that was used? I _am_ interested in knowing that my parent was or was-not selected for sampling and the (upper bound on) its probability. I am not interested in head sampling using fixed probability.

The reason I wrote "upper bound on" is that it is easy to imagine a sampling scheme combining head and tail sampling policies. We might use a 10% head sample to lower the overhead of tracing, and then decide which among those 10% of spans should be recorded in a tail sampler. The tail sampler will correctly compute the outcome probability and apply it as a `sampling.probability` attribute, but the probability is expected to be <= 10%. Therefore the value we propagate in-band is an upper-bound on the probability that the parent was recorded.

Kanwaldeep:
Yes agreed that sampling probability is different than upper bound of the sampling policy.

This proposal is about having the first app/component in the distributed trace calculating the probability and propagating the value so the next app/component can honor it. This is to align apps  using different algorithms for Probability Sampler.

jmacd:
Could we make clear why a tracing system would prefer to propagate probability and then use it downstream, as opposed to propagating a decision (true/false) for the downstream system to honor? (I'm not sure I know the answer.)

Kanwaldeep:
The reason we need probability is the components downstream are using different algorithm for sampling. So let's assume  head component A started the trace and sampled in 60% of the traces and then next component B has sampling policy of 30% but is using different sampling algorithm. In this scenario how can we make sure that the 30% of the traces sampled in by component B are subset of the 60% traces sampled in by component A.

text/trace/0107-sampling-priority.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Sampling hints exchange
+
+OpenTelemetry SDK has an infrastructure to expose sampling hints as span attributes today
+via
+[SamplingResult](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#shouldsample)
+class.
+
+This recommendation suggests OpenTelemetry will allow to return an updated
+`tracestate` that will be used for a newly created `Span`. This will allow
```
yurishkuro, line 41: I think this is mixing two different concerns, the API of the sampler result and the serialization format of the context. Not everybody is going to use W3C TraceContext for propagation format, so the mechanism of returning the probability should be somehow separate from the propagation format (i.e. the sampling probability may be recorded in the Span or SpanContext).

text/trace/0107-sampling-priority.md
```diff
@@ -0,0 +1,81 @@
...
+Increasingly more apps participate in distributed tracing. With the 
+standardized wire format for context propagation, there is bigger chance that 
+algorithm for sampling chosen by one app may not match another app. 
+Since coordination of sampling algorithms across many apps not always possible, 
+OpenTelemetry must provide a way to share sampling hints between apps.
+
+Beyond sampling hints infrastructure, it is beneficial for the community to
+agree on standard behavior and hints exposed by probability sampler.
+
+The `sampling.priority` property is used by many vendors for various purposes.
```
yurishkuro, line 24: Did you mean `sampling.probability`?  `sampling.priority` is a standard OpenTracing span tag that can be used to influence/override sampling decisions. In the context of this OTEP "priority" doesn't seem to make sense.

text/trace/0107-sampling-priority.md
yurishkuro: Is this behavior even relevant to this OTEP? I think _whichever_ algorithm is used by the sampler, as long as it can express the result via a "probability" value (not every sampler can, e.g. rate limiting), the actual calculation is not relevant.

text/trace/0107-sampling-priority.md
```diff
@@ -0,0 +1,81 @@
...
+It is suggested to extend the sampler behavior to expose a `tracestate` field
+called `sampling.priority` with the integer value `[0-1000]` that will indicate
+sampling priority of the current span. The sampling.priority is calculated by rounding 
+the output of the hash function by factor of 1000. For example calculated probability 
+of 0.461 the sampling.priority would be 461. 
+
+This will allow to align sampling algorithms between various components.
+Especially for the transition scenarios where components are using different 
+versions of the SDKs with different sampling algorithms. The components would 
+respect the sampling.priority if passed in to make sampling decision.
```
yurishkuro, line 59: Sorry, I don't follow this reasoning. If the framework already has a capability to propagate context, then the sampling decision itself is going to be propagated via span flag. Downstream services do not need to make that sampling decision again. If they do (e.g. they don't trust the caller), then I am not sure how the probability of upstream decision is helpful (unless the trace id hashing algorithms are the same, which as the intro says cannot be guaranteed).

text/trace/0107-sampling-priority.md
```diff
@@ -0,0 +1,81 @@
...
+
+Beyond the transition scenarios, propagation of `sampling.priority` will allow
+to build more sophisticated logic of probability calculation that may be
+customized to be consistent across properties beyond `trace-id`.
```
yurishkuro, line 63: This is very hand-wavy; please either describe these "more sophisticated calculations" or let's not mention them. If we introduce a new feature, it needs to have a concrete use case with well-understood implementation (especially since the project's aspiration goal is to converge implementations onto OTEL SDKs).

text/trace/0107-sampling-priority.md
Kanwaldeep: The proposal is to address the difference in algorithms. If the upstream component's sampling priority is passed to the downstream app, then the downstream app can use that value to make sampling decision instead of re-calculating with a different algorithm. https://github.com/open-telemetry/oteps/pull/107#issuecomment-636279795

text/trace/0107-sampling-priority.md
Kanwaldeep: In the proposal the intent is similar to sampling.priority in OpenTracing. So if the trace is sampled in by first component it passes the priority to next component to make sampling decision.

text/trace/0107-sampling-priority.md
Kanwaldeep: Agreed. We can make changes for this.

jmacd:
I can't re-review this because it still uses "sampling.priority".

Kanwaldeep:
> I can't re-review this because it still uses "sampling.priority".

I have not yet changed the name as it is conflicting with this open PR - https://github.com/open-telemetry/opentelemetry-specification/pull/570. In this PR my understanding is sampling.probability is the configured probability of the sampler. In this spec we want to use the calculated probability for the span.

lizthegrey:
> The reason we need probability is the components downstream are using different algorithm for sampling. So let's assume head component A started the trace and sampled in 60% of the traces and then next component B has sampling policy of 30% but is using different sampling algorithm. In this scenario how can we make sure that the 30% of the traces sampled in by component B are subset of the 60% traces sampled in by component A.

This, and also systems that depend upon dynamically sampling traces to generate metrics rather than a separate metrics system, need to know how many spans one span represents when counting up spans, computing percentiles of latency, etc.

jmacd:
I do not see any disagreement with https://github.com/open-telemetry/opentelemetry-specification/pull/570.  

The calculated probability of the span equals the configured probability of the sampler. If you have another interpretation, please clarify?

lizthegrey:
> I do not see any disagreement with [open-telemetry/opentelemetry-specification#570](https://github.com/open-telemetry/opentelemetry-specification/pull/570).
> 
> The calculated probability of the span equals the configured probability of the sampler. If you have another interpretation, please clarify?

Yup, we're in vigorous agreement.

jmacd:
Note that we are discussing sampling in the metrics API here: https://github.com/open-telemetry/oteps/pull/113

In that proposal, I've encouraged us to use the term "sample_count" to describe inverse probability. I'm willing to accept a "sample_probability" field instead, containing the inverse.

Kanwaldeep:
> I do not see any disagreement with [open-telemetry/opentelemetry-specification#570](https://github.com/open-telemetry/opentelemetry-specification/pull/570).
> 
> The calculated probability of the span equals the configured probability of the sampler. If you have another interpretation, please clarify?

Definitely seems like I am missing something :(. The probability sampler calculates the hash function and checks if it is within the configured probability of the sampler. For example Sampled-in = hash(trace-id) / max_long  < sampling-ratio. So the output of hash(trace-id)/ max_long is what we want to propagate. My understanding is the sampling-ratio is the configured probability of the sampler.

jmacd:
The value of `hash(trace-id) / max_long` is a presumably a uniform random number in the range [0,1] (assuming a perfect hash function). The equation `hash(trace-id) / max_long < sampling_probability` will be true with probability `sampling_probability`.  If you consider one span, the expected number of spans that will be recorded equals `sampling_probability`, so with 100% sampling you expect 1 output sample span per arriving span. With 50% sampling you expect 0.5 sample spans per arriving span.

If you consider `1 / sampling_probability` arriving spans, the expected number of sample spans equals `sampling_probability * 1 / samping_probability == 1`, and that is why when we see one sample span we want to know its `sampling_probability` (or the inverse).

After the sampling decision is made, the value of `hash(trace-id) / max_long` is not relevant, `sampling_probability` (or the inverse) is what matters to the downstream system.

Kanwaldeep:
So if you only pass sampling_probability say 0.6 then it's not clear to me how the following scenario can be handled.

So let's assume head component A started the trace and sampled in 60% of the traces and then next component B has sampling policy of 30% but is using different sampling algorithm. In this scenario how can we make sure that the 30% of the traces sampled in by component B are subset of the 60% traces sampled in by component A.

The idea of the proposal is Component B checks the passed probability and makes the decision. So if the passed value is 0.6 then Component B would not sample it in. But if the output of the hash function in component A was 0.25 and that value was passed to Component B then it would sample this trace in as it is using the calculated probability by A to make the decision.

jmacd:
Now I understand the meaning of "priority", I apologize for taking so long to get to this point.

There are a lot of ways to organize a sampling scheme. The ones I'm most familiar with allow the root in a trace to make the sampling decision, and all the children will respect it. In this case, the children don't need to know the probability.

Your example describes a coordinated sampling scheme, where it is important to know _how_ the decision was made, for the purpose of coordination. Sending the proposed "priority" (hash(trace_id) / max_long) is one way to achieve that, but then all uses of this value have to follow this one scheme, right?

I believe elsewhere @yurishkuro has described how Jaeger conveys this information as two attributes: (1) the sampling probability, (2) the sampling technique. In your example, we could transmit the `sample.probability` as a value in the range [0,1] along with a description of the coordination scheme you've used, which in this case means a descriptor for "consistent hashing of trace_id". For example:
```
sample.probability=0.5
sample.scheme=cityhash(traceid)
```

The recipient must know the scheme, and then can apply the same hash to compute the priority value that you're interested in.

Kanwaldeep:
> Your example describes a coordinated sampling scheme, where it is important to know _how_ the decision was made, for the purpose of coordination. Sending the proposed "priority" (hash(trace_id) / max_long) is one way to achieve that, but then all uses of this value have to follow this one scheme, right?

They could configure not to use the passed in sampling.priority. But yes all the components use the output of the algorithm as calculated by the root span.

Kanwaldeep:
> The recipient must know the scheme, and then can apply the same hash to compute the priority value that you're interested in.

I will wait for  @yurishkuro  for more details on this. But passing the sample.scheme would require all the downstream components to handle implementation for various schemes to calculate the probability. My 2 cents: Passing sampling.priority seems much more efficient for this use case.

jmacd:
It looks like Jaeger uses another meaning for the word "priority", see here: https://github.com/open-telemetry/oteps/pull/115#discussion_r439100765

@yurishkuro there is a question for you above.

text/trace/0107-sampling-priority.md
yurishkuro: I don't think it's similar. In OpenTracing the word "priority" is a hint from the user. Here it's an output from the sampler.

yurishkuro:
Not sure what the question is. Jaeger does not pass sampling parameters downstream. The decision is made once, represented by a single bit in the flags, and respected by all children. The `sampler.type` and `sampler.param` tags are reported with the span out of band.

jmacd:
Thanks @yurishkuro, sorry it wasn't a proper question. I couldn't find information about `sampler.type` and `sampler.param`. 

@Kanwaldeep I wonder if this use of a `tracestate` attribute even needs to be standardized. The system you are describing will have to be aware of this on either end, right? At the very least, I believe the word "priority" has caused considerable confusion and I wonder if there might be a better choice.

Kanwaldeep:
> @Kanwaldeep I wonder if this use of a `tracestate` attribute even needs to be standardized. The system you are describing will have to be aware of this on either end, right? At the very least, I believe the word "priority" has caused considerable confusion and I wonder if there might be a better choice.

I was thinking sampling.probability.output as this would make it clear. By making this a standard we make it easier for teams to adopt OTel SDK if they are already using other SDKs.

Kanwaldeep:
Based on discussion in Sampling WG meeting we are working on PR for:
- Sampler should be able to add/modify keys on tracestate.
- Plugins to have access tracestate. This way, new samplers are not blocked by having to agree on the meanings of common terms like “priority.”

Oberon00:
Hi! What are "plugins" here? You mean tracestate should be publicly accessible? I think it is.

ericmustin:
👋 Hope all is well. Just popping in from otel-js to echo what @lizthegrey offered here https://github.com/open-telemetry/oteps/pull/107#issuecomment-641385424.  This PR's approach, or something like [otep-specification #570](https://github.com/open-telemetry/opentelemetry-specification/pull/570) would be tremendously helpful for vendor specific exporters/systems which rely on the sampled traces/spans themselves to generate metrics (like hits/latency/etc) rather than a separate metrics system.  I won't pretend to have enough knowledge about the Spec to say whether this info should live on `span.attribute` or `span.context.traceState`, but just that having enough information to infer a span's `sample.count/weight/etc` at the exporter level would be quite useful for many practical use cases.

Oberon00:
>  I won't pretend to have enough knowledge about the Spec to say whether this info should live on span.attribute or span.context.traceState

This is the crucial difference though. Samplers can already set normal attributes today. Is this enough for you? The key difference is that normal attributes are not propagated in-band, while traceState is (e.g. in the W3C tracestate HTTP header).

ericmustin:
> This is the crucial difference though. Samplers can already set normal attributes today. Is this enough for you? The key difference is that normal attributes are not propagated in-band, while traceState is (e.g. in the W3C tracestate HTTP header).

@Oberon00 In-band means, within the same process right? Basically, would that attribute be propagated from the parent to a non-root span? It seems like at the moment (at least in practice) that is not the case, and what [PR #570](https://github.com/open-telemetry/opentelemetry-specification/pull/570) is trying to address, no? 

 -  https://github.com/open-telemetry/opentelemetry-specification/blame/efb73763c63ff5a2e14e7f30735509f3139c82bd/specification/trace/sdk.md#L122

In-Band (if i understand the term correctly) propagation is the main issue for the use case I have in mind, since the idea is that currently the default `ProbabilitySampler` leaves exporters blind to the sampling decision on any downstream spans they receive.

Oberon00:
> In-band means, within the same process

No, it means in the same communication "band" that the process would also use without instrumentation to communicate with other processes. E.g. if a process normally makes an HTTP request to some other service, with in-band propagation, headers would be injected into that HTTP request.

ericmustin:
@Oberon00 ah ok gotcha, thank you for the clarification. So, at the most basic level, what matters more _for me at least_, is in-process propagation. This seems more reasonably accomplished by [PR #570](https://github.com/open-telemetry/opentelemetry-specification/pull/570). Does that help clarify things? Hmm TL;DR I should've commented on [PR #570](https://github.com/open-telemetry/opentelemetry-specification/pull/570) instead, but i felt the the use case described in the comments here was worth highlighting.

text/trace/0107-sampling-priority.md
lizthegrey: Large scale systems are _especially_ in need of having one trace stand for many traces; thus I worry about artificially capping at 1000 sample rate.

text/trace/0107-sampling-priority.md
lizthegrey: Agree, that the discussion in https://github.com/open-telemetry/opentelemetry-specification/pull/611 has people split between using the "random bits" of the UUID from traceid generation, vs. hashing the whole traceid again just in case the trace id was not randomly generated.

PR #108, opened 2020-05-21
Metric naming conventions

tedpennings:
This proposal follows the discussion we've been having in the metrics SIG.

There are TODOs in the doc. Many sections require expertise in a specific domain; I'm very eager for discussion and suggestions.

This is written from my perspective as a UI/product engineer who consumes and presents metric data. I think is a bit different from the perspective of an instrumentation engineer or data platform engineer. Feedback extra extra welcome from y'all in those roles, and submitted with the caveat that there will be impedances in this regard.


tedpennings:
Meta question: is the usage of `ie` and `eg` awkward or academic?

tedpennings:
I think we will want to include the measures / data types for the metrics prescribed in this document. Can someone take a pass at that? I am not as familiar with the nomenclature and options in this regard 😳

text/metrics/0001-naming-conventions.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Purpose
+
+Metric names and labels are the primary read-interface to metric data. The names and taxonomy need to be understandable and discoverable during routine exploration -- and this becomes critical during incidents.
+
+## Guidelines
+
+Namespace similar metrics together. Define top-level namespaces for common metric categories: for OS, like CPU and network; for app runtimes, like the JVM. This aids in discovery and adhoc comparison
+
+Provide consistent names for common label. Provides discoverability and disambiguation similar to metric names.
```
jkwatson, line 11:
```suggestion
Provide consistent names for common labels. Provides discoverability and disambiguation similar to metric names.
```

tedpennings:
Prior work:
* [Excellent spreadsheet of OS metrics](https://docs.google.com/spreadsheets/d/11qSmzD9e7PnzaJPYRFdkkKbjTLrAKmvyQpjBjpJsR2s/edit#gid=0) -- I think this came from Splunk?
* [Spreadsheet that I drafted](https://docs.google.com/spreadsheets/d/1WlStcUe2eQoN1y_UF7TOd6Sw7aV_U0lFcLk5kBNxPsY/edit#gid=0) a few weeks ago that's largely similar. Lots of discussion questions. No labels in this spreadsheet, which is not what we want

text/metrics/0001-naming-conventions.md
```diff
@@ -0,0 +1,137 @@
...
+
+It is recommended that per-core labels should not be reported by default to reduce cardinality; a user should opt-in with via configuration.
+
+#### Network
+
+`system.network` is the core metric name.
+
+Standard labels include, non-exhaustively,
+* `network.sent`
+* `network.received`
```
jkwatson, line 54: I don't think these should be labels. Does it makes sense to have an aggregate across both of these labels? I don't think that looking at the "Net network traffic" makes much sense. The instrument names should be, I think, `system.network.sent` and `system.network.received`.

text/metrics/0001-naming-conventions.md
```diff
@@ -0,0 +1,137 @@
+# Metric naming conventions
```
jkwatson, line 1: Should this be "Metric instrument naming conventions" since the thing that instrumentation will do is create instruments, not metrics?

jkwatson:
Before we get to specific names, I'd like to see a meta-OTEP with the guiding principles for instrument naming and measurement labeling, so we can all agree to that before we start digging into individual names. This PR briefly mentions some guiding principles, and then dives into system metrics. Separating those two concerns into separate OTEPs will, I think, make it easier to discuss them in isolation, rather than muddling the discussion of general guidelines with specifics.

Thoughts?

tedpennings:
@jkwatson that's a great point re generic conventions vs specific recommendations. I found it very difficult to talk about the generic conventions without examples, which is what brought me down this path. I will give it some thought. 

Ultimately I think we will want a single document -- or extensive examples in the primary document with an appendix for further details.

There's a lot to discuss with both topics, so it would definitely be easier to discuss one thing at a time.

jkwatson:
That's fair. having examples that are outside of system metrics would be helpful, then. Things like http calls, database calls, and rpc calls will help add depth to the discussion, I think.

text/metrics/0001-naming-conventions.md
bogdandrutu: We already discuss this at long, and we concluded that it does make sense to have a Sum of these value to get all the network traffic.

text/metrics/0001-naming-conventions.md
```diff
@@ -0,0 +1,147 @@
...
+Standard usage labels include, non-exhaustively,
+* `cpu.idle`
+* `cpu.user`
+* `cpu.sys`
+* `cpu.real`
+* `cpu.iowait`
+* `cpu.nice`
+* `cpu.interrupt`
+* `cpu.softirq`
+* `cpu.steal`
```
bogdandrutu, line 35: I am confused here, because a label has a key and a value and you are proposing only `values` (or I am just super confused)?

text/metrics/0001-naming-conventions.md
```diff
@@ -0,0 +1,147 @@
...
+* `runtime.java.heap_alloc`
+* `runtime.java.gc` with labels `gc.count` and `gc.time`
+
+#### Node
+
+All Node runtime metrics should be within the `runtime.node.*` namespace.
+
+* `runtime.node.gc` with labels `pause.time` and `pause.count`
+* `runtime.node.heap_alloc` with labels `heap.total_size`, `heap.available_size`, `heap.used_heap_size` **TODO** confirm
+* `runtime.node.event_loop` with TBD labels **TODO**
```
bogdandrutu, line 127: Please use `nodejs` :)) `Node` is too generic and may confuse with kubernets.node :))

jrcamp:
> Prior work:
> 
> * [Excellent spreadsheet of OS metrics](https://docs.google.com/spreadsheets/d/11qSmzD9e7PnzaJPYRFdkkKbjTLrAKmvyQpjBjpJsR2s/edit#gid=0) -- I think this came from Splunk?

@james-bebbington from Google gets the credit for the awesome spreadsheet but we (Splunk) have been working with him on defining the set of host metrics.

bogdandrutu:
@tedpennings I think we should do what @jkwatson proposed which is to define how the naming structure looks like for metrics and for labels.

text/metrics/0001-naming-conventions.md
tedpennings: 63443ef19282b4764b130eece22fd2082f635780

text/metrics/0001-naming-conventions.md
tedpennings: Just the label keys, sorry if that was unclear.

text/metrics/0001-naming-conventions.md
```diff
@@ -0,0 +1,149 @@
...
+Provide consistent names for common label. Provides discoverability and disambiguation similar to metric names.
+
+Avoid semantic ambiguity. Use namespaced metric names in cases where similar metrics have significantly different implementations across the holistics of metrics. For example, every garbage collected runtime has a slightly different strategies and measures. Using common metric names for GC, not namespaces by the runtime, could create dissimilar comparisons and confusion for end users. Measures of operating system memory are similar.
+
+## Conventions
+
+All metrics have a limited set of common labels:
+* `system.host`
+
+### Operation System
```
quentinmit, line 20: We should be careful to distinguish OS-specific and OS-agnostic metrics and labels. For example, memory metrics might have different label values depending on OS - Linux will have a different set of types of memory (private, slab, etc.) than Windows. So maybe we could add a generic statement - "standard metric names and labels should be OS-agnostic".

text/metrics/0001-naming-conventions.md
quentinmit: "cpu.sys" is a combination of key and value, not just a key.

text/metrics/0001-naming-conventions.md
```diff
@@ -0,0 +1,149 @@
...
+
+#### Memory
+
+`system.memory` is the core metric name.
+
+Standard labels include, non-exhaustively,
+* `memory.free`
+* `memory.resident`
+* `memory.shared`
+* `memory.private`
```
quentinmit, line 75: This is a bad example because private and resident overlap; they shouldn't both be in the same metric.

text/metrics/0001-naming-conventions.md
```diff
@@ -0,0 +1,149 @@
...
+All runtime metric names should be reported with a namespace that includes the name of the runtime, eg, `runtime.go.*`.
+
+#### Go
+
+All Go runtime metrics should be within the `runtime.go.*` namespace.
+
+Common metrics include:
+* `runtime.go.goroutines`
+* `runtime.go.heap_alloc`
+* `runtime.go.gc` with labels `gc.count`
```
quentinmit, line 111: What is the meaning of "gc.count"? The number of objects found? The number of GC cycles?

text/metrics/0001-naming-conventions.md
```diff
@@ -0,0 +1,149 @@
...
+
+#### Java
+
+All Java runtime metrics should be within the `runtime.java.*` namespace.
+
+* `runtime.java.threads` with optional labels for thread pools, eg, `runtime.java.thread_pool.[name]`
+* `runtime.java.heap_alloc`
+* `runtime.java.gc` with labels `gc.count` and `gc.time`
```
quentinmit, line 119: `count` and `time` have different units. Does it really make sense for both to be in one metric?

text/metrics/0001-naming-conventions.md
tedpennings: `cycles` -- will update.

object counts can be relevant in some runtimes, too. (I believe python does this with generational object counts as a diagnostic for GC.)

text/metrics/0001-naming-conventions.md
tedpennings: Yeah, actually I intended it that way because it introduces this complexity. I think we need to consider how we want to describe this in a non-overlapping way, and how we want to provide the more granular data elsewhere.

text/metrics/0001-naming-conventions.md
quentinmit: I think I don't follow the purpose of this label. If the metric is reporting the number of GC cycles that have happened, why does it have a label at all? Why isn't the metric just "runtime.go.gc.cycle_count" with no labels?

text/metrics/0001-naming-conventions.md
quentinmit: I think every time there are two quantities that don't make sense to add together those quantities should be separate metrics; that means that, in this example, "resident", "shared", and "private" should all be separate metrics (I think it's possible for memory to be in two of the three categories simultaneously on Linux).

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,151 @@
...
+Provide consistent names for common label. Provides discoverability and disambiguation similar to metric names.
+
+"As a rule of thumb, [aggregations] over all the dimensions of a given metric should be meaningful," as Prometheus recommends.
+
+Avoid semantic ambiguity. Use namespaced metric names in cases where similar metrics have significantly different implementations across the holistics of metrics. For example, every garbage collected runtime has a slightly different strategies and measures. Using common metric names for GC, not namespaces by the runtime, could create dissimilar comparisons and confusion for end users. Measures of operating system memory are similar.
+
+## Conventions
+
+All metrics have a limited set of common labels:
+* `system.host`
```
tigrannajaryan, line 20: Unclear why `system.host` is singled out. Is there something special about it or this is just an example?

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,151 @@
...
+Standard usage labels include, non-exhaustively,
+* `cpu.idle`
+* `cpu.user`
+* `cpu.sys`
+* `cpu.real`
+* `cpu.iowait`
+* `cpu.nice`
+* `cpu.interrupt`
+* `cpu.softirq`
+* `cpu.steal`
```
tigrannajaryan, line 37: Aren't labels a key/value pair? Which is the key and which is the value here?

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,151 @@
...
+All runtime metric names should be reported with a namespace that includes the name of the runtime, eg, `runtime.go.*`.
+
+#### Go
+
+All Go runtime metrics should be within the `runtime.go.*` namespace.
+
+Common metrics include:
+* `runtime.go.goroutines`
+* `runtime.go.heap_alloc`
+* `runtime.go.gc` with labels `gc.count`
```
tigrannajaryan, line 113: What does `runtime.go.gc` measure? Is it a duration? What does `gc.count` label do?

jrcamp:
> @jkwatson that's a great point re generic conventions vs specific recommendations. I found it very difficult to talk about the generic conventions without examples, which is what brought me down this path. I will give it some thought.
> 
> Ultimately I think we will want a single document -- or extensive examples in the primary document with an appendix for further details.
> 
> There's a lot to discuss with both topics, so it would definitely be easier to discuss one thing at a time.

I agree that I think we need to split it up. Host metrics are a topic in and of itself. I think examples are useful but I don't think we should have the meta discussion about metrics while also trying to design host metrics at the same time.

Some of the meta topics to agree on are:

1. What is the format of a metric? How are they namespaced? Do they have a relation to semantic convention types? Are they foo.bar or foo/bar?
2. What is part of a metric name vs a label?
3. What determines what labels are part of the resource and which are part of the metric?
4. Should there be guidelines on how metric descriptions are written?

I think it's fine to makeup some examples using common things like CPU and network to discuss the above but I think they should just be for demonstration purpose and as part of that not be trying to define that those are the exact metric names and labels that will be used.

tedpennings:
Thanks for all of the feedback, and discussion at the working group meetings. Substantial revisions have been made. Guidance on specific metrics has been removed (for a future doc). 

There are a few areas that are left open for discussion -- that discussion is very welcome. @jrcamp, you raise some, and I'd love to any copy or outlines of content you'd like to contribute to the document.

Please re-review and provide your feedback -- thank you! 🙌  @jrcamp @jkwatson @tigrannajaryan @quentinmit @bogdandrutu

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,52 @@
...
+
+Provide consistent names for common labels. This aids in discoverability and disambiguates similar labels to metric names.
+
+"As a rule of thumb, **aggregations** over all the dimensions of a given metric should be meaningful," as Prometheus recommends.
+
+Avoid semantic ambiguity. Use prefixed metric names in cases where similar metrics have significantly different implementations across the breadth of all existing metrics. For example, every garbage collected runtime has a slightly different strategies and measures. Using common metric names for GC, not namespaced by the runtime, could create dissimilar comparisons and confusion for end users. (For example, prefer `runtime.java.gc*` over `runtime.gc*`.) Measures of many operating system metrics are similar.
+
+For conventional metrics or metrics that have their units covered by OpenTelemetry metadata (eg `Metric.WithUnit` in Go), do not include the units in the metric name. Units may be included when it provides additional meaning to the metric name. Metrics must, above all, be understandable and usable.
+
+## End of document
```
tedpennings, line 23: The below is for discussion. It's easier to have threaded discussion with placeholder points from the main content.

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,56 @@
...
+- one common label name or metric-specific label names, eg, for CPU metric, **kind**:idle or **cpu**:idle
+
+* What about things that overlap with tracing span data like upstream/downstream callers or originating systems?
+
+## Bike shed
+
+* Separators
+  * namespace separators, eg, `runtime.go`
+  * word-token separtors inside a metric name, eg, `heap_alloc`
+* Fixed gear? 10 speed?
```
jkwatson, line 54: Belt drive with an IGH

tedpennings:
ps: ack to markdownlint. will fix once it's in its final draft form.

text/metrics/0108-naming-conventions.md
tedpennings: Resolving now -- this will be covered in a future doc

text/metrics/0001-naming-conventions.md
tedpennings: Resolving now -- specifics about metric names and labels will be covered in a future doc

text/metrics/0001-naming-conventions.md
tedpennings: Resolving now -- specifics about metric names and labels will be covered in a future doc

This is definitely a good point and I will circle back to you when it comes time to answer this question. It's a hard problem, I think.

text/metrics/0001-naming-conventions.md
tedpennings: Resolving now -- specifics about metric names and labels will be covered in a future doc

text/metrics/0108-naming-conventions.md
tedpennings: Resolving now -- specifics about metric names and labels will be covered in a future doc. Either `cpu` or `kind` would be the label name. Do you want this doc to cover that, or should we save it for the doc with the list of specific metrics?

text/metrics/0001-naming-conventions.md
tedpennings: Good question. I'm going to save this for the doc that lists specific metrics. Resolving for now.

tedpennings:
I resolved some comments that reflect our desire to keep this document general purpose, focused on naming conventions. Questions about specific names and labels should be handled in the next doc that details common metrics, like OS/Host metrics. We want to keep this about conventions and best practices, rather than implementation.

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,55 @@
...
+Labels, key:value
+- one common label name or metric-specific label names, eg, for CPU metric, **kind**:idle or **cpu**:idle
+
+* What about things that overlap with tracing span data like upstream/downstream callers or originating systems?
+
+## Bike shed
+
+* Separators
+  * namespace separators, eg, `runtime.go`
+  * word-token separtors inside a metric name, eg, `heap_alloc`
```
jkwatson, line 52: My vote, FWIW, is dot-separate the namespace from the name, and underscore the rest. 

But, I don't feel strongly about this. What I do feel strongly about is that there is an obvious programmatic way to map between what we decide and other schemes that might be needed for various backends.  If we were to decide (for instance) to double-underscore rather than dots for separating the namespace, that would be totally a-ok with me.

tedpennings: I'm with you. `runtime.go.goroutines` and `runtime.go.heap_alloc`. Names are merely examples, but it's the dots and the underscores.

text/metrics/0001-naming-conventions.md
jkwatson: I'll trust that the network monitoring experts are fine with this. Seems like a weird thing to sum together, but /shrug

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,55 @@
+# Metric instrument naming conventions
+
+## Purpose
+
+Names and labels for metric instruments are primarily how humans interact with metric data -- users rely on these names to build dashboards and perform analysis. The names and hierarchical structure need to be understandable and discoverable during routine exploration -- and this becomes critical during incidents.
```
MrAlias, line 5: I think it would help to clarify also that this doc is intended to act as a framework for subsequent semantic convention PRs.

Maybe something like:

```suggestion
Names and labels for metric instruments are primarily how humans interact with metric data -- users rely on these names to build dashboards and perform analysis. The names and hierarchical structure need to be understandable and discoverable during routine exploration -- and this becomes critical during incidents.

To ensure these goals and consistency in future metric naming standards, this outlines a meta-standard for these names.
```

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,55 @@
...
+
+Provide consistent names for common labels. This aids in discoverability and disambiguates similar labels to metric names.
+
+"As a rule of thumb, **aggregations** over all the dimensions of a given metric should be meaningful," as Prometheus recommends.
+
+Avoid semantic ambiguity. Use prefixed metric names in cases where similar metrics have significantly different implementations across the breadth of all existing metrics. For example, every garbage collected runtime has a slightly different strategies and measures. Using common metric names for GC, not namespaced by the runtime, could create dissimilar comparisons and confusion for end users. (For example, prefer `runtime.java.gc*` over `runtime.gc*`.) Measures of many operating system metrics are similar.
+
+For conventional metrics or metrics that have their units covered by OpenTelemetry metadata (eg `Metric.WithUnit` in Go), do not include the units in the metric name. Units may be included when it provides additional meaning to the metric name. Metrics must, above all, be understandable and usable.
```
MrAlias, line 21: This should more explicitly state its normative guidance:

```suggestion
For conventional metrics or metrics that have their units included in OpenTelemetry metadata (eg `metric.WithUnit` in Go), SHOULD NOT include the units in the metric name. Units may be included when it provides additional meaning to the metric name. Metrics MUST, above all, be understandable and usable.
```

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,55 @@
...
+
+The below text is for threaded discussion and will be deleted when ready to merge.
+
+Goal heading out of this document: when this merges, start writing specific semantic conventions for os, database, etc. Forthcoming docs will address specific metric: Network, OS, OS-agnostic
+
+
+
+## TODO TODO Working Area and Outstanding Questions TODO TODO
+
+### Fundamental questions - do we need to address this?
```
MrAlias, line 33: > do we need to address this?

I don't think these need to be addressed, at least not here.

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,55 @@
...
+
+The hierarchical structure of metrics defines the namespacing. Supporting OpenTelemetry artifacts define the metric structures and hierarchies for some categories of metrics, and these can assist decisions when creating future metrics.
+
+Provide consistent names for common labels. This aids in discoverability and disambiguates similar labels to metric names.
+
+"As a rule of thumb, **aggregations** over all the dimensions of a given metric should be meaningful," as Prometheus recommends.
+
+Avoid semantic ambiguity. Use prefixed metric names in cases where similar metrics have significantly different implementations across the breadth of all existing metrics. For example, every garbage collected runtime has a slightly different strategies and measures. Using common metric names for GC, not namespaced by the runtime, could create dissimilar comparisons and confusion for end users. (For example, prefer `runtime.java.gc*` over `runtime.gc*`.) Measures of many operating system metrics are similar.
```
MrAlias, line 19:
```suggestion
Semantic ambiguity SHOULD be avoided. Use prefixed metric names in cases where similar metrics have significantly different implementations across the breadth of all existing metrics. For example, every garbage collected runtime has a slightly different strategies and measures. Using common metric names for GC, not namespaced by the runtime, could create dissimilar comparisons and confusion for end users. (For example, prefer `runtime.java.gc*` over `runtime.gc*`.) Measures of many operating system metrics are similar.
```

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,55 @@
...
+Metric names and labels exist within a single universe and a single hierarchy. Metric names and labels should be considered within the universe of all existing metric names. When defining new metric names and labels, consider the prior art of existing standard metrics and metrics from frameworks/libraries.
+
+Nest similar metrics together in a hierarchy based on their usage. Define a top-level hierarchy for common metric categories: for OS metrics, like CPU and network; for app runtimes, like GC internals. Libraries and frameworks should nest their metrics into a hierarchy as well. This aids in discovery and adhoc comparison. This allows a user to find similar metrics given a certain metric.
+
+The hierarchical structure of metrics defines the namespacing. Supporting OpenTelemetry artifacts define the metric structures and hierarchies for some categories of metrics, and these can assist decisions when creating future metrics.
+
+Provide consistent names for common labels. This aids in discoverability and disambiguates similar labels to metric names.
+
+"As a rule of thumb, **aggregations** over all the dimensions of a given metric should be meaningful," as Prometheus recommends.
```
MrAlias, line 17:
```suggestion
["As a rule of thumb, **aggregations** over all the dimensions of a given metric **SHOULD** be meaningful,"](https://prometheus.io/docs/practices/naming/#metric-names) as Prometheus recommends.
```

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Guidelines
+
+Metric names and labels exist within a single universe and a single hierarchy. Metric names and labels should be considered within the universe of all existing metric names. When defining new metric names and labels, consider the prior art of existing standard metrics and metrics from frameworks/libraries.
+
+Nest similar metrics together in a hierarchy based on their usage. Define a top-level hierarchy for common metric categories: for OS metrics, like CPU and network; for app runtimes, like GC internals. Libraries and frameworks should nest their metrics into a hierarchy as well. This aids in discovery and adhoc comparison. This allows a user to find similar metrics given a certain metric.
+
+The hierarchical structure of metrics defines the namespacing. Supporting OpenTelemetry artifacts define the metric structures and hierarchies for some categories of metrics, and these can assist decisions when creating future metrics.
+
+Provide consistent names for common labels. This aids in discoverability and disambiguates similar labels to metric names.
```
MrAlias, line 15:
```suggestion
Common labels SHOULD be consistently named. This aids in discoverability and disambiguates similar labels to metric names.
```

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Purpose
+
+Names and labels for metric instruments are primarily how humans interact with metric data -- users rely on these names to build dashboards and perform analysis. The names and hierarchical structure need to be understandable and discoverable during routine exploration -- and this becomes critical during incidents.
+
+## Guidelines
+
+Metric names and labels exist within a single universe and a single hierarchy. Metric names and labels should be considered within the universe of all existing metric names. When defining new metric names and labels, consider the prior art of existing standard metrics and metrics from frameworks/libraries.
+
+Nest similar metrics together in a hierarchy based on their usage. Define a top-level hierarchy for common metric categories: for OS metrics, like CPU and network; for app runtimes, like GC internals. Libraries and frameworks should nest their metrics into a hierarchy as well. This aids in discovery and adhoc comparison. This allows a user to find similar metrics given a certain metric.
```
MrAlias, line 11:
```suggestion
Associated metrics SHOULD be nested together in a hierarchy based on their usage. Define a top-level hierarchy for common metric categories: for OS metrics, like CPU and network; for app runtimes, like GC internals. Libraries and frameworks should nest their metrics into a hierarchy as well. This aids in discovery and adhoc comparison. This allows a user to find similar metrics given a certain metric.
```

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,55 @@
+# Metric instrument naming conventions
+
+## Purpose
+
+Names and labels for metric instruments are primarily how humans interact with metric data -- users rely on these names to build dashboards and perform analysis. The names and hierarchical structure need to be understandable and discoverable during routine exploration -- and this becomes critical during incidents.
+
+## Guidelines
+
+Metric names and labels exist within a single universe and a single hierarchy. Metric names and labels should be considered within the universe of all existing metric names. When defining new metric names and labels, consider the prior art of existing standard metrics and metrics from frameworks/libraries.
```
MrAlias, line 9:
```suggestion
Metric names and labels exist within a single universe and a single hierarchy. Metric names and labels MUST be considered within the universe of all existing metric names. When defining new metric names and labels, consider the prior art of existing standard metrics and metrics from frameworks/libraries.
```

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,55 @@
...
+
+## TODO TODO Working Area and Outstanding Questions TODO TODO
+
+### Fundamental questions - do we need to address this?
+* What is a metric?
+* What separates a metric from a label?
+
+### General questions
+
+~Instrumentation library != namespace~ (see note above)
```
dyladan, line 39: It is not clear which node is referred to as `namespace` is rather ambiguous here and nothing else mentions instrumentation libraries.

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,55 @@
...
+
+Labels, key:value
+- one common label name or metric-specific label names, eg, for CPU metric, **kind**:idle or **cpu**:idle
+
+* What about things that overlap with tracing span data like upstream/downstream callers or originating systems?
+
+## Bike shed
+
+* Separators
+  * namespace separators, eg, `runtime.go`
```
dyladan, line 51: We already have a namespace which is the instrumentation library name. Use of namespace here to mean something different is confusing.

text/metrics/0108-naming-conventions.md
dyladan, line 19: Use of the term `namespace` here is confusing for the same reason.

tedpennings: Great point -- fixed in f2c990e ("divided by the runtime")

tedpennings:
I incorporated all of @MrAlias's excellent suggestions, and I think this PR now reflects the discussion we've had. I'd like to everyone to take another look for final review.

tedpennings:
PS I followed the steps for cla/linuxfoundation but it's still failing. I used my @lightstep.com email (tpennings@). I think it should just be whitelisted for the domain? Happy to do whatever to fix it -- please let me know -- maybe @jmacd or @tedsuo would know?

tedpennings:
I signed it

text/metrics/0001-naming-conventions.md
tedpennings: Removed specific metric naming guidance from this doc -- this will be handled in a future otep.

text/metrics/0001-naming-conventions.md
tedpennings: Removed specific metric naming guidance from this doc -- this will be handled in a future otep.

text/metrics/0108-naming-conventions.md
tedpennings: This came up a couple times in the SIG and there was no dissent. Marking as resolved and moving forward.

text/metrics/0108-naming-conventions.md
tedpennings: Revised, thanks for pointing this out!

tedpennings: This was from a discussion section of the doc and isn't part of the final doc. Resolving.

text/metrics/0108-naming-conventions.md
tedpennings: This was from a discussion section of the doc and is not part of the final doc. Resolving.

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Guidelines
+
+Namespace similar metrics together. Define top-level namespaces for common metric categories: for OS, like CPU and network; for app runtimes, like the JVM. This aids in discovery and adhoc comparison
+
+Provide consistent names for common label. Provides discoverability and disambiguation similar to metric names.
+
+"As a rule of thumb, [aggregations] over all the dimensions of a given metric should be meaningful," as Prometheus recommends.
+
+Avoid semantic ambiguity. Use namespaced metric names in cases where similar metrics have significantly different implementations across the holistics of metrics. For example, every garbage collected runtime has a slightly different strategies and measures. Using common metric names for GC, not namespaces by the runtime, could create dissimilar comparisons and confusion for end users. Measures of operating system memory are similar.
```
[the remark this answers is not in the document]
tedpennings: Revised, thanks!

andrewhsu:
I see opentelemetry community has switched to `EasyCLA` https://github.com/open-telemetry/community/pull/373

And this PR currently has `EasyCLA` checks green. Can this merge even though `cla/linuxfoundation` PR checks are red? From my personal experience, older `cla/linuxfoundation` checks are a pain to setup.

jmacd:
Metrics approvers, please review.

jmacd:
I wonder if we should say something about plural vs singular names. There is a related discussion here: https://github.com/open-telemetry/opentelemetry-specification/pull/657

jmacd:
See this extremely relevant spec issue: https://github.com/open-telemetry/opentelemetry-specification/pull/657

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,23 @@
...
+
+The hierarchical structure of metrics defines the namespacing. Supporting OpenTelemetry artifacts define the metric structures and hierarchies for some categories of metrics, and these can assist decisions when creating future metrics.
+
+Common labels SHOULD be consistently named. This aids in discoverability and disambiguates similar labels to metric names.
+
+["As a rule of thumb, **aggregations** over all the dimensions of a given metric **SHOULD** be meaningful,"](https://prometheus.io/docs/practices/naming/#metric-names) as Prometheus recommends.
+
+Semantic ambiguity SHOULD be avoided. Use prefixed metric names in cases where similar metrics have significantly different implementations across the breadth of all existing metrics. For example, every garbage collected runtime has slightly different strategies and measures. Using a single set of metric names for GC, not divided by the runtime, could create dissimilar comparisons and confusion for end users. (For example, prefer `runtime.java.gc*` over `runtime.gc.*`.) Measures of many operating system metrics are similar.
```
aabmass, line 21: I think this makes sense semantically but creates an issue in discoverability. I could see myself searching for e.g. [load average](https://en.wikipedia.org/wiki/Load_(computing)) and not knowing whether to look at `system.cpu.load_average` or `system.unix.cpu.load_average` or `system.linux.cpu.load_average` for example. The convention of "use prefixed metric names in cases where similar metrics have significantly different implementations across the breadth of all existing metrics" makes a lot of sense when creating metrics. When discovering them though, it requires the user to know the breadth of all existing metrics to determine if what they're looking for is prefixed by the OS/runtime or not 😨. I guess there is a trade-off between discoverability and ambiguity?

jmacd: My hopes are pinned on autocompletion support to help users discover metrics.

jmacd:
@bogdandrutu I'm not sure if you were planning to apply the per-signal approvers changes in this repository? Either way, this has enough approvals from the metrics approvers and could be merged.

aabmass:
Was there any discussion around versioning in names? I suppose this OTEP covers it implicitly since you could prefix everything, like `v1.*`, but is this a common enough need to add guidance on?

This could be important for any backends that have a strict schema for metrics and their labels (e.g. Google Cloud Monitoring with [MetricDescriptors](https://cloud.google.com/monitoring/api/v3/metrics-details#intro-metdesc)) to prevent breaking changes.

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,23 @@
+# Metric instrument naming conventions
```
yurishkuro, line 1: this document reads more like naming guidelines, not conventions. Conventions would actually include recommendations for common patterns, like `runtime.***`.

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Guidelines
+
+Metric names and labels exist within a single universe and a single hierarchy. Metric names and labels MUST be considered within the universe of all existing metric names. When defining new metric names and labels, consider the prior art of existing standard metrics and metrics from frameworks/libraries.
+
+Associated metrics SHOULD be nested together in a hierarchy based on their usage. Define a top-level hierarchy for common metric categories: for OS metrics, like CPU and network; for app runtimes, like GC internals. Libraries and frameworks should nest their metrics into a hierarchy as well. This aids in discovery and adhoc comparison. This allows a user to find similar metrics given a certain metric.
+
+The hierarchical structure of metrics defines the namespacing. Supporting OpenTelemetry artifacts define the metric structures and hierarchies for some categories of metrics, and these can assist decisions when creating future metrics.
+
+Common labels SHOULD be consistently named. This aids in discoverability and disambiguates similar labels to metric names.
```
yurishkuro, line 17: I am not sure how to interpret this w.r.t. service name. For example, [Prometheus][] guidelines specifically state

> For metrics specific to an application, the prefix is usually the application name itself.

In contrast, at Uber we structure all metrics to have a set of common infra labels (like zone/datacenter) that includes `service` name. So if have two services A and B and they both emit a metric with name "blah", then because it comes from different services there's no guarantee that the semantics of such metric are identical.

[Prometheus]: https://prometheus.io/docs/practices/naming/

text/metrics/0108-naming-conventions.md
```diff
@@ -0,0 +1,23 @@
...
+
```
yurishkuro, line 18: > Common labels SHOULD be consistently named

What does "common labels" mean? Common in what context/scope? A single service? An organization?

bogdandrutu:
@jmacd there are some comments left open, will merge once this is done.

@tedpennings please fix all open comments.

tedpennings:
@bogdandrutu will take a look later tonight, thanks! 

sorry for delay, I had family medical things come up over the past few days. things are better now, and I'm catching up on work now.

text/metrics/0108-naming-conventions.md
jmacd: #119 is coupled with this document with some conventions. https://github.com/open-telemetry/opentelemetry-specification/pull/657 describes a conventional naming pattern.

Originally this document contained conventions, and we split the work apart. @tedpennings would you substitute "guidelines" in for "conventions"?

text/metrics/0108-naming-conventions.md
jmacd: (I assume this comment relates to line 15.) 

The Prometheus guidelines, which we are following here, say that same-named metrics should be meaningful to aggregate (maybe not sensible, just meaningful). If you have a metric named "requests", maybe, with a service label, it's meaningful to aggregate those across services. So if "blah" is used across services, the different appearances **ought** to mean the same thing.

You didn't say whether you think the way Uber does this is good or not? At Lightstep we did in fact use the application prefix on our "requests" metric, and most of the time I'm glad I don't have to filter by service to build a service-specific dashboard. But once in a while, I want a generic dashboard and I can't just group by service, so it's not always clear what is best.

text/metrics/0108-naming-conventions.md
jmacd: I was hoping to avoid a debate over naming label keys. For example, you have a label key named "service" and have used it on some metrics, and I have a label key named "service" and have used it on some different metrics. How are we to know that those labels are not the same? The answer would be to add namespacing of labels. I recall the OpenCensus guidelines were to prefix your label names with a DNS prefix that you own. So I might have a `lightstep.com/service` label and you might have an `uber.com/service` label. For this to create a good user experience, I'd like the DNS prefix to not display by default. What would you like to see, @yurishkuro?

bogdandrutu:
@tedpennings friendly ping :)

justinfoote:
In the metrics SIG meeting on July 9, I was volunteered to help wrap up this PR.  I'm just now diving into it, but I'll try to get open questions/issues resolved before Thursday July 16.

text/metrics/0108-naming-conventions.md
justinfoote: I'm not sure if there's something actionable about this line.  
As it is written, it makes sense to me.  We're saying that, across broad categories of metrics, labels that mean the same thing (that is: common labels) should be named the same thing.  

This seems like enough detail for this guideline.  More detail than this belongs in semantic conventions, in my opinion.

text/metrics/0108-naming-conventions.md
justinfoote: The way I read this guidance is: if we have a label that should be added to many different categories of metric instrument, and that label's semantic meaning is the same across all those categories, its name should be consistent.  

The most obvious example I can think of would be _status_, whose value will be a CanonicalSpanStatus.  

As a user, I would find it intuitive when searching my metrics in my UI to always find the success/failure information under the same `status` label.

I'm not sure I understand the example `service` label.  Would it be the name of the service being instrumented?  If so, perhaps we would want some semantic conventions around how to apply Resource attributes as metric labels.

If this is the case, I'm not sure if we need to change this line. Is this guidance not clear enough? What wording would make our meaning more clear?

text/metrics/0108-naming-conventions.md
justinfoote: Changed to guidelines in 8b88b36

justinfoote:
@tc, It seems that we have broad approval on this OTEP, with five approvals so far. The only outstanding issue is with the specific definition of "common" labels.  

I'd like to propose that we merge this OTEP and sort out the details in a spec at a later date.

justinfoote:
@open-telemetry/technical-committee, It seems that we have broad approval on this OTEP, with five approvals so far. The only outstanding issue is with the specific definition of "common" labels.  

I'd like to propose that we merge this OTEP and sort out the details in a spec at a later date.

text/metrics/0108-naming-conventions.md
jmacd: I think it's safe to say we can merge this and debate this topic again as we modify the specification.

jmacd:
@open-telemetry/specs-metrics-approvers please merge.

PR #111, opened 2020-06-03
Added auto resource detection proposal

james-bebbington:
This OTEP proposes a mechanism to support automatic detection of resource information, including a default implementation to detect resource information from an environment variable. This is largely based on existing work in OpenCensus & the current implementation is OpenTelemetry JS SDK.

**Related issues/PRs:**

- Resources: define string values and keys as urlencoded: https://github.com/open-telemetry/opentelemetry-specification/pull/505
- Resource initialization from environment variable: https://github.com/open-telemetry/opentelemetry-specification/issues/535
- Standardize environment variables between SDK implementations: https://github.com/open-telemetry/opentelemetry-specification/issues/572
- Add Library Info to trace/metric provider: https://github.com/open-telemetry/opentelemetry-go/issues/587 (not included in this OTEP but closely related)

Note this is my first time creating an OTEP. I'm not aware of all the previous discussion on this topic (couldn't find any directly related previous work), but figured I'd go ahead and create an OTEP to start a discussion at least, as it seems like there is currently a gap in the specification.

### DIFF
--- a/text/0111-auto-resource-detection.md
+++ b/text/0111-auto-resource-detection.md
@@ -1,39 +1,68 @@
 # Automatic Resource Detection
 
-Describes a mechanism to support auto-detection of resource information.
+Introduce a mechanism to support auto-detection of resources.
 
 ## Motivation
 
-Resource information, i.e. attributes associated with the entity producing telemetry, can currently be supplied to tracer and meter providers or appended in custom exporters. In addition to this, it would be useful to have a mechanism to automatically detect resource information from the host (e.g. from an environment variable or from aws, gcp, etc metadata) and apply this to all kinds of telemetry. This will in many cases prevent users from having to manually configure resource information.
+Resource information, i.e. attributes associated with the entity producing
+telemetry, can currently be supplied to tracer and meter providers or appended
+in custom exporters. In addition to this, it would be useful to have a mechanism
+to automatically detect resource information from the host (e.g. from an
+environment variable or from aws, gcp, etc metadata) and apply this to all kinds
+of telemetry. This will in many cases prevent users from having to manually
+configure resource information.
 
-Note there are some existing implementations of this already in the SDKs (see [below](#prior-art-and-alternatives)), but nothing currently in the specification.
+Note there are some existing implementations of this already in the SDKs (see
+[below](#prior-art-and-alternatives)), but nothing currently in the
+specification.
 
 ## Explanation
 
-In order to apply auto-detected resource information to all kinds of telemetry, a user will need to configure which resource detector(s) they would like to run (e.g. AWS EC2 detector).
+In order to apply auto-detected resource information to all kinds of telemetry,
+a user will need to configure which resource detector(s) they would like to run
+(e.g. AWS EC2 detector).
 
-If multiple detectors are configured, and more than one of these successfully detects a resource, the resources will be merged according to the Merge interface already defined in the specification, i.e. the earliest matched resource's attributes will take precedence. Each detector may be run in parallel, but to ensure deterministic results, the resources must be merged in the order the detectors were added.
+If multiple detectors are configured, and more than one of these successfully
+detects a resource, the resources will be merged according to the Merge
+interface already defined in the specification, i.e. the earliest matched
+resource's attributes will take precedence. Each detector may be run in
+parallel, but to ensure deterministic results, the resources must be merged in
+the order the detectors were added.
 
-A default implementation of a detector that reads resource data from the `OTEL_RESOURCE` environment variable will be included in the SDK. The environment variable will contain of a list of key value pairs, and these are expected to be represented in a format similar to the [W3C Correlation-Context](https://github.com/w3c/correlation-context/blob/master/correlation_context/HTTP_HEADER_FORMAT.md#header-value), except that additional semi-colon delimited metadata is not supported, i.e.: `key1=value1,key2=value2`. If the user does not specify any resource, this detector will be run by default.
+A default implementation of a detector that reads resource data from the
+`OTEL_RESOURCE` environment variable will be included in the SDK. The
+environment variable will contain of a list of key value pairs, and these are
+expected to be represented in a format similar to the [W3C
+Correlation-Context](https://github.com/w3c/correlation-context/blob/master/correlation_context/HTTP_HEADER_FORMAT.md#header-value),
+except that additional semi-colon delimited metadata is not supported, i.e.:
+`key1=value1,key2=value2`. If the user does not specify any resource, this
+detector will be run by default.
 
-Custom resource detectors related to specific environments (e.g. specific cloud vendors) must be implemented as packages separate to the core SDK, and users will need to import these separately.
+Custom resource detectors related to specific environments (e.g. specific cloud
+vendors) must be implemented as packages separate to the core SDK, and users
+will need to import these separately.
 
 ## Internal details
 
-As described above, the following will be added to the Resource SDK specification:
+As described above, the following will be added to the Resource SDK
+specification:
 
 - An interface for "detectors", to retrieve resource information
-- Specification for a global function to merge resources returned by a set of detectors
-- Details of the "from environment variable" detector implementation as described above
-- Specification that default detection (from environment variable) runs once on startup, and is used by all tracer & meter providers by default if no custom resource is supplied
-
-This is a relatively small proposal so it is easiest to explain the details with a code example:
+- Specification for a global function to merge resources returned by a set of
+  detectors
+- Details of the "from environment variable" detector implementation as
+  described above
+- Specification that default detection (from environment variable) runs once on
+  startup, and is used by all tracer & meter providers by default if no custom
+  resource is supplied
 
 ### Usage
 
-The following example in Go creates a tracer and meter provider that uses resource information automatically detected from AWS or GCP:
+The following example in Go creates a tracer and meter provider that uses
+resource information automatically detected from AWS or GCP:
 
-Assumes a dependency has been added on the `otel/api`, `otel/sdk`, `otel/awsdetector`, and `otel/gcpdetector` packages.
+Assumes a dependency has been added on the `otel/api`, `otel/sdk`,
+`otel/awsdetector`, and `otel/gcpdetector` packages.
 
 ```go
 resource, _ := sdkresource.Detect(ctx, 5 * time.Second, awsdetector.ec2, gcpdetector.gce)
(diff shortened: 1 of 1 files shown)


text/0111-auto-resource-detection.md
```diff
@@ -0,0 +1,110 @@
...
+
+The Detector interface will simply return a Resource:
+
+```go
+type Detector interface {
+    Detect(ctx context.Context) (*Resource, error)
+}
+```
+
+If a detector is not able to detect a resource, it must return an empty resource so that the result of each call to `Detect` can be merged.
```
nilebox, line 65: `nil` resource should work fine for Go as well.
In other languages it might be different, as long as it leads to the same result.

I think there is also a difference between empty and missing resource in OTLP? See https://github.com/open-telemetry/opentelemetry-proto/blob/b54688569186e0b862bf7462a983ccf2c50c0547/opentelemetry/proto/metrics/v1/metrics.proto#L30

text/0111-auto-resource-detection.md
```diff
@@ -0,0 +1,110 @@
...
+- One alternative option to this proposal would be to supply a resource provider directly to the metric or trace provider instead of, or as an additional option to, a static resource. That would result in marginally simpler setup code where the user doesn't need to call `AutoDetect` themselves.
+- In this proposal, no resource detection will happen by default. It may be preferable to make trace / metric providers use the default resource provider if no resource is supplied. This would be in line with the behaviour of the current Java implementation - see [below](#prior-art-and-alternatives)
+- In the case of an error at resource detection time, another alternative would be to start a background thread to retry following some strategy, but it's not clear that there would be much value in doing this, and it would add considerable unnecessary complexity.
```
nilebox, line 92: If resource detection is a must, it should just crash an application probably to make the behavior predictable.
Otherwise we may run into same metrics having different resource attributes from time to time / not having resource at all, which will be problematic for aggregation.

text/0111-auto-resource-detection.md
```diff
@@ -0,0 +1,110 @@
+# Automatic Resource Detection
+
+Describes a mechanism to support auto-detection of resource information.
+
+## Motivation
+
+Resource information, i.e. attributes associated with the entity producing telemetry, can currently be supplied to tracer and meter providers or appended in custom exporters. In addition to this, it would be useful to have a mechanism to automatically detect resource information from the host (e.g. from an environment variable or from aws, gcp, etc metadata) and apply this to traces and metrics. This will in many cases prevent users from having to manually configure resource information.
```
jrcamp, line 7: > and apply this to traces and metrics

And logs.

text/0111-auto-resource-detection.md
```diff
@@ -0,0 +1,110 @@
...
+
+Note there are some existing implementations of this already in the SDKs (see [below](#prior-art-and-alternatives)), but nothing currently in the specification.
+
+## Explanation
+
+In order to apply auto-detected resource information to traces & metrics, a user will need to:
+
+- Configure which resource detector(s) they would like to run (e.g. AWS EC2 detector)
```
jrcamp, line 15: Is explicit configuration required? Could we run all cloud providers and take the first one that succeeds? If none succeed then the Detect() call fails.

text/0111-auto-resource-detection.md
```diff
@@ -0,0 +1,110 @@
...
+
+If multiple detectors are configured, and more than one of these successfully detects a resource, the resources will be merged according to the Merge interface already defined in the specification, i.e. the earliest matched resource's attributes will take precedence. Each detector may be run in parallel, but to ensure deterministic results, the resources must be merged in the order the detectors were added.
+
+A default implementation of a detector that reads resource data from the `OTEL_RESOURCE` environment variable will be included in the SDK. The environment variable will contain of a list of key value pairs, and these are expected to be represented in a format similar to the [W3C Correlation-Context](https://github.com/w3c/correlation-context/blob/master/correlation_context/HTTP_HEADER_FORMAT.md#header-value), i.e.: `key1=value1,key2=value2`. This detector must always be configured as the first detector by default.
+
+Custom resource detectors related to specific environments (e.g. specific cloud vendors) must be implemented outside of the SDK, and users will need to import these separately.
```
jrcamp, line 24: Resource detection is fairly trivial for all the major cloud platforms. You make an HTTP call to some well-known endpoint and get JSON or some plain text data. I don't think we necessarily need to use third party libraries which may bring in more dependencies than we care for. This would also simplify the number of dependencies for users as it'd all live in the SDK.

text/0111-auto-resource-detection.md
```diff
@@ -0,0 +1,110 @@
...
+
+The following example in Go creates a trace provider that uses resource information automatically detected from AWS or GCP:
+
+Assumes a dependency has been added on the `otel api`, `otel sdk`, `otel awsdetector`, and `otel gcpdetector` packages.
+
+```go
+rp := resource.NewMustProvider()
+rp.AddDetectors(awsdetector.EC2, gcpdetector.GCE, gcpdetector.GKE)
+resource := rp.AutoDetect(ctx)
+tp := sdktrace.NewProvider(sdktrace.WithResource(resource))
```
jrcamp, line 48: I think it's good to be able to have it explicit like this but would be nice to have a simple out of the box default assuming autodetection would work based on above:

`sdktrace.NewProvider(sdktrace.DetectResource())` or `sdktrace.NewProvider(sdktrace.MustDetectResource())`

text/0111-auto-resource-detection.md
james-bebbington: I probably should have added this as an Open question. I've suggested a similar approach to what was in the OpenCensus spec which dicates that cloud-vendor specific detection must not be included in the SDK (and thus would require some configuration), but that may not be necessary. As you point out below, the code to detect instance metadata is usually trivial: lookup an environment variable(s) or get a text response from a known endpoint.

If we do allow cloud vendor-specific detectors to be included in the SDK that raises a few related questions:

- Do we want to enforce that vendor libraries cannot be included in detectors?
- Are we open to any cloud providers adding detection code here as long as they follow the above rule? (other than just major ones)
- Do we attempt detection for all cloud providers by default? (including non-major ones)

text/0111-auto-resource-detection.md
james-bebbington: Hmm need to confirm if that is an important distinction. Note the current typescript implemention returns an empty resource.

text/0111-auto-resource-detection.md
james-bebbington: I like this suggestion - it makes up for one of the trade-offs I'd mentioned below, and makes initialization much simpler.

I added your code below but allowing detectors to be provided as parameters (even if we do require configuration of detectors, this could still be initialized in one line). Depending on the outcome of the thread above, I may remove the extra parameters.

text/0111-auto-resource-detection.md
james-bebbington: Update all similar reference in this doc to "all kinds of telemetry"

text/0111-auto-resource-detection.md
jrcamp: > Do we want to enforce that vendor libraries cannot be included in detectors?

Yes IMO

> Are we open to any cloud providers adding detection code here as long as they follow the above rule? (other than just major ones)

Yes as long as they add it to OT libraries for all languages.

> Do we attempt detection for all cloud providers by default? (including non-major ones)
If yes, then are there any cases where we would need to let users override this?

Yeah this does seem like it could be tricky given the number of cloud providers and that they're on the rise. Could limit it by some factors, like "must have X% market share" to be included in autodetection but any legitimate provider can still add a provider as a fallback if they don't meet that threshold?

I think ideally we would run all autodetectors in parallel with a relatively small time limit (we use 1s by default in SignalFx agent but I don't have any hard data supporting what the correct value should be).

text/0111-auto-resource-detection.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+If a detector is not able to detect a resource, it must return an uninitialized resource such that the result of each call to `Detect` can be merged.
+
+#### Provider
+
+A Provider will have a function to add detectors & to retrieve the auto-detected reosurce information:
+
+```go
+type Provider struct {
```
yurishkuro, line 78: Seems like Provider is just a composite Detector. Fewer concepts is better.

text/0111-auto-resource-detection.md
yurishkuro: can always add an option `detector := detectorBuilder.crashOnError(true).build()`

text/0111-auto-resource-detection.md
yurishkuro: I would start with requiring explicit enumeration of detectors by the user. It can always be extended to an auto-detector.

text/0111-auto-resource-detection.md
MrAlias: Adding generic and configurable detectors as a package would be useful. Providing not only the detector interface, but also including configurable Environment Variable and HTTP detectors could be broadly useful.

For example,

```
type Resourcer interface {
	Resource(io.Reader) (*Resource, error)
}

type HTTPDetector struct {
	Client *http.Client
	Request http.Request
	Resourcer  Resourcer
}

func (d HTTPDetector) Detect(ctx context.Context) (*Resource, error) {
	resp, err := d.Client.Do(d.Request)
	if err != nil {
		return nil, err
	}
	defer resp.Body.Close()
	return d.Resourcer.Resource(resp.Body)
}
```

text/0111-auto-resource-detection.md
MrAlias: I think @yurishkuro is correct here, and it was alluded to in the duplicate examples you gave above.

If we just went with the second approach:

```go
￼tp := sdktrace.NewProvider(sdktrace.MustDetectResource(awsdetector.EC2, gcpdetector.GCE, gcpdetector.GKE)) // or DetectResource
￼```

There would not be a need to add another provider.

File: text/0111-auto-resource-detection.md [hunk 8]
```diff
@@ -0,0 +1,115 @@
...
+
+### Existing OpenTelemetry implementations
+
+- Resource detection implementation in JS SDK [here](https://github.com/open-telemetry/opentelemetry-js/tree/master/packages/opentelemetry-resources): The JS implementation is very similar to this proposal. This proposal adds a resource provider instead of just having a global `DetectResources` function. In addition, vendor specific resource detection code is currently in the resource package, so this would need to be separated.
+- Environment variable resource detection in Java SDK [here](https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/src/main/java/io/opentelemetry/sdk/resources/EnvVarResource.java): This implementation does not currently include a detector interface, and this detector is used by default for trace and meter providers (so this proposal would introduce a breaking change in its current form)
+
+## Open questions
+
+- Does this interfere with any other upcoming specification changes related to resources?
+- If custom detectors need to live outside the core repo, what is the expectation regarding where they should be hosted?
```
MrAlias, line 111: I think this could follow the same model we have for vendor exporters, in that they are hosted in a vendor repository.

text/0111-auto-resource-detection.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Future possibilities
+
+When the Collector is run as an agent, the same interface, shared with the Go SDK, could be used to append resource information detected from the host to all kinds of telemetry in a Processor (probably as an extension to the existing Resource Processor). This would require a translation from the SDK resource to the collector's internal representation of a resource.
```
MrAlias, line 115: :100:

text/0111-auto-resource-detection.md
james-bebbington: I think that's captured by "These options will be provided as separate interfaces to let the user decide how to recover from failure, e.g. `DetectResource` & `MustDetectResource`"

text/0111-auto-resource-detection.md
james-bebbington: For the most part I agree, so I've removed this from the otep.

One disadvantage of removing this is that for each trace / metric provider you create, you will need to supply the list of detectors you want to run (rather than supplying a resource or provider) .. but that's not a huge issue, as you can just declare a list of detectors and re-use this (see the new example code)

text/0111-auto-resource-detection.md
james-bebbington: If we have default detectors, they should just be part of the SDK imo (as per the env variable detector mentioned in this OTEP). A generic HTTP detector could be useful as an implementation detail, but probably doesn't need to be mentioned in the spec.

text/0111-auto-resource-detection.md
james-bebbington: Note the overarching question here is:

"Is it okay for cloud-vendor specific detectors to live in the SDK directly as long as they don't include any third party libraries?"

I'll wait for a couple more opinions before updating the OTEP but if there is general consensus that this is okay, then I'll get rid of the notes about custom detectors living in separate packages.

text/0111-auto-resource-detection.md
james-bebbington: 👍 I agree that would make sense. Will update based on the outcome of the discussion above.

text/0111-auto-resource-detection.md
```diff
@@ -0,0 +1,92 @@
...
+
+The following example in Go creates a tracer and meter provider that uses resource information automatically detected from AWS or GCP:
+
+Assumes a dependency has been added on the `otel/api`, `otel/sdk`, `otel/awsdetector`, and `otel/gcpdetector` packages.
+
+```go
+resources := resource.Detector[]{awsdetector.EC2, gcpdetector.GCE, gcpdetector.GKE}
+tp := sdktrace.NewProvider(sdktrace.MustDetectResource(resources)) // or DetectResource (see below)
+mp := push.New(..., push.MustDetectResource(resources))
+```
```
jmacd, line 41: I think we should keep the Resource building separate from the SDKs. Instead of a `WithXXX` argument accepting detectors, why not have a resource library that simply builds a resource using the detectors you give it, and then pass the built up Resource to the SDK? IOW I'd prefer if both Tracing and Metric SDKs continue to support `WithResource`.

```
  resource := resource.Must(resource.Detect(awsdetector.EC2, gcpdetector.GCE))
  tp := sdktrace.NewProvicer(sdktrace.WithResource(resource))
```

jmacd: A related remark: caching inside the detectors sounds like more trouble than it's worth. I would advocate for users to simply evaluate the resource detectors they need and merge manually.

james-bebbington: I originally had something closer to your suggestion, but thought passing in detectors directly would result in a slightly simpler API for users. What's the main reason(s) for wanting resource building to be separate from the SDK? The SDKs can always support both `WithResource` & `DetectResource` (maybe having both options adds some confusion?)

I do get your point on caching though. That's definitely one issue with the current proposal.

MrAlias: For reference, the removal was motivated here: https://github.com/open-telemetry/oteps/pull/111#discussion_r435617411

jmacd: OK, sounds good.

text/0111-auto-resource-detection.md
```diff
@@ -0,0 +1,92 @@
...
+In the case of one or more detectors raising an error, there are two reasonable options:
+
+1. Ignore that detector, and continue with a warning (likely meaning we will continue without expected resource information)
+2. Crash the application (raise a panic)
+
+These options will be provided as separate interfaces to let the user decide how to recover from failure, e.g. `DetectResource` & `MustDetectResource`
+
+## Trade-offs and mitigations
+
+- In the case of an error at resource detection time, another alternative would be to start a background thread to retry following some strategy, but it's not clear that there would be much value in doing this, and it would add considerable unnecessary complexity.
```
AndrewAXue, line 74: I don't believe this will work, since you state that "the resources must be merged in the order the detectors were added". If there were a detector B after the one that failed, call it A, B would have to wait for A's detection to be complete. So there's no real value in making A do detection in a background thread if it blocks B anyways.

text/0111-auto-resource-detection.md
james-bebbington: I've actually ended up changing this to be exactly as Josh suggested. See the third bullet point in the trade off / mitigations section for reasoning

text/0111-auto-resource-detection.md
james-bebbington: FYI I ended up implementing a [resource detection processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/309) in the Collector to unblock some work I needed to get done rather than waiting for this OTEP, but happy to look into refactoring this in the future if we still think its worthwhile

text/0111-auto-resource-detection.md
```diff
@@ -0,0 +1,101 @@
+# Automatic Resource Detection
+
+Describes a mechanism to support auto-detection of resource information.
```
Oberon00, line 3: This should be a one-line summary that describes the changed introduced by this document, not the document itself.

```suggestion
Introduce a mechanism to support auto-detection of resources.
```

text/0111-auto-resource-detection.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Motivation
+
+Resource information, i.e. attributes associated with the entity producing telemetry, can currently be supplied to tracer and meter providers or appended in custom exporters. In addition to this, it would be useful to have a mechanism to automatically detect resource information from the host (e.g. from an environment variable or from aws, gcp, etc metadata) and apply this to all kinds of telemetry. This will in many cases prevent users from having to manually configure resource information.
```
Oberon00, line 7: Please add line breaks at least every sentence, otherwise it is impossible to comment on specific parts of the paragraph.

james-bebbington: 👍 - updated this to use a line wrapper extension

text/0111-auto-resource-detection.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Components
+
+#### Detector
+
+The Detector interface will simply return a Resource:
+
+```go
+type Detector interface {
+    Detect(ctx context.Context) (*Resource, error)
```
jkwatson, line 84: What is the context used for here?

jkwatson:
FYI, in the java project, we have something similar for auto-detecting AWS resource: https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/AwsResource.java

text/0111-auto-resource-detection.md
james-bebbington: This is just a go specific thing, where context always has to be passed around manually. For most other languages, ignore it. I should probably change these snippets to psuedo-code.

Oberon00: The telemetry context (which is indistinguishable from Go's context because it got re-used for that) is very meaningful for OpenTelemetry, so requiring it or not is an important distinction.

carlosalberto: I suggest we remove it from here, as it seems to be a Go-specific thing ;)

carlosalberto:
Just as @MrAlias I think this is great progress, and we can iron out the details as we go.

text/0111-auto-resource-detection.md
james-bebbington: > The telemetry context (which is indistinguishable from Go's context because it got re-used for that) is very meaningful for OpenTelemetry, so requiring it or not is an important distinction.

That is a good point and I had not thought about that. I had only included the context in this code snippet for the regular Go usage of context. I don't think there's any requirement to pass the telemetry context into this function, but I'm happy to learn if there is a use case for that.

> I suggest we remove it from here, as it seems to be a Go-specific thing ;)

I removed the code snippets (other than the `Usage` example) & replaced with slightly clearer wording.

james-bebbington:
We currently have three approvals, but before we merge this I'd be keen to get opinions on these important questions to inform how I write up the specification:

1. Is it okay for vendor specific detectors to be included in the SDK?
2. Do we want to run any/all resource detection by default?

For number 2 I could see a strong case being made for either way. As per @jrcamp's comments above, for the majority of customers & vendors, I can imagine that running AWS + Azure + GCP detection by default on startup would be extremely useful. If we implement this, however, it would be difficult to ever remove it.

Oberon00:
I have a fundamental question for this PR: Is a new API-package-level API really needed for auto-detecting resources? I think resource detection should happen in contrib packages, so that you can do (pseudo-Java):

```java
OpenTelemetry.setupTracing(AwsAutoDetector.detect().merge(K8SAutoDetector.detect()).merge(...));
```

It would be nice if K8SAutoDetector, AwsAutoDetector, etc. would implement some interface and were themselves autdetected, but then you should still be able to do this in a contrib package like this:

```java
OpenTelemetry.setupTracing(ClasspathAutoDetector.detectAndMergeAll());
```

The only crucial API-level thing is to have some `setupTracing`/`setupResources` function that is comfortable to use.

Oberon00:
It would also be great if vendor-SDKs could profit from this, so I would immediately vote for moving the Resource data structure (only) back to the API level.

text/0111-auto-resource-detection.md
```diff
@@ -0,0 +1,178 @@
...
+`key1=value1,key2=value2`. If the user does not specify any resource, this
+detector will be run by default.
+
+Custom resource detectors related to specific environments (e.g. specific cloud
+vendors) must be implemented as packages separate to the core SDK, and users
+will need to import these separately.
+
+## Internal details
+
+As described above, the following will be added to the Resource SDK
```
Oberon00, line 47: As stated in https://github.com/open-telemetry/oteps/pull/111#issuecomment-656034077, I'm not sure if this should be added directly to the Resource SDK spec. I thinklanguages should be encouraged to implement this as a separate, optional package.

text/0111-auto-resource-detection.md
```diff
@@ -0,0 +1,178 @@
...
+
+### Components
+
+#### Detector
+
+The `Detector` interface will simply contain a `Detect` function that returns a
+Resource.
+
+The `Detect` function should contain a mechanism to timeout and cancel the
+request. If a detector is not able to detect a resource, it must return an
```
Oberon00, line 81: > the request

Which request? First time the word "request" appears in this OTEP. I think resource detection must not perform network I/O, as resources are immutable and thus initializing resources must happen before telemetry can be collected. A timeout is out of the question.

One way to lift this restriction would be to specify a way to do this asynchronously, and delay the first call to export() until resources are detected. But this warrants a whole new OTEP. For now, I'd say anything which could require a timeout is out of the question as input for resources.

text/0111-auto-resource-detection.md
```diff
@@ -0,0 +1,178 @@
...
+The SDK will also provide a global `Detect` function. This will take a timeout
+duration and a set of detectors that should be run and merged in order as
+described in the intro, and return a resource.
+
+### Error Handling
+
+In the case of one or more detectors raising an error, there are two reasonable
+options:
+
+1. Ignore that detector, and continue with a warning (likely meaning we will
```
Oberon00, line 96: Please compare https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/error-handling.md

text/0111-auto-resource-detection.md
Aneurysm9: I'm not sure that this is something that needs to be in the spec, but at least in the Go SDK this timeout/cancel mechanism would be provided by the `context` referenced in [this comment](https://github.com/open-telemetry/oteps/pull/111#discussion_r448612875).  I think that's why it would be important to include the `context` even if it were not being used as an OpenTelemetry Context.

james-bebbington:
> I have a fundamental question for this PR: Is a new SDK-package-level API really needed for auto-detecting resources? I think resource detection should happen in contrib packages, so that you can do (pseudo-Java):
> 
> ...

That's a really good point. I have thought about this comment for quite a while now and can't think of any great counter arguments. I suppose the spec could still provide guidelines on how detect functions should be implemented? There is also still the question of which detectors the SDK should provide and/or run by default.

> It would also be great if vendor-SDKs could profit from this, so I would immediately vote for moving the Resource data structure (only) back to the API level (from the SDK level).

On a related note, I notice that the Span & Metric Exporter interfaces exist in the SDK rather than the API so the OpenTelemetry exporters can't be re-used by vendor-SDKs. To me that seems like a similar, but larger, issue?

text/0111-auto-resource-detection.md
james-bebbington: > Which request? First time the word "request" appears in this OTEP. I think resource detection must not perform network I/O, as resources are immutable and thus initializing resources must happen before telemetry can be collected.

Changed the wording to `operation` since not all detectors would not be creating any kind of request. I don't think it's unreasonable for detectors to perform network I/O synchronously if it is expected to be (relatively) quick, especially if a user specifically pulls in that package. Indeed I think would not be uncommon for looking up resource information: I think the convention by all major cloud vendors is to provide instance metadata via `http://169.254.169.254`

> One way to lift this restriction would be to specify a way to do this asynchronously, and delay the first call to export() until resources are detected. But this warrants a whole new OTEP. For now, I'd say anything which could require a timeout is out of the question as input for resources.

I don't think this would be worth the complexity. We'd have to make sure the resource information is applied to any "pending" telemetry before being exported which would not be trivial.

text/0111-auto-resource-detection.md
james-bebbington: I don't think there's anything here that contradicts those guidelines given that detection code runs on initialization.

Usually we would "assume that users would prefer to lose telemetry data rather than have the library significantly change the behavior of the instrumented application", but given that this code runs on initialization, this seems to come under pont no. 2: "The API or SDK may _fail fast_ and cause the application to fail on initialization".

Having the detection code fail fast, and leaving it up to the user to handle this if desired seems reasonable?

Oberon00: Sorry, that was a rather lazy and vague comment from me 😄 I guess what struck me as odd here is in what detail you describe error handling here. I think most of the specification does at most say "must not throw exceptions" and otherwise this document is the only thing that has more details.

text/0111-auto-resource-detection.md
Oberon00: Good point with it being the users choice. I just hope that most resources can be provided via things like environment variables or local files. For example, cloud instance metadata could probably queried by the backend if the resource contains some sort of instance ID?

SergeyKanzhelev:
Approvals from 3 companies and 2 TC members. Merging. @james-bebbington mentioned that some open comments might be addressed when merged into specification repo

PR #113, opened 2020-06-03
Integrate Exemplars with Metrics SDK

cnnradams:
This OTEP defines exemplars within OpenTelemetry and specifies behaviour for exemplars for the default set of aggregations: [Proposal doc](https://docs.google.com/document/d/1XpSCxzt5Q7pAIv6EeNt5x4jbqYO6pERVBO02sJvvMws/edit?ts=5ed6bf8e#)

### DIFF
--- a/text/metrics/0113-exemplars.md
+++ b/text/metrics/0113-exemplars.md
@@ -1,0 +1,94 @@
+# Integrate Exemplars with Metrics
+
+This OTEP adds exemplar support to aggregations defined in the Metrics SDK.
+
+## Definition
+
+Exemplars are example data points for aggregated data. They provide specific context to otherwise general aggregations. For histogram-type metrics, exemplars are points associated with each bucket in the histogram giving an example of what was aggregated into the bucket. Exemplars are augmented beyond just measurements with references to the sampled trace where the measurement was recorded and labels that were attached to the measurement.
+
+## Motivation
+
+Defining exemplar behaviour for aggregations allows OpenTelemetry to support exemplars in Google Cloud Monitoring.
+
+Exemplars provide a link between metrics and traces. Consider a user using a Histogram aggregation to track response latencies over time for a high QPS server. The histogram is composed of buckets based on the speed of the request, for example, "there were 55 requests that took 400-500 milliseconds". The user wants to troubleshoot slow requests, so they would need to find a trace where the latency was high. With exemplars, the user is able to get an exemplar trace from a high latency bucket, an exemplar trace from a low latency bucket, and compare them to figure out the reason for the high latency.
+
+Exemplars are meaningful for all aggregations where relevant traces can provide more context to the aggregation, as well as when exemplars can display specific information not otherwise shown in the aggregation (for example, the full set of labels where they otherwise might be aggregated away).
+
+## Internal details
+
+An exemplar is defined as:
+
+```
+message Exemplar {
+  // Numerical value of the measurement that was recorded. Only one of these two fields is
+  // used for the data, depending on its type
+  double double_value = 0;
+  int64 int64_value = 1;
+  
+  // Exact time that the measurement was recorded
+  fixed64 time_unix_nano = 2;
+
+  // 'label:value' map of all labels that were provided by the user recording the measurement
+  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 3;
+
+  // Span ID of the current trace [Optional]
+  string span_id = 4;
+
+  // Trace ID of the current trace [Optional]
+  string trace_id = 5;
+}
+```
+
+Exemplar collection should be enabled through an optional parameter, and when not enabled, there should be no collection/logic performed related to exemplars. This is to ensure that when necessary, aggregations are as high performance as possible.
+
+[#347](https://github.com/open-telemetry/opentelemetry-specification/pull/347) describes a set of standard aggregations in the metrics SDK. Here we describe how exemplars could be implemented for each aggregation.
+
+### Exemplar behaviour for standard aggregations
+
+#### HistogramAggregator
+
+Every bucket in the HistogramAggregator MUST (when enabled) maintain a list of exemplars whose values are within the boundaries of the bucket. Implementations should attempt to retain at least one exemplar per bucket, with a preference for exemplars with a sampled trace context and exemplars that were recorded later in the time period. They should also not retain an unbounded number of exemplars.
+
+#### Sketch
+
+A Sketch aggregation should maintain a list of exemplars whose values are spaced out across the distribution. There is no specific number of exemplars that should be retained (although the amount should not be unbounded), but the implementation should pick exemplars that represent as much of the distribution as possible. Preference should be given to exemplars with a sampled trace context. (Specific details not defined, see open questions.)
+
+#### Gauge
+
+Most (if not all) Gauges operate asynchronously and do not ever interact with traces. Since the value of a Gauge is the last measurement (essentially the other parts of an exemplar), exemplars are not worth implementing for Gauge.
+
+#### Exact
+
+The Exact aggregator does not aggregate measurements. If exemplars are enabled, implementations may attach a separate exemplar to each measurement in an exact aggregation including the trace context and full set of labels.
+
+Exemplars will always be retrieved from aggregations (by the exporter) as a list of Exemplar objects.
+
+## Trade-offs and mitigations
+
+Performance (in terms of memory usage and to some extent time complexity) is the main concern of implementing exemplars. However, by making recording exemplars optional, there should be minimal overhead when exemplars are not enabled.
+
+## Prior art and alternatives
+
+Exemplars are implemented in [OpenCensus](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/Exemplars.md#exemplars), but only for HistogramAggregator. This OTEP is largely a port from the OpenCensus definition of exemplars, but it also adds exemplar support to other aggregators.
+
+[Cloud monitoring API doc for exemplars](https://cloud.google.com/monitoring/api/ref_v3/rpc/google.api#google.api.Distribution.Exemplar)
+
+## Open questions
+
+- Exemplars usually refer to a span in a sampled trace. While using the collector to perform tail-sampling, the sampling decision may be deferred until after the metric would be exported. How do we create exemplars in this case?
+
+- We don’t have a strong grasp on how the sketch aggregator works in terms of implementation - so we don’t have enough information to design how exemplars should work properly.
+
+- The spec doesn't yet define a standard set of aggregations, just default aggregations for standard metric instruments. Since exemplars are always attached to particular aggregations, it's impossible to fully specify the behavior of exemplars.
+
+### Which aggregations should include exemplars?
+
+There are other aggregations that can benefit from exemplars, but they do not have well defined exemplar implementations and they are not supported by any known exporter. Should these be included in the OTEP or should they be left out?:
+
+#### Counter
+
+Exemplars give value to counter aggregations by tying metric and trace data together. When enabled, the aggregator will retain a small bounded list of exemplars at each checkpoint, containing at least the minimum and maximum value measurements whose trace context was sampled. Measurements should only be retained if there is a sampled trace context when the measurement was recorded.
+
+#### MinMaxSumCount
+
+The aggregator should maintain a list of at least two exemplars (when enabled), one near the maximum value and one near the minimum value. Preference should be given to exemplars with sampled traces, and if those are not available then the actual min and max values should be used.
--- a/text/metrics/0113-exemplars.md
+++ b/text/metrics/0113-exemplars.md
@@ -32,10 +32,10 @@ message RawValue {
   repeated opentelemetry.proto.common.v1.StringKeyValue labels = 3;
 
   // Span ID of the current trace
-  optional string span_id = 4;
+  optional bytes span_id = 4;
 
   // Trace ID of the current trace
-  optional string trace_id = 5;
+  optional bytes trace_id = 5;
 
   // When sample_count is non-zero, this exemplar has been chosen in a statistically
   // unbiased way such that the exemplar is representative of `sample_count` individual events
@@ -59,7 +59,7 @@ A Sketch aggregator SHOULD maintain a list of exemplars whose values are spaced 
 
 #### Last-Value
 
-Most (if not all) Last-Value aggregators operate asynchronously and do not ever interact with context. Since the value of a Last-Value is the last measurement (essentially the other parts of an exemplar), exemplars are not worth implementing for Gauge.
+Most (if not all) Last-Value aggregators operate asynchronously and do not ever interact with context. Since the value of a Last-Value is the last measurement (essentially the other parts of an exemplar), exemplars are not worth implementing for Last-Value.
 
 #### Exact
 
--- a/text/metrics/0113-exemplars.md
+++ b/text/metrics/0113-exemplars.md
@@ -19,7 +19,7 @@ Exemplars are meaningful for all aggregations where relevant traces can provide 
 An exemplar is defined as:
 
 ```
-message Exemplar {
+message RawValue {
   // Numerical value of the measurement that was recorded. Only one of these two fields is
   // used for the data, depending on its type
   double double_value = 0;
@@ -31,15 +31,19 @@ message Exemplar {
   // 'label:value' map of all labels that were provided by the user recording the measurement
   repeated opentelemetry.proto.common.v1.StringKeyValue labels = 3;
 
-  // Span ID of the current trace [Optional]
-  string span_id = 4;
+  // Span ID of the current trace
+  optional string span_id = 4;
 
-  // Trace ID of the current trace [Optional]
-  string trace_id = 5;
+  // Trace ID of the current trace
+  optional string trace_id = 5;
+
+  // When sample_count is non-zero, this exemplar has been chosen in a statistically
+  // unbiased way such that the exemplar is representative of `sample_count` individual events
+  optional double sample_count = 6;
 }
 ```
 
-Exemplar collection should be enabled through an optional parameter, and when not enabled, there should be no collection/logic performed related to exemplars. This is to ensure that when necessary, aggregations are as high performance as possible.
+Exemplar collection should be enabled through an optional parameter (disabled by default), and when not enabled, there should be no collection/logic performed related to exemplars. This is to ensure that when necessary, aggregations are as high performance as possible. Aggregations should also have a parameter to determine whether exemplars should only be collected if they are recorded during a sampled trace, or if tracing should have no effect on which exemplars are sampled. This allows aggregations to prioritize either the link between metrics and traces or the statistical significance of exemplars, when necessary.
 
 [#347](https://github.com/open-telemetry/opentelemetry-specification/pull/347) describes a set of standard aggregations in the metrics SDK. Here we describe how exemplars could be implemented for each aggregation.
 
@@ -47,11 +51,11 @@ Exemplar collection should be enabled through an optional parameter, and when no
 
 #### HistogramAggregator
 
-Every bucket in the HistogramAggregator MUST (when enabled) maintain a list of exemplars whose values are within the boundaries of the bucket (for buckets containing a population of one or more). Implementations SHOULD attempt to retain at least one exemplar per bucket, with a preference for exemplars with a sampled trace context and exemplars that were recorded later in the time period. They SHOULD NOT retain an unbounded number of exemplars.
+The HistogramAggregator MUST (when enabled) maintain a list of exemplars whose values are distributed across all buckets of the histogram (there should be one or more exemplars in every bucket that has a population of at least one sample-able measurement). Implementations SHOULD NOT retain an unbounded number of exemplars.
 
 #### Sketch
 
-A Sketch aggregation SHOULD maintain a list of exemplars whose values are spaced out across the distribution. There is no specific number of exemplars that should be retained (although the amount SHOULD NOT be unbounded), but the implementation SHOULD pick exemplars that represent as much of the distribution as possible. Preference SHOULD be given to exemplars with a sampled trace context. (Specific details not defined, see open questions.)
+A Sketch aggregation SHOULD maintain a list of exemplars whose values are spaced out across the distribution. There is no specific number of exemplars that should be retained (although the amount SHOULD NOT be unbounded), but the implementation SHOULD pick exemplars that represent as much of the distribution as possible. (Specific details not defined, see open questions.)
 
 #### Last-Value
 
@@ -59,9 +63,27 @@ Most (if not all) Gauges operate asynchronously and do not ever interact with tr
 
 #### Exact
 
-The Exact aggregator does not aggregate measurements. If exemplars are enabled, implementations may attach a separate exemplar to each measurement in an exact aggregation including the trace context and full set of labels.
+The Exact aggregation will function by maintaining a list of `RawValue`s, which contain all of the information exemplars would carry. Therefore the Exact aggregation will not need to maintain any exemplars.
 
-Exemplars will always be retrieved from aggregations (by the exporter) as a list of Exemplar objects.
+#### Counter
+
+Exemplars give value to counter aggregations in two ways: One, by tying metric and trace data together, and two, by providing necessary information to re-create the input distribution. When enabled, the aggregator will retain a bounded list of exemplars at each checkpoint, sampled from across the distribution of the data. Exemplars should be sampled in a statistically significant way.
+
+#### MinMaxSumCount
+
+Similar to Counter, MinMaxSumCount should retain a bounded list of exemplars that were sampled from across the input distribution in a statistically significant way.
+
+#### Custom Aggregations
+
+Custom aggregations MAY support exemplars by maintaining a list of exemplars that can be retrieved by exporters. Custom aggregations should select exemplars based on their usage by the connected exporter (for example, exemplars recorded for Stackdriver should only be retained if they were recorded within a sampled trace).
+
+Exemplars will always be retrieved from aggregations (by the exporter) as a list of RawValue objects. They will be communicated via a
+
+```
+optional repeated RawValue exemplars = 6
+```
+
+attribute on the `metric_descriptor` object.
 
 ## Trade-offs and mitigations
 
@@ -80,15 +102,3 @@ Exemplars are implemented in [OpenCensus](https://github.com/census-instrumentat
 - We don’t have a strong grasp on how the sketch aggregator works in terms of implementation - so we don’t have enough information to design how exemplars should work properly.
 
 - The spec doesn't yet define a standard set of aggregations, just default aggregations for standard metric instruments. Since exemplars are always attached to particular aggregations, it's impossible to fully specify the behavior of exemplars.
-
-### Which aggregations should include exemplars?
-
-There are other aggregations that can benefit from exemplars, but they do not have well defined exemplar implementations and they are not supported by any known exporter. Should these be included in the OTEP or should they be left out?:
-
-#### Counter
-
-Exemplars give value to counter aggregations by tying metric and trace data together. When enabled, the aggregator will retain a small bounded list of exemplars at each checkpoint, containing at least the minimum and maximum value measurements whose trace context was sampled. Measurements SHOULD be retained only if there is a sampled trace context when the measurement was recorded.
-
-#### MinMaxSumCount
-
-The aggregator should maintain a list of at least two exemplars (when enabled), one near the maximum value and one near the minimum value. Preference SHOULD be given to exemplars with sampled traces, and if those are not available then the actual min and max values SHOULD be used.
--- a/text/metrics/0113-exemplars.md
+++ b/text/metrics/0113-exemplars.md
@@ -16,7 +16,7 @@ Exemplars are meaningful for all aggregations where relevant traces can provide 
 
 ## Internal details
 
-An exemplar is defined as:
+An exemplar is a `RawValue`, which is defined as:
 
 ```
 message RawValue {
@@ -43,11 +43,11 @@ message RawValue {
 }
 ```
 
-Exemplar collection should be enabled through an optional parameter (disabled by default), and when not enabled, there should be no collection/logic performed related to exemplars. This is to ensure that when necessary, aggregations are as high performance as possible. Aggregations should also have a parameter to determine whether exemplars should only be collected if they are recorded during a sampled trace, or if tracing should have no effect on which exemplars are sampled. This allows aggregations to prioritize either the link between metrics and traces or the statistical significance of exemplars, when necessary.
+Exemplar collection should be enabled through an optional parameter (disabled by default), and when not enabled, there should be no collection/logic performed related to exemplars. This is to ensure that when necessary, aggregators are as high performance as possible. Aggregators should also have a parameter to determine whether exemplars should only be collected if they are recorded during a sampled trace, or if tracing should have no effect on which exemplars are sampled. This allows aggregations to prioritize either the link between metrics and traces or the statistical significance of exemplars, when necessary.
 
-[#347](https://github.com/open-telemetry/opentelemetry-specification/pull/347) describes a set of standard aggregations in the metrics SDK. Here we describe how exemplars could be implemented for each aggregation.
+[#347](https://github.com/open-telemetry/opentelemetry-specification/pull/347) describes a set of standard aggregators in the metrics SDK. Here we describe how exemplars could be implemented for each aggregator.
 
-### Exemplar behaviour for standard aggregations
+### Exemplar behaviour for standard aggregators
 
 #### HistogramAggregator
 
@@ -55,15 +55,15 @@ The HistogramAggregator MUST (when enabled) maintain a list of exemplars whose v
 
 #### Sketch
 
-A Sketch aggregation SHOULD maintain a list of exemplars whose values are spaced out across the distribution. There is no specific number of exemplars that should be retained (although the amount SHOULD NOT be unbounded), but the implementation SHOULD pick exemplars that represent as much of the distribution as possible. (Specific details not defined, see open questions.)
+A Sketch aggregator SHOULD maintain a list of exemplars whose values are spaced out across the distribution. There is no specific number of exemplars that should be retained (although the amount SHOULD NOT be unbounded), but the implementation SHOULD pick exemplars that represent as much of the distribution as possible. (Specific details not defined, see open questions.)
 
 #### Last-Value
 
-Most (if not all) Gauges operate asynchronously and do not ever interact with traces. Since the value of a Gauge is the last measurement (essentially the other parts of an exemplar), exemplars are not worth implementing for Gauge.
+Most (if not all) Last-Value aggregators operate asynchronously and do not ever interact with context. Since the value of a Last-Value is the last measurement (essentially the other parts of an exemplar), exemplars are not worth implementing for Gauge.
 
 #### Exact
 
-The Exact aggregation will function by maintaining a list of `RawValue`s, which contain all of the information exemplars would carry. Therefore the Exact aggregation will not need to maintain any exemplars.
+The Exact aggregator will function by maintaining a list of `RawValue`s, which contain all of the information exemplars would carry. Therefore the Exact aggregator will not need to maintain any exemplars.
 
 #### Counter
 
@@ -73,9 +73,9 @@ Exemplars give value to counter aggregations in two ways: One, by tying metric a
 
 Similar to Counter, MinMaxSumCount should retain a bounded list of exemplars that were sampled from across the input distribution in a statistically significant way.
 
-#### Custom Aggregations
+#### Custom Aggregators
 
-Custom aggregations MAY support exemplars by maintaining a list of exemplars that can be retrieved by exporters. Custom aggregations should select exemplars based on their usage by the connected exporter (for example, exemplars recorded for Stackdriver should only be retained if they were recorded within a sampled trace).
+Custom aggregators MAY support exemplars by maintaining a list of exemplars that can be retrieved by exporters. Custom aggregators should select exemplars based on their usage by the connected exporter (for example, exemplars recorded for Stackdriver should only be retained if they were recorded within a sampled trace).
 
 Exemplars will always be retrieved from aggregations (by the exporter) as a list of RawValue objects. They will be communicated via a
 
@@ -83,7 +83,7 @@ Exemplars will always be retrieved from aggregations (by the exporter) as a list
 optional repeated RawValue exemplars = 6
 ```
 
-attribute on the `metric_descriptor` object.
+attribute on the `Metric` object.
 
 ## Trade-offs and mitigations
 
--- a/text/metrics/0113-exemplars.md
+++ b/text/metrics/0113-exemplars.md
@@ -75,7 +75,7 @@ Similar to Counter, MinMaxSumCount should retain a bounded list of exemplars tha
 
 #### Custom Aggregators
 
-Custom aggregators MAY support exemplars by maintaining a list of exemplars that can be retrieved by exporters. Custom aggregators should select exemplars based on their usage by the connected exporter (for example, exemplars recorded for Stackdriver should only be retained if they were recorded within a sampled trace).
+Custom aggregators MAY support exemplars by maintaining a list of exemplars that can be retrieved by exporters. Custom aggregators should select exemplars based on their usage by the connected exporter (for example, exemplars recorded for Google Cloud Monitoring should only be retained if they were recorded within a sampled trace).
 
 Exemplars will always be retrieved from aggregations (by the exporter) as a list of RawValue objects. They will be communicated via a
 
--- a/text/metrics/0113-exemplars.md
+++ b/text/metrics/0113-exemplars.md
@@ -47,13 +47,13 @@ Exemplar collection should be enabled through an optional parameter, and when no
 
 #### HistogramAggregator
 
-Every bucket in the HistogramAggregator MUST (when enabled) maintain a list of exemplars whose values are within the boundaries of the bucket. Implementations should attempt to retain at least one exemplar per bucket, with a preference for exemplars with a sampled trace context and exemplars that were recorded later in the time period. They should also not retain an unbounded number of exemplars.
+Every bucket in the HistogramAggregator MUST (when enabled) maintain a list of exemplars whose values are within the boundaries of the bucket (for buckets containing a population of one or more). Implementations SHOULD attempt to retain at least one exemplar per bucket, with a preference for exemplars with a sampled trace context and exemplars that were recorded later in the time period. They SHOULD NOT retain an unbounded number of exemplars.
 
 #### Sketch
 
-A Sketch aggregation should maintain a list of exemplars whose values are spaced out across the distribution. There is no specific number of exemplars that should be retained (although the amount should not be unbounded), but the implementation should pick exemplars that represent as much of the distribution as possible. Preference should be given to exemplars with a sampled trace context. (Specific details not defined, see open questions.)
+A Sketch aggregation SHOULD maintain a list of exemplars whose values are spaced out across the distribution. There is no specific number of exemplars that should be retained (although the amount SHOULD NOT be unbounded), but the implementation SHOULD pick exemplars that represent as much of the distribution as possible. Preference SHOULD be given to exemplars with a sampled trace context. (Specific details not defined, see open questions.)
 
-#### Gauge
+#### Last-Value
 
 Most (if not all) Gauges operate asynchronously and do not ever interact with traces. Since the value of a Gauge is the last measurement (essentially the other parts of an exemplar), exemplars are not worth implementing for Gauge.
 
@@ -87,8 +87,8 @@ There are other aggregations that can benefit from exemplars, but they do not ha
 
 #### Counter
 
-Exemplars give value to counter aggregations by tying metric and trace data together. When enabled, the aggregator will retain a small bounded list of exemplars at each checkpoint, containing at least the minimum and maximum value measurements whose trace context was sampled. Measurements should only be retained if there is a sampled trace context when the measurement was recorded.
+Exemplars give value to counter aggregations by tying metric and trace data together. When enabled, the aggregator will retain a small bounded list of exemplars at each checkpoint, containing at least the minimum and maximum value measurements whose trace context was sampled. Measurements SHOULD be retained only if there is a sampled trace context when the measurement was recorded.
 
 #### MinMaxSumCount
 
-The aggregator should maintain a list of at least two exemplars (when enabled), one near the maximum value and one near the minimum value. Preference should be given to exemplars with sampled traces, and if those are not available then the actual min and max values should be used.
+The aggregator should maintain a list of at least two exemplars (when enabled), one near the maximum value and one near the minimum value. Preference SHOULD be given to exemplars with sampled traces, and if those are not available then the actual min and max values SHOULD be used.


MrAlias:
Relevant to this: https://github.com/open-telemetry/opentelemetry-proto/issues/81

jmacd:
This is great. I added a bunch of comments describing how I think we can leave room for statistical interpretation based on exemplars in the future.

jmacd:
Another related remark: In the extreme case of supporting one exemplar per metric event, can the protocol degenerate into the raw structured event encoding that has been requested?

https://github.com/open-telemetry/opentelemetry-specification/issues/617

I hadn't thought about this before making comments on the document above, but I think it might be possible to combine these ideas. If the value of an aggregation can optionally include raw values in addition to another aggregation, then raw values are the exemplars. @mxiamxia ^^^

mxiamxia:
> Another related remark: In the extreme case of supporting one exemplar per metric event, can the protocol degenerate into the raw structured event encoding that has been requested?
> 
> [open-telemetry/opentelemetry-specification#617](https://github.com/open-telemetry/opentelemetry-specification/issues/617)
> 
> I hadn't thought about this before making comments on the document above, but I think it might be possible to combine these ideas. If the value of an aggregation can optionally include raw values in addition to another aggregation, then raw values are the exemplars. @mxiamxia ^^^

Hi @jmacd, If I get you right, we can store the raw event values as ```Exemplar``` along with the aggregations. For example, for any aggregated metrics data point, we can attach all the raw data as list of Exemplars (by preference in SDK) within a collection interval to aggregated data point. So we can have both raw data values and the original purpose of Exemplar, is it right? Thanks. 

> Exact
> The Exact aggregator does not aggregate measurements. If exemplars are enabled, implementations may attach a separate exemplar to each measurement in an exact aggregation including the trace context and full set of labels.

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,94 @@
...
+
+Exemplar collection should be enabled through an optional parameter, and when not enabled, there should be no collection/logic performed related to exemplars. This is to ensure that when necessary, aggregations are as high performance as possible.
+
+[#347](https://github.com/open-telemetry/opentelemetry-specification/pull/347) describes a set of standard aggregations in the metrics SDK. Here we describe how exemplars could be implemented for each aggregation.
+
+### Exemplar behaviour for standard aggregations
+
+#### HistogramAggregator
+
+Every bucket in the HistogramAggregator MUST (when enabled) maintain a list of exemplars whose values are within the boundaries of the bucket. Implementations should attempt to retain at least one exemplar per bucket, with a preference for exemplars with a sampled trace context and exemplars that were recorded later in the time period. They should also not retain an unbounded number of exemplars.
```
MrAlias, line 50:
```suggestion
Every bucket in the HistogramAggregator MUST (when enabled) maintain a list of exemplars whose values are within the boundaries of the bucket (for buckets containing a population of one or more). Implementations SHOULD attempt to retain at least one exemplar per bucket, with a preference for exemplars with a sampled trace context and exemplars that were recorded later in the time period. They SHOULD NOT retain an unbounded number of exemplars.
```

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### Sketch
+
+A Sketch aggregation should maintain a list of exemplars whose values are spaced out across the distribution. There is no specific number of exemplars that should be retained (although the amount should not be unbounded), but the implementation should pick exemplars that represent as much of the distribution as possible. Preference should be given to exemplars with a sampled trace context. (Specific details not defined, see open questions.)
```
MrAlias, line 54:
```suggestion
A Sketch aggregation SHOULD maintain a list of exemplars whose values are spaced out across the distribution. There is no specific number of exemplars that should be retained (although the amount SHOULD NOT be unbounded), but the implementation SHOULD pick exemplars that represent as much of the distribution as possible. Preference SHOULD be given to exemplars with a sampled trace context. (Specific details not defined, see open questions.)
```

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### Gauge
```
MrAlias, line 56: Nit: the linked PR does talk about `Gauge`, however, as an aggregation I'm not sure that is or ever was a thing. Current thinking is to refer to this aggregation as last-value or last-value-wins.

```suggestion
#### Last-Value
```

If consistency with that PR is preferred, feel free to disregard.

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,94 @@
...
+
+Most (if not all) Gauges operate asynchronously and do not ever interact with traces. Since the value of a Gauge is the last measurement (essentially the other parts of an exemplar), exemplars are not worth implementing for Gauge.
+
+#### Exact
+
+The Exact aggregator does not aggregate measurements. If exemplars are enabled, implementations may attach a separate exemplar to each measurement in an exact aggregation including the trace context and full set of labels.
+
+Exemplars will always be retrieved from aggregations (by the exporter) as a list of Exemplar objects.
```
MrAlias, line 64: :+1:

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,94 @@
...
+
+[Cloud monitoring API doc for exemplars](https://cloud.google.com/monitoring/api/ref_v3/rpc/google.api#google.api.Distribution.Exemplar)
+
+## Open questions
+
+- Exemplars usually refer to a span in a sampled trace. While using the collector to perform tail-sampling, the sampling decision may be deferred until after the metric would be exported. How do we create exemplars in this case?
+
+- We don’t have a strong grasp on how the sketch aggregator works in terms of implementation - so we don’t have enough information to design how exemplars should work properly.
+
+- The spec doesn't yet define a standard set of aggregations, just default aggregations for standard metric instruments. Since exemplars are always attached to particular aggregations, it's impossible to fully specify the behavior of exemplars.
```
MrAlias, line 82: There is the added complication of the current push to have the default SDK support custom aggregations as well.

It might be useful to add a section above on how custom aggregations are left to their own to produce useful exemplars. And, possibly, how that would look.

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Prior art and alternatives
+
+Exemplars are implemented in [OpenCensus](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/Exemplars.md#exemplars), but only for HistogramAggregator. This OTEP is largely a port from the OpenCensus definition of exemplars, but it also adds exemplar support to other aggregators.
+
+[Cloud monitoring API doc for exemplars](https://cloud.google.com/monitoring/api/ref_v3/rpc/google.api#google.api.Distribution.Exemplar)
+
+## Open questions
+
+- Exemplars usually refer to a span in a sampled trace. While using the collector to perform tail-sampling, the sampling decision may be deferred until after the metric would be exported. How do we create exemplars in this case?
```
MrAlias, line 78: Doubt there is a great solution here, but possibly:

- the processing engine could purge the related exemplar if available.
- add a field to signify the span is included in an exemplar.

:shrug:

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,94 @@
...
+
+There are other aggregations that can benefit from exemplars, but they do not have well defined exemplar implementations and they are not supported by any known exporter. Should these be included in the OTEP or should they be left out?:
+
+#### Counter
+
+Exemplars give value to counter aggregations by tying metric and trace data together. When enabled, the aggregator will retain a small bounded list of exemplars at each checkpoint, containing at least the minimum and maximum value measurements whose trace context was sampled. Measurements should only be retained if there is a sampled trace context when the measurement was recorded.
+
+#### MinMaxSumCount
+
+The aggregator should maintain a list of at least two exemplars (when enabled), one near the maximum value and one near the minimum value. Preference should be given to exemplars with sampled traces, and if those are not available then the actual min and max values should be used.
```
MrAlias, line 94:
```suggestion
The aggregator should maintain a list of at least two exemplars (when enabled), one near the maximum value and one near the minimum value. Preference SHOULD be given to exemplars with sampled traces, and if those are not available then the actual min and max values SHOULD be used.
```

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,94 @@
...
+
+- The spec doesn't yet define a standard set of aggregations, just default aggregations for standard metric instruments. Since exemplars are always attached to particular aggregations, it's impossible to fully specify the behavior of exemplars.
+
+### Which aggregations should include exemplars?
+
+There are other aggregations that can benefit from exemplars, but they do not have well defined exemplar implementations and they are not supported by any known exporter. Should these be included in the OTEP or should they be left out?:
+
+#### Counter
+
+Exemplars give value to counter aggregations by tying metric and trace data together. When enabled, the aggregator will retain a small bounded list of exemplars at each checkpoint, containing at least the minimum and maximum value measurements whose trace context was sampled. Measurements should only be retained if there is a sampled trace context when the measurement was recorded.
```
MrAlias, line 90:
```suggestion
Exemplars give value to counter aggregations by tying metric and trace data together. When enabled, the aggregator will retain a small bounded list of exemplars at each checkpoint, containing at least the minimum and maximum value measurements whose trace context was sampled. Measurements SHOULD be retained only if there is a sampled trace context when the measurement was recorded.
```

MrAlias:
> @jmacd I don't see you comments?

Oh, :facepalm:, they're in the linked Google doc.

MrAlias:
> Another related remark: In the extreme case of supporting one exemplar per metric event, can the protocol degenerate into the raw structured event encoding that has been requested?
> 
> [open-telemetry/opentelemetry-specification#617](https://github.com/open-telemetry/opentelemetry-specification/issues/617)
> 
> I hadn't thought about this before making comments on the document above, but I think it might be possible to combine these ideas. If the value of an aggregation can optionally include raw values in addition to another aggregation, then raw values are the exemplars. @mxiamxia ^^^

Agreed, I was thinking the same about merging raw metric exports and this message type.

What about calling the exemplar message `Raw` or `RawValues`?

text/metrics/0113-exemplars.md
cnnradams: Agreed, exemplar support in custom aggregations would have to be rewritten for each aggregation if exemplars stay coupled to aggregations. We could pull exemplars out a level to function seperately from aggregations, letting the user plug in exemplar systems to custom aggregations (as well as define their own), I wrote a bit about it in a comment on the doc.

cnnradams:
> > Another related remark: In the extreme case of supporting one exemplar per metric event, can the protocol degenerate into the raw structured event encoding that has been requested?
> > [open-telemetry/opentelemetry-specification#617](https://github.com/open-telemetry/opentelemetry-specification/issues/617)
> > I hadn't thought about this before making comments on the document above, but I think it might be possible to combine these ideas. If the value of an aggregation can optionally include raw values in addition to another aggregation, then raw values are the exemplars. @mxiamxia ^^^
> 
> Agreed, I was thinking the same about merging raw metric exports and this message type.
> 
> What about calling the exemplar message `Raw` or `RawValues`?

If I'm understanding this correctly, in any case where we export raw measurements it would use the `RawValue` type, including for exemplars? That sounds good to me :)

text/metrics/0113-exemplars.md
MrAlias: > brings the scope of Exemplars outside of just the SDK into the API, which we might not want

Agreed, I don't think we want that.

Maybe it just goes without saying, that if you are writing a custom aggregator it's up to you to annotate appropriately with exemplars. Likely, it will only be the vendor or user who wrote the custom aggregation that will use it, so this is not an OTel issue.

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,94 @@
...
+  // Numerical value of the measurement that was recorded. Only one of these two fields is
+  // used for the data, depending on its type
+  double double_value = 0;
+  int64 int64_value = 1;
+  
+  // Exact time that the measurement was recorded
+  fixed64 time_unix_nano = 2;
+
+  // 'label:value' map of all labels that were provided by the user recording the measurement
+  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 3;
```
mxiamxia, line 32: Should we consider add a field to indicate the original Metrics instrument type (one of six)?

cnnradams: I assume that parameter would belong on the Aggregator instead, since Exemplars won't leave the scope of Aggregator and each aggregator has a single associated metric instrument (although I could be mistaken)

mxiamxia: We have all the metric instrument info in SDK side. but If we'll store all the raw metric events in Exemplars and pass it to Collector. I may want to keep the same info in Exemplar and Collector will be able to retrieve it.

cnnradams: fair point, it just feels like a lot of unnecessary space for an attribute that will be the same for every exemplar in the list. If the instrument type was attached to the aggregator, the exporter could grab it and send it separately alongside the exemplars, which to me makes more sense. (there is no case where an aggregator is a raw list of exemplars, it will have an aggregator object wrapping it that can have the instrument property. so the information should always get to at least the exporter)

jmacd: The instrument type should be part of the metric descriptor, but the exemplars are associated with aggregations. We intend to support derived aggregations in the protocol, to support the Views API. I agree that we should not include parts of the instrument descriptor in the exemplar, but then how should we tie the output of an aggregation to the descriptor of the instrument that was used? I believe this is a question for the Views API design.

text/metrics/0113-exemplars.md
jmacd: I'd say that being selected as an exemplar should tell the tail sampler that a decision has been made.

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,94 @@
...
+
+  // Span ID of the current trace [Optional]
+  string span_id = 4;
+
+  // Trace ID of the current trace [Optional]
+  string trace_id = 5;
+}
```
jmacd, line 39: For me what's most important is that:
(1) exemplars carry an optional statistical weight
(2) exemplars can be included alongside the output of any aggregation
(3) multiple exemplars are supported.

Today we have

```
message Metric {
  // metric_descriptor describes the Metric.
  MetricDescriptor metric_descriptor = 1;

  // Data is a list of one or more DataPoints for a single metric. Only one of the
  // following fields is used for the data, depending on the type of the metric defined
  // by MetricDescriptor.type field.
  repeated Int64DataPoint int64_data_points = 2;
  repeated DoubleDataPoint double_data_points = 3;
  repeated HistogramDataPoint histogram_data_points = 4;
  repeated SummaryDataPoint summary_data_points = 5;
}
```

I'm suggesting that exemplars should be an optional repeated field alongside the other 4 value types.

jmacd: I should have added: this "statistical weight" should be expressed as an adjusted count.  
```
   // When sample_count is non-zero, this exemplar has been chosen in a statistically
   // unbiased way such that the exemplar is representative of `sample_count` individual events.
   optional double sample_count = 10;
```

jmacd: (The "non-zero" statement above allows non-probabilistically selected exemplars to be included.)

text/metrics/0113-exemplars.md
jmacd: Using the proposed "sample_count" field described in a comment above, systems will be able to make meaningful interpretation of the exemplars (regardless of how they were selected). At the protocol level, I do not think we need to specify more than that.

This is a question about what the default SDK should be specified to do. There I think we can find some agreement:

- exemplars are disabled by default
- exemplars are enabled by indicating how many exemplars to collect per instrument
- turning on exemplars means asking for arbitrary selection (i.e., not using unbiased sampling)
- the histogram aggregator is expected to produce per-bucket exemplars when enabled, so you should configure exemplars in multiples of the number of buckets
- the protocol may support optional statistical count, but no default behavior takes advantage of it.

This sounds problematic. One of the problems is that it's not widely known how we can compute statistical samples. 

In https://github.com/open-telemetry/opentelemetry-specification/issues/608 there is a long discussion about creating a new instrument to capture a sum of individual measurements and a distribution of those measurements. Hypothetical name: CountRecorder. With statistically-sampled exemplars, we can get approximately the same output from a simple Counter. Let me explain how.

Let's say I use the Priority Sampling algorithm, because it's very simple to implement: https://dl.acm.org/doi/10.1145/1314690.1314696. 

The Counter instrument will be configured to collect 1000 exemplars using weighted sampling, with no more than 10 exemplars per distinct label set. Today's SDK has two distinct phases: in the Accumulator the aggregator combines events for a single label set, then the Integrator combines those results.

In the Accumulator (aggregating per label set), the Sum aggregator is used with an additional size=10 sampler to select exemplars using weighted sampling. Each `Add(N)` event on the aggregator will call the sampler with Weight=N. After the interval is complete, each Aggregator will contain <=10 exemplars. The sampler gives me an _output_ weight for each individual, and the computed `sample_count` described above is equal to `output_weight / N`, i.e., the ratio of output weight to input weight.  If there are fewer than 100 distinct label sets, we can simply use the exemplars at this stage because there are fewer than 1000 in total. If there are more than 100 distinct label sets, the exemplars must be resampled (again using weighted sampling).

In the case where there were more than 100 distinct label sets, some will be missing from the output. This may sound like a source of error, but an unbiased sampling algorithm is expected to introduce errors. Averaging the results across multiple windows will correct these errors. I like this sort of approach because it limits cardinality but conveys approximate full dimensionality.

I'm happy to leave the optional "sample_weight" out for now, because someone will have to demonstrate it in code. This will only start to matter to me when OTLP is more widely used, so it can wait.

text/metrics/0113-exemplars.md
cnnradams: Added the sample_count parameter, and mentioned that an 
```
optional repeated RawValue exemplars = 6
``` 
parameter would be included on the metric_descriptor object. I wasn't sure about that originally because I was thinking in terms of the client SDK implementation (ie exemplars are computed in the aggregator, so how can it be pulled out?) and not in terms of the proto. It makes sense now, and I agree with it. 

I also revised the wording of the aggregations to allow for statistical sampling, and moved the 'preference for traces' out into a parameter.

text/metrics/0113-exemplars.md
cnnradams: @MrAlias added a section talking about custom aggregations

@jmacd agreed with all of the points except for the last one, I think default behaviour might as well select exemplars statistically when possible (and use the parameter). I don't think we should go about defining statistical implementation details for the aggregations in this OTEP, but making sure we don't block it and aggregation descriptions encourage it is a reasonable goal, imo.

cnnradams:
I realize having both a doc and a markdown file is not a good idea since I have to update both. So going to make the markdown the source of truth and not update the doc anymore for consistency

text/metrics/0113-exemplars.md
jmacd: Sounds good.

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,104 @@
...
+  fixed64 time_unix_nano = 2;
+
+  // 'label:value' map of all labels that were provided by the user recording the measurement
+  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 3;
+
+  // Span ID of the current trace
+  optional string span_id = 4;
+
+  // Trace ID of the current trace
+  optional string trace_id = 5;
```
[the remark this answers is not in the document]
jmacd: +1. That occurred to me too.

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,104 @@
...
+## Internal details
+
+An exemplar is defined as:
+
+```
+message RawValue {
+  // Numerical value of the measurement that was recorded. Only one of these two fields is
+  // used for the data, depending on its type
+  double double_value = 0;
+  int64 int64_value = 1;
```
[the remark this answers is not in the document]
jmacd: I think it would have to be the same type as the descriptor of the enclosing message.

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,104 @@
...
+
+A Sketch aggregation SHOULD maintain a list of exemplars whose values are spaced out across the distribution. There is no specific number of exemplars that should be retained (although the amount SHOULD NOT be unbounded), but the implementation SHOULD pick exemplars that represent as much of the distribution as possible. (Specific details not defined, see open questions.)
+
+#### Last-Value
+
+Most (if not all) Gauges operate asynchronously and do not ever interact with traces. Since the value of a Gauge is the last measurement (essentially the other parts of an exemplar), exemplars are not worth implementing for Gauge.
+
+#### Exact
+
+The Exact aggregation will function by maintaining a list of `RawValue`s, which contain all of the information exemplars would carry. Therefore the Exact aggregation will not need to maintain any exemplars.
```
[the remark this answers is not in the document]
jmacd: Yes. We are trying to let "Aggregator" refer to the code that calculates an "Aggregation", which is a data value.

text/metrics/0113-exemplars.md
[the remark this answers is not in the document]
cnnradams: What would a `SpanContext` contain? Just the span_id and trace_id?

text/metrics/0113-exemplars.md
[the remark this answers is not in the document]
cnnradams: My bad, was misusing both definitions throughout the otep. fixed.

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,104 @@
...
+
+Custom aggregations MAY support exemplars by maintaining a list of exemplars that can be retrieved by exporters. Custom aggregations should select exemplars based on their usage by the connected exporter (for example, exemplars recorded for Stackdriver should only be retained if they were recorded within a sampled trace).
+
+Exemplars will always be retrieved from aggregations (by the exporter) as a list of RawValue objects. They will be communicated via a
+
+```
+optional repeated RawValue exemplars = 6
+```
+
+attribute on the `metric_descriptor` object.
```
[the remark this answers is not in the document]
cnnradams: Changed it to `Metric`, one level up from the actual aggregation data points as desired but not part of the descriptor (I think if we have metadata separate from the exemplar list that tells us how it was collected, it could go alone in the `MetricDescriptor`, but we don't right now so everything should be in `Metric`)

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,104 @@
...
+Exemplars provide a link between metrics and traces. Consider a user using a Histogram aggregation to track response latencies over time for a high QPS server. The histogram is composed of buckets based on the speed of the request, for example, "there were 55 requests that took 400-500 milliseconds". The user wants to troubleshoot slow requests, so they would need to find a trace where the latency was high. With exemplars, the user is able to get an exemplar trace from a high latency bucket, an exemplar trace from a low latency bucket, and compare them to figure out the reason for the high latency.
+
+Exemplars are meaningful for all aggregations where relevant traces can provide more context to the aggregation, as well as when exemplars can display specific information not otherwise shown in the aggregation (for example, the full set of labels where they otherwise might be aggregated away).
+
+## Internal details
+
+An exemplar is defined as:
+
+```
+message RawValue {
```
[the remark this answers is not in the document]
cnnradams: We redefined Exemplar as RawValue so we could use the same data type for all measurements instead of just exemplars (eg for exact aggregator). Will try to reword `An Exemplar is defined as` better

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,104 @@
...
+  // Numerical value of the measurement that was recorded. Only one of these two fields is
+  // used for the data, depending on its type
+  double double_value = 0;
+  int64 int64_value = 1;
+  
+  // Exact time that the measurement was recorded
+  fixed64 time_unix_nano = 2;
+
+  // 'label:value' map of all labels that were provided by the user recording the measurement
+  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 3;
```
jmacd, line 32: I've become concerned that the labels are going to be duplicated between a data point and its exemplars. If we have a structure where exemplars are siblings of the data points, that is. I've been wondering if we could have a structure like:
```
Metric {
  Descriptor
  Labeled {
     repeated Labels string
     repeated Int64DataPoint int64_data_points
     repeated DoubleDataPoint double_data_points
     repeated HistogramDataPoint histogram_data_points
     repeated SummaryDataPoint summary_data_points
     repeated Exemplar exemplars
  }
}
```

This way we would not repeat the common labels that were part of the aggregation in the exemplar and the data point.

jmacd:
There is high-level agreement that we should merge this PR and continue to debate the finer points as we update the protocol. @bogdandrutu @lzchen @jkwatson @cijothomas please consider.

text/metrics/0113-exemplars.md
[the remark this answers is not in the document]
cnnradams: Not sure I understand, #114 defines trace_id and span_id as parameters, right? How is that any different than what's done here?

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,104 @@
...
+
+The HistogramAggregator MUST (when enabled) maintain a list of exemplars whose values are distributed across all buckets of the histogram (there should be one or more exemplars in every bucket that has a population of at least one sample-able measurement). Implementations SHOULD NOT retain an unbounded number of exemplars.
+
+#### Sketch
+
+A Sketch aggregator SHOULD maintain a list of exemplars whose values are spaced out across the distribution. There is no specific number of exemplars that should be retained (although the amount SHOULD NOT be unbounded), but the implementation SHOULD pick exemplars that represent as much of the distribution as possible. (Specific details not defined, see open questions.)
+
+#### Last-Value
+
+Most (if not all) Last-Value aggregators operate asynchronously and do not ever interact with context. Since the value of a Last-Value is the last measurement (essentially the other parts of an exemplar), exemplars are not worth implementing for Gauge.
```
jkwatson, line 62: Probably shouldn't mention the `Gauge` here, since that's not currently an OTel term.

cnnradams: missed that one from when I switched earlier. fixed

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,104 @@
...
+
+The Exact aggregator will function by maintaining a list of `RawValue`s, which contain all of the information exemplars would carry. Therefore the Exact aggregator will not need to maintain any exemplars.
+
+#### Counter
+
+Exemplars give value to counter aggregations in two ways: One, by tying metric and trace data together, and two, by providing necessary information to re-create the input distribution. When enabled, the aggregator will retain a bounded list of exemplars at each checkpoint, sampled from across the distribution of the data. Exemplars should be sampled in a statistically significant way.
+
+#### MinMaxSumCount
+
+Similar to Counter, MinMaxSumCount should retain a bounded list of exemplars that were sampled from across the input distribution in a statistically significant way.
```
jkwatson, line 74: Would it make sense to always retain an exemplar for the min, and the max?

cnnradams: Keeping the min & max was the original definition when I first made the OTEP 😬 At some point I changed it to this definition for the sake of allowing the exemplars to be used statistically (consistent with other aggregators), which min + max would not work with.

I think a common theme that I've been seeing is there is two uses of exemplars that benefit from different ways of collecting them; one is viewing singular exemplars manually where we prefer having traces attached and exemplars more relevant to the aggregator, and the other is using a larger group of exemplars to determine information about the input distribution, common tags, etc, which requires statistically sampled exemplars. Maybe there is a way to satisfy both sides with one definition for aggregators like MMSC, but I can't think of it.

jkwatson: Heh. ok. I'm totally fine with this for now...it was just an idea (and obviously one that had already been considered!).  👍

text/metrics/0113-exemplars.md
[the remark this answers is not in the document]
jmacd: I mean that instead of having dedicated fields for SpanID and TraceID, we could use labels as in `spanid=12345678` and `traceid=2390487123984320`.

Logically, I think of these two fields as just labels. This would give us fewer fields.

text/metrics/0113-exemplars.md
jmacd: This is a fine point. (I'm still thinking about this.)

jmacd:
As discussed in the SIG call today, one of my latest concerns here is that the proposed structure for an Exemplar is nearly identical to the structure of `Int64DataPoint` and `DoubleDataPoint`. This is one reason I asked if `span_id` and `trace_id` could just be labels, because if you treat those as labels, there's only one new field that distinguishes an exemplar from an ordinary data point: the `sample_count`. 

I'm trying to think how I'd like to see these unified. This comment https://github.com/open-telemetry/oteps/pull/113#discussion_r439026331 brings me back to an earlier conversation about how to represent MinMaxSumCount aggregation in OTLP. One answer was to use a Summary with quantiles at 0 and 100%. But another answer would be to treat these four values as four data points: two of them are exemplars (MIN and MAX) and two of them are aggregations (COUNT and SUM). 

This has me thinking that we're missing some field in the OTLP structure to say the kind of aggregation or exemplar selection that produced data point. I'm starting to think that with two new fields in the Int64DataPoint and DoubleDataPoint structures, we could unify these ideas.

The structure would look like:

```
// Int64DataPoint is a single data point in a timeseries that describes the time-varying
// values of a int64 metric.
message Int64DataPoint {
  // The set of labels that uniquely identify this point (in addition to any common labels in 
  // the enclosing Metric). This may include "traceid" and "spanid" labels for correlation with
  // Tracing.
  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1;

  // If this is an aggregation, not a raw point, the timestamp of the beginning of the interval.
  // For Deltas: This will be the same as a prior data point's `time_unix_nano`
  // For Cumulatives: This will be the same as the prior data point's `start_time_unix_nano`
  fixed64 start_time_unix_nano = 2;

  // If this is a raw point, the timestamp when the point was generated.
  // If this is an aggregation, this is the timestamp at the end of the interval.
  fixed64 time_unix_nano = 3;

  // value itself.
  int64 value = 4;

  // If this is a raw point, this indicates the representivity
  double sample_count = 5;

  // THE KIND OF POINT IS NEW HERE
  PointKind point_kind = 6;
}

enum PointKind {
   // AGGREGATIONS

   SUM
   COUNT

   // EXEMPLARS

   // Sample is a statistically meaningful exemplar, its sample_count is set.
   SAMPLE
   // MIN and MAX are not statistical samples, they have sample_count unset.
   MIN
   MAX
}
```

I wrote "(in addition to any common labels in the enclosing Metric)" about labels in these points, because I think it's likely that many of the entries for a Metric have some labels in common. The scenario I'm thinking of is where you've configured in-process aggregation by two label keys ("A" and "B") so that a number of aggregation values will have two labels in common. However, the exemplars will have all their labels, meaning they will repeat the ones used in the aggregation. It suggests a structure like in the comment here: https://github.com/open-telemetry/oteps/pull/113#discussion_r438950970

```
Metric {
  Descriptor
  repeated Labeled {
     // These are common labels
     repeated Labels string
     // Some of these points will have no extra labels, because they are aggregations
     // and some of these points will have extra labels, because they are exemplars associated
     // with the common labels used in the aggregations.
     repeated Int64DataPoint int64_data_points
     ...
  }
}
```

text/metrics/0113-exemplars.md
jmacd: I'm resolving this remark in favor of https://github.com/open-telemetry/oteps/pull/113#issuecomment-642992182

jmacd:
Regarding the above comment, I'd like to request input from people that have been involved with the Metric Views API both here as well as in OpenCensus. I do not see any way in the current OTLP protocol to encode information about which aggregation has been applied. In today's protocol how would you know the difference between a Sum and a LastValue aggregation? Possibly this is decided based on knowing whether the descriptor is a "grouping" or "adding" kind (which distinguishes Sum and LastValue), but this won't help us if we want to distinguish "LastValue" vs "RandomValue". 

Is something missing from the OTLP protocol? How does the suggestion above (https://github.com/open-telemetry/oteps/pull/113#issuecomment-642992182) look? @bogdandrutu @c24t @rghetia @cijothomas @lzchen

text/metrics/0113-exemplars.md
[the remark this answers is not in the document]
bogdandrutu: Hmm.. in case of Histogram/Summary we don't have that. Either add it or we need to do something else. But happy to discuss this in the proto change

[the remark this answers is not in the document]
cnnradams: Fair point, I think having an `InputType` or similar on the metric descriptor would make the most sense. It could also be different depending on if people approve of [this](https://github.com/open-telemetry/oteps/pull/113#issuecomment-642992182) comment or not. Happy to talk about this in proto as well

text/metrics/0113-exemplars.md
jmacd: OK, I have thoughts. In https://github.com/open-telemetry/opentelemetry-specification/issues/636 I've begun to propose that the default aggregator ValueRecorder become MinMaxLastSumCount, which I think of as being "like" 3 non-probabilistic exemplars (all extreme values) and two aggregations.  This could be represented as 5 aggregations, in the inexpensive case. It could be replaced by 2 aggregations and 3 exemplars. These would be the standard output of the instrument, not considering any additional randomly selected exemplars, which would have probability values associated with them.

jmacd: Therefore: I think this is a protocol question. I think it would be nice if we had the option to include full tags in "exemplar form" for MIN/MAX/LAST or not, depending on configuration, because the protocol made it so.

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,104 @@
...
+
+#### Custom Aggregators
+
+Custom aggregators MAY support exemplars by maintaining a list of exemplars that can be retrieved by exporters. Custom aggregators should select exemplars based on their usage by the connected exporter (for example, exemplars recorded for Stackdriver should only be retained if they were recorded within a sampled trace).
```
c24t, line 78:
```suggestion
Custom aggregators MAY support exemplars by maintaining a list of exemplars that can be retrieved by exporters. Custom aggregators should select exemplars based on their usage by the connected exporter (for example, exemplars recorded for Google Cloud Monitoring should only be retained if they were recorded within a sampled trace).
```

Nit, but the naming may not matter for the purpose of the OTEP.

c24t:
> I do not see any way in the current OTLP protocol to encode information about which aggregation has been applied. In today's protocol how would you know the difference between a Sum and a LastValue aggregation?

As I understand it, this is a known limitation of today's protocol and something we'll need to change to support nonstandard aggregations, with or without exemplars. The discussion from https://github.com/open-telemetry/opentelemetry-proto/issues/81, https://github.com/open-telemetry/opentelemetry-proto/issues/125, and https://github.com/open-telemetry/opentelemetry-proto/pull/137 is relevant here.

> This has me thinking that we're missing some field in the OTLP structure to say the kind of aggregation or exemplar selection that produced data point

I think we're missing something too, but I don't think the right solution is to overload `Int64DataPoint`/`DoubleDataPoint`. The structure is close to what we'd need for exemplars, but the semantics are different enough that I think it would cause more harm than good to shoehorn both kinds of data into the same type. I think we should prefer more types over interpreting the same type differently depending on `point_kind`.

It looks like there are a lot of questions about the implementation that need to be solved in the proto (Do we still need `Histogram` and `Summary` types? Should these types represent both raw and aggregated data? Do we need separate types for each aggregator, and if so how do we handle nonstandard aggregations?), but there's enough support for this description of exemplars to move from the OTEP to the spec.

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,104 @@
...
+
+Exemplars are implemented in [OpenCensus](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/Exemplars.md#exemplars), but only for HistogramAggregator. This OTEP is largely a port from the OpenCensus definition of exemplars, but it also adds exemplar support to other aggregators.
+
+[Cloud monitoring API doc for exemplars](https://cloud.google.com/monitoring/api/ref_v3/rpc/google.api#google.api.Distribution.Exemplar)
+
+## Open questions
+
+- Exemplars usually refer to a span in a sampled trace. While using the collector to perform tail-sampling, the sampling decision may be deferred until after the metric would be exported. How do we create exemplars in this case?
+
+- We don’t have a strong grasp on how the sketch aggregator works in terms of implementation - so we don’t have enough information to design how exemplars should work properly.
```
bsherrod, line 102: Apologies in advance if this is off-topic, but I'm trying to find a detailed spec on sketches in OT, and this is the closest I've found.

https://blog.acolyer.org/2019/09/06/ddsketch/ suggests at least one way that sketches can work, with code to implement found here: https://github.com/DataDog/datadog-agent/tree/master/pkg/quantile

The spec seems to be a bit vague about what the exemplar means in the case of sketches: is it the upper bound of the bucket?  the lower bound?   

Similarly, the content of the bucket, for histogram sketches, is ambiguous: are they cumulative (https://en.wikipedia.org/wiki/Histogram#Cumulative_histogram)?  For reference, Datadog histogram sketches are not cumulative, but Prometheus histogram sketches are cumulative (https://prometheus.io/docs/practices/histograms/#apdex-score).  It would be great if we could nail down these details in the spec, so that receivers and exporters can be implemented.

jmacd: Thank you @bsherrod. This is indeed not quite the topic of this OTEP, but it is very important and you are correct that there is something missing with regards to sketches. I am currently working on a few related matters. In the OTel-Go repository, there are several PRs associated with enabling exporters to choose whether they want cumulative or delta-oriented exports (the first attempt in a series here: https://github.com/open-telemetry/opentelemetry-go/pull/799). In the specification, I'm responsible for finishing the SDK specification--there's a WIP but it's out-of-date and doesn't answer the question about cumulative vs. delta (see the [WIP](https://github.com/open-telemetry/opentelemetry-specification/pull/347).

These matters ^^^ are mostly about specification, and I think we're pretty close to resolving them. There is something that we need to do in OTLP to support sending sketches, and then work to support sketches in the collector. I have done some research, but not enough research into good ways of encoding sketch data structures. In the OTel-Go repository, I've used the DDSketch implementation to implement approximate quantiles for Summary exports, but this isn't really how we want to export sketches.

There are bound to be many ways to encode sketches, and the protocol will want to address that. One mechanism that I favor is T-digest: the output of T-digest can be expressed as a set of weighted data points, which might be a useful representation--similar to a histogram but explicitly an approximate structure. A set of weighted data points is also similar in representation to a set of exemplars, so I guess your question is not really off topic after all. (I would propose that "set of exemplars" is a viable way to represent a sketch aggregation, but without an algorithm I can point to for computing this, I'll just leave this an open question.)

jmacd: I would invite you, if you're interested, to file an issue about this for OTLP in the https://github.com/open-telemetry/opentelemetry-proto repository.

jmacd:
I think this is good to submit. My last comment on a potential restructuring that would avoid duplication of labels can be ignored, I think: https://github.com/open-telemetry/oteps/pull/113#issuecomment-642992182

I discussed the matter with several people who agreed that in the long run, OTLP should have a way to avoid duplication and repetition of label sets in general and that we can ignore this question for now.

text/metrics/0113-exemplars.md
bsherrod: Thank you @jmacd !  I'll read the links you've added here and follow up.

jmacd: Re: Delta vs Cumulative aggregations, here's some progress: https://github.com/open-telemetry/opentelemetry-go/pull/840

jmacd:
@bogdandrutu It's the same story with this PR. We have enough metrics approvers' approvals.

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,104 @@
...
+
+A Sketch aggregator SHOULD maintain a list of exemplars whose values are spaced out across the distribution. There is no specific number of exemplars that should be retained (although the amount SHOULD NOT be unbounded), but the implementation SHOULD pick exemplars that represent as much of the distribution as possible. (Specific details not defined, see open questions.)
+
+#### Last-Value
+
+Most (if not all) Last-Value aggregators operate asynchronously and do not ever interact with context. Since the value of a Last-Value is the last measurement (essentially the other parts of an exemplar), exemplars are not worth implementing for Last-Value.
+
+#### Exact
+
+The Exact aggregator will function by maintaining a list of `RawValue`s, which contain all of the information exemplars would carry. Therefore the Exact aggregator will not need to maintain any exemplars.
```
cijothomas, line 66: Exact aggregator stores every recorded measurements, but does it also have traceid/spanid as well?

cnnradams: yes since it stores them as `RawValue`s, which have a `trace_id` and `span_id` property

cijothomas: Ok. I think the spec about Exact aggregator is not finalized or this is an upcoming change.

The current proto doesn't show an option to store traceid/spanid.
https://github.com/open-telemetry/opentelemetry-proto/blob/master/opentelemetry/proto/metrics/v1/metrics.proto

cnnradams: My bad, I think I got your question backwards. I don't think there is any mention of Exact in the proto right now if I'm not mistaken. I realize now that it's a bit misworded, but essentially I'm trying to say in the otep that the proto would communicate Exact aggregators through a list of `RawValue`s, either through the list of exemplars on the metrics object or in some other way, but either way it would have a span + trace id.

text/metrics/0113-exemplars.md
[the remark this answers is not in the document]
jmacd: Continue here: https://github.com/open-telemetry/opentelemetry-proto/pull/159

text/metrics/0113-exemplars.md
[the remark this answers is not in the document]
jmacd: Continue here: https://github.com/open-telemetry/opentelemetry-proto/pull/159

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,104 @@
...
+
+The HistogramAggregator MUST (when enabled) maintain a list of exemplars whose values are distributed across all buckets of the histogram (there should be one or more exemplars in every bucket that has a population of at least one sample-able measurement). Implementations SHOULD NOT retain an unbounded number of exemplars.
+
+#### Sketch
+
+A Sketch aggregation SHOULD maintain a list of exemplars whose values are spaced out across the distribution. There is no specific number of exemplars that should be retained (although the amount SHOULD NOT be unbounded), but the implementation SHOULD pick exemplars that represent as much of the distribution as possible. (Specific details not defined, see open questions.)
+
+#### Last-Value
+
+Most (if not all) Gauges operate asynchronously and do not ever interact with traces. Since the value of a Gauge is the last measurement (essentially the other parts of an exemplar), exemplars are not worth implementing for Gauge.
```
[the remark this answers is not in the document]
jmacd: Continue here: https://github.com/open-telemetry/opentelemetry-proto/pull/159

text/metrics/0113-exemplars.md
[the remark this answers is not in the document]
jmacd: Continue here: https://github.com/open-telemetry/opentelemetry-proto/pull/159

text/metrics/0113-exemplars.md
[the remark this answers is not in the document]
jmacd: Continue here: https://github.com/open-telemetry/opentelemetry-proto/pull/159

text/metrics/0113-exemplars.md
jmacd: Continue here: https://github.com/open-telemetry/opentelemetry-proto/pull/159

jmacd:
This has enough metrics reviewer approvals. Please merge. We have to keep interns unblocked!

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,104 @@
+# Integrate Exemplars with Metrics
+
+This OTEP adds exemplar support to aggregations defined in the Metrics SDK.
+
+## Definition
+
+Exemplars are example data points for aggregated data. They provide specific context to otherwise general aggregations. For histogram-type metrics, exemplars are points associated with each bucket in the histogram giving an example of what was aggregated into the bucket. Exemplars are augmented beyond just measurements with references to the sampled trace where the measurement was recorded and labels that were attached to the measurement.
```
yurishkuro, line 7: > Exemplars are example data points for aggregated data.

Are they this "generic" thing, or are they "traces"? The proto schema below suggests the latter. For example, could I store "customer id" as exemplar, so that I could answer the question "which sample customer IDs have latencies in this bucket"?

text/metrics/0113-exemplars.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Prior art and alternatives
+
+Exemplars are implemented in [OpenCensus](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/Exemplars.md#exemplars), but only for HistogramAggregator. This OTEP is largely a port from the OpenCensus definition of exemplars, but it also adds exemplar support to other aggregators.
+
+[Cloud monitoring API doc for exemplars](https://cloud.google.com/monitoring/api/ref_v3/rpc/google.api#google.api.Distribution.Exemplar)
+
+## Open questions
+
+- Exemplars usually refer to a span in a sampled trace. While using the collector to perform tail-sampling, the sampling decision may be deferred until after the metric would be exported. How do we create exemplars in this case?
```
yurishkuro, line 100: can we address this question from the resolved discussion?

text/metrics/0113-exemplars.md
jmacd: The RawValue representation is a generic way to represent sampled metric events. There will some SDK-specific/custom selection logic that may decide to select only exemplars that have trace context, or they can decide to focus on the distribution of customer IDs.  The customer ID would be represented by a label value.

When the aggregator is a histogram:
The SDK can select samples using fixed-size uniform selection on a per-bucket basis, or it can select items probabilistically so as to produce an expected number of exemplars per bucket that is equal (the latter is likely to have better coverage in the case where there are empty buckets--this can be accomplished using Weighted Sampling and inverse-probability weights, for example).

Let's suppose you configure exemplar selection to choose 100 exemplars per bucket per collection period.  If the selection is unbiased and the sample_count fields are accurate, you will be able to summarize the contribution to each bucket by up to 100 customers.

Suppose it's a Counter producing a Sum aggregation, instead of a histogram. You could use exemplars selected from the Counter to summarize the contribution to a sum by customer ID. There are lots of ways to sample, and I believe this representation will support a large number of useful approaches.

text/metrics/0113-exemplars.md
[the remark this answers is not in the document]
jmacd: The WIP https://github.com/open-telemetry/opentelemetry-proto/pull/162 specifies that RawValue messages may be used in two ways. 
1. As exemplars that add additional information to SCALAR, HISTOGRAM, and SUMMARY data points.
2. As raw data in its own right.

The SDK spec discusses three types of aggregation that can be represented as a scalar value: Sum, LastValue, and Exact. The exact representation would use RawValues with sample_count == 1.

text/metrics/0113-exemplars.md
jmacd: I don't feel that this is the place to describe a fancy SDK approach to this problem. This question leads to arbitrarily complex approaches that are also found in the discussion about tail sampling itself. How should we decide to propagate a trace-is-sampled bit in-band when making child spans during a span lifetime? It's almost the same question.

A simple approach would be to maintain a per-span sample of metric events and buffer metric data until the span ends.

Another approach would use the statistics of the spans that are being selected by the tail sampler to form an unequal probability sampling scheme. Select sample metric events that are likely to be associated with spans that match the tail-sampling decision. E.g., if tail latency is used to select exemplars, and a high correlation is observed between latency and label X, then use label X to boost sample weight on metric events. This leads to a speculative approach where you try to choose exemplars that will have interesting traces.

text/metrics/0113-exemplars.md
yurishkuro: Customer ID as a label value would kill most metric backends. To me that's the whole point of exemplars, to allow associating samples of high-cardinality values with metrics. I am fine if we limit this high-cardinality dimension to trace IDs for now, but I am not seeing a "generic" solution here that would support exemplars on other dimensions.

cnnradams: I think the comment [above](https://github.com/open-telemetry/oteps/pull/113#commitcomment-39806707) had the same concern as you? Would adding correlation context as an attribute on `RawValue` that can have the customer ID solve the problem?

jmacd: The SDK has built-in support for aggregation so that high-cardinality labels can be eliminated before they reach most metric backends. The Sum, Histogram, or Summary that you export can be aggregated so that customer_id does not appear in the aggregation value. 

Exemplars selected from the same series of events (that were summarized without customer_id) can include the customer_id, and the exemplars may be used to approximate the distribution of customer_id and other dimensions that were aggregated away in the Sum, Histogram or Summary value. One of the nice properties of the approach described here is that by limiting the number of exemplars, we limit cardinality reported in a single collection interval. For example, you could select 100 exemplars and even if there are 1000 actual customer_ids, you will collect at most 100 distinct values, and if chosen probabilistically, we can expect to recover the customer_ids that were most representative of the actual distribution (i.e., the "heavy hitters", the top of the distribution).

I want to emphasize that the API and the Protocol should not discourage the use of high-cardinality metrics. Given that I see exemplars as exactly the tool for addressing high cardinality, I'm confused by:

> I am fine if we limit this high-cardinality dimension to trace IDs for now

What do you think we should restrict?

yurishkuro: The issue I have with this schema is that it makes no distinction between regular label dimensions (which should survive aggregation) and the exemplar dimensions like customer-id. Only trace id is explicitly separated as exemplar dimensions. That makes it very easy for a user to shoot themselves in the foot and send an explosion of dimensions to the backend. The only way to avoid it is by carefully defining custom aggregation rules in the SDK and explicitly defining which labels should be treated as time series dimensions vs. exemplar attributes. While it minimizes the API surface, I feel that it makes the API more dangerous to use. Why not allow specifying exemplar labels explicitly from the beginning, and keep them separate from regular labels?

jmacd: Thanks for clarifying. I understand the concern, now.

By the way, an earlier draft of the metrics API allowed the application writer to recommend aggregation dimensions, by the name "Recommended Keys". It was removed: https://github.com/open-telemetry/opentelemetry-specification/issues/463. The reason these keys were recommended is that we do not believe the author of the code knows which labels the system or the operator wants to monitor. If we ask the developer decide which dimensions are for aggregation and which are for exemplars, we make a semantic distinction out of a performance limitation (and not a universal one, as far as I know).

There is a practical reason to support arbitrary labels and deal with them through configuration: this is the natural thing to do when creating Metric events from Spans. Span attributes simply become Metric labels. We are adding a semantic convention to cover duration measurements: https://github.com/open-telemetry/opentelemetry-specification/pull/657

The span-to-metrics issue is discussed here: https://github.com/open-telemetry/opentelemetry-specification/issues/381

One way to address your concern would be to set the default to aggregate over zero dimensions, so that all labels are exemplar labels unless configured otherwise.

bogdandrutu:
@jmacd can I merge this or you need answers/actions for your comments?

text/metrics/0113-exemplars.md
jmacd: @bogdandrutu I would like your opinion on this topic. We introduced `RecommendedKeys()` to address a perceived need in Prometheus, since Prometheus clients actually enforce pre-declared label keys. We discovered that the Prometheus protocol does not have any such restriction, which made it appealing to remove a feature. In the (dog)statsd world, it's common to add labels as needed. In modern terminology, we had created (proposed) Metrics Processors named "defaultkeys" that would use the developer-provided recommended keys, and named "ungrouped" that would use all the keys when exporting metrics. Removing recommended keys brought us back to a single basic metrics processor.

With this proposal, we begin to see a "Sampler API" for metrics, that is one that takes a full set of labels, applies a sampling decision (whether to select an exemplar or not) and then returns the set of labels to use for aggregation

If we have a choice between:

(1) asking the user to choose which labels are significant for aggregation and which are not
(2) making it really easy to configure which labels are used for aggregation

I would absolutely prefer the second choice--whether aggregation is configured by a dynamic configuration API, by a static configuration API, or by hard-coding a View in your main() function, any of these should be viable and relatively easy, and all of these are more appealing to me than asking the user to distinguish two kinds of label.

text/metrics/0113-exemplars.md
jmacd: @cnnradams Would you be willing to add an answer to this question? If head sampling, the logic for selecting trace contexts that are also being sampled is simple. If tail sampling, the logic for selecting metric samples has to be coordinated with tracing, delayed, or somehow speculative--and this decision is practically the same as deciding to what to tell your child in a span before the span is finished.

cnnradams: I missed this 😬 

To the depth that this OTEP goes, yes, this question is answered by "either the tail sampler needs to pick traces with exemplar choices in mind, or exemplars will need to be picked without a guarantee that they will have a trace". But the actual details of this still need to be worked out as far as I'm aware. I can't really mark this as answered now that its merged, so 🤷 

> Another approach would use the statistics of the spans that are being selected by the tail sampler to form an unequal probability sampling scheme.

How would you have knowledge of the spans that were chosen to be sampled when that decision was made in a different process without your input?

text/metrics/0113-exemplars.md
jmacd: @bogdandrutu Thanks for merging, but I think we should capture this discussion or at least address the question.

PR #114, opened 2020-06-08
add conventions for log correlation

zenmoto:
This OTEP is to add basic conventions for correlating logs with traces and resources by modifying existing logging formats.


text/logs/0114-log-correlation.md
```diff
@@ -0,0 +1,137 @@
...
+
+Two types of correlation need to happen to tie a log record into its full 
+execution context. The first is Request Correlation, which ties the record to 
+operations that were occurring when the record was created. The second is 
+Resource Correlation, which associates the entry with where the event occurred 
+such as a host, a pod, or a virtual machine.
+
+### Request Correlation
+
+Trace correlation is achieved primarily with two values- a trace identifier and
```
tigrannajaryan, line 22: Are "Trace correlation" and "Request correlation" interchangeable and equivalent terms?

text/logs/0114-log-correlation.md
```diff
@@ -0,0 +1,137 @@
...
+current scope do not need to be added to the log entry- doing so would duplicate 
+transmission of those values. As one span is likely associated with several (or 
+many) log entries, it is more efficient to transmit those attributes with the 
+span once rather than many times with each log entry. As a result, for most 
+purposes the goal is to set the three values of traceid, spanid, and traceflags 
+into each log entry that should be associated with that span.
+
+| Field      | Required | Format  
+| :--------- | :------- | :--------------------------------------------------
+| traceid    | Yes      | 16-byte numeric value as Base16-encoded string
```
tigrannajaryan, line 39: This refers to base-16 encoded string. However, what if my logging medium is binary and allows proper representation of arbitrary by sequences? Do I still have to use base-16 strings or I can just emit the bytes? If this recommendation is for text medium (such as text log files) it may be worth calling out specifically.

text/logs/0114-log-correlation.md
```diff
@@ -0,0 +1,137 @@
...
+identifier and a span identifier corresponds to a specific scope of work. In the 
+tracing API, that scope can also contain attributes and events that describe 
+what work the program being traced was currently performing.
+
+In most cases when traces and logs are used in tandem, the attributes of the 
+current scope do not need to be added to the log entry- doing so would duplicate 
+transmission of those values. As one span is likely associated with several (or 
+many) log entries, it is more efficient to transmit those attributes with the 
+span once rather than many times with each log entry. As a result, for most 
+purposes the goal is to set the three values of traceid, spanid, and traceflags
```
tigrannajaryan, line 34: "traceflags" is not explained anywhere above. I think it would be useful to tell what it is about.

text/logs/0114-log-correlation.md
```diff
@@ -0,0 +1,137 @@
...
+result, resource correlation may not happen at the log entry level, so we may or 
+may not put the resource correlation information in the log entry itself.
+
+Resource information may be managed as part of the log ingestion process- for 
+example a Docker logging driver will know which container logs came from. 
+Full resource information may also not be available, as when logs are 
+aggregated by syslog or a similar system that has less context available to it. 
+As a result resource correlation information in the logs entries themselves 
+should be considered optional. If included, the resource information should 
+follow the semantic conventions for resources.
```
tigrannajaryan, line 67: It would be good to link to the semantic convention document so that it is clear this refers to OpenTelemetry conventions.

text/logs/0114-log-correlation.md
```diff
@@ -0,0 +1,137 @@
...
+log entries. When the context is embedded in a log entry, the key-value
+pairs should be placed in a ‘correlation’ namespace. Where key-value pairs are 
+supported, embed the correlation key as “correlation.key_name”. In JSON or 
+other formats that allow nested structures, the key-value pairs should be 
+placed in an object named ‘correlation’ at the top of the object.
+
+
+## Internal details
+
+### Encoding
```
tigrannajaryan, line 89: The content of this section are 2 examples. Perhaps use "Examples" as the section title instead of "Encoding"?

text/logs/0114-log-correlation.md
```diff
@@ -0,0 +1,137 @@
...
+extraction mechanism. These types of extractions may have the advantage of being 
+less verbose, but they also have the disadvantage of requiring setting up a 
+custom extraction process, and may be more fragile. Since this approach is 
+vendor-dependent, there is little guidance that can be provided by 
+OpenTelemetry.
+
+## Trade-offs and mitigations
+
+As a convention, adherence to these standards may have some variation, and some
+of this advice may be difficult to implement in certain circumstances.
```
tigrannajaryan, line 121: It is not quite clear what this means. What level of variation is acceptable? Typically, where it matters we use RFC kewords MAY, SHOULD, MUST, etc to specify how strict the particular part of the spec is and what variation is allowed.

text/logs/0114-log-correlation.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Prior art and alternatives
+
+Elastic Common Schema [has standards](https://www.elastic.co/guide/en/ecs/current/ecs-tracing.html#ecs-tracing) 
+for adding trace information to JSON log formats, but they do not support the 
+full OT correlation model
```
tigrannajaryan, line 127:
```suggestion
full OpenTelemetry correlation model.
```

text/logs/0114-log-correlation.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Open questions
```
tigrannajaryan, line 129: If there are no open questions best to remove the section.

text/logs/0114-log-correlation.md
zenmoto: At the moment sampled is the only flag defined in the W3C spec, so it's effectively the same.

text/logs/0114-log-correlation.md
zenmoto: I'll clarify, thanks.

text/logs/0114-log-correlation.md
```diff
@@ -0,0 +1,137 @@
...
+
+A Correlation Context is a set of key-value pairs that is shared amongst the 
+spans of a distributed trace. Like span attributes, the correlation context may 
+already be carried with span information, so duplicating this information may be 
+redundant. In certain cases it may be important to associate this context with 
+log entries. When the context is embedded in a log entry, the key-value
+pairs should be placed in a ‘correlation’ namespace. Where key-value pairs are 
+supported, embed the correlation key as “correlation.key_name”. In JSON or 
+other formats that allow nested structures, the key-value pairs should be 
+placed in an object named ‘correlation’ at the top of the object.
```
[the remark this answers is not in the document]
zenmoto: This document isn't trying to describe how the selection or decision to include correlation attributes happens, just where it should go if it *is* included. As to the `correlation.tenancy` vs `tenancy`, the intent is to be able to determine whether or not a field is part of the correlation context or not. I understand the concern about the extra bytes, but I'm not seeing a good solution that's shorter but still keeps the values grouped together and retains clarity of where those values are coming from (I don't want to change "correlation", but it doesn't really have any good abbreviations). I might be missing something simple here though.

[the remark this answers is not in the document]
yurishkuro: In order to make the selection, there must be some configuration mechanism. Part of that mechanism can be the renaming of the fields (with `correlation.blah` being the default).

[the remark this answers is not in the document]
zenmoto: This document is only describing the output, so there is no configuration here. We are looking for a convention that describes where we can look for them when we are reading log entries.

text/logs/0114-log-correlation.md
tigrannajaryan: I don't have an answer, no preference here. Not sure if have a standardized terminology for this already. Perhaps check the rest of the spec and see if there is one.
If there is none let's choose one and stick with it.

bogdandrutu:
@open-telemetry/specs-approvers @open-telemetry/specs-logs-approvers  please review this.

text/logs/0114-log-correlation.md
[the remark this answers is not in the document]
yurishkuro: I understand, but I am disagreeing with this convention as too bulky and not user-friendly. The users are both the producers and consumers of the correlation fields. Imposing a `correlation.` prefix doesn't present enough benefits for me.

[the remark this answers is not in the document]
zenmoto: I understand your point too. If we changed to 'context', would that add confusion? We could use a 'ctx' abbreviation if we did, I just don't want to introduce confusion.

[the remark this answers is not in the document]
yurishkuro: I could get behind `ctx.*` recommendation.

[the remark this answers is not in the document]
zenmoto: Changed.

PatrikSteuer:
https://github.com/open-telemetry/opentelemetry-specification/pull/1283 changes the format of `traceid`, `spanid`, `traceflags` to `trace_id`, `span_id`, and `trace_flags` which seems to be related.

kamalmarhubi:
it also seems that correlation context should be renamed to baggage? cf https://github.com/open-telemetry/opentelemetry-specification/issues/536 and https://github.com/open-telemetry/opentelemetry-specification/pull/857

text/logs/0114-log-correlation.md
```diff
@@ -0,0 +1,149 @@
...
+purposes the goal is to set the three values of traceid, spanid, and traceflags
+into each log entry that should be associated with that span.
+
+Request correlation fields only make sense when the log event occurred within
+an open span, so the fields in the table below are only required when the log
+event is to be correlated.
+
+| Field      | Required | Format
+| :--------- | :------- | :--------------------------------------------------
+| traceid    | Yes      | 16-byte numeric value as Base16-encoded string
```
owais, line 45: At least in Python, the log format requires all fields to be present on the log record object or it throws an exception. Also, the log format cannot be changes dynamically depending on whether an active span (or fields on log record) is present or not. So in this case we can either leave the fields empty so that the formatted logs look like `trace_id= span_id= ` or we can set them to zero so it looks like `trace_id=0 span_id=0`. Can we clarify what values the fields should be set to in case no active span is found in the context?

text/logs/0114-log-correlation.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Correlation Context
+
+A Correlation Context is a set of key-value pairs that is shared amongst the
+spans of a distributed trace. Like span attributes, the correlation context may
+already be carried with span information, so duplicating this information may be
+redundant. In certain cases it may be important to associate this context with
+log entries. When the context is embedded in a log entry, the key-value
+pairs should be placed in a 'ctx' namespace. Where key-value pairs are
+supported, embed the correlation key as “correlation.key_name”. In JSON or
```
yurishkuro, line 94:
```suggestion
supported, embed the correlation key as “ctx.key_name”. In JSON or
```

text/logs/0114-log-correlation.md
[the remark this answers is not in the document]
yurishkuro: Thanks. Resolving.

text/logs/0114-log-correlation.md
yurishkuro: @owais what would you prefer? Also, it seems like a fairly specific use case, could it be left to whoever configures the logger on how they want to represent the lack of trace context?

owais: I have a slight preference for `0`.

>  could it be left to whoever configures the logger on how they want to represent the lack of trace context?

It could but then backends will have to deal with multiple ways of representing nil values (0 vs '' vs null in JSON, etc). It would be nicer if the spec recommended one thing when the field cannot be omitted completely so analysis tools can have a standard way of detecting it.

tigrannajaryan:
Most of the proposals in this PR are already part of the specification. This OTEP is actually someone out of sync / behind with the specification. I suggest that we close the OTEP and any additional clarifications necessary are submitted as issues/PRs against specification repo.

mars64:
> Most of the proposals in this PR are already part of the specification. This OTEP is actually someone out of sync / behind with the specification. I suggest that we close the OTEP and any additional clarifications necessary are submitted as issues/PRs against specification repo.

I know this is an aged pull here, but just wanted to point out that at least as of today the [open-telemetry/opentelemetry-specification/specification/logs/overview.md](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/overview.md#trace-context-in-legacy-formats) doc still refers to this pull for definitions on log correlation behavior. 

For posterity, I _think_ what you're saying is that folks interested learning more/contributing to the log correlation spec should instead refer to and issue requests against the [data-model.md](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md) doc/area, is that right?

tedsuo:
@tigrannajaryan the spec still refers to this OTEP: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/README.md#trace-context-in-legacy-formats

Should we go ahead and approve it?

tigrannajaryan:
> @tigrannajaryan the spec still refers to this OTEP: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/README.md#trace-context-in-legacy-formats
> 
> Should we go ahead and approve it?

This OTEP's ideas were incorporated in to the spec, so it served its purpose, however the spec slightly deviated from this OTEP in nuances. I will close this OTEP and it can stay for posterity. The spec probably needs slight re-wording to ensure it does not imply that the OTEP is the source of truth.

tigrannajaryan:
> > @tigrannajaryan the spec still refers to this OTEP: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/README.md#trace-context-in-legacy-formats
> > Should we go ahead and approve it?
> 
> This OTEP's ideas were incorporated in to the spec, so it served its purpose, however the spec slightly deviated from this OTEP in nuances. I will close this OTEP and it can stay for posterity. The spec probably needs slight re-wording to ensure it does not imply that the OTEP is the source of truth.

@zenmoto do you think you can condense this OTEP into 1-2 paragraphs that we can incorporate into the spec? Linking to the OTEP is a bit misleading (e.g. OTEP uses `traceid`, but we went with `trace_id` in the spec), so I think it is best to avoid linking to the OTEP.

zenmoto:
The important parts of this OTEP have made their way into the spec. Closing this PR as it's outdated.

PR #115, opened 2020-06-11
[feat] Allow additional sampling hooks

toumorokoshi:
Hi,

This OTEP is trying to consolidate a few discussions around sampling and deferred span creation. There's a few issues that have been in the spec issues for a while, I thought it would be good to maybe raise this as an OTEP.


text/trace/115-add-sample-to-span-end.md
```diff
@@ -0,0 +1,182 @@
...
+The span's sampling decision would be determined by the intersection of the
+sampling decision made by either ShouldSample call:
+
+| First ShouldSample result | Second ShouldSample result | Final Result       |
+|---------------------------|----------------------------|--------------------|
+| NOT_RECORD                | N/A (Don't Call)           | NOT_RECORD         |
+| RECORD_AND_SAMPLED        | N/A (Don't Call)           | RECORD_AND_SAMPLED |
+| RECORD                    | NOT_RECORD                 | NOT_RECORD         |
+| RECORD                    | RECORD                     | RECORD             |
+| RECORD                    | RECORD_AND_SAMPLED         | RECORD_AND_SAMPLED |
```
yurishkuro, line 108: I would actually go the other way (as we do in Jaeger clients): No can be upgraded to Yes, but Yes cannot be downgraded.

text/trace/115-add-sample-to-span-end.md
```diff
@@ -0,0 +1,182 @@
...
+       result in a net increase of system resources relative to an approach that
+       allows optimizations.
+
+### Alternative: more hooks for the sampler to update it's decision
+
+[spec #307](https://github.com/open-telemetry/opentelemetry-specification/issues/307) brings up
+a few more hooks to modify sampling decisions. This is similar to the pros and
+cons of the proposal, with an added pro of requiring more incremental effort
+on determining sampler decisions, and a con of requiring more complexity on
+the sampler's part.
```
yurishkuro, line 165: I described another important pro of multi-hook approach: https://github.com/open-telemetry/opentelemetry-specification/issues/307#issuecomment-642954936

text/trace/115-add-sample-to-span-end.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Internal details
+
+### Idempotence of samplers
+
+The calling of shouldSample multiple times can result in inconsistent results
+if the sampler is truly random. As such, the samplers will need to be idempotent
+based on the input passed. For example, a probability sampler would need to
+use data present in the passed parameters, such as the span id, to return sampling
+decisions.
```
yurishkuro, line 118: not necessarily. In Jaeger, the [sampling decision][] explicitly contains a `Retryable` flag. A probabilistic sampler would return `retryable=false`, because it does not depend on the content of the span and therefore there's no point calling it more than once (in fact it's detrimental since it will skew the weight of the probability, which breaks inference of other statistics). We also support a [PrioritySampler][] that can combine multiple simpler samplers and short-circuit their execution once all retryable decisions have been exhausted. In practice this allows efficient sampling in the simplest cases, while also supporting customized sampling policies.

[sampling decision]: https://github.com/jaegertracing/jaeger-client-go/blob/master/sampler_v2.go#L18
[PrioritySampler]: https://github.com/jaegertracing/jaeger-client-go/blob/f2bdef7faf35d024087d9225d88a202a86e21a31/x/priority_sampler.go#L66

text/trace/115-add-sample-to-span-end.md
toumorokoshi: I think that works in most cases, with the exception of NOT_RECORD. 

As I understand it, some implementations are taking a NOT_RECORD initial decision as a way to insert in a noop span, which will not be able to record further values or changes. So that is irreversible:

https://github.com/open-telemetry/opentelemetry-specification/issues/530#issuecomment-605980247

Assuming that constraint, the only thing that changes in the table is that RECORD + NOT_RECORD => RECORD

yurishkuro: > As I understand it, some implementations are taking a NOT_RECORD initial decision as a way to insert in a noop span

That's fine in the implementation does not support up-sampling at all, as in the most basic implementation of upfront sampling where the first decision is permanent. If they want to support _any form of delayed sampling_, noop span simply doesn't make sense.

Allowing partial downsampling (excluding some subtree from the trace) makes it pretty difficult to correctly interpret effective sampling probability when deriving statistics from sampled data.

jmacd: I thought one of the primary goals of the sampling API was to support attaching a noop span some of the time for users that do not want to cost of delayed sampling. What I see here works as intended.

yurishkuro: the mechanism in Jaeger that I described does not prevent the use of noop, because of the `retryable` flag.

jmacd: It sounds like we should be thinking of the first decision as 3-valued (yes, no, maybe), where "maybe" corresponds with a retryable No decision in Jaeger?

I support this proposal and I think it will resolve https://github.com/open-telemetry/opentelemetry-specification/issues/620 too.

yurishkuro: yes, with the caveat that `retryable` in Jaeger is local to the process, Jaeger does not support "maybe" between processes like Zipkin does (although without reverse propagation I don't know what benefit Zipkin gets from "maybe"). So if we want to support "maybe" between processes, it's a different design discussion from what `retryable` provides in Jaeger samplers.

jmacd: Great. I was not considering cross-process coordination in this way. A good solution that problem would be quite a sight.

text/trace/115-add-sample-to-span-end.md
toumorokoshi: I'll amend this section with that option.

text/trace/115-add-sample-to-span-end.md
toumorokoshi: Just an FYI I've read through this thread and trying to gather a few thoughts around it.

text/trace/115-add-sample-to-span-end.md
toumorokoshi: added to the otep!

text/trace/115-add-sample-to-span-end.md
toumorokoshi: done.

toumorokoshi: I was doing some thinking of this concept, and I'm debating adding this to the OTEP, as it might be required to ensure optimal performance.

The clear use case is for a sampler that is truly random, doesn't depend on the contents of the span, and makes the decision once and doesn't need to make it again.

The second use case is when there is some level of complex logic to determine a span decision. For example, a sampler that depends on various http route attributes to rate limit them (e.g. http path + method + status code). Assuming that the final interface does provide more hooks for samplers to update their decision (e.g. on every attribute update), then it would be valuable to notify the span when no further shouldSample calls should be made. 

This could be deferred to a further OTEP, but to your point this does allow more complex chaining of samplers as you described with PrioritySampler.

text/trace/115-add-sample-to-span-end.md
toumorokoshi: So I think that, as has been alluded, there is no robust solution for ensuring that a mutating samplling decision can be honored in all child spans, propagated or not. At least with this type of sampling. Aside from samplers who make decisions on variables are independent of the contents of the span, there will be spans that will not have all attributes until right when the span should end, which in turn means that samplers could be written that could modify sampling decisions right up until the end of the span, well after all children have been created.

A good example I can come up with is a sampler that leverages HTTP status code: that will not be known until after all application code, which means that a modified sampling decision when that attribute is set will be well after all remote requests.

If one wants to enable significant flexibility to the sampling logic, I think that has to be handled at the tail, which has the ability to see all spans and attributes in a trace, and make a decision from that. 

So anything we do here is best-effort. And I do believe there's a class of samplers that can do an almost perfect job: ones that depend on attributes that are set before business logic. For example, samplers that are dependent on http method + http path.

With all that said, I think I'm going to change this proposal to be very similar to https://github.com/open-telemetry/opentelemetry-specification/issues/307, effectively enabling updated decisions when any data around the span has been modified. And I'll clarify that this cannot ensure consistent sampling decisions across all spans in a trace, as is the case today.

> It sounds like we should be thinking of the first decision as 3-valued (yes, no, maybe), where "maybe" corresponds with a retryable No decision in Jaeger?

I'm going to do some thinking around the most appropriate semantics, but I do believe something that states a tentative result (needed for propagation) along with some ability to state whether the decision should be re-evaluated, is needed.

jmacd: I think we could add a little detail about sampling probability -- in the case where an in-band message is sent in one of these speculative ("maybe") states, you may only know an upper bound on the sampling probability. For example, a head sampler is configured with 10% sampling and a tail sampler is configured to further adjust the final probabilities. The in-band sampling probability (0.1) is an upper bound.  Sampling probability can be adjusted down (toward greater "sample count") by a later decision, but the probability cannot be adjusted up (toward lower "sample count").

toumorokoshi: Can you clarify the value of the sampling probability? Is that to provide information for future samplers to make a more informed decision?

That feels like a valuable component (and something #107) might address, but unrelated to the introduction of a "maybe" state. I think to enable this OTEP of multiple sampling decision points, only the "maybe" state is needed.

toumorokoshi: sorry to be clear I think it's a good idea and happy to write another OTEP after this one (or support an existing one), but want to keep scope specific to the problem to help drive a decision.

jmacd: It's fine to leave this detail out--I wasn't trying to derail your progress. I was anticipating that someone would raise an objection that we don't know what value to propagate for sampling probability when the in the "maybe" state, but I think we can answer that question. The value of propagating sampling probability is to support some kind of rate limiting in the downstream service.

toumorokoshi: yes, makes sense!

I've updated the document to advocate for @yurishkuro's suggestion of a retryable value instead (which I've named shouldRetry based on the expected behavior, but that's up for debate).

The rationale was due to the concern about what "MAYBE" should be. I view there being design tradeoffs with the two choices:

* MAYBE -> propagate yes results in more spans than desired sent to the backend. This is a good idea for systems that favor completeness over the span consumer's resource utilization.
* MAYBE -> propagate no results in fewer spans sent to the backend, effectively missing spans. This is better for systems that favor resource utilization over completeness.

As such, I would say it's better to leave that decision as an exercise for the span processor, and have the SDK enable both options.

text/trace/115-add-sample-to-span-end.md
```diff
@@ -0,0 +1,204 @@
...
+
+These methods will consume both the span, as well as the field or state that
+was changed. This enables optimizations in the case that samplers only depend
+on the data in the modification to make a decision, as well as more complex
+sample checks that may require all fields of the span.
+
+onEnd is also provided, as an optimization for samplers who may opt for
+running the sample decision after no further updates to the span are possible.
+
+## Samples return shouldRetry
```
yurishkuro, line 117:
```suggestion
## Samplers return shouldRetry
```

text/trace/115-add-sample-to-span-end.md
```diff
@@ -0,0 +1,204 @@
...
+
+Samplers will return an additional "shouldRetry" boolean value, that indicates
+whether the sample should be called again. Once shouldRetry has returned false,
+further sample hooks will not be called.
+
+shouldRetry is only valid when paired with a RECORD or RECORD_AND_SAMPLED result.
+A NOT_RECORD result results in no further calls to the sampler, regardless of
+the shouldRetry value.
```
yurishkuro, line 125: This is the opposite of how Jaeger samplers behave - a YES decision is final, a NO decision may be retryable.

toumorokoshi: Yes, this is different a bit from Jaeger. The intention was to enable the SDK to make a decision around how to handle "MAYBE" decisions (whether to consider them a Yes or No), as well as allow the SDK / Sampler make the decision on whether a Yes or No is final.

More rationalle here: https://github.com/open-telemetry/oteps/pull/115#discussion_r442645146

toumorokoshi:
@jmacd @yurishkuro I'd like to do whatever I can to help move this forward. I'm itching to do some other OT stuff but don't want to leave work unfinished.

Would you mind giving more feedback and calling out what you feel like are blockers, or approving? I feel your approval would go a long way in the viability of this OTEP.

yurishkuro:
@toumorokoshi I like the overall direction, but my concern is with the handling of what you called "MAYBE" state. Specifically, per https://github.com/open-telemetry/oteps/pull/115/files#r446273010, I would not be able to implement Jaeger SDK behavior, because this OTEP takes the opposite approach, without conclusive arguments imo. What if we made this aspect configurable?

> shouldRetry is only valid when paired with a RECORD or RECORD_AND_SAMPLED result.
A NOT_RECORD result results in no further calls to the sampler, regardless of
the shouldRetry value.

toumorokoshi:
> @toumorokoshi I like the overall direction, but my concern is with the handling of what you called "MAYBE" state. Specifically, per https://github.com/open-telemetry/oteps/pull/115/files#r446273010, I would not be able to implement Jaeger SDK behavior, because this OTEP takes the opposite approach, without conclusive arguments imo. What if we made this aspect configurable?
> 
> > shouldRetry is only valid when paired with a RECORD or RECORD_AND_SAMPLED result.
> > A NOT_RECORD result results in no further calls to the sampler, regardless of
> > the shouldRetry value.

Thanks @yurishkuro for looking!

I think I personally am ambivalent about the concept of recording / not recording, but the rationalle for this behavior was to allow the usage of noop spans if the initial decision is to not record. I'm having trouble finding the original ticket, I'll see if I can find it.

Looking through the Jaeger code linked, I don't think the concept of "record" exists: A span is never replaced with a dummy version, it is always recording values.

If that is the case, I think one could just not use the "NOT_RECORD" state at all in a sampler to achieve a Jaeger client-like behavior. The Jaeger-style sampler would return on the following decisions:

- "RECORD" -> sampled: False, retryable: True
- "RECORD_AND_SAMPLED" -> sampled: true, retryable: False

## Alternative: remove NOT_RECORD state

An option is to also just remove the concept of a NOT_RECORD state, and by extension not inject noop spans at all. I personally feel this is a good idea for other reasons:

- SDK users having to reason about edge cases where attributes are not recorded or stored
- seeing inconsistent performance of functions based on recording decision)

Either way I believe Jaeger-client style behavior is achievable.

yurishkuro:
The reason Jaeger does not use true no-op spans is because regardless of the sampling decision for trace collection users usually still want unique trace ID for every trace, which can be used for other correlation purposes (e.g. tagging logs). I don't see how that's possible to achieve with no-op spans unless one still allocates unique instances of no-op spans pointing to distinct `SpanContext`'s (which is not how no-op works in OpenTracing, there no-op span is a singleton that does not incur allocation cost). Jaeger SDKs do support _effective no-op_ because when the span is not sampled, all its write operations are short-circuited to do nothing (but it does require an atomic read of a flag, so slightly more expensive than real no-op span). Strictly speaking, it's possible to implement another span type that has true no-op functions, as long as the decision is known at span creation.

> Looking through the Jaeger code linked, I don't think the concept of "record" exists: A span is never replaced with a dummy version, it is always recording values.

That's not completely accurate - Jaeger SDKs have another hidden sampling state called "finalized", which is set to true as soon as the sampler returns non-retryable decision, which makes the most common cases and most common samplers much more efficient, since a finalized sampling means the write ops (e.g. span.setAttribute) do not need to be echoed into sampler callbacks anymore. I think it's kind of related to RECORD mode in OTel:

| Jaeger: `decision.sample` | Jaeger: `decision.retryable` | OTEL decision |
| ---- | ------- | -------- |
| false | true | RECORD |
| false | false | NOT_RECORD |
| true | false | RECORD_AND_SAMPLED |

However, the additional differences are:
  * in Jaeger the sampled state is shared between all spans of a given trace that are not finished in the given process (which allows sampling decisions based on tags on children spans affect parent span)
  * the "finalized" state is explicitly set to true if the parent span context was remote, i.e. downstream services don't attempt to make changes to sampling (which means it's possible that parent A starts non-sampled, calls child B that does not record its data, then A gets an attribute that trips sampling to on, and child C will be recording/sampling data).

Is it worth including these in the OTEP?  The main objective of that design in Jaeger was graceful fallback onto "normal" behavior where you only have simplistic samplers saying yes/no and making it efficient (i.e. short-circuiting the rest of the callbacks).

toumorokoshi:
@yurishkuro thanks for the thorough reply!

I believe the Jaeger behavior as you described it is pretty much how OpenTelemetry works today:

- effective no-op spans: write short-circuit, but a custom SpanContext is passed to allow a unique traceId. 
- the choice for a no-op span is based on the initial sampling choice (which is the only path that will cause an OTEL decision to be "NOT_RECORD")

I think I understand your concern a bit better now: the sticking point is the ability for the Sampler to return back a "NOT_RECORD" result. If you want a no decision to be retryble, then I would just implement the Sampler to never return NOT_RECORD, so either "RECORD" or "RECORD_AND_SAMPLED".

But looking at the mechanics, I feel like this could all be simplified by having "record" by a calculation of the sample decision and the retryable decision. So slightly modifying your table, we could have is_recording be a function:

Jaeger: decision.sample / OTEL sampled | Jaeger: decision.retryable / OTEL shouldRetry | OTEL is_recording() return value
-- | -- | --
false | true | true
false | false | false
true | false | true
true | true | true

This would be a modification of the "SamplingResult" struct in the SDK. This would also still enable the mostly no-op spans that you described.

> Is it worth including these in the OTEP? The main objective of that design in Jaeger was graceful fallback onto "normal" behavior where you only have simplistic samplers saying yes/no and making it efficient (i.e. short-circuiting the rest of the callbacks).

I think the two additional behaviors you called out are really good additions to the spec, but can be tackled outside of this OTEP (which is just advocating for additional sampling hooks).

I'm happy to keep filing follow-up PRs and discussions to get all the behavioral issues smoothed out. Unless you feel like this is a blocker for your own approval, I'd like to commit to following up separately on those instead. 

Regarding the risks if we go that route:

- the shared sampling choice in a trace is not possible (or at least not defined in the current spec). To enable jeager-client like behavior, this would have to be addressed
- the "finalized" state being honored will be possible with the APIs exposed, since they will consume the span and thus the SpanContext. I.E. This behavior could be replicated by a custom sampler. To make it possible in the SDK as well, I think this would be a matter of adding a flag to the built-in samplers to honor the parent's sampling choices.

text/trace/115-add-sample-to-span-end.md
```diff
@@ -0,0 +1,204 @@
...
+- span links
+
+The sampler would incorporate additional hooks, one to match each modification
+of events on an existing span:
+
+* onUpdateName, called by updateName
+* onSetAttribute, called by setAttribute
+* onAddEvent, called by addEvent
+* onSetStatus, called by setStatus
+* onEnd, called at the end of the span
```
lzchen, line 107: Is calling `onEnd` equivalent to `Alternative: Always sample at the end`?

text/trace/115-add-sample-to-span-end.md
```diff
@@ -0,0 +1,204 @@
...
+This implementation should address the following issues:
+
+- [remove warning on trace.updateName](https://github.com/open-telemetry/opentelemetry-specification/issues/468)
+- [remove warning on trace.updateName PR](https://github.com/open-telemetry/opentelemetry-specification/pull/506)
+- [Allow samplers to be called during different moments in the Span lifetime](https://github.com/open-telemetry/opentelemetry-specification/issues/307)
+- [Sampling decision is too late to gain much performance](https://github.com/open-telemetry/opentelemetry-specification/issues/620)
+
+### Tradeoffs of additional sample hooks
+
+- Pro: enables setting the recording decision early, which can skip additional
```
lzchen, line 144: I'm a bit confused. How does modifying the sampling decision during one of the hooks (let's say onEnd) set the recording decision early, compared to right now in where the decision is made upon creation?

text/trace/115-add-sample-to-span-end.md
```diff
@@ -0,0 +1,204 @@
...
+       processing for instrumentations that use the isRecording field.
+- Pro: enables setting the recording decision early, which can be used
```
lzchen, line 146: How can the decision to use a cheaper Span implementation (such as noop span) upon a hook decision be made without having a span created already? Unless I am misunderstanding something, if we aren't using deferred span creation, wouldn't a "real" span be created if the initial sampling decision (upon starting the span) was set to true?

e.g.
```
CustomSampler():
...
should_sample(...):
   return True
...
onUpdateName(span, name):
    return False
...
tracer_provider = TracerProvider(sampler=CustomSampler())
span = tracer.start_as_current_span("oldname") -> always sampled, so real Span is created
span.updateName("new name") -> what happens here? The real Span is already created, so how do we "save on processing"? How is this "setting the recording decision early", compared to setting it on creation?
```

text/trace/115-add-sample-to-span-end.md
toumorokoshi: yes, although the alternative is lesser featured (since there's fewer hooks to update sampling decisions).

text/trace/115-add-sample-to-span-end.md
toumorokoshi: I would say this is a pro relative to other ways to address the problem above, rather than a pro vs the current solution.

To your point, the current spec sets the decision as early as possible.

tedsuo:
@toumorokoshi I'm closing this since the Sampler API has changed significantly since this OTEP was opened. Please update or open a new PR if you are still interested in this.

PR #116, opened 2020-06-16
Move 0097-log-data-model.md to text/logs directory

tigrannajaryan:
This places it in the right directory according to new
structure for this repo.


bogdandrutu:
Using admin power since this is just a move.

PR #117, opened 2020-06-17
Add Last value to MinMaxSumCount OTLP data point a.k.a. "MMLSC"

jmacd:
This proposal was the outcome of a discussion in https://github.com/open-telemetry/opentelemetry-specification/issues/636.

It combines LastValue aggregation into the existing MinMaxSumCount aggregator, which will support exporting Gauges to Prometheus and Statsd over OTLP.

### DIFF
--- a/text/metrics/0117-min-max-last-sum-count-aggregator.md
+++ b/text/metrics/0117-min-max-last-sum-count-aggregator.md
@@ -49,117 +49,69 @@ conveying quantile information.  The `Summary` value type is not
 mergeable in general, so it is problematic to use it with
 MinMaxSumCount aggregations, which are mergeable.  In general we do
 not want MinMaxSumCount values to be exposed as Prometheus Summary
-values by default, although this is the current behavior in the
-OpenTelemetry Collector.
+values by default, although this does not preclude using the
+SummaryDataPoint struct in the current draft of OTLPK.
 
-An OTLP value type for the new MinMaxLastSumCount aggregation will be
-added to the `Metric` as follows:
+This proposal is summarized by the following change to
+SummaryDataPoint.
 
 
 ```
-message Metric {
-  // [descriptor and 4 existing value types fields]
-
-  // MinMaxLastSumCount values go here:
-  repeated MinMaxLastSumCountDataPoint minmaxlastsumcount_data_points = 6;
-}
-
-// Defines a metric type and its schema.
-message MetricDescriptor {
-
-  // [name, description, and unit fields]
-
-  // Type is the type of values a metric has.
-  enum Type {
-    // [ ... 7 existing values ]
-
-    // MINMAXLASTSUMCOUNT_INT64 indicates a MinMaxSumCountDataPoint of int64 values.
-    MINMAXLASTSUMCOUNT_INT64 = 7;
-
-    // MINMAXLASTSUMCOUNT_DOUBLE indicates a MinMaxSumCountDataPoint of double values.
-    MINMAXLASTSUMCOUNT_DOUBLE = 8;
-  }
-
-  // [ ... ]
-}
-
-// MinMaxLastSumCountDataPoint is a mergeable summary of a timeseries of individual values.
-//
-// This type contains an important optimization to avoid sending min, max, sum, and count
-// values when there was a single value in the stream with count equal to 1.0, in which
-// case only the last-value field is needed.
-message MinMaxSumCountDataPoint {
-  // The set of labels that uniquely identify this timeseries.
-  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1;
-
-  // start_time_unix_nano is the time when the cumulative value was reset to zero.
-  // This is used for Counter type only. For Gauge the value is not specified and
-  // defaults to 0.
-  //
-  // The cumulative value is over the time interval (start_time_unix_nano, time_unix_nano].
-  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
-  //
-  // Value of 0 indicates that the timestamp is unspecified. In that case the timestamp
-  // may be decided by the backend.
-  fixed64 start_time_unix_nano = 2;
-
-  // time_unix_nano is the moment when this value was recorded.
-  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
-  fixed64 time_unix_nano = 3;
-
-  // last_time_unix_nano is the moment when the latest value was captured during
-  // the interval described by [start_time_unix_nano, time_unix_nano].
-  fixed64 last_time_unix_nano = 4;
-
-  // last_int64 is the last value as an integer.  This field MUST be set when the 
-  // metric descriptor type is MINMAXLASTSUMCOUNT_INT64.
-  int64 last_int64 = 5;
-
-  // min_int64 is the minimum value as an integer, when set.  If this field is
-  // not set and last_int64 is set, the minimum value equals the value of last_int64.
-  int64 min_int64 = 6;
-
-  // max_int64 is the maximum value as an integer, when set.  If this field is
-  // not set and last_int64 is set, the maximum value equals the value of last_int64.
-  int64 max_int64 = 7;
-
-  // last_double is the last value as an integer.  This field MUST be set when the 
-  // metric descriptor type is MINMAXLASTSUMCOUNT_DOUBLE.
-  double last_double = 8;
-
-  // min_double is the minimum value as an integer, when set.  If this field is
-  // not set and last_double is set, the minimum value equals the value of last_double.
-  double min_double = 9;
-
-  // max_double is the maximum value as an integer, when set.  If this field is
-  // not set and last_double is set, the maximum value equals the value of last_double.
-  double max_double = 10;
-
-  // sum is the sum of individual values captured during the interval
-  // described by [start_time_unix_nano, time_unix_nano].  This is a floating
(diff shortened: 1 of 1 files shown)


text/metrics/0117-min-max-last-sum-count-aggregator.md
```diff
@@ -0,0 +1,192 @@
...
+  // This MUST be omitted when the value is exactly 1.0.  When this value
+  // is not set, it MUST be assumed to equal 1.0.
+  double count = 12;
+}
+```
+
+This message type is specified so that:
+
+- the count can be omitted when it equals 1.0
+- the minimum and maximum can be omitted when they are equal to the last value
```
lzchen, line 161: Are these optimizations for lessening number of fields defined?

text/metrics/0117-min-max-last-sum-count-aggregator.md
```diff
@@ -0,0 +1,192 @@
...
+instruments, send data through OTLP, and expose the LastValue as a
+Gauge without loss of the information commonly used by other exporters
+(i.e., Min, Max, Sum, Count).
+
+## Internal details
+
+This proposal covers how the data should be represented in OTLP.
+
+Prior to this proposal, the MinMaxSumCount value has been exposed as a
+`Summary` value, which is a general-purpose value type meant for
```
lzchen, line 47: Are `Summary` values not able to be merged because of the percentiles?

text/metrics/0117-min-max-last-sum-count-aggregator.md
jmacd: Yes. My goal was to ensure that when there is only one value input, the corresponding OTLP contains only one value. This will apply to all asynchronous instruments and helps address a concern that MinMaxLastSumCount is inefficient for asynchronous instruments.

text/metrics/0117-min-max-last-sum-count-aggregator.md
jmacd: Yes. This is the reason why Summary values are generally discouraged in Prometheus--they are not mergeable while Histograms are.

text/metrics/0117-min-max-last-sum-count-aggregator.md
```diff
@@ -0,0 +1,192 @@
...
+  //
+  // This MUST be omitted when the value is exactly 1.0.  When this value
+  // is not set, it MUST be assumed to equal 1.0.
+  double count = 12;
+}
+```
+
+This message type is specified so that:
+
+- the count can be omitted when it equals 1.0
```
aabmass, line 160: I noticed it says count MUST be omitted above, should it be strict or no?

text/metrics/0117-min-max-last-sum-count-aggregator.md
aabmass: I just want to make sure I understand, it looks like there are only three different cases:
- Multiple different values were measured, need all of MMLSC
- Only one value measured, only need `last`
- Multiple of the same value measured, need `last` and  `count`

If that's right, would it be cleaner to strictly require the OTLP follows those conventions instead of saying "can be omitted"?

jmacd: > Multiple of the same value measured, need last and count

I was thinking more about a case where a single item has been sampled and its sample count is > 1. I realize now, as discussed in https://github.com/open-telemetry/opentelemetry-proto/issues/73, that it's probably clearer to separate an integer count of the actual number of data items from a floating point multiplier to indicate how much sampling was applied. I will rework this text and coordinate with https://github.com/open-telemetry/opentelemetry-proto/pull/162.

text/metrics/0117-min-max-last-sum-count-aggregator.md
jmacd: I agree with "strictly require the OTLP follows those conventions" above. I'll update this.

text/metrics/0117-min-max-last-sum-count-aggregator.md
```diff
@@ -0,0 +1,192 @@
...
+- the minimum and maximum can be omitted when they are equal to the last value
+- the sum value can omitted when it equals the last value times the count.
```
MrAlias, line 162: It seems like there could be implementations of this proto that are not able to distinguish between unset and a zero value right? If that is the case, it might not be a good idea to allow omission as it could introduce ambiguity.

text/metrics/0117-min-max-last-sum-count-aggregator.md
```diff
@@ -0,0 +1,192 @@
+# Replace MinMaxSumCount aggregator with MinMaxLastSumCount aggregator
```
MrAlias, line 1: Was the decision to put "Last" in the middle intentional? It's a bikeshed, but on first read it could be interpreted as "last sum". Wondering if `MinMaxSumCountLast` was considered?

bogdandrutu:
Please rebase and fix the CI

text/metrics/0117-min-max-last-sum-count-aggregator.md
MrAlias: Take away from the Metric SIG is MMLSC is fine.

jmacd:
I'd like to very carefully separate my response to this into two parts, one related to `ValueRecorder`, and one about `ValueObserver`. The discussion about `ValueObserver` is more difficult, so let's find agreement about `ValueRecorder` before moving on.

There has been an issue raised by multiple early adopters of OTel Metrics who noticed that a ValueRecorder would be exposed as a Summary in Prometheus.  For example, https://github.com/open-telemetry/opentelemetry-go/issues/875 and https://github.com/open-telemetry/opentelemetry-specification/issues/636.

Prometheus summary is not the desired default, and the alternatives are to ensure a Histogram or a Gauge (scalar) is the default. We are already explicitly choosing the less-expensive option for all instruments, which is why MMSC has been specified as the default for ValueRecorder. And there are good explanations for this, and some vendors really want this. If you're collecting metrics on a short time interval, the MMSC summary is the maximally-useful-fixed-size way to expose a distribution. Any quantile information other than 0 (min) and 1 (max) requires variable-size to encode precisely (in a mergeable way), so we defined MMSC as a fixed-size summarization.

But when we speak of summarizing distributions, we are generally eliminating the time dimension. We know the min, max, sum and count, but we know nothing about the relationship between these values and the time variable. Adding a last-value to the summarization adds one piece of information about the time dimension, with which:

- Prometheus gets the default its users want (L)
- Other vendors still get the information they want (MMSC, e.g., New Relic)
- Knowing the difference between Mean == (S/C) and Last (L) tells us a bound on variance (this is more technical than we really care about, but Chebyshev's inequality gives us some information here IIUC).

Now to your points:

> new type of aggregation

This is a change to MMSC, not a new aggregation. Adding one value to MMSC makes it MMLSC. We were already encoding MMSC using a Summary data point, L just adds one value in that encoding and this PR makes Min/Max fields to reduce the encode size.

> Arguments can be made to expose any combination of sum/count/min/max/last_value

I don't think that these arguments will succeed. The definition of MMLSC is that all five fields are always included in the Summary. I would state that any use of Summary must fill in all five fields. The use of quantiles / percentiles is optional, orthogonal, and complementary to MMLSC.

> This OTEP specify that Prometheus needs the last value (which is true), but it only needs that in some scenarios (most likely for ValueObserver)

I think you are implying that Histogram should be the default for ValueRecorder+Prometheus, but we haven't specified the default aggregation to depend on the exporter. So I assert that Prometheus wants the Last value by default in all scenarios.

If ValueRecorder is used to record latencies, recording the Last value gives meaningful information about the stream of events, but it does discard the sum, count, min, and maximum values. It's less information, and it's susceptible to systematic errors (e.g., due to a correlation between the collection event and the last value measured) but on a long time scale you can plot the last value and the result is meaningful. Not necessarily as useful as a histogram, but cheaper. Not necessarily as useful as a MMSC summary on short time scales, but more useful than a MMSC summary on long time scales.

So, I view this as a good compromise for ValueRecorder. I expect the additional value will compress reasonably well so that MMLSC encoding will not be much larger than MMSC. Users will have to configure ValueRecorder to use Histogram when they want the more-expensive option, otherwise exporters will use MMSC or L depending on the conventions in that system.

jmacd:
I will not write such a lengthy explanation for ValueObserver at this juncture, as the points in that debate are somewhat different, and I have one new idea that may help settle this. First we should decide on a term for ASYNCHRONOUS as discussed here: https://github.com/open-telemetry/opentelemetry-proto/pull/168#discussion_r456544105, after which I will argue that LastValue for an Observer instrument should be uniform-randomly selected from the observed values of a single callback. This preserves the semantics of LastValue used in the argument above ("recording the Last value gives meaningful, statistically-viable information about the stream of events").

text/metrics/0117-min-max-last-sum-count-aggregator.md
jmacd: I have simplified this proposal and made it a clear requirement to follow the optimization (i.e., encode only a single value when count == 1).

text/metrics/0117-min-max-last-sum-count-aggregator.md
jmacd: Done.

text/metrics/0117-min-max-last-sum-count-aggregator.md
jmacd: I simplified this proposal and clarified that we do not need to distinguish between omitted and zero values.

jmacd:
@open-telemetry/specs-metrics-approvers See also https://github.com/open-telemetry/opentelemetry-proto/pull/170

PR #119, opened 2020-06-19
Standard system metrics and semantic conventions

aabmass:
See https://github.com/open-telemetry/opentelemetry-specification/issues/651. This OTEP proposes some standard system metric names as well as semantic conventions for naming system/runtime metrics. This mostly follows the work done in #108 and the Collector. I left a few TODOs and open questions, the biggest things being standard runtime metrics and process metrics.

### DIFF
--- a/text/0119-standard-system-metrics.md
+++ b/text/0119-standard-system-metrics.md
@@ -57,72 +57,70 @@ The following standard metric names should be used in libraries instrumenting sy
 #### `system.cpu.`
 
 **Description:** System level processor metrics.
-|Name                  |Units  |Instrument       |Label Key|Label Values                       |
-|----------------------|-------|-----------------|---------|-----------------------------------|
-|system.cpu.time       |seconds|SumObserver      |state    |idle, user, system, interrupt, etc.|
-|                      |       |                 |cpu      |1 - #cores                         |
-|system.cpu.utilization|%      |UpDownSumObserver|state    |idle, user, system, interrupt, etc.|
-|                      |       |                 |cpu      |1 - #cores                         |
+|Name                  |Units  |Instrument Type  |Value Type|Label Key|Label Values                       |
+|----------------------|-------|-----------------|----------|---------|-----------------------------------|
+|system.cpu.time       |seconds|SumObserver      |Double    |state    |idle, user, system, interrupt, etc.|
+|                      |       |                 |          |cpu      |1 - #cores                         |
+|system.cpu.utilization|1      |UpDownSumObserver|Double    |state    |idle, user, system, interrupt, etc.|
+|                      |       |                 |          |cpu      |1 - #cores                         |
 
 #### `system.memory.`
 
 **Description:** System level memory metrics.
-|Name                     |Units|Instrument       |Label Key|Label Values            |
-|-------------------------|-----|-----------------|---------|------------------------|
-|system.memory.usage      |bytes|UpDownSumObserver|state    |used, free, cached, etc.|
-|system.memory.utilization|%    |UpDownSumObserver|state    |used, free, cached, etc.|
+|Name                     |Units|Instrument Type  |Value Type|Label Key|Label Values            |
+|-------------------------|-----|-----------------|----------|---------|------------------------|
+|system.memory.usage      |bytes|UpDownSumObserver|Int64     |state    |used, free, cached, etc.|
+|system.memory.utilization|1    |UpDownSumObserver|Double    |state    |used, free, cached, etc.|
 
 #### `system.swap.`
 
 **Description:** System level swap/paging metrics.
-|Name                    |Units|Instrument       |Label Key|Label Values|
-|------------------------|-----|-----------------|---------|------------|
-|system.swap.usage       |pages|UpDownSumObserver|state    |used, free  |
-|system.swap.utilization |%    |UpDownSumObserver|state    |used, free  |
-|system.swap.page\_faults|1    |SumObserver      |type     |major, minor|
-|system.swap.paging\_ops |1    |SumObserver      |type     |major, minor|
-|                        |     |                 |direction|in, out     |
+|Name                    |Units |Instrument Type  |Value Type|Label Key|Label Values|
+|------------------------|------|-----------------|----------|---------|------------|
+|system.swap.usage       |pages |UpDownSumObserver|Int64     |state    |used, free  |
+|system.swap.utilization |1     |UpDownSumObserver|Double    |state    |used, free  |
+|system.swap.page\_faults|faults|SumObserver      |Int64     |type     |major, minor|
+|system.swap.paging\_ops |ops   |SumObserver      |Int64     |type     |major, minor|
+|                        |      |                 |          |direction|in, out     |
 
 #### `system.disk.`
 
 **Description:** System level disk performance metrics.
-|Name                        |Units|Instrument |Label Key|Label Values|
-|----------------------------|-----|-----------|---------|------------|
-|system.disk.io<!--notlink-->|bytes|SumObserver|device   |(identifier)|
-|                            |     |           |direction|read, write |
-|system.disk.ops             |1    |SumObserver|device   |(identifier)|
-|                            |     |           |direction|read, write |
-|system.disk.time            |s    |SumObserver|device   |(identifier)|
-|                            |     |           |direction|read, write |
-|system.disk.merged          |1    |SumObserver|device   |(identifier)|
-|                            |     |           |direction|read, write |
+|Name                        |Units  |Instrument Type|Value Type|Label Key|Label Values|
+|----------------------------|-------|---------------|----------|---------|------------|
+|system.disk.io<!--notlink-->|bytes  |SumObserver    |Int64     |device   |(identifier)|
+|                            |       |               |          |direction|read, write |
+|system.disk.ops             |ops    |SumObserver    |Int64     |device   |(identifier)|
+|                            |       |               |          |direction|read, write |
+|system.disk.time            |seconds|SumObserver    |Double    |device   |(identifier)|
+|                            |       |               |          |direction|read, write |
 
 #### `system.filesystem.`
 
 **Description:** System level filesystem metrics. *I think usage/utilization should be consolidated into `system.disk`. Any filesystem specifics are probably scoped to the OS.*
-|Name                         |Units|Instrument       |Label Key|Label Values        |
-|-----------------------------|-----|-----------------|---------|--------------------|
-|system.filesystem.usage      |bytes|UpDownSumObserver|device   |(identifier)        |
-|                             |%    |                 |state    |used, free, reserved|
-|system.filesystem.utilization|     |UpDownSumObserver|device   |(identifier)        |
-|                             |     |                 |state    |used, free, reserved|
+|Name                         |Units|Instrument Type  |Value Type|Label Key|Label Values        |
+|-----------------------------|-----|-----------------|----------|---------|--------------------|
+|system.filesystem.usage      |bytes|UpDownSumObserver|Int64     |device   |(identifier)        |
+|                             |     |                 |          |state    |used, free, reserved|
+|system.filesystem.utilization|1    |UpDownSumObserver|Double    |device   |(identifier)        |
+|                             |     |                 |          |state    |used, free, reserved|
 
 #### `system.net.`
 
 **Description:** System level network metrics.
-|Name                       |Units|Instrument       |Label Key|Label Values                                                                                  |
-|---------------------------|-----|-----------------|---------|----------------------------------------------------------------------------------------------|
-|system.net.dropped\_packets|1    |SumObserver      |device   |(identifier)                                                                                  |
-|                           |     |                 |direction|transmit, receive                                                                             |
-|system.net.packets         |1    |SumObserver      |device   |(identifier)                                                                                  |
-|                           |     |                 |direction|transmit, receive                                                                             |
-|system.net.errors          |1    |SumObserver      |device   |(identifier)                                                                                  |
-|                           |     |                 |direction|transmit, receive                                                                             |
-|system<!--notlink-->.net.io|bytes|SumObserver      |device   |(identifier)                                                                                  |
-|                           |     |                 |direction|transmit, receive                                                                             |
-|system.net.connections     |1    |UpDownSumObserver|device   |(identifier)                                                                                  |
-|                           |     |                 |protocol |tcp, udp, [others](https://en.wikipedia.org/wiki/Transport_layer#Protocols)                   |
-|                           |     |                 |state    |[e.g. for tcp](https://en.wikipedia.org/wiki/Transmission_Control_Protocol#Protocol_operation)|
+|Name                       |Units      |Instrument Type  |Value Type|Label Key|Label Values                                                                                  |
+|---------------------------|-----------|-----------------|----------|---------|----------------------------------------------------------------------------------------------|
+|system.net.dropped\_packets|packets    |SumObserver      |Int64     |device   |(identifier)                                                                                  |
+|                           |           |                 |          |direction|transmit, receive                                                                             |
+|system.net.packets         |packets    |SumObserver      |Int64     |device   |(identifier)                                                                                  |
+|                           |           |                 |          |direction|transmit, receive                                                                             |
+|system.net.errors          |errors     |SumObserver      |Int64     |device   |(identifier)                                                                                  |
+|                           |           |                 |          |direction|transmit, receive                                                                             |
+|system<!--notlink-->.net.io|bytes      |SumObserver      |Int64     |device   |(identifier)                                                                                  |
+|                           |           |                 |          |direction|transmit, receive                                                                             |
+|system.net.connections     |connections|UpDownSumObserver|Int64     |device   |(identifier)                                                                                  |
+|                           |           |                 |          |protocol |tcp, udp, [others](https://en.wikipedia.org/wiki/Transport_layer#Protocols)                   |
+|                           |           |                 |          |state    |[e.g. for tcp](https://en.wikipedia.org/wiki/Transmission_Control_Protocol#Protocol_operation)|
 
 #### OS Specific System Metrics - `system.{os}.`
 
@@ -136,11 +134,10 @@ Runtime environments vary widely in their terminology, implementation, and relat
 
 #### Runtime Specific Metrics - `runtime.{environment}.`
 
-Runtime level metrics specific to a certain runtime environment should be prefixed with `runtime.{environment}.` and follow the semantic conventions outlined in [Semantic Conventions](#semantic-conventions).
+Runtime level metrics specific to a certain runtime environment should be prefixed with `runtime.{environment}.` and follow the semantic conventions outlined in [Semantic Conventions](#semantic-conventions). For example, Go runtime metrics use `runtime.go.` as a prefix.
+
+Some programming languages have multiple runtime environments that vary significantly in their implementation, for example [Python has many implementations](https://wiki.python.org/moin/PythonImplementations). For these languages, consider using specific `environment` prefixes to avoid ambiguity, like `runtime.cpython.` and `runtime.pypy.`.
 
 ## Open questions
 
-- Are there any common runtime metrics worth enumerating in [Standard Runtime Metrics](#standard-runtime-metrics---runtime)?
 - Should the individual runtimes have their specific naming conventions in the spec?
-- Process level metrics would be valuable, but how can one choose the processes to instrument? There is also the issue of labels for a process; if PID is included, the cardinality will grow unbounded since PIDs can always be different.
-- Establishing units for instruments of the same unit family e.g. seconds vs. milliseconds. [Prometheus recommends using base units for compatibility](https://prometheus.io/docs/practices/naming/#base-units).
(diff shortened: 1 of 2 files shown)


text/0119-standard-system-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+#### `system.disk.`
+
+**Description:** System level disk performance metrics.
+|Name                        |Units|Instrument |Label Key|Label Values|
+|----------------------------|-----|-----------|---------|------------|
+|system.disk.io<!--notlink-->|bytes|SumObserver|device   |(identifier)|
+|                            |     |           |direction|read, write |
+|system.disk.ops             |1    |SumObserver|device   |(identifier)|
+|                            |     |           |direction|read, write |
+|system.disk.time            |s    |SumObserver|device   |(identifier)|
```
james-bebbington, line 95:
```suggestion
|system.disk.time            |seconds|SumObserver|device   |(identifier)|
```
(to be consistent with above)

text/0119-standard-system-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+|                            |     |           |direction|read, write |
+
+#### `system.filesystem.`
+
+**Description:** System level filesystem metrics. *I think usage/utilization should be consolidated into `system.disk`. Any filesystem specifics are probably scoped to the OS.*
+|Name                         |Units|Instrument       |Label Key|Label Values        |
+|-----------------------------|-----|-----------------|---------|--------------------|
+|system.filesystem.usage      |bytes|UpDownSumObserver|device   |(identifier)        |
+|                             |%    |                 |state    |used, free, reserved|
+|system.filesystem.utilization|     |UpDownSumObserver|device   |(identifier)        |
```
james-bebbington, line 107:
```suggestion
|                             |     |                 |state    |used, free, reserved|
|system.filesystem.utilization|%    |UpDownSumObserver|device   |(identifier)        |
```

text/0119-standard-system-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+|system<!--notlink-->.net.io|bytes|SumObserver      |device   |(identifier)                                                                                  |
+|                           |     |                 |direction|transmit, receive                                                                             |
+|system.net.connections     |1    |UpDownSumObserver|device   |(identifier)                                                                                  |
+|                           |     |                 |protocol |tcp, udp, [others](https://en.wikipedia.org/wiki/Transport_layer#Protocols)                   |
+|                           |     |                 |state    |[e.g. for tcp](https://en.wikipedia.org/wiki/Transmission_Control_Protocol#Protocol_operation)|
+
+#### OS Specific System Metrics - `system.{os}.`
+
+System level metrics specific to a certain operating system should be prefixed with `system.{os}.` and follow the hierarchies listed above for different entities like CPU, memory, and network. For example, metrics pertaining to Linux inodes would appear under `system.linux.filesystem.inodes.*`, reusing the `filesystem` name proposed above.
```
james-bebbington, line 129: Do we care about metrics that are consistent over a number of operating systems, but not all, e.g. would we have?

- `system.linux.filesystem.inodes.*`
- `system.freebsd.filesystem.inodes.*`
- `system.osx.filesystem.inodes.*`

text/0119-standard-system-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+
+#### Runtime Specific Metrics - `runtime.{environment}.`
+
+Runtime level metrics specific to a certain runtime environment should be prefixed with `runtime.{environment}.` and follow the semantic conventions outlined in [Semantic Conventions](#semantic-conventions).
+
+## Open questions
+
+- Are there any common runtime metrics worth enumerating in [Standard Runtime Metrics](#standard-runtime-metrics---runtime)?
+- Should the individual runtimes have their specific naming conventions in the spec?
+- Process level metrics would be valuable, but how can one choose the processes to instrument? There is also the issue of labels for a process; if PID is included, the cardinality will grow unbounded since PIDs can always be different.
```
james-bebbington, line 145: Process metadata (PID, etc) should presumably be recorded as a resource so doesn't need to be defined here, but I think it would be useful to define a list of standard process metrics & labels.

text/0119-standard-system-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+- Establishing units for instruments of the same unit family e.g. seconds vs. milliseconds. [Prometheus recommends using base units for compatibility](https://prometheus.io/docs/practices/naming/#base-units).
```
james-bebbington, line 146: That list looks sensible to me

text/0119-standard-system-metrics.md
jrcamp: Something feels off about having the platform type in the metric name. Feels a little bit like when the unit type is in the metric name. Might be useful to have platform as a resource label though.

The only reason I could see for needing `system.$platform` is if there were conflicting metrics where they `system.foo` means something different on Linux than Windows but I think we're trying to avoid that situation anyway.

aabmass: I also am not sure on this. #108 mostly gives examples for runtimes like `runtime.java.*`, here's the relevant section:
> use prefixed metric names in cases where similar metrics have significantly different implementations across the breadth of all existing metrics

I agree it's not ambiguous to have `system.filesystem.inodes`, but is it valuable to have everything under the common top level `system.*` be relevant to all platforms?

jrcamp: If the metrics were _truly_ platform specific it might make sense to namespace the metric. But generally I think they fall into either Windows or non-Windows, as Jame's point out. Is it useful to distinguish between system.linux.filesystem.inodes and system.freebsd.filesystem.inodes? I'm not sure that is useful.

text/0119-standard-system-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+
+System level metrics specific to a certain operating system should be prefixed with `system.{os}.` and follow the hierarchies listed above for different entities like CPU, memory, and network. For example, metrics pertaining to Linux inodes would appear under `system.linux.filesystem.inodes.*`, reusing the `filesystem` name proposed above.
+
+### Standard Runtime Metrics - `runtime.`
+
+---
+
+Runtime environments vary widely in their terminology, implementation, and relative values for a given metric. For example, Go and Python are both garbage collected languages, but comparing heap usage between the two runtimes directly is not meaningful. For this reason, this OTEP does not propose any standard top-level runtime metrics. See [OTEP #108](https://github.com/open-telemetry/oteps/pull/108/files) for additional discussion.
+
+#### Runtime Specific Metrics - `runtime.{environment}.`
```
jkwatson, line 137: does `environment` here map roughly to `language runtime` ?  So, for Java you would use `runtime.jvm...` for instrument names?

text/0119-standard-system-metrics.md
```diff
@@ -0,0 +1,146 @@
+# Standard names for system/runtime metrics
```
jkwatson, line 1: I think we need to be super clear that we're specifying the names of the `instruments`, and the labels for the recordings, not actual metrics, in this proposal.

text/0119-standard-system-metrics.md
aabmass: Just to add a counterexample where metrics of the same name can vary in implementation, Linux apparently [calculates load average differently](https://en.wikipedia.org/wiki/Load_(computing)#Unix-style_load_calculation) than many other unix systems. Since the magnitude of metric values would be different across OSes, I think a separate namespace would be called for.

Just looking at more prior art as well, Prometheus Node Exporter has a [support table](https://github.com/prometheus/node_exporter#enabled-by-default) showing a lot of metrics are not available on all unixes. These aren't necessarily ambiguous so I suppose its ok to not namespace them.

text/0119-standard-system-metrics.md
aabmass: Yes, or for languages with implementations that differ a lot something like `runtime.pypy` vs `runtime.cpython`.

jkwatson: Probably worth an example, then, that demonstrates, this.  👍

text/0119-standard-system-metrics.md
aabmass: Definitely, but what is the difference exactly? Just looking at [this part of the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#metric-instruments) where it mentions the instrument names, I took it that the instruments are Counter, UpDownCounter, etc. and that a metric uses the instrument to take measurements.

jkwatson: When you acquire an Instrument from a Meter, you give it a name. This generally ends up, at the end of the metric pipeline, resulting in metrics with the same name as the instrument. But, they're not the same thing. Most importantly, there is nothing in the API called "Metric", just Instruments.  

See here for the rules for assigning a name to an instrument: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#instrument-naming-requirements

text/0119-standard-system-metrics.md
jrcamp: Yeah I'm not sure what to do about load. I just feel like if most metrics start with system.{filesystem,cpu,etc.} because they're normalized but some start with system.$platform it'll be kind of weird but I hear you. I don't have a good answer.

jmacd: Is load average something that people really want to monitor? I would prefer to monitor utilization (i.e., rate of CPU seconds used per second), a value I expect to range between 0 and NumCPUs.

text/0119-standard-system-metrics.md
jlegoff: I also found it confusing that this OTEP references the host metrics receiver, but if I look at the [code](https://github.com/open-telemetry/opentelemetry-collector/blob/master/receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper.go) I don't see instruments being used. Would it make sense to use instruments there?

text/0119-standard-system-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+  * Makes use of labels, similar to the Collector.
+  * [Overview of collected metrics](https://docs.google.com/spreadsheets/d/1r50cC9ass0A8SZIg2ZpLdvZf6HmQJsUSXFOu-rl4yaY/edit#gid=0).
+- **TODO: Java**
+- **TODO: Opencensus**
+
+## Semantic Conventions
+
+The following semantic conventions aim to keep naming consistent across different metrics. Not all possible metrics are covered by these conventions, but they provide guidelines for most of the cases in this proposal:
+
+- **usage** - an instrument that measures an amount used out of a known total amount should be called `entity.usage`. For example, `system.filesystem.usage` for the amount of disk spaced used. A measure of the amount of an unlimited resource consumed is differentiated from **usage**. This may be time, data, etc. *(I'm open to adjusting this to unit names or something else).*
```
jlegoff, line 43: I know it's a trick question, but what do we call an entity? Is there any relationship with the resource concept also defined in OTEL? If that's the case, should we also add the resource type to the prefix? For instance, `host.system.filesystem.usage`.

text/0119-standard-system-metrics.md
jkwatson: The collector consumes metrics, after being aggregated, not instruments, so I think that still makes sense.

text/0119-standard-system-metrics.md
jmacd: Metric events carry a Resource just the same as Spans do. Resource attributes are merged with metric event labels in the export pipelines, so in your example you might have a `host=hostname` label.

text/0119-standard-system-metrics.md
jmacd: (This is somewhat off topic.)

The idea of using instruments in the collector sounds a lot like using the OTel-Go SDK to generate metrics inside the collector, which could then be "exported" from the SDK into a collector processing pipeline. I don't think this will be worth the complexity, however in the discussion about adding a [Statsd receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/290) we are consuming _unaggregated_ metrics and want to feed them into a collector processing pipeline. 

(We have discussed for Statsd how you could use parts of the OTel-Go SDK to implement this aggregation, mainly the `Aggregator` classes are re-usable. However, I'd prefer not to perform this aggregation in the Statsd receiver--instead I'd like to leverage the proposed [Metrics Transform Processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/332).)

text/0119-standard-system-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+---
+
+#### `system.cpu.`
+
+**Description:** System level processor metrics.
+|Name                  |Units  |Instrument       |Label Key|Label Values                       |
+|----------------------|-------|-----------------|---------|-----------------------------------|
+|system.cpu.time       |seconds|SumObserver      |state    |idle, user, system, interrupt, etc.|
+|                      |       |                 |cpu      |1 - #cores                         |
+|system.cpu.utilization|%      |UpDownSumObserver|state    |idle, user, system, interrupt, etc.|
```
jmacd, line 64: By "%", do you mean values are in the range 0-100?
I think we should use the ratio value, range 0-1. I think this falls under the rule used elsewhere in this document:

> Prometheus recommends using base units for compatibility

text/0119-standard-system-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+|system.swap.page\_faults|1    |SumObserver      |type     |major, minor|
+|system.swap.paging\_ops |1    |SumObserver      |type     |major, minor|
+|                        |     |                 |direction|in, out     |
+
+#### `system.disk.`
+
+**Description:** System level disk performance metrics.
+|Name                        |Units|Instrument |Label Key|Label Values|
+|----------------------------|-----|-----------|---------|------------|
+|system.disk.io<!--notlink-->|bytes|SumObserver|device   |(identifier)|
```
jmacd, line 91: I wonder how often a user will meaningfully aggregate this metric and combine both the `read` and `write` values. It's a question of whether this is a meaningful enough aggregation that we should use a single metric. I'd lean in the other direction, myself. Could we use one metric for read and one for write? (The same for all the bi-directional metrics in this document.)

text/0119-standard-system-metrics.md
aabmass: @jmacd the [Collector](https://github.com/open-telemetry/opentelemetry-collector/tree/master/receiver/hostmetricsreceiver/internal/scraper/loadscraper) sends load average as does the [Prometheus Node Exporter](https://github.com/prometheus/node_exporter#enabled-by-default), but I'm not sure if that's a good justification.

@jrcamp I agree. What do you think of `$platform.system.*`?

text/0119-standard-system-metrics.md
aabmass: @jkwatson thank you for explaining. I think my confusion is between the instrument vs. an instance of the instrument. In the Python API right now, it is not this way; we have [`Meter.create_metric() -> Metric`](https://github.com/open-telemetry/opentelemetry-python/blob/8be986cd6c94368ed769149022a7862f0ca47f35/opentelemetry-api/src/opentelemetry/metrics/__init__.py#L338-L347).

text/0119-standard-system-metrics.md
aabmass: I am not sure on this one, could something like ratio of reads/writes be useful? Just want to copy [that prometheus recommendation](https://prometheus.io/docs/practices/naming/#metric-names) we keep going back to here:
> As a rule of thumb, either the sum() or the avg() over all dimensions of a given metric should be meaningful (though not necessarily useful). If it is not meaningful, split the data up into multiple metrics. For example, having the capacity of various queues in one metric is good, while mixing the capacity of a queue with the current number of elements in the queue is not.

I think disk IO falls into the category of meaningful but not necessarily useful? The sum would be the total number of bytes

jmacd: I agree with your logic. Let's keep it a single metric.

text/0119-standard-system-metrics.md
jrcamp: @aabmass what would we do about inodes where it's on pretty much every platform but windows? Would on linux and freebsd it be linux.system.filesystem.inodes and freebsd.system.filesystem.inodes?

jmacd:
Please resolve comments that have been addressed. We think this is ready to merge after the current feedback is addressed

bogdandrutu:
Also please rebase

text/0119-standard-system-metrics.md
aabmass: @jrcamp, I'm not sure. I will change this example to something unambiguous, and then capture this discussion in the "Open Questions" section.

text/0119-standard-system-metrics.md
bogdandrutu: If we use ratio value we should not use the unit "%". If we use the unit "%" we should have values 0-100 :)

text/0119-standard-system-metrics.md
jmacd: @aabmass are you still planning to add this discussion? Thanks.

aabmass: I moved it to "Open Questions". I don't think we've reached a conclusion still

jmacd: We agreed to merge this and move on to the specification.

jmacd:
@open-telemetry/specs-metrics-approvers please consider this PR.

text/0119-standard-system-metrics.md
```diff
@@ -0,0 +1,150 @@
...
+## Open questions
+
+- Should the individual runtimes have their specific naming conventions in the spec?
+- Is it ok to include instruments specific to an OS (or OS family) under a top-level prefix, as long as they are unambiguous? For example, naming inode related instruments, which of the below is preferred?
+  1. Top level: `system.filesystem.inodes.*`
+  2. UNIX family level: `system.unix.filesystem.inodes.*`
+  3. One for each UNIX OS: `system.linux.filesystem.inodes.*`, `system.freebsd.filesystem.inodes.*`, `system.netbsd.filesystem.inodes.*`, etc.
+- Should instrument names be versioned e.g. `v1.system.cpu.usage`?
+  * Some telemetry backends use a strict schema for metrics and their labels (e.g. Google Cloud Monitoring with [MetricDescriptors](https://cloud.google.com/monitoring/api/v3/metrics-details#intro-metdesc)), so changes to the instruments proposed in this document could break ingestion of new telemetry data.
+  * Should the exporters for such backends be responsible for handling this? How could this be implemented in the exporter without additional information?
```
aabmass, line 150: I opened a spec issue for this, it looks like a strong "no". I will remove this and update

jmacd: The use of strict schemas is a very Prometheus-ey thing, not common in the Statsd world. I would prefer if systems could perform some aggregation to transform data into the schema they want as opposed to not recognizing matching time series. For example: your system expects to monitor a system metric w/ four labels, but it arrives with 5 labels? You can aggregate away one label and get the result you need. In any case, it doesn't sound like this issue is blocking this OTEP.

aabmass: Here's the relevant issue https://github.com/open-telemetry/opentelemetry-specification/issues/703

text/0119-standard-system-metrics.md
```diff
@@ -0,0 +1,147 @@
...
+|                           |           |                 |          |direction|transmit, receive                                                                             |
+|system.net.errors          |errors     |SumObserver      |Int64     |device   |(identifier)                                                                                  |
+|                           |           |                 |          |direction|transmit, receive                                                                             |
+|system<!--notlink-->.net.io|bytes      |SumObserver      |Int64     |device   |(identifier)                                                                                  |
+|                           |           |                 |          |direction|transmit, receive                                                                             |
+|system.net.connections     |connections|UpDownSumObserver|Int64     |device   |(identifier)                                                                                  |
```
dmitryax, line 121: Not sure if it was discussed before. It seems a bit inconsistent that we use a short version `net` instead of `network` here, but full names like `memory` and `filesystem` are used for other metrics. Also I see the `network` already used in Collector https://github.com/open-telemetry/opentelemetry-collector/blob/master/receiver/hostmetricsreceiver/hostmetrics_receiver_test.go#L54-L58. Do we need to update it here or in the collector? /cc @james-bebbington

jmacd:
@open-telemetry/specs-metrics-approvers please review.

text/0119-standard-system-metrics.md
aabmass: @dmitryax thanks, changing it to "network" for consistency.

james-bebbington:
I'm pretty sure this OR is ready to merge but for the spec repo, t would be good to add a convention for process counts (with "state" = running / inactive)

text/0119-standard-system-metrics.md
```diff
@@ -0,0 +1,149 @@
...
+---
+
+#### `system.cpu.`
+
+**Description:** System level processor metrics.
+|Name                  |Units  |Instrument Type  |Value Type|Label Key|Label Values                       |
+|----------------------|-------|-----------------|----------|---------|-----------------------------------|
+|system.cpu.time       |seconds|SumObserver      |Double    |state    |idle, user, system, interrupt, etc.|
+|                      |       |                 |          |cpu      |1 - #cores                         |
+|system.cpu.utilization|1      |UpDownSumObserver|Double    |state    |idle, user, system, interrupt, etc.|
```
bogdandrutu, line 64: s/UpDownSumObserver/ValueObserver

PR #121, opened 2020-06-25
Remote configuration for the SDK proposal

churanchen:
This proposal asks for approval to create an experimental feature for remote configuration of the SDK, per [this] (https://github.com/open-telemetry/opentelemetry-specification/tree/master/experimental) framework. This experiment specifically focuses on configuring metric collection.

It is related to this PR: https://github.com/open-telemetry/opentelemetry-proto/pull/155

Spec: [OpenTelemetry.Configuration.Service.pdf](https://github.com/open-telemetry/oteps/files/4833020/OpenTelemetry.Configuration.Service.pdf)


churanchen:
I signed it

SergeyKanzhelev:
@tigrannajaryan thank you for review! Very good topics are raised. Do you agree with this project as an experimental project to be moved forward?

It is hidden in the otep description, but the motivation is to get this to POC and than decide on how to move it to the main product: 

> It is intended to seek approval to develop a proof of concept.

text/0121-config-service.md
```diff
@@ -0,0 +1,161 @@
...
+
+One way to configure metric export schedules is to, instead of pushing the metrics, use a pull mechanism to have metrics pulled whenever wanted. This has been implemented for [Prometheus](https://github.com/open-telemetry/opentelemetry-go/pull/751).
+
+The alternative is to stick with the status quo, where the SDK has a [fixed collection period](https://github.com/open-telemetry/opentelemetry-go/blob/34bd99896311a81cf843475779cae2e1c05e6257/sdk/metric/controller/push/push.go#L72-L76) and a fixed sampling rate.
+
+## Open questions
+
+- As mentioned [here](https://github.com/open-telemetry/opentelemetry-proto/pull/155#issuecomment-640582048). what happens if a malicious/accidental config change overwhelms the application/monitoring system? Is it the responsibility of the user to be cautious while making config changes? Should we automatically decrease telemetry exporting if we can detect performance problems?
+
+## Future possibilities
```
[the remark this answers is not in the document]
SergeyKanzhelev: I like your suggestion on thinking of these pipelines separately. If we do so - it is quite logical, that collector may be configured to participate in config pipeline. But the default mode may be a simple proxy of configs back to whoever is asking for it.

text/0121-config-service.md
```diff
@@ -0,0 +1,161 @@
...
+
+The user, when instrumenting their application, can configure the SDK with the endpoint of their remote configuration service, the associated Resource and a default config we revert to if we fail to read from the configuration service.
+
+The user must then set up the config service. This MUST be done through the collector, which can be set up to expose an arbitrary configuration service implementation. Depending on implementation, this allows the collector to either act as a stand-alone configuration service, or as a bridge to remote configurations of the user's monitoring and tracing backend by 'translating' the monitoring backend's protocol to comply with the Open Telemetry configuration protocol.
```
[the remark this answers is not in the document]
SergeyKanzhelev: +1, @churanchen can you please fix MUST to should or rephrase it in terms - "if user decide to use collector as a config service, than she MUST..."

text/0121-config-service.md
```diff
@@ -0,0 +1,161 @@
...
+  }
+  TraceConfig trace_config = 3;
+
+  // Optional. The client is suggested to wait this long (in seconds) before
+  // pinging the configuration service again.
+  int32 suggested_wait_time_sec = 4;
+}
+```
+
+The SDK will periodically read a config from the service using GetConfig. If it fails to do so, it will just use either the default config or the most recent successfully read config. If it reads a new config, it will apply it.
```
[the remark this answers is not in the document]
SergeyKanzhelev: +1 @churanchen I suggest to add a section in alternative consideration about possible mechanisms of config pull/push. And the decision can be based on perf measurement

text/0121-config-service.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Trade-offs and mitigations
+
+This feature will be implemented purely as an experiment, to demonstrate its viability and usefulness. More investigation can be done after a rough prototype is demonstrated.
+
+As mentioned [here](https://github.com/open-telemetry/opentelemetry-proto/pull/155#issuecomment-640582048), the configuration service can be a potential attack vector for an application instrumented with Open Telemetry, depending on what we allow in the protocol. This can be mitigated since this proposal only allows the SDK to read from a configuration service managed by the collector. Additionally, we can highlight in the remote configuration protocol that for future changes, caution is needed in terms of the sorts of configurations we allow.
+
+## Prior art and alternatives
```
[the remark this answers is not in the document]
SergeyKanzhelev: +1 Good thing to explore in this experiment

tigrannajaryan:
> @tigrannajaryan thank you for review! Very good topics are raised. Do you agree with this project as an experimental project to be moved forward?
> 
> It is hidden in the otep description, but the motivation is to get this to POC and than decide on how to move it to the main product:
> 
> > It is intended to seek approval to develop a proof of concept.
> 
> I think this POC may uncover some limitations of current SDKs config story.

Yes, definitely, I think remote configuration is a very useful addition. Assuming this an experiment I do not see anything that prevents us from moving forward. For production use this will need more refinement but it can be done after the PoC.

text/0121-config-service.md
```diff
@@ -0,0 +1,161 @@
...
+
+One way to configure metric export schedules is to, instead of pushing the metrics, use a pull mechanism to have metrics pulled whenever wanted. This has been implemented for [Prometheus](https://github.com/open-telemetry/opentelemetry-go/pull/751).
```
c24t, line 149: This is an alternative to the current push-based approach to metrics, but not really an alternative to the proposal here. Echoing @tigrannajaryan's comment above, it sounds like the important alternative to note here is pushing vs. polling the config.

On its face, It's pretty surprising that we'd ask end users to implement a remote configuration service which the collector (or SDK) would poll, instead of adding something like a `setConfig` API method to the collector. Why did you decide on polling here?

text/0121-config-service.md
```diff
@@ -0,0 +1,161 @@
...
+
+It is related to [this pull request](https://github.com/open-telemetry/opentelemetry-proto/pull/155)
+
+## Motivation
+
+During normal use, users may wish to collect metrics every 10 minutes. Later, while investigating a production issue, the same user could easily increase information available for debugging by reconfiguring some of their processes to collect metrics every 30 seconds. Because this change is centralized and does not require redeploying with new configurations, there is lower friction and risk in updating the configurations.
+
+## Explanation
+
+This OTEP is a request for an experimental feature [open-telemetry/opentelemetry-specification#62](https://github.com/open-telemetry/opentelemetry-specification/pull/632). It is intended to seek approval to develop a proof of concept. This means no development will be done inside either the Open Telemetry SDK or the collector. Since this will be implemented in [opentelemetry-go-contrib](https://github.com/open-telemetry/opentelemetry-go-contrib) and [opentelemetry-collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib), all of this functionality will be optional.
```
c24t, line 13:
```suggestion
This OTEP is a request for an experimental feature [open-telemetry/opentelemetry-specification#62](https://github.com/open-telemetry/opentelemetry-specification/pull/632). It is intended to seek approval to develop a proof of concept. This means no development will be done inside either the OpenTelemetry SDK or the collector. Since this will be implemented in [opentelemetry-go-contrib](https://github.com/open-telemetry/opentelemetry-go-contrib) and [opentelemetry-collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib), all of this functionality will be optional.
```

yurishkuro:
I am bothered by the wording in the motivation:

> This OTEP is a **request** for an experimental feature... It is intended to **seek approval** to develop a **proof of concept**.

OTEL is supposed to be a collaborative project, not a rigid process. No "approval" is required to develop experimental features or proofs of concept.

OTEP is an "enhancement proposal". You're proposing new functionality and the OTEP is used to discuss the design decisions and pros & cons, not to seek approval.

It is also a good practice to cite prior art. This is not a new idea to have a remotely controlled configuration, Jaeger SDKs have implemented it from day 1.

text/0121-config-service.md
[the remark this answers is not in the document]
churanchen: Done! We'd originally limited it to the collector to limit scope/possible security problems, but this makes sense to remove.

text/0121-config-service.md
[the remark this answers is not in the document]
churanchen: Good points. Mentioned here: https://github.com/open-telemetry/oteps/blob/613b463da66d52bd18a3b2a1b376279ec78e381a/text/0121-config-service.md#prior-art-and-alternatives

text/0121-config-service.md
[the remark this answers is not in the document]
churanchen: We did think about a push mechanism. Our rationale for a polling mechanism is that it might be more scalable if you have a central configuration service that is read from many processes, rather than the service having to push to many processes. However, this is something that definitely needs a harder examination.

text/0121-config-service.md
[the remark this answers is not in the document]
churanchen: As far as our prototype works so far, we intend the collector to only be a proxy.

For the future, if the collector can be configured, maybe individual collectors should also be labelled (if they are not already), in the way resources are currently. There may be a large-scale use case where there are "layers" of collectors, and it's necessary to configure individual collectors. Then, a label given to each configured process can also be used to distinguish between SDK and collector.

text/0121-config-service.md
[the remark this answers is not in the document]
churanchen: If it is acceptable to the OpenTelemetry maintainers, in line with the experiment framework for new features (https://github.com/open-telemetry/opentelemetry-specification/tree/master/experimental), would it be a good idea to have an experimental folder in opentelemetry-proto for experimental protocols? Or is this too invasive?

@SergeyKanzhelev

text/0121-config-service.md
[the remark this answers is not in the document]
churanchen: Oh shoot, thanks catching this. Should have fixed it everywhere.

text/0121-config-service.md
churanchen: Our primary rationale was the large-scale use case where the collector would be responsible for pushing configs to many instrumented processes. It seemed more scalable to have the SDK read from the configuration service.

In a situation where the collector is extended to act as a configuration service or act as a proxy for remote configuration built into a monitoring backend, hopefully the end user won't be need to do much extra work relative to setting up a mechanism that calls a `setConfig` method on the collector.

The trade-offs for pushing vs. polling is definitely something worth thinking about though.

churanchen:
I'm sorry I didn't make things clear! We're not exactly seeking approval to experiment, but rather approval to develop an experimental feature per this [framework](https://github.com/open-telemetry/opentelemetry-specification/tree/master/experimental) that belongs in the contrib repositories, but not part of the core OpenTelemetry implementation. I edited the PR description to clear up this bad wording.
 
> It is also a good practice to cite prior art. This is not a new idea to have a remotely controlled configuration, Jaeger SDKs have implemented it from day 1.

Added it! Sorry about that, I'm not very familiar with Jaeger.

text/0121-config-service.md
```diff
@@ -0,0 +1,161 @@
...
+  bytes last_known_fingerprint = 2;
+}
+```
+
+While the response will look like this:
+
+```
+message ConfigResponse {
+
+  // Optional. The fingerprint associated with this ConfigResponse. Each change
```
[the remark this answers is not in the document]
churanchen: Our intention for the fingerprint field was to have an identifier for configs that the user can set and use if they wish, similar to the metadata field. The idea being that there might be some fields that are nice and useful to include, but that we shouldn't force the end users to have in their configs.

churanchen:
Thank you for all the questions! A lot of good points were brought up!

text/0121-config-service.md
```diff
@@ -0,0 +1,165 @@
...
+        MIN_10 = 600;
+        MIN_30 = 1800;
+        HR_1 = 3600;
+        HR_2 = 7200;
+        HR_4 = 14400;
+        HR_12 = 43200;
+        DAY_1 = 86400;
+        DAY_7 = 604800;
+      }
+      CollectionPeriod period = 3;
```
jmacd, line 103: Is this enum needed, if the number is in seconds? Given that the values are in seconds, it implies that unlisted values (e.g., 2 days == 172800) may be used. If unlisted values can be used, do we need these constants?

How is NONE meant to be used?

text/0121-config-service.md
```diff
@@ -0,0 +1,165 @@
...
+
+      // Optional. Additional opaque metadata associated with the schedule.
+      // Interpreting metadata is implementation specific.
+      bytes metadata = 4;
```
jmacd, line 107: This looks difficult to use. How do you see it being used?

text/0121-config-service.md
```diff
@@ -0,0 +1,161 @@
...
+
+In the future, we will add per-metric configuration. Each Schedule also has inclusion_patterns and exclusion_patterns. Any metrics that match any of the inclusion_patterns and do not match any of the exclusion_patterns will be exported every CollectionPeriod (ie. every minute). A component will be added that can export metrics that match a pattern from one Schedule at that Schedule's collection period, while exporting other metrics that match the patterns of another Schedule at that other Schedule's collection period.
+
+The collector will support a new interface for a DynamicConfig service that can be used by an SDK, allowing a custom implementation of the configuration service protocol described above, to act as an optional bridge between an SDK and an arbitrary configuration service. This interface can be implemented as a shim to support accessing remote configurations from arbitrary backends. The collector is configured to expose an endpoint for requests to the DynamicConfig service, and returns results on that endpoint.
```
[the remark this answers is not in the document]
wtong98: Great points. Currently in our prototype, the service is implemented as an extension on the collector (like zpages, healthcheck). If enabled, the user has the option of specifying one of two (or more, potentially) backends from which the collector gets its configs:
*  file backend: a local file containing the metric and trace configs is served. This option corresponds to the usage you mentioned in which the collector itself is the configuration service. The file is actively monitored, so any updates to the file will be propagated to clients upon their next update request.
* remote backend: the collector receives config updates from a remote backend that implements the server-side of the gRPC service. This remote backend could be another collector, or it can be third-party. In this way, the remote backend fufills the bridging concept.

I like your idea of implementing the process as a pipeline. We chose an extension originally as a simple way to cleanly develop a prototype, but as the implementation grows, it may be nice to organize it into a full pipeline.

For specific details, feel free to check out a [PR introducing the extension](https://github.com/vmingchen/opentelemetry-collector-contrib/pull/1) and [another that adds the remote backend](https://github.com/vmingchen/opentelemetry-collector-contrib/pull/2). Both are made on our intern host's fork of the project, and will be moved to the upstream contrib repo when finished.

Does this approach sound viable? Let me know what you think!

text/0121-config-service.md
churanchen: This is a sort of catch-all field for metadata that is used by a specific monitoring back-end implementation but that we do not specifically include in the protocol. Say a monitoring system has the notion of network traffic priority. It might be useful to relay this to the instrumented application, or for the instrumented application to relay this back to the back-end. It's up to the user if/how they want to use this field.

text/0121-config-service.md
churanchen: The enum is used to enforce the property that all larger periods must be multiples of the smaller ones. This might be useful in that we can batch the export of metrics with less frequent CollectionPeriods with those with more frequent CollectionPeriods. 

Just using an int definitely definitely provides more flexibility. Let me know what you think!

So I deleted the comment left with `NONE`. The intention of `NONE` is just to say "do not export". Might be useful for the user to have an option they can toggle to stop exports for the user. `NONE` can be gotten rid of if it seems unnecessary.

text/0121-config-service.md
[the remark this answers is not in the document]
tigrannajaryan: @churanchen I think it is a good idea to also clearly mark the Protobuf messages in opentelemetry-proto as experimental.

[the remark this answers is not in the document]
churanchen: I agree. If this is the approach we go through with, we can add an `Experimental_` prefix to all our messages in our past PR.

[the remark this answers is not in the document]
churanchen: To clarify, is that what you meant by marking the Protobuf messages?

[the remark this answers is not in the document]
tigrannajaryan: Either that or put them in a separate directory/package (or even do both to be very explicit).

churanchen:
Thanks for the additional feedback! We've tried to address it as best we can, but let us know if we're missing or misunderstood something!

churanchen:
@jmacd @yurishkuro Let us know if you have more concerns!

text/0121-config-service.md
```diff
@@ -0,0 +1,178 @@
...
+
+It is related to [this pull request](https://github.com/open-telemetry/opentelemetry-proto/pull/155)
+
+## Motivation
+
+During normal use, users may wish to collect metrics every 10 minutes. Later, while investigating a production issue, the same user could easily increase information available for debugging by reconfiguring some of their processes to collect metrics every 30 seconds. Because this change is centralized and does not require redeploying with new configurations, there is lower friction and risk in updating the configurations.
+
+## Explanation
+
+This OTEP is a request for an experimental feature [open-telemetry/opentelemetry-specification#62](https://github.com/open-telemetry/opentelemetry-specification/pull/632). It is intended to seek approval to develop a proof of concept. This means no development will be done inside either the OpenTelemetry SDK or the collector. Since this will be implemented in [opentelemetry-go-contrib](https://github.com/open-telemetry/opentelemetry-go-contrib) and [opentelemetry-collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib), all of this functionality will be optional.
```
yurishkuro, line 13:
```suggestion
This OTEP is a proposal for an experimental feature [open-telemetry/opentelemetry-specification#62](https://github.com/open-telemetry/opentelemetry-specification/pull/632), to be developed as a proof of concept. This means no development will be done inside either the OpenTelemetry SDK or the collector. Since this will be implemented in [opentelemetry-go-contrib](https://github.com/open-telemetry/opentelemetry-go-contrib) and [opentelemetry-collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib), all of this functionality will be optional.
```

text/0121-config-service.md
```diff
@@ -0,0 +1,178 @@
...
+
+The user, when instrumenting their application, can configure the SDK with the endpoint of their remote configuration service, the associated Resource and a default config we revert to if we fail to read from the configuration service.
```
yurishkuro, line 15:
```suggestion
The user, when instrumenting their application, can configure the SDK with the endpoint of their remote configuration service, the associated Resource, and a default config to be used if fail to read from the configuration service.
```

text/0121-config-service.md
```diff
@@ -0,0 +1,178 @@
...
+
+The user must then set up the config service. This can be done through the collector, which can be set up to expose an arbitrary configuration service implementation. Depending on implementation, this allows the collector to either act as a stand-alone configuration service, or as a bridge to remote configurations of the user's monitoring and tracing backend by 'translating' the monitoring backend's protocol to comply with the OpenTelemetry configuration protocol.
+
+## Internal details
+
+In the future, it is intended to add per-metric configuration. For example, this would allow the user to collect 5xx server error counts ever minute, and CPU usage statistics every 10 minutes. The remote configuration protocol was designed with this in mind, meaning that it includes more details than simply the metric collection period.
```
yurishkuro, line 21:
```suggestion
In the future, we intend to add per-metric configuration. For example, this would allow the user to collect 5xx server error counts ever minute, and CPU usage statistics every 10 minutes. The remote configuration protocol was designed with this in mind, meaning that it includes more details than simply the metric collection period.
```

text/0121-config-service.md
```diff
@@ -0,0 +1,178 @@
...
+A request to the config service will look like this:
+
+```
+message ConfigRequest{
+
+  // Required. The resource for which configuration should be returned.
+  opentelemetry.proto.resource.v1.Resource resource = 1;
+
+  // Optional. The value of ConfigResponse.fingerprint for the last configuration
+  // a resource received that was successfully applied.
```
yurishkuro, line 40:
```suggestion
  // that the caller received and successfully applied.
```

text/0121-config-service.md
yurishkuro: Since the server supplies this configuration to the SDK, it can have whichever validation on multiples that it wants, without having that reflected in the protocol schema. So +1 to just have a duration field. The field name should reflect that it is in seconds.

text/0121-config-service.md
yurishkuro: I would prefer to remove this. Some of the scenarios you describe are general enough and could be expressed via proper schema in the future. For pure custom vendor extensions I don't think `bytes metadata` is a good format. Doesn't proto allow custom Any types? And does the custom extension even need to be part of this proto? It will require custom SDKs, and there are other ways custom SDKs can access this.

text/0121-config-service.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Internal details
+
+In the future, it is intended to add per-metric configuration. For example, this would allow the user to collect 5xx server error counts ever minute, and CPU usage statistics every 10 minutes. The remote configuration protocol was designed with this in mind, meaning that it includes more details than simply the metric collection period.
+
+Our remote configuration protocol will support this call:
+
+```
+service DynamicConfig {
+  rpc GetConfig (ConfigRequest) returns (ConfigResponse);
```
yurishkuro, line 27: I don't particularly like bundling all configs like that into a single RPC call. For example, Uber currently instruments all apps with OpenTracing and Tally for traces/metrics respectively. We may have a migration path to swap tracing instrumentation for OpenTelemetry, but replacing Tally instrumentation is going to be a huge effort that's not worth it. So we would only need remote configurability for the tracing part of OTEL SDK.

I would prefer separating these into different services, DynamicTracingConfig, DynamicMetricsConfig, etc.

text/0121-config-service.md
```diff
@@ -0,0 +1,178 @@
...
+
+      // Metrics with names that match at least one rule in the inclusion_patterns are
+      // targeted by this schedule. Metrics that match at least one rule from the
+      // exclusion_patterns are not targeted for this schedule, even if they match an
+      // inclusion pattern.
+
+      // For this iteration, since we only want one Schedule that applies to all metrics,
+      // we will not check the inclusion_patterns and exclusion_patterns.
+      repeated Pattern inclusion_patterns = 1;
+      repeated Pattern exclusion_patterns = 2;
```
yurishkuro, line 82: nits: 
  1. because exclusion patterns have higher priority per the comment above, I would move it to the first position.
  2. I am not sure this grammar is expressive enough. Is there any prior art for this?

text/0121-config-service.md
churanchen: I definitely agree that users should not be forced into configuring both metrics and tracing if they only need to configure one. What if the protocol supported having MetricConfig and TraceConfig be optional? Uber, for example, could choose to remotely configure tracing, but not have a metrics config at all.

I worry in the future, if remote configuration is adopted, that it may get unwieldy if we have a DynamicTracingConfig, DynamicMetricsConfig, DynamicLoggingConfig (if OpenTelemetry supports this in the future), etc. where if a user wants configure multiple parts of the SDK, multiple calls need to be made to a configuration service.

What do you think? Does this sound okay, or would you still prefer having separate services?

yurishkuro: We need to see what other people think. Personally, I am not too bothered by multiple calls. It also allows different TTLs for different types of configuration.

text/0121-config-service.md
churanchen: Done!

text/0121-config-service.md
churanchen: 1. Changed!
2. The choice to use exact match and starts with was made because they are more efficient to apply and clearer to read than more expressive alternatives, and have in practice been the only required pattern types for specifying metric names in Google's internal metric collection configuration protocol. However we are open to also adding a more expressive grammar such as regex or glob as a third option for a pattern type, so that monitoring backends can optionally choose to support a more expressive grammar without affecting efficiency for the simpler pattern types.

text/0121-config-service.md
churanchen: Hmm, how about we change DynamicConfig -> DynamicMetricsConfig, and just leave out DynamicTracingConfig for this OTEP. Our OTEP is more about developing a proof-of-concept for the idea of remote configuration, and it doesn't go into detail about tracing. If remote configuration gets adopted into OpenTelemetry, a DynamicTracingConfig service can be added later.

yurishkuro: +1

churanchen: Done!

text/0121-config-service.md
churanchen: Removed! This is still something we might be interested in later, but it's not necessary for a proof-of-concept.

churanchen:
Would welcome any more reviews!

jmacd:
@churanchen Could you resolve the issues that you feel are resolve (and/or indicate which are resolved)? This helps reviewers.

bogdandrutu:
@yurishkuro is this ready to be merged in your opinion?

@SergeyKanzhelev should we have a way to mark that this is experimental vs not?

yurishkuro:
I have no objections to merging. I have not had experience with metrics library that requires external configuration, so aside from the general config mechanism I don't have an opinion on the metrics-specific details of this proposal.

text/0121-config-service.md
```diff
@@ -0,0 +1,127 @@
+# A Dynamic Configuration Service for the SDK
```
bogdandrutu, line 1: Consider to move this to `experimental/...` this way we know that it is experimental.

churanchen: Done! @SergeyKanzhelev, would it be a good idea to move the z-pages otep there as well?

PR #122, opened 2020-06-27
HTTP/JSON Encoded for OTLP

tensorchen:
Protobuf is a relatively big dependency, which some clients are not willing to take. For example, webjs, iOS/Android (in some scenarios, the size of the installation package is limited, do not want to introduce protobuf dependencies). Plain JSON is a smaller dependency and is built in the standard libraries of many programming languages.

This OTEP will be consistent with the OTEP0099(OTLP/HTTP) specification except that the payload will use JSON instead of protobuf.

### DIFF
--- a/text/0000-otlp-http-json.md
+++ b/text/0000-otlp-http-json.md
@@ -6,7 +6,8 @@ This is a proposal to add HTTP Transport extension supporting json serialization
 ## Table of Contents
 
 * [Motivation](#motivation)
-* [OTLP/HTTP+JSON Protocol Details](#otlphttp-protocol-details)
+* [OTLP/HTTP+JSON Protocol Details](#otlphttpjson-protocol-details)
+  * [JSON Mapping](#json-mapping)
   * [Request](#request)
   * [Response](#response)
     * [Success](#success)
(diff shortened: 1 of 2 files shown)


tensorchen:
@open-telemetry/specs-approvers  please review.

text/0000-otlp-http-json.md
```diff
@@ -0,0 +1,146 @@
...
+    * [Throttling](#throttling)
+    * [All Other Responses](#all-other-responses)
+  * [Connection](#connection)
+  * [Parallel Connections](#parallel-connections)
+
+## Motivation
+
+Protobuf is a relatively big dependency, which some clients are not willing to
+take. For example, webjs, iOS/Android (in some scenarios, the size of the installation package is limited, do not want to introduce protobuf dependencies). Plain JSON is a smaller dependency and is built in the standard
+libraries of many programming languages.
```
tigrannajaryan, line 24: Please use consistent line-breaking. Either do hard breaks at certain line length or don't do hard breaks at all.

text/0000-otlp-http-json.md
```diff
@@ -0,0 +1,146 @@
...
+The default URL path for requests that carry trace data is `/v1/traces` (for
+example the full URL when connecting to "example.com" server will be
+`https://example.com/v1/traces`). The request body is a JSON-encoded
+[`ExportTraceServiceRequest`](https://github.com/open-telemetry/opentelemetry-proto/blob/e6c3c4a74d57f870a0d781bada02cb2b2c497d14/opentelemetry/proto/collector/trace/v1/trace_service.proto#L38)
+message.
+
+The default URL path for requests that carry metric data is `/v1/metrics` and the
+request body is a JSON-encoded
+[`ExportMetricsServiceRequest`](https://github.com/open-telemetry/opentelemetry-proto/blob/e6c3c4a74d57f870a0d781bada02cb2b2c497d14/opentelemetry/proto/collector/metrics/v1/metrics_service.proto#L35)
+message.
```
tigrannajaryan, line 47: I assume this is copied verbatim from OTEP0099, except it says "JSON-encoded" instead of "ProtoBuf-encoded", right? (It is good to copy, I just want to confirm it is the only difference).

text/0000-otlp-http-json.md
```diff
@@ -0,0 +1,146 @@
+# OTLP: HTTP+JSON Transport Extension for OTLP
```
tigrannajaryan, line 1: Nit: this is not a transport extension but an additional encoding type for existing OTLP/HTTP transport.
Perhaps reword slightly:
```suggestion
# OTLP: JSON Encoding for OTLP/HTTP
```

carlosalberto:
Hey @tensorchen 

The (brand new, still undocumented) flow for OTEPS is to first create an issue in https://github.com/open-telemetry/opentelemetry-specification, discuss whether it's a breaking/polemical change, and

a) If it's a relatively straightforward change, work directly in the Specification repo to reflect the proposed changes.
b) If it is indeed deemed as change that needs an OTEP, discuss it here (as you did).

Would you mind creating an issue for this in https://github.com/open-telemetry/opentelemetry-specification so we discuss it there (and we possibly later re-take this OTEP)? Since you already wrote this proposal, you can mention it in such issue ;)

Thanks in advance!

text/0000-otlp-http-json.md
tensorchen: Yes, this is the only difference.

bogdandrutu:
@open-telemetry/specs-approvers please make sure you review this :)

PR #126, opened 2020-07-15
A proposal for SDK configurations for metric aggregation (Basic Views)

jkwatson:
This proposal is less than a full-blown Views API. The goal is just to give some flexibility to operators and exporter authors around how metrics should be aggregated by the SDK.


text/0126-Configurable-Metric-Aggregations.md
```diff
@@ -0,0 +1,95 @@
...
+consider a non-programmatic configuration option.
+
+## Prior art and alternatives
+
+* Prior Art is probably mostly in the [OpenCensus Views](https://opencensus.io/stats/view/) system.
+* Another [OTEP](https://github.com/open-telemetry/oteps/pull/8) attempted to address building a Views API.
+
+## Open questions
+
+There has been some question about whether a custom aggregation should be allowable for a Counter type
```
tsloughter, line 87: I'll put a concern of mine with custom aggregators on Counters here with an example, it is Erlang but my have parallels to other languages people will see and provide feedback as well:

An instrument has a kind and a number kind, for this case we'll have an integer counter. If we know that the recordings to this instrument will always be treated as a sum then a potential optimization is available when using `bind_instrument`. 

Erlang has shared mutable counters (https://erlang.org/doc/man/counters.html) so a `bind` could return a reference to the shared mutable counter for instrument `I`. Each `bind` in any process would return the same reference for `I`.

If `I` is a counter but could have any and multiple aggregators that reference it then a `bind` could not return a shared mutable counter but must record the value the same as with `ValueRecorder` instruments.

I suppose that technically at the time of `bind` the aggregators configured for `I` could be checked and if the only existing one is the `Sum` aggregator it can return the shared mutable counter... But it feels like we are wasting valuable information that makes for a more efficient handling of the instruments. In OpenCensus there was only the equivalent of `ValueRecorder` so this wasn't a consideration, with OpenTelemetry I saw opportunity for optimizations.

jkwatson: If we truly want to make this restriction, I'm not opposed to having a specific option create an error if your SDK can't support it. For example, in java right now, not all aggregations are supported for all instrument types, so this kind of error seems like a possible escape clause, per SDK.

text/0126-Configurable-Metric-Aggregations.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Explanation
+
+I propose a new SDK feature, available on the interface of the SDK's MeterProvider implementation, to configure
+the batching strategies and aggregations that will be used by the SDK when metric recordings are made.
+
+The basic API has two parts.
+
+* InstrumentSelector - Enables specifying the selection of one or more instruments for the configuration to apply to.
+  - Selection options include: the instrument type (Counter, ValueRecorder, etc), and a regex for instrument name.
```
tsloughter, line 31: I get why users need to be able to set things like whether to use a delta or cumulative for all `Counter` instruments and the default aggregator for all `ValueRecorder` instruments, but the selector pattern with regexes feels overly powerful. I would have expected this to be simply SDK configuration like `default_value_recorder_aggregator: ddsketch`.

jkwatson: I think there are solid use-cases for only using an expensive aggregation for a subset of named instruments. eg. I'm running an http service, and I want details on only the ValueRecorders with a given naming pattern like "http.*.duration". Whereas, you might want just a very simple summary aggregation for any other timed operations in your application, or something that isn't a timed operation at all, like the size of incoming http request bodies.

tsloughter: Agreed about only sometimes wanting certain expensive aggregations, whether it fits a pattern or not is what I'm unsure about. Another example would be wanting specific distribution buckets. I would just have expected them to be set on a specific aggregation of an instrument and not through a regex.

jkwatson: well, you can give the specific instrument name as the regex. that would work fine in this scheme. :)

jkwatson: My rationale is that the operator won't necessarily know all the details of all the instruments that are baked into the instrumentation an app is using. But, they should be able to use general semantic conventions to pick out what they want. I don't feel super strongly about this, though. What is your concern with this? That it's a footgun, or that regex is horrible and should die in a fire? Or something else?

tsloughter: The footgun :). But maybe it isn't so dangerous.

tsloughter: Will bring up in the meeting now but putting here for record.

In the case of views each view has a name. In this selection process is this a) creating an aggregation for each metric matched individually or aggregating them together (I assumed the former) b) you can't give an individual name to each "view" selection is done this way.

text/0126-Configurable-Metric-Aggregations.md
jmacd: What wasted information is there? The reason I mention the Counter and Histogram configuration is that it's a common scenario where you both want to monitor a size distribution and treat the sum as a metric. Deciding whether to use the faster behavior (default case) or the slower behavior (e.g., histogram) is supposed to be an option, but the default case should be the efficient one.

text/0126-Configurable-Metric-Aggregations.md
jmacd: @tsloughter I like this:

> `default_value_recorder_aggregator: ddsketch`

In the experimental branch I hacked on (briefly), [here](https://github.com/jmacd/opentelemetry-go/blob/jmacd/configurable/sdk/metric/integrator/configurable/configurable.go#L42) is the Config struct that has a Defaults sections w/ aggregation by instrument kind, a Views w/ aggregation by instrument name, followed by an Aggregations section that contains Labels/Aggregator configs.

text/0126-Configurable-Metric-Aggregations.md
tsloughter: @jmacd the fact that a counter is a counter is only used for setting the default aggregation. It feels more like when creating an instrument the user can just set its default aggregation and no need for multiple instrument kinds. The fact that we have instrument kinds I think gives the impression that the act of recording is specialized for each one.

I don't like the idea of the Erlang/Elixir SDK acting differently from other languages (at least not where it is absolutely required) so if this is the direction the SDK keeps I'll have it act the same as the Erlang OpenCensus implementation where measure recording is basically the same as this.

jmacd: The Counter is not only used to set the default aggregation, it's used to imply ADDING and MONOTONIC which tell you how you can use the information regardless of which aggregation is applied. So if you choose to use a Histogram aggregation with a Counter instrument, you still know that the sum indicates a meaningful rate (because of ADDING and MONOTONIC).

jkwatson:
@open-telemetry/specs-metrics-approvers Please take a look at this. If there are no objections to this proposal - in principle, not in the exact details, please approve, so we can merge and get to work on having a more concrete specification for it.

james-bebbington:
Overall, this proposal looks really good. Nice and to the point 😄 

I do have some concerns, however, around not providing the ability to have multiple aggregations per instrument for GA:

- From what I understand, OpenCensus supported the ability to export multiple aggregations for a single instrument (through Views). Not supporting this in OpenTelemetry would make migration difficult (or impossible) for users who are taking advantage of this functionality.
- We have use cases at Google where being able to export multiple aggregations per instrument is critical, e.g. A data service is used by various client services. Each of these client services exports stats broken down by specific labels that they inject upstream via Correlation Context.

It seems like it would be fairly straightforward to extend this proposal to include the ability to have multiple aggregations per instrument, and I can't see any blockers for adding that in the future. Is the main reason for not doing that now to avoid implementation complexity?

I think I'd feel a little more comfortable if this proposal included a few more details about how this would likely be extended to support multi-aggregation use cases in the future, as I think it's something we would want to make sure gets included shortly after GA. 

---

Additional Thoughts / Considerations for this OTEP:

1. What happens if you configure two default aggregations for the same instrument? Does this raise an error or the second just silently take preference? What are the trade-offs there?
2. Do we want to provide any way for an instrumentation library, e.g. gRPC, to provide recommended default aggregations?

tsloughter:
While I also have metrics at work that utilize multi-aggregation on the same measure what really struck me about the last comment was the "provide recommended default aggregations" line.

I hadn't thought about yet how I'd convert the metrics in my grpc library to opentelemetry.

Right now it exports default views (aggregations) the user can choose to subscribe to specific ones in the list, all or none: https://github.com/tsloughter/grpcbox/blob/master/src/grpcbox_oc_stats.erl#L76

Maybe multi-aggregates has to come in a separate PR because it also needs to allow specifying the labels to keep per-aggregation.

jkwatson:
I don't know why an instrumentation library would ever want to recommend default aggregations. How can the grpc instrumentation writer know anything about backends or how the data needs to be delivered to one?

jkwatson:
@james-bebbington Can you elaborate on this? I don't understand how this is applies. 

> We have use cases at Google where being able to export multiple aggregations per instrument is critical, e.g. A data service is used by various client services. Each of these client services exports stats broken down by specific labels that they inject upstream via Correlation Context.

tsloughter:
> I don't know why an instrumentation library would ever want to recommend default aggregations. How can the grpc instrumentation writer know anything about backends or how the data needs to be delivered to one?

The library author definitely knows what metrics and aggregations may be of interest to the user. In the link I gave it shows client and server aggregations for sizes, latencies, number of requests.

OC has specs for default views for common protocols like http, grpc, sql, etc https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md

jkwatson:
> > I don't know why an instrumentation library would ever want to recommend default aggregations. How can the grpc instrumentation writer know anything about backends or how the data needs to be delivered to one?
> 
> The library author definitely knows what metrics and aggregations may be of interest to the user. In the link I gave it shows client and server aggregations for sizes, latencies, number of requests.
> 
> OC has specs for default views for common protocols like http, grpc, sql, etc https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md

Metrics, sure, but aggregations are going to depend on the backend, aren't they?

tsloughter:
What do you mean by backend? Aside from a user wanting to define their own distribution buckets based on their particular services latencies I'm not sure what you mean.

Do you mean like if a view is a histogram the user will want/need to specify the particular histogram impl to use?

jkwatson:
The aggregation generates metrics, which get sent somewhere (the telemetry backend). Those backends support different shapes of data being sent to them. Some may support histograms, some may not. Some may support MMSC summaries, some may not. In general, the instrumentation author can't possibly know where the data will end up, and hence can't tell people what aggregations to use. 

In my mental model, instrumentation uses Instruments with from the API and gives them meaningful names (generally based on semantic conventions). Those instruments have default aggregations assigned to them. If the operator's telemetry backend doesn't want the defaults, then the operator should use this proposed API to change those aggregations.

tsloughter:
Oh, I took it as up to the exporter or collector to deal with metrics in shapes that backends can't handle. Might not be perfect and still entail a user to override a default to use X instead of Y but that in the end the intention was to not have, "what aggregations does my backend support", be a regular thing the user should have to think about when defining an aggregation.

jkwatson:
> Oh, I took it as up to the exporter or collector to deal with metrics in shapes that backends can't handle. Might not be perfect and still entail a user to override a default to use X instead of Y but that in the end the intention was to not have, "what aggregations does my backend support", be a regular thing the user should have to think about when defining an aggregation.

But, an exporter can't turn a MMSC summary into a histogram, or a sum into a set of quantiles, right? In addition to capabilities, the operator might say "it's really important to get lots of details on metric xyz, and I know my backend supports quantiles, so I'll configure that aggregation, with these settings". Again, this isn't something that an instrumentation author can have any knowledge of.

tsloughter:
> or a sum into a set of quantiles, right? 

Right, though I don't like having that in the first place :)

> "I know my backend supports quantiles, so I'll configure that aggregation, with these settings". Again, this isn't something that an instrumentation author can have any knowledge of.

True, but the instrumentation author does know what are likely useful aggregations, such as a distribution of latencies per method/route, and can make this simple for the user to enable if Otel provides the right tools.

tsloughter:
I don't know if I should propose a separate OTEP or what but I'm realizing now I think instruments and aggregations are too tied together.

Thinking about how a view would just be a name, description, aggregation, label list and instrument name I realized this it also needs to be defined as where the checkpoint is and not the instrument.

My proposal would make this change as well as removal any aggregation from the instrument definition. Instead, the instrument kind informs the SDK of what default aggregation to use for creating a default view per instrument.

Recording a measurement to an instrument updates each view that is bound to it (preferably which is also subscribed to an exporter, but that can be an optimization for later), the accumulator checkpoints each view and the integrator does any necessary merging -- not sure the integrator and merging is actually necessary if the recording to a view does the dropping of labels itself?

So if you create a counter and have the default SDK installed it will create a view by the same name that keeps all the labels in each recording and uses a sum aggregation.

---

Hm, thought of another reason I find changing the aggregation on a counter funny. An `UpDownCounter` instrument is monotonic, but what does this mean outside a sum aggregation? If it was an array aggregation and defined as monotonic a user would expect each value to be greater than the last in the array, but that isn't necessarily the case since the `UpDownCounter` is guaranteed monotonic only in the case of summing. Just seems odd and actually makes it more complicated if we are expect to have to define aggregations that do and don't make sense depending on the chosen instrument kind.

jkwatson:
@tsloughter So...it sounds like you are not willing to approve this proposal, then?

I personally don't think we should try to re-do our instrument model before GA.

tsloughter:
> So...it sounds like you are not willing to approve this proposal, then?

I'm leaning that way, yea. But, my review also doesn't count towards the 4 approvals anyway :)

And yea, I don't know if anyone else would get behind my tweaks to the model. It would keep the same API in place, I think, just the SDK would change to describe updating the view's value and checkpointing the view, instead of active instrument.

text/0126-Configurable-Metric-Aggregations.md
```diff
@@ -0,0 +1,97 @@
...
+ InstrumentSelector instrumentSelector = InstrumentSelector.newBuilder()
+  .instrumentType(InstrumentType.COUNTER)
+  .build();
+
+ // create a specification of how you want the metrics aggregated:
+ Specification specification =
+      specification.create(Aggregations.minMaxSumCount(), Temporality.DELTA);
+
+ //register the configuration with the MeterSdkProvider
+ meterProvider.registerAggregation(instrumentSelector, specification);
```
cijothomas, line 63: Just confirming - this API (or equivalent) is part of the SDK only, right? Library authors will simply use the API, with no control or care about which aggregation or temporality is applied.

text/0126-Configurable-Metric-Aggregations.md
```diff
@@ -0,0 +1,97 @@
...
+ // get a handle to the MeterSdkProvider (note, this is concrete name of the default SDK class in java, not a general SDK)
+ MeterSdkProvider meterProvider = OpenTelemetrySdk.getMeterProvider();
+
+ // create a selector to select which instruments to customize:
+ InstrumentSelector instrumentSelector = InstrumentSelector.newBuilder()
+  .instrumentType(InstrumentType.COUNTER)
+  .build();
+
+ // create a specification of how you want the metrics aggregated:
+ Specification specification =
```
cijothomas, line 59: The text earlier uses the term `Configuration` as opposed to `Specification` in java example. Would be great if they are using similar terms to make it easier to understand.

text/0126-Configurable-Metric-Aggregations.md
```diff
@@ -0,0 +1,97 @@
...
+OpenTelemetry's architecture separates the concerns of instrumentation and operation. The Metric Instruments
+provided by the Metric API are all defined to have a default aggregation. And, by default, aggregations are
+performed with all Labels being used to define a unit of aggregation. Although this is a good default
+configuration for the SDK to provide, more configurability is needed.
+
+There are 3 main use-cases that this proposal is intended to address:
+
+1) The application developer/operator wishes to use an aggregation other than the default provided by the SDK
+for a given instrument or set of instruments.
+2) An exporter author wishes to inform the SDK what "Temporality" (delta vs. cumulative) the resulting metric
```
cijothomas, line 16: Don't see how this PR is explaining how this is achieved. There is now way (in this PR) for an ExporterAuthor to tell the SDK that it expects delta or cumulative, right? The end user must know whether the exporter he is using is expecting delta or cumulative, and use that knowledge to setup Temporality right?

text/0126-Configurable-Metric-Aggregations.md
jkwatson: yes, the intent is SDK-only

text/0126-Configurable-Metric-Aggregations.md
jkwatson: I can update the java example. It was done based on an earlier version. :)

text/0126-Configurable-Metric-Aggregations.md
jkwatson: This is a good point. I had been assuming that exporters could get access to the "global" SDK instance the same way that any other SDK user could. But, if there is no global SDK in a given language implementation, we may need to provide a mechanism for exporters to get access explicitly to the SDK that they are being used by. I think this is related to the open question about whether Temporality is something that should only be specifiable by an exporter.

text/0126-Configurable-Metric-Aggregations.md
jkwatson: done!

text/0126-Configurable-Metric-Aggregations.md
jkwatson: I added an open question to capture this.

cijothomas: Ok. I'd prefer if Exporters don't have to retrieve a Global SDK, and change its settings. Exporter can expose a method like `GetExpectedTemporality()`, and end user can read this property from the exporter, and use that to set the SDK config. So that exporters do what it is supposed to do - export, and not do anything with SDK settings/config.

Anyway - this is not blocking the PR. I approve.

james-bebbington:
> James Bebbington Can you elaborate on this? I don't understand what this means, or how it applies to a single instrument in a single process.

I was being intentionally vague as it was an overly complex example I was providing. I can give more details if you like but don't want to derail the discussion here.

I think there are probably better examples that can be found from OpenCensus, e.g. the latency instrument is used to record both a count and a distribution (maybe not a big deal as I guess you could just have two instruments for this).

I don't think it really matters anyway as long as we agree this is a use case we want to support eventually and have a path to get there.

james-bebbington:
Re the question around instrumentation plugins needing to provide default aggregations, here's an example from OC to hopefully explain my thoughts. A typical plugin seems to:

- Insert a bunch of labels into context (e.g. at gRPC or HTTP request start).
  - Concrete example: HTTP instrumentation plugin includes "path" label: https://github.com/census-instrumentation/opencensus-go/blob/6ddd4bcc9c808594ec82377ce4323c3f7913be6d/plugin/ochttp/client_stats.go#L42
- These labels will get automatically injected into measurements (not yet implemented in OT)
- Define default views that only export some instruments and some labels
  - Concrete example: HTTP instrumentation plugin does not export "path" label by default: https://github.com/census-instrumentation/opencensus-go/blob/aad2c527c5defcf89b5afab7f37274304195a6b2/plugin/ochttp/stats.go#L167

That allows you to capture more data than most people need but not expose it by default. If a user, for example, wants to use your HTTP instrumentation library and include "path" info, they just define their own view.

Even though we don't yet have automatic label injection in OT (presumably this is coming?), I think the same still applies. You want to write a plugin that can capture a lot of dimensions, but only export some by default.

Btw, I wasn't a contributor on OpenCensus so would be great if someone with more details could jump in 😄  

> I personally don't think we should try to re-do our instrument model before GA.

I do support this btw, but just want to make sure we have thought of

text/0126-Configurable-Metric-Aggregations.md
jmacd: FYI I solved this problem in https://github.com/open-telemetry/opentelemetry-go/pull/840 for OTel-Go as follows. 
1. The Processor gets an ExportKindSelector for determining the ExportKind of a specific descriptor
2. The Exporter implements ExportKindSelector with its own preferences
3. CheckpointSet.ForEach requires you to pass the ExportKindSelector again

The processor calls the ExportKindSelector when it encounters new metrics. The exporter calls ForEach with the specific export it wants to view. The selector passed in (1) can return multiple kinds to allow multiple sub-exporters. The selector passed in (3) must return a single kind when reading values in the exporter.

```
// Exporter handles presentation of the checkpoint of aggregate
// metrics.  This is the final stage of a metrics export pipeline,
// where metric data are formatted for a specific system.
type Exporter interface {
	// Export is called immediately after completing a collection
	// pass in the SDK.
	//
	// The Context comes from the controller that initiated
	// collection.
	//
	// The CheckpointSet interface refers to the Processor that just
	// completed collection.
	Export(context.Context, CheckpointSet) error

	// ExportKindSelector is an interface used by the Processor
	// in deciding whether to compute Delta or Cumulative
	// Aggregations when passing Records to this Exporter.
	ExportKindSelector
}

// ExportKindSelector is a sub-interface of Exporter used to indicate
// whether the Processor should compute Delta or Cumulative
// Aggregations.
type ExportKindSelector interface {
	// ExportKindFor should return the correct ExportKind that
	// should be used when exporting data for the given metric
	// instrument and Aggregator kind.
	ExportKindFor(*metric.Descriptor, aggregation.Kind) ExportKind
}

// CheckpointSet allows a controller to access a complete checkpoint of
// aggregated metrics from the Processor.  This is passed to the
// Exporter which may then use ForEach to iterate over the collection
// of aggregated metrics.
type CheckpointSet interface {
	// ForEach iterates over aggregated checkpoints for all
	// metrics that were updated during the last collection
	// period. Each aggregated checkpoint returned by the
	// function parameter may return an error.
	//
	// The ExportKindSelector argument is used to determine
	// whether the Record is computed using Delta or Cumulative
	// aggregation.
	//
	// ForEach tolerates ErrNoData silently, as this is
	// expected from the Meter implementation. Any other kind
	// of error will immediately halt ForEach and return
	// the error to the caller.
	ForEach(ExportKindSelector, func(Record) error) error

	// Locker supports locking the checkpoint set.  Collection
	// into the checkpoint set cannot take place (in case of a
	// stateful processor) while it is locked.
	//
	// The Processor attached to the Accumulator MUST be called
	// with the lock held.
	sync.Locker

	// RLock acquires a read lock corresponding to this Locker.
	RLock()
	// RUnlock releases a read lock corresponding to this Locker.
	RUnlock()
}
```

jmacd:
I believe selecting and configuring labels to inject from the distributed context should be an eventual goal of the Views API. This is something OpenCensus did support, but I was not directly involved.

text/0126-Configurable-Metric-Aggregations.md
```diff
@@ -0,0 +1,98 @@
...
+This does not intend to deliver a full "Views" API, although it may eventually form the basis for one. The goal here is
+simply to allow configuration of the batching and aggregation by operators and exporter authors.
+
+This does not intend to specify the exact interface for providing these configurations, nor does it
+consider a non-programmatic configuration option.
+
+## Prior art and alternatives
+
+* Prior Art is probably mostly in the [OpenCensus Views](https://opencensus.io/stats/view/) system.
+* Another [OTEP](https://github.com/open-telemetry/oteps/pull/8) attempted to address building a Views API.
```
c24t, line 83:
```suggestion
* Another [OTEP](https://github.com/open-telemetry/oteps/pull/89) attempted to address building a Views API.
```

text/0126-Configurable-Metric-Aggregations.md
```diff
@@ -0,0 +1,98 @@
...
+2. Should the requesting of DELTA vs. CUMULATIVE be only available via an exporter-only API, rather than generally available to all operators?
+3. Is regex-based name matching too broad and dangerous? Would the alternative (having to know the exact name of all instruments to configure) be too onerous?
+4. Is there anything in this proposal that would make implementing a full Views API (i.e. having multiple, named aggregations per instrument) difficult?
+5. How should an exporter interact with the SDK for which it is configured, in order to change aggregation settings?
+
+## Future possibilities
+
+What are some future changes that this proposal would enable?
+
+- A full-blown views API, which would allow multiple "views" per instrument. It's unclear how an exporter would specify which one it wanted, or if it would all the generated metrics.
```
lzchen, line 97: There is a rough prototype of the Views API that just got merged for [Python](https://github.com/open-telemetry/opentelemetry-python/pull/596). Right now the ability for multiple views per instrument is possible. Each view can specify the aggregation type, default set of label keys to aggregate by and aggregator configuration (histogram buckets and such). Something that hasn't been discussed yet is the temporality of the instruments (which is configured on the batcher and an SDK property, so currently all instruments  will have the same behaviour). I'm not  sure if there is need of new mechanisms exposed to the user to construct these configurations when Views achieve something similar (maybe they will turn into the Views API eventually anyways?).

Also, to maybe open up the floor for addressing open questions 3 and 4, could it possible to have the views handle all of the instrument related configuration that the exporter doesn't need to specify, and just have the temporality be controlled by the SDK (since exporters need to know about it). So for 3, the granularity of control for those configuration fields would be per instrument, and the granularity for temporality would be via regex. Although it might be too complex/too much configuration for a user to have to do. Maybe there is some way we can have the Views API handle all of the configuration (although then the exporter would have to know about it).

jkwatson: I'd rather not discuss a full-blown Views API in this OTEP. This OTEP is intentionally much more limited in scope, as I believe it is more achievable cross-language to be delivered in the GA timeframe.

Details for the open questions should be resolved in the specification PR. I'd like general consensus on an approach, via OTEP before we dive into the details, which should be done in the specification.

jmacd:
@open-telemetry/specs-metrics-approvers Please review if you have not already approved.
@jkwatson and reviewers, please close issues that are resolved.

text/0126-Configurable-Metric-Aggregations.md
jkwatson: I think this question is outside of the scope of this particular OTEP.  The concern about Counters is in the open questions. :)

text/0126-Configurable-Metric-Aggregations.md
```diff
@@ -0,0 +1,98 @@
...
+ MeterSdkProvider meterProvider = OpenTelemetrySdk.getMeterProvider();
+
+ // create a selector to select which instruments to customize:
+ InstrumentSelector instrumentSelector = InstrumentSelector.newBuilder()
+  .instrumentType(InstrumentType.COUNTER)
+  .build();
+
+ // create a configuration of how you want the metrics aggregated:
+ Configuration configuration =
+      Configuration.create(Aggregations.minMaxSumCount(), Temporality.DELTA);
```
bogdandrutu, line 60: To move us to the end goal can we add the concept of "View" here, that contains for the moment only these properties?

I think otherwise we will have troubles moving forward with the View proposal.

jkwatson: I'm very happy to do that. [my prototype in java calls them views ;]

bogdandrutu:
Would like to see this merged with #128

text/0126-Configurable-Metric-Aggregations.md
jkwatson: I updated the text to not be so afraid to use the word "View".  :) please take a look.

text/0126-Configurable-Metric-Aggregations.md
```diff
@@ -0,0 +1,99 @@
...
+
+The basic API has two parts.
+
+* InstrumentSelector - Enables specifying the selection of one or more instruments for the configuration to apply to.
+  - Selection options include: the instrument type (Counter, ValueRecorder, etc), and a regex for instrument name.
+  - If more than one option is provided, they are considered additive.
+  - Example: select all ValueRecorders whose name ends with ".duration".
+* View - configures how the batching and aggregation should be done.
+  - 3 things can be specified: The aggregation (Sum, MinMaxSumCount, Histogram, etc), the "temporality" of the batching,
+    and a set of pre-defined labels to consider as the subset to be used for aggregations.
```
bogdandrutu, line 37: I have in my mind this optimization that I want you to not make me hard to achieve when the full views is approved. There needs to be 2 lists of labels (labels from the instruments record calls, labels from correlation context). So probably I would just suggest to make sure we clearly specify that these are labels from the record calls.

Another suggestion that I have is that later (most likely) we want to do label rename (or do we?) maybe not.

So maybe a design like:

Labels: List<LabelKey>. Where then in the LabelKey we can have a source property, as well as maybe an "exported name" or other things, may not be a bad idea.

Right now we can start with only one string being in the LabelKey but later we can build on top of that.

What do you think?

text/0126-Configurable-Metric-Aggregations.md
```diff
@@ -0,0 +1,99 @@
...
+    - Note: "temporality" can be one of "DELTA" and "CUMULATIVE" and specifies whether the values of the aggregation
+      are reset after a collection is done or not, respectively.
+  - If not all are specified, then the others should be considered to be requesting the default.
+  - Examples:
+    - Use a MinMaxSumCount aggregation, and provide delta-style batching.
+    - Use a Histogram aggregation, and only use two labels "route" and "error" for aggregations.
+    - Use a quantile aggregation, and drop all labels when aggregating.
+
+In this proposal, there is only one configuration associated with each selector.
```
bogdandrutu, line 46: one "view"?

text/0126-Configurable-Metric-Aggregations.md
```diff
@@ -0,0 +1,99 @@
...
+
+As a concrete example, in Java, this might look something like this:
+
+```java
+ // get a handle to the MeterSdkProvider (note, this is concrete name of the default SDK class in java, not a general SDK)
+ MeterSdkProvider meterProvider = OpenTelemetrySdk.getMeterProvider();
+
+ // create a selector to select which instruments to customize:
+ InstrumentSelector instrumentSelector = InstrumentSelector.newBuilder()
```
bogdandrutu, line 55: Probably at least the "meter name" must be provided. This may get into too many details that we can cover in the specs, if that is the intention maybe just remind this :))

text/0126-Configurable-Metric-Aggregations.md
```diff
@@ -0,0 +1,99 @@
...
+  .instrumentType(InstrumentType.COUNTER)
+  .build();
+
+ // create a configuration of how you want the metrics aggregated:
+ View view =
+      View.create(Aggregations.minMaxSumCount(), Temporality.DELTA);
```
bogdandrutu, line 61: You want to provide the labels as well here correct?

text/0126-Configurable-Metric-Aggregations.md
bogdandrutu, line 46: Does that mean if two selectors match the same instrument only one View is active?

text/0126-Configurable-Metric-Aggregations.md
```diff
@@ -0,0 +1,99 @@
+# A Proposal For SDK Support for Configurable Batching and Aggregations (Basic Views)
+
+Add support to the default SDK for the ability to configure Metric Aggregations.
+
+## Motivation
```
bogdandrutu, line 5: I can see also another config, simpler:
* Change default temporality for instrument type - separated than view or selector.
* Enable/Disable an instrument default aggregation.

Maybe this are out of scope, but kind of think that they may be configured separately than the views.

text/0126-Configurable-Metric-Aggregations.md
jkwatson: I was thinking that you could provide just the instrument type, and get all of them (as a way to set the default).

text/0126-Configurable-Metric-Aggregations.md
jkwatson: right, with some sort of precedence to be determined later. (see my java prototype for an example).

text/0126-Configurable-Metric-Aggregations.md
jkwatson: I was thinking that the "disable" use-case could just be done by setting the aggregation to a no-op aggregator.

text/0126-Configurable-Metric-Aggregations.md
jkwatson: yeah, this example wasn't intended to be comprehensive, but more of a sketch of how it might look.

text/0126-Configurable-Metric-Aggregations.md
jkwatson: the labels part of this is definitely the most fuzzy in my head. I'm not 100% sure of what is needed at this level. I was thinking initially of just having the configuration contain the set of labels to reduce down to, and not consider renaming or re-mapping or anything fancy like that. 

So, assuming the instrumentation author has provided labels [a,b,c,d,e] the view might be configured just with [b,e].

text/0126-Configurable-Metric-Aggregations.md
jkwatson: done.

text/0126-Configurable-Metric-Aggregations.md
bogdandrutu: Worried that having this "only one" view requirement and relying on that to enable/disable default aggregations, or even doing some merging of "Views" may be too tricky and too restrictive for the future.

I would like you to think how would you do this if full views API is supported.

Bogdan self brainstorming:

Maybe having a way to retrieve all the configured views (InstrumentDescriptor + View) and allow:
* "Delete" - the user will use this to disable default aggregation for example
* You can clone/change the retrieved "View" after deleting it and set it back to update the "temporality"

It seems a lot of work to just disable "default aggregations". Maybe treat "default views" a bit special, and offer specific APIs to manipulate "all of them" like:
* "default views temporality" - these APIs may include selectors to facilitate.
* "disable all default views"
But this way we make it clear that these "DefaultViewsConfiguration" API refers to only the default views and you no longer need the "one view only" requirement.

bogdandrutu: Another way is to have a "strategy" param to the API: replace all (only supported option now).

jkwatson: I like the idea of considering this OTEP the "default views configuration". Maybe we can just simplify it down to just selecting by instrument type for this?

bogdandrutu: I think the selector can be fully defined, the API should include in the name that this applies only to default views, also operation that we accept:
* Update temporality for selected instruments
* Disable/Delete default views for selected instruments.

Changing labels is a complex operation that we should probably not support on the default views, recommended way should be deleted the default view and add a new view.

bogdandrutu: I can be convinced that labels can be also an operation available for default views, but probably we can discuss that later. Let's start with something then incremently add more stuff

bogdandrutu: Also if for selector only type is important now, we can start only with that. But these decisions I think can be made in the specs

james-bebbington:
Spoke with @bogdandrutu offline. I think his comments above alleviate most of the concerns I had around supporting multiple views in the future.

The only question not yet addressed was around supporting instrumentation of libraries that want to provide default "default views" (i.e. default aggregations / label sets) along with the instrumented metrics. There are a couple of possible options for this:

1. Put these functions in the OT API so they can be used in libraries (adds complexity)
2. Libraries supply a yaml file to specify recommended views that can be read by the OT official SDK
3. Put functions to define recommended views (not to install views) in a separate api extension that can be consumed by OT SDKs

jkwatson:
> Spoke with @bogdandrutu offline. I think his comments above alleviate most of the concerns I had around supporting multiple views in the future.
> 
> The only question not yet addressed was around supporting instrumentation of libraries that want to provide default "default views" (i.e. default aggregations / label sets) along with the instrumented metrics. There are a couple of possible options for this:
> 
> 1. Put the functions described in this OTEP in the OT API so they can be used in libraries (adds a lot of complexity)
> 2. Libraries supply a yaml file to specify recommended views that can be read by the OT official SDK
> 3. Put functions to define recommended views (not to actually install them) in a separate api extension that can be consumed by OT SDKs

I don't want this OTEP to address the issue of API-level views. That is a separate concern, and I don't want to muddy the waters at this point.

text/0126-Configurable-Metric-Aggregations.md
jkwatson: Agree with leaving details to the specs. I'd be happy to drop label support for now, to keep things ultra-simple and build from there. Do you want this OTEP updated to reflect that, or is that something we can wrangle out in the specs?

bogdandrutu: Just a sentence that say details still to be polished in specs is more than enough

jkwatson: I added a few more open questions to capture the discussion, and a note that we will resolve them in the spec. Sound ok?

bogdandrutu:
@jkwatson how can we make sure this otep and #128 will get us to one place?

jkwatson:
> @jkwatson how can we make sure this otep and #128 will get us to one place?

Fantastic question. 2 options that I see. 

1) Have this one merge, and have @tsloughter update #128 to assume this one exists.
2) Have @tsloughter tweak #128 to include the content here as a "phase 1" approach, and have 128 be the PR that gets merged (closing this one).

I'm totally happy either way.

tsloughter:
I'm fine with either of those as well. Which one is likely unimportant since it is the actual spec that developers looking to implement this will be relying on.

text/0126-Configurable-Metric-Aggregations.md
```diff
@@ -0,0 +1,102 @@
...
+consider a non-programmatic configuration option.
+
+## Prior art and alternatives
+
+* Prior Art is probably mostly in the [OpenCensus Views](https://opencensus.io/stats/view/) system.
+* Another [OTEP](https://github.com/open-telemetry/oteps/pull/89) attempted to address building a Views API.
+
+## Open questions (to be resolved in an official specification)
+
+1. Should custom aggregations should be allowable for all instruments? How should an SDK respond to a request for a non-supported aggregation?
```
MrAlias, line 88:
```suggestion
1. Should custom aggregations be allowable for all instruments? How should an SDK respond to a request for a non-supported aggregation?
```

text/0126-Configurable-Metric-Aggregations.md
```diff
@@ -0,0 +1,102 @@
...
+2. Should the requesting of DELTA vs. CUMULATIVE be only available via an exporter-only API, rather than generally available to all operators?
+3. Is regex-based name matching too broad and dangerous? Would the alternative (having to know the exact name of all instruments to configure) be too onerous?
+4. Is there anything in this proposal that would make implementing a full Views API (i.e. having multiple, named aggregations per instrument) difficult?
```
MrAlias, line 91: I feel like this question needs to be answered before this is merged. If the answer is "yes" this could be moving us away from our goal instead of towards it.

jkwatson: So, if anyone thinks that this proposal will stop views from happening, please speak up now, so we can resolve it ASAP.

bogdandrutu:
Decision during the metrics SIG was to merge this OTEP as a starting point and #128 will be changed to build on top of this.

PR #127, opened 2020-08-04
Update SignalFx Event/Log conventions

keitwb:
Per my conversation with @tigrannajaryan, we have determined that event dimensions are
more appropriate as attributes since a single logical Resource could emit events with
disparate dimension sets.  Also, we need to be able to unambiguously reverse a LogRecord to
a SignalFx Event, so properties are nested in a specially keyed map inside Attributes.

### DIFF
--- a/text/logs/0097-log-data-model.md
+++ b/text/logs/0097-log-data-model.md
@@ -734,20 +734,20 @@ Rest of SDIDs -> Attributes["syslog.*"]</td>
   <tr>
     <td>Category</td>
     <td>enum</td>
-    <td>Describes where the event originated and why. SignalFx specific concept. Example: AGENT. </td>
+    <td>Describes where the event originated and why. SignalFx specific concept. Example: AGENT. If this attribute is not present on the SignalFx Event, it should be set to the null attribute value in the LogRecord -- this will allow unambigous identification of SignalFx events when they are represented as LogRecords.</td>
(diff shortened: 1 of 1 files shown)


bogdandrutu:
To me this "vendor" specific mapping seems unreasonable in the OTEP. As we try to keep the project vendor neutral would really recommend to move this where the exporting to that vendor backend happens.

tigrannajaryan:
Correct, the purpose of showing the mappings in the OTEP was only to facilitate the discussion. We do not need to treat the examples as part of the spec, especially given that the mappings are vendor specific and vendors are free to implement it the way they think fits best the semantics of their protocol or data model.
I am fine if we want to update this OTEP to reflect SignalFx's understanding or just keep the OTEP as is. Ben wanted to keep the OTEP up-to-date, but if we don't think it is necessary then we can close this PR.
Happy to follow what TC recommends to do.

bogdandrutu:
Not strong opinion, but I think usually oteps should not be seen as the up-to-date state. Usually in oteps we make some assumptions or some decisions that we will not be able to keep up-to-date, specs should be up-to-date always. Otherwise I feel we have a problem because we need to update two places always.

tigrannajaryan:
Sounds good, close this PR then and keep the OTEP as is?

bogdandrutu:
Up to you for this specific PR. But in general because of the otep 4 approvals requirements this kind of changes are going to be hard to make.

PR #128, opened 2020-08-04
proposal for Views in the SDK

tsloughter:
This proposal has each instrument associated with a default view when created and views replace the default aggregation associated with an instrument.

### DIFF
--- a/text/metrics/0128-instrument-multi-aggregators.md
+++ b/text/metrics/0128-instrument-multi-aggregators.md
@@ -1,0 +1,82 @@
+# SDK Proposal: Default and Configurable Aggregations for Instruments
+
+This proposal builds off the work of John in
+[OTEP-126](https://github.com/open-telemetry/oteps/pull/126) and Chris in
+[OTEP-89](https://github.com/open-telemetry/oteps/pull/89).
+
+The unique part of this proposal is that all metric exports are based on
+views. Each `Instrument` has a default `View` created when the `Instrument` is
+created.
+
+## Definitions
+
+* Measurement: The value recorded along with a set of label keys and values.
+* Instrument: The thing a `Measurement` is associated with when recorded. Each
+  has a name, description, instrument kind and unit.
+* Aggregation: Defines how a set of measurements for an instrument are to be
+  combined for export.
+* View: An aggregation on an `Instrument` and some metadata (a name,
+  description, etc).
+
+## View
+
+* name: The name used when exporting metrics. 
+* description: String value describing what the aggregated measurements represent.
+* aggregation: The implementation of combining measurements to use.
+* labels: The labels to filter down to from the labels included in the measurement.
+* exporter: Each `view` can be associated with an exporter. By default it is set
+  to the default SDK exporter, if it is overridden as some form of `undefined`
+  value (depending on the language) this `view` is not exported and does not
+  need to be updated when a `measurement` is recorded for an `instrument`. This
+  leaves open the possibility of more complex export piplines in the SDK, as
+  opposed to simply being a boolean value.
+
+### Default Views
+
+An `Instrument` is more than simply the metadata for a metric that
+will be recorded. `Instruments` have default `views` defined for them. This
+eases the use by an end user as opposed to OpenCensus where a `View` was
+required to be created by the user for any measurement the user wants exported.
+
+The default view for an instrument uses the same name and description as the
+instrument and the default aggregation defined for the kind of instrument it
+is. The table in the [Metrics API
+specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#interpretation)
+provides the default aggregation per instrument kind:
+
+| **Name** | Instrument kind | Function(argument) | Default aggregation | Notes |
+| ----------------------- | ----- | --------- | ------------- | --- |
+| **Counter**             | Synchronous additive monotonic | Add(increment) | Sum | Per-request, part of a monotonic sum |
+| **UpDownCounter**       | Synchronous additive | Add(increment) | Sum | Per-request, part of a non-monotonic sum |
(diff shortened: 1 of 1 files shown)


text/metrics/0128-instrument-multi-aggregators.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+### Additional Views
+
+An instrument can have multiple views associated with it:
+
+```
+// Meter.new_view(InstrumentName, InstrumentKind, ViewName, ViewDescription, Aggregation, Labels)
+Meter.register_view(InstrumentName, InstrumentKind, "http.latency_histogram", "Histogram of HTTP request
+latencies", HdrHistogram, [HttpMethod])
```
jkwatson, line 76: I don't see the exporter in this call. At the top, you say an exporter can be associated with a view. How does that work?

text/metrics/0128-instrument-multi-aggregators.md
```diff
@@ -0,0 +1,82 @@
...
+
+Through the SDK the aggregation used by a kind of instrument can be overridden
+for all instruments of that kind or by name and kind of the instrument:
+
+```
+\\ for all instruments of a specific kind
+Meter.set_default_aggregation(ValueRecorder, TDigest)
+
+\\ for a specific instrument
+Meter.set_default_aggregation("http.latency", ValueRecorder, Distribution{Buckets=[0,100,200,500,1000]})
```
jkwatson, line 66: What about the "temporality" of the aggregation output (delta vs. cumulative). Where is that specified? Or is it a completely separate concern? Or is that something that the aggregation itself needs to be configured with?

jkwatson:
I had a couple of questions, but I generally like this. What I really like about it is that it can be implemented in 2 phases. 1) implement the default view support, and 2) implement full view support. The default view support maps reasonably well to my proposal in #126 (aside from the temporality question), but then can be built upon for full views.

text/metrics/0128-instrument-multi-aggregators.md
tsloughter: Ah yea, I switched it to not a boolean after having written this. But either way, it would just use the default in a function like this.

I kind of want `exporter` to for now just be treated like a boolean (either `none` or `default`) since the SDK PR only says multiple exporters could come, not that they exist today.

text/metrics/0128-instrument-multi-aggregators.md
tsloughter: Good question. Every time I think about it I don't see how it can't be somehow based on the exporter. But is that possible... I suppose it kind of is since this form of view technically has the exporter to use defined for it, so it could ask the exporter which to use?

jkwatson: Can you add an 'open questions' section to capture these... open questions? I'd like to make sure they're at least captured as things we need to figure out in the spec, but after the OTEP is approved.

tsloughter: Yup, will do, thanks.

text/metrics/0128-instrument-multi-aggregators.md
```diff
@@ -0,0 +1,82 @@
...
+* View: An aggregation on an `Instrument` and some metadata (a name,
+  description, etc).
+
+## View
+
+* name: The name used when exporting metrics. 
+* description: String value describing what the aggregated measurements represent.
+* aggregation: The implementation of combining measurements to use.
+* labels: The labels to filter down to from the labels included in the measurement.
+* exporter: Each `view` can be associated with an exporter. By default it is set
```
bogdandrutu, line 27: Do we need this now? This can be done in the exporter itself if needed. Trying to limit the work needed initially.

tsloughter:
@jkwatson yup, exactly, and I see implementing the default view support as more of a refactor of existing metric SDKs and not necessarily much new code.

text/metrics/0128-instrument-multi-aggregators.md
tsloughter: Could be. There will be far more views than exporters, and likely usually 1, so N views to 1 exporter, making configuration in the exporter be based on saying what of the N views to export while configuration on the view is a limited set of exporters to pick from. Just seems easier to put configuration on the thing that has fewer options to choose from.

text/metrics/0128-instrument-multi-aggregators.md
```diff
@@ -0,0 +1,88 @@
...
```
lzchen, line 27: How will specifying the exporter as part of the view work with in the whole  record -> collect -> export pipeline? Is specifying the exporter through views the recommended way of exporting now (instead of a PushController -> start_pipeline mechanism). If this is the case, will I need to define a View with an exporter even though I just want to export a simple metric (with no View functionality)?

tsloughter: @lzchen this proposal is that any instrument is only aggregated and thus exportable through a view, but there is a default view created automatically for every created instrument, so it results in the same outcome as we have today without it but is now through a view.

lzchen: @tsloughter 
What is the default SDK exporter? Also, if a user wants to use something other than the default exporter, they have to create a View?

tsloughter: A view is always used, even if only using a single exporter.

The "default exporter" in the case of only a single exporter would be just that exporter. I don't know what the spec would actually look like for supporting multiple exporters, this is currently a "todo" it seems in the metrics SDK PR. I figure it'd be kind of language specific. In Erlang we'd probably just either have users give a single exporter or a map of exporters where use of the atom `default` is given for the exporter to use when none is given to the view definition.

lzchen: So the default exporter would be something that users set correct? And then Views (if exporter is not specified), would use that/pick that up somehow?

tsloughter: Correct.

PR #133, opened 2020-08-18
Span Status Improvements (without errors)

mwear:
We recently discussed OTEP #123 at the error working group and agreed it introduced some good ideas, but they were coupled with error reporting. Exception data was handled by https://github.com/open-telemetry/opentelemetry-specification/pull/697 and the `successHint` inspired this `error.hint` PR: https://github.com/open-telemetry/opentelemetry-specification/pull/822.

There are currently proposals to remove Span.status (see: https://github.com/open-telemetry/opentelemetry-specification/issues/706). If we decide to keep it, this adaption of OTEP-123 could be an improvement worth considering.

See the [original OTEP](https://github.com/open-telemetry/oteps/pull/123/) for commentary and rationale.


text/0123-improve-span-status-api.md
```diff
@@ -95,55 +95,32 @@ translation effort from the native representation and capture at least the key
...
-3. SuccessHint - An optional boolean that represents the span author's best
-guess whether this status represents a successful or failed operation, however
-they choose to define those terms. For well-known status types I'd suggest the
-hint be ignored but for user-defined status types this is likely the only clue
-whether the span should be surfaced in a UI as being abnormal or failed.
+1. Domain - The name of the domain the status data applies to such as "HTTP",
+   "gRPC", "HRESULT", "POSIX". The list is end-user-extensible but common status
+   type names should be standardized.
+   (Perhaps there is already some standardization we could borrow?)
+2. Code - An integer status code. Can be combined with an status message. Either
```
Oberon00: Integer is insufficient. E.g. POSIX codes are symbolic names (like ENOENT) which do not have standardized integer values (e.g. ENOENT might have a different value on Linux than on BSD)

text/0123-improve-span-status-api.md
```diff
@@ -95,55 +95,32 @@ translation effort from the native representation and capture at least the key
...
+   a code or message are required.
+3. Message - A string status message. Can be combined with a status code. Either
+   a code or message are required.
```
arminru:
```suggestion
   a code or message is required.
```

text/0123-improve-span-status-api.md
```diff
@@ -154,46 +131,20 @@ events. The choice of storage may have some modest effects on memory usage
...
-additional error types.
-
-The current proposal only gives one string which can be used either for a 
+  user becomes unclear which one to pick. In general I expect this to be mitigated
+  with documented conventions, API defaults, and the potentially canonicalizing
+  data anywhere in the processing pipeline.
+
+The current proposal only gives one string which can be used either for a
 freeform message or a textual status code. Adding a 2nd string to the
 StatusData would allow both to be collected side by side. This is another example of
```
arminru: Same as above.

text/0123-improve-span-status-api.md
```diff
@@ -210,16 +161,9 @@ valuable with potentially a slight
 
 There have been a few past attempts to make improvements here:
 
-- [open-telemetry/oteps#69](https://github.com/open-telemetry/oteps/pull/69)
-- [#427](https://github.com/open-telemetry/opentelemetry-specification/pull/427)
-- [#432](https://github.com/open-telemetry/opentelemetry-specification/pull/432)
-- [#521](https://github.com/open-telemetry/opentelemetry-specification/pull/521)
-- [#599](https://github.com/open-telemetry/opentelemetry-specification/issues/599)
+- [open-telemetry/oteps#123](https://github.com/open-telemetry/oteps/pull/123)
```
arminru:
```suggestion
- [open-telemetry/oteps#123](https://github.com/open-telemetry/oteps/pull/123) on which this OTEP is based
- [open-telemetry/opentelemetry-specification#697](https://github.com/open-telemetry/opentelemetry-specification/pull/697) for reporting exceptions
```

text/0123-improve-span-status-api.md
```diff
@@ -95,55 +95,32 @@ translation effort from the native representation and capture at least the key
...
 StatusData qualified with the StatusType and some separator character. For
-example StatusData alone might create names like "FileNotFoundException",
-"503", "E_FAIL (0x80004005)", "SyntaxError on line 405: Did you forget a
-semicolon?", and "BadQuery".
-
-Exceptions could have progressive level of detail drilling into messages,
-stack traces, inner exceptions, links to source, etc if the exporter serialized
-sufficient data but how and whether that occurs is out-of-scope in this design. 
+example StatusData alone might create names like "503", "E_FAIL (0x80004005)",
+"Status Code 12: Unimplemented".
```
arminru: This needs to be updated since `StatusData` was renamed to `Code` and `StatusType` to `Domain` above and `Message` was introduced as a separate parameter.

tigrannajaryan:
> There are currently proposals to remove Span.status (see: [open-telemetry/opentelemetry-specification#706](https://github.com/open-telemetry/opentelemetry-specification/issues/706)). If we decide to keep it, this adaption of OTEP-123 could be an improvement worth considering.

What is the Error Working Group position on this? Do we remove Status? Do we keep it but improve? Do we have 2 opposing proposals because Error Working Group was not able to come to a concensus on this?

I would like to have more clarity on this otherwise I am not sure which of the opposing proposals to actively pursue. The OTEP to remove Status is also submitted: https://github.com/open-telemetry/oteps/pull/134

Just to be clear: I am happy if the goal is to discuss 2 proposals in parallel and chose a winning one, just want to understand if that is the intent here.

iNikem:
We had much more agreement on removing current status code, than on its replacement. So we want to remove Google RPC code bases status before GA, that is what my otep is about. And separately we discuss what should we put instead of it. Either before GA or after. And for that there are several options. One is this PR. Another is open-telemetry/opentelemetry-specification#822

tigrannajaryan:
I think we need to require having a reasonable replacement for conveying erroneous state in the Span. I believe this is needed before GA. Ideally we will _first_ know what is this replacement _before_ removing the Status. I am also happy if the answer is that we don't need _any_ replacement, but I would like that analysis to also be done before removal of the Status.

iNikem (commented):
This proposal definitely solves some of the problems of the current Span.Status. But it leaves one (probably the most) important question open: how to quickly determine if span represents potential failure?

text/0131-improve-span-status-api.md
```diff
@@ -0,0 +1,238 @@
+# Improve Span Status API
+
+Allow the Span Status API to represent more kinds of status
+
+## Motivation
+
+Right now OpenTelemetry Status is defined as an enumeration of gRPC status
```
bogdandrutu, line 7: It is not gRPC :)) I think I keep saying this. They are Google's codes which are used in gRPC but they are used in http APIs as well that Google expose.

text/0131-improve-span-status-api.md
```diff
@@ -0,0 +1,238 @@
...
+
+We should aim for representations that require minimal-to-no end-developer
+translation effort from the native representation and capture at least the key
+numeric/string data developers are likely to search for or relate to. I suggest
+status is this set of information:
+
+1. Domain - The name of the domain the status data applies to such as "HTTP",
+   "gRPC", "HRESULT", "POSIX". The list is end-user-extensible but common status
+   type names should be standardized.
+   (Perhaps there is already some standardization we could borrow?)
```
bogdandrutu, line 101: Google did the same thing, and at the end they decided to remove this and support only what they called "canonical status" because custom translation were hard to keep track of.

One thing that was very important was the notion of "IsOk" on the Status itself, every domain has it's own codes, but in a lot of time was important to know if it is ok or not :)

iNikem:
@mwear Should this be closed in favor of #136 ?

tedsuo:
@mwear please reopen if you change your mind about this

PR #134, opened 2020-08-19
Remove Span.Status

iNikem:
Addresses open-telemetry/opentelemetry-specification#706


text/0134-remove-span-status.md
```diff
@@ -0,0 +1,88 @@
...
+and during Errors Working Group meetings a lot of concerns were raised about suitability of the current `Span.Status`
+field encoded as Google RPC code for recording a result of arbitrary monitored operations from various domains
+(http requests, database queries, file system operations etc). It is hard to define clear and umambigous mapping
+from those domain to a fixed enum originated from RPC.
+
+Therefore the proposal is to remove the `Span.Status` field from Trace API and OTLP.
+
+## Internal details
+
+Several areas are affected by this change and will require the coordinate effort to make this happen.
```
tigrannajaryan, line 25:
```suggestion
Several areas are affected by this change and will require a coordinated effort to make this happen.
```

text/0134-remove-span-status.md
```diff
@@ -0,0 +1,88 @@
...
+This will allow for older pipelines, still relying on `Span.Status` to continue to work with as little disruption as possible.
+The exact duration of this transition period will be left for Collector's maintainers to decide.
+
+* If incoming OTLP data has `Span.Status` present, that will NOT be removed and will be passed on as-is.
+* If incoming OTLP data has `Span.Status` present, Collector will translate that to temporary semantic attributes as follows:
+
+|Status|Tag Key| Tag Value|
+|--|--|--|
+|StatusCanonicalCode | `otel.deprecated_status_code` | Name of the code, if that differs from ‘OK’ |
+|Message *(optional)* | `otel.deprecated_status_description` | `{message}` |
```
tigrannajaryan, line 70: Since the field is called `Status.message` I suggest:
```suggestion
|Message *(optional)* | `otel.deprecated_status_message` | `{message}` |
```

text/0134-remove-span-status.md
```diff
@@ -0,0 +1,88 @@
...
+which has `Span.Status` present, and new data without Status.
+This will allow for older pipelines, still relying on `Span.Status` to continue to work with as little disruption as possible.
+The exact duration of this transition period will be left for Collector's maintainers to decide.
+
+* If incoming OTLP data has `Span.Status` present, that will NOT be removed and will be passed on as-is.
+* If incoming OTLP data has `Span.Status` present, Collector will translate that to temporary semantic attributes as follows:
+
+|Status|Tag Key| Tag Value|
+|--|--|--|
+|StatusCanonicalCode | `otel.deprecated_status_code` | Name of the code, if that differs from ‘OK’ |
```
tigrannajaryan, line 69: What value do we set if it is 'OK'? Do we omit the `otel.deprecated_status_code` attribute?

text/0134-remove-span-status.md
```diff
@@ -0,0 +1,88 @@
...
+* When receiving data in Jaeger format, no special handling will take place.
+Jaeger `error` tag will be translated in the usual way to OTLP `error` attribute.
+See `Open questions` below.
+* When exporting data into Jaeger format, if a new temporary attribute `otel.deprecated_status_code` is present,
+then Jaeger `error` tag will be set to `true`.
+See `Open questions` below.
+
+## Open questions
+
+Neither OpenTelemetry Specification nor Protocol currently assigns any special meaning to the `error` attribute mentioned above
```
tigrannajaryan, line 86: I am not quite sure what happens if Status is removed but we do not (yet) have the "error"-like attribute. How will the errors be indicated? Do we lose this capability? Or the intent is to wait until "error" indicator (whatever it is) is first added to the specification and only after that we remove Status?

text/0134-remove-span-status.md
iNikem: Yes, that was my idea. And probably should skip setting `otel.deprecated_status_message` as well

text/0134-remove-span-status.md
iNikem: Current proposal is to remove status right now, file a follow up issue and work with Errors WG to resolve it. Either via one of the already existing proposals, like open-telemetry/opentelemetry-specification#822, or a new one.

text/0134-remove-span-status.md
tigrannajaryan: Sounds good, please make it explicit in the text.

text/0134-remove-span-status.md
tigrannajaryan: Got it. I see in this OTEP that you already thought through the transition process, which is great, but I need to mention this for visibility: we must be careful with Collector changes. There are people who use OpenTelemetry Collector in production already. All intermediate states of the Collector need to be fully functional.

iNikem:
@open-telemetry/technical-committee Can this be merged now? Then I can proceed doing actual changes.

text/0134-remove-span-status.md
```diff
@@ -0,0 +1,89 @@
...
+No changes needed.
+
+### Trace SDK exporters
+
+Remove translation rules pertaining to `Span.Status`.
+
+### Span semantic conventions
+
+* From HTTP conventions, remove `Status` section.
+* In gRPC conventions, replace the requirement to set span status with the requirement to set span attribute `grpc.status`
```
Oberon00, line 43: Should the OTLP exporter (optionally) do the translation of grpc and HTTP status attributes to deprecated_status for now? Or is this not worth it?

text/0134-remove-span-status.md
```diff
@@ -0,0 +1,89 @@
...
+### Collector
+
+Collector will have a transition period during which it has a special handling both for old data,
+which has `Span.Status` present, and new data without Status.
+This will allow for older pipelines, still relying on `Span.Status` to continue to work with as little disruption as possible.
+The exact duration of this transition period will be left for Collector's maintainers to decide.
+
+* If incoming OTLP data has `Span.Status` present, that will NOT be removed and will be passed on as-is.
+* If incoming OTLP data has `Span.Status` present and it is not equal to `OK`,
+Collector will translate that to temporary semantic attributes as follows:
```
Oberon00, line 66: Is the previous point not enough? Do we need to duplicate the status as attributes?

text/0134-remove-span-status.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Collector
+
+Collector will have a transition period during which it has a special handling both for old data,
+which has `Span.Status` present, and new data without Status.
+This will allow for older pipelines, still relying on `Span.Status` to continue to work with as little disruption as possible.
+The exact duration of this transition period will be left for Collector's maintainers to decide.
+
+* If incoming OTLP data has `Span.Status` present, that will NOT be removed and will be passed on as-is.
+* If incoming OTLP data has `Span.Status` present and it is not equal to `OK`,
```
Oberon00, line 65: Should a non-empty message also be translated independently of the status code?

text/0134-remove-span-status.md
```diff
@@ -0,0 +1,89 @@
...
+|Message *(optional)* | `otel.deprecated_status_message` | `{message}`|
+
+* When receiving data in Zipkin format, no special handling will take place.
+All Zipkin tags present will be translated verbatim to OTLP attributes.
+* When exporting data into Zipkin format, no special handling will take place.
+All OTLP attributes will be translated verbatim to Zipkin tags.
+`Span.Status` if present, will NOT be translated.
+* When receiving data in Jaeger format, no special handling will take place.
+Jaeger `error` tag will be translated in the usual way to OTLP `error` attribute.
+See `Open questions` below.
```
Oberon00, line 80:
```suggestion
Jaeger `error` tag will be translated in the usual way to an OTLP error attribute
(see [Open questions](#open-questions)).
```

The use of code formatting was confusing me when reading this, because that made it sound like the attribute actually exists with that name.

text/0134-remove-span-status.md
```diff
@@ -0,0 +1,89 @@
...
+* When exporting data into Jaeger format, if a new temporary attribute `otel.deprecated_status_code` is present,
```
Oberon00, line 81:
```suggestion
* When exporting data into Jaeger format, if a new temporary attribute `otel.deprecated_status_code` is present
and does indicate a non-OK status,
```

Oberon00 (approved):
I have some questions about details, but all in all I agree to remove Status.

text/0134-remove-span-status.md
iNikem: As one of the reasons to remove `status` was the ambiguity of such translations (at least in http part), I don't think it is worth doing. And ideally we will not produce this Google codes based status anymore.

text/0134-remove-span-status.md
iNikem: Yes, because new OTLP receivers will not read `status` field from the wire. But we still may want to pass this information if it was available in the Collector.

text/0134-remove-span-status.md
iNikem: Message is part of the `Status`, how can it be independent?

Oberon00: I mean, it makes sense to add deprecated_status when "incoming OTLP data has `Span.Status` present and it is not equal to `OK`", but it may make sense to independently set deprecated_message when "incoming OTLP data has `Span.Status` present and **message non-empty**`"

text/0134-remove-span-status.md
Oberon00: Do we expect code behind new OTLP receivers that don't read (deprecated_)status anymore to still need status?

text/0134-remove-span-status.md
Oberon00: Make sense.

text/0134-remove-span-status.md
iNikem: I think having message without code will be confusing, because current state is exactly the opposite: you may have code without message.

text/0134-remove-span-status.md
iNikem: Don't know, that is way we leave it there for the time being. Collector maintainers may decide when they want to remove this translation.

text/0134-remove-span-status.md
Oberon00: But you may also have a message with an OK code. If you are worried about confusion, then set the status code additionally when the message is non-empty.

text/0134-remove-span-status.md
carlosalberto: @tigrannajaryan It seems open-telemetry/opentelemetry-specification#822 will be effectively merged soon (the `error.hint` Span attribute at least). Would that be enough for the Collector?

carlosalberto:
@open-telemetry/specs-approvers Please review this OTEP. We need one more approval to proceed (and it's possible some of you haven't had time to pay attention to check this OTEP out ;) )

arminru (approved):

Oberon00 (approved):

text/0134-remove-span-status.md
tigrannajaryan: I am reluctant to give an approval until I clearly understand what we replace this by.

If https://github.com/open-telemetry/opentelemetry-specification/pull/822 is the replacement can we wait for it be merged first and update this OTEP to make use of it? It is an open question in this OTEP and is one of the central problems that this OTEPs needs to solve, but remains partially unsolved for now.

tigrannajaryan: @iNikem To be clear, I am not opposed to this OTEP, but I think it could be improved and an open question can be hopefully closed if https://github.com/open-telemetry/opentelemetry-specification/pull/822 is accepted.

We do not typically alter OTEPs after they are merged, so I don't know if a follow-up PR that would make the improvement is a reasonable approach. Because of this I also think that it may make sense to wait for https://github.com/open-telemetry/opentelemetry-specification/pull/822 and then make any required changes to this OTEP.

If you do not think that https://github.com/open-telemetry/opentelemetry-specification/pull/822 can have any significant impact on this OTEP then we don't need to wait for it and can assume https://github.com/open-telemetry/opentelemetry-specification/pull/822 does not exist. However in that case the way I perceive this PR changes a bit, because it seems to leave us without standardized error reporting approach and I think we need such approach.

bogdandrutu: I think the otep may change as a result of resolving any open question, but that doesn't mean we should stop this. I think the main point of this otep is the removing of span status and an action item is required to think about if we need an equivalent of `error.hint` or not separately.

Trying to make progress so we don't get stuck, but I can see your point as well, if you suggest that we really need a way to signal error from the application (which I don't believe we need) then I understand why we should block this until we have that.

iNikem: @tigrannajaryan can you express your opinion in open-telemetry/opentelemetry-specification#822? If you think that it may resolve your hesitation about this PR

bogdandrutu: > I am not against moving forward with this OTEP, but I believe it will have an important open question and will be unactionable in the Collector if it does not take into account either the existence of "error.hint" or the lack of whatever else needs to replace Span Status. IMO, it will mean we will need another OTEP to correct this OTEP once "error.hint" is merged or rejected.

Not sure you need an OTEP for that, that is a semantic convention, and collector will use it.

tigrannajaryan: @bogdandrutu sure, that probably works too.

tigrannajaryan:
All, I don't want to be the person who blocks this OTEP. My only worry is that "error.hint" PR is not going to be merged due to disagreements and we will remove the Span status without having a reasonable replacement.

If everyone else in this thread thinks that it is not a problem and we will find a solution to that I am happy to approve and go ahead with this OTEP.

tedsuo:
@iNikem  Now that [OTEP 136](https://github.com/open-telemetry/oteps/blob/master/text/trace/0136-error_flagging.md) is merged, this PR may be closed.

iNikem:
Closed in favor of #136

PR #135, opened 2020-08-21
Introduce sampling score and propagate it with the trace

lmolkova:
This is a reincarnation of https://github.com/open-telemetry/oteps/pull/107.

The spec described consistent sampling between existing tracing tools that use vendor-specific sampling algorithms and OTel and enables an upgrade path that vendors may use to onboard customers on otel without breaking cross-tracing-tool traces.

The delta OTEP introduces is relatively small:

1. Add `SamplingResult.Tracestate` field: sampler should be able to assign a
   new tracestate for to-be-created span
2. Add convention for `sampling.score` attribute on span (let's discuss attribute vs field).
**[Update] after review**
3. Add notion of `SamplingScoreGenerator` that is capable of calculating float score from sampling parameters.
   It  has `TraceIdRatioGenerator`, `RandomGenerator`, and possible other implementations.
   - Change `TraceIdRatioBased` sampler to use corresponding generator and serve as generic probability sampler with configurable score generation approach.
4. Add `ExternalScoreSampler` implementation of `Sampler`. It's created with probability value and implementation of `SamplingScoreGenerator`.

With this OTEP we are trying to come up with long-term plan for interop between Microsoft SDKs and services and OpenTelemetry-enabled apps. We want to make sure this solution exists, while implementation can wait.

The specification changes can wait till post-GA, however, I've heard several requests for p1 before GA (ability to modify `tracestate` by Sampler) - https://github.com/open-telemetry/opentelemetry-specification/issues/856.


text/trace/0107-sampling-score.md
```diff
@@ -0,0 +1,217 @@
...
+
+Consistent sampling decision made in each app of a distributed trace is
+important for better user experience of trace analysis. Consistency is achieved
+by following means:
+
+1. Same hashing algorithms used across all apps in a trace.
+   Coordination of sampling algorithms across multiple apps not always possible:
+   for example existing components in a system use vendor-specific
+   tracing tool (pre-OpenTelemetry and update is hard to justify) while there
+   is a desire to use OpenTelemetry for new components.
```
bogdandrutu, line 34: How is this new solution helping with this problem?

text/trace/0107-sampling-score.md
```diff
@@ -0,0 +1,217 @@
+# Associating sampling score with the trace
+
+Enable consistent sampling across distributed application with different
+sampling rates and probability calculation algorithms.
+
+## TL;DR
+
+**Score** is a floating point number associated with
+the trace. It's calculated when trace starts and flows in the `tracestate`,
+it's used by samplers to make consistent sampling decisions.
```
bogdandrutu, line 10: May want to clarify that this is not the initial sampling probability, but a score that combined with the local probability determines if a trace is sampled or not.

text/trace/0107-sampling-score.md
```diff
@@ -0,0 +1,217 @@
...
+(or trace-flags).
+
+*Score* is not related to sampling *rate* (aka *probability* which represents
+sampler's configuration not specific to trace).
+
+## Motivation
+
+The goal is to enable a mechanism for consistent (best effort) sampling
+between services with different sampling rates and different probability
+calculation algorithms (for interoperability with existing tracing tools).
```
bogdandrutu, line 24: You are referring couple of times that the trace id based algorithm does not work well enough, can you add in the motivation section the reason why?

text/trace/0107-sampling-score.md
```diff
@@ -0,0 +1,217 @@
...
+   sampling rates across different components.
+
+## Explanation
+
+Sampling propabaility is generated by the first service to make sampling
+decision. It's a random float number in [0, 1] range.
+Score is stamped on the span and also propagated further within `tracestate`.
+
+Next service reads score from `tracestate` (instead of calculating it from
+trace-id) and compares it with its sampling rate to make sampling decision.
```
bogdandrutu, line 48: Why is this better?

text/trace/0107-sampling-score.md
```diff
@@ -0,0 +1,217 @@
...
+
+Score is also exposed through span attributes. Vendors can leverage it
+to sort traces based on their completeness: the lower the value of score is,
+the higher the chance it was sampled it by each component.
```
bogdandrutu, line 52: Do we always need to have a score? How do we calculate the score if a custom sampler is used like always on? How does the score interact with custom samplers that do rate based sampling 1 trace every 1 second? How does the backend know when score was used to do sampling decision or just ignored?

text/trace/0107-sampling-score.md
```diff
@@ -0,0 +1,217 @@
...
+2. Sampling flag propagated from the head component/app is used by downstream
+   apps to sample in a given trace.
+   It requires to trust upstream decision and does not allow to have different
+   sampling rates across different components.
+
+## Explanation
+
+Sampling propabaility is generated by the first service to make sampling
+decision. It's a random float number in [0, 1] range.
+Score is stamped on the span and also propagated further within `tracestate`.
```
bogdandrutu, line 45: probably we need a sampler name + sampler details. So here would be "ExternalScoreSampler(0.5 /*probabilty*/, 0.12343 /*score*/)"

text/trace/0107-sampling-score.md
```diff
@@ -0,0 +1,217 @@
...
+`ExternalScoreSampler`(name TBD) is configured by user. Within `ShouldSample`
+callback sampler
+  - generates random float score (6-9 digits) in [0, 1] interval
+  - makes sampling decision by comparing generated score to configured rate
+  - if decision is `RECORD` (or `RECORD_AND_SAMPLED`), sampler adds
+    `sampling.score` attribute to attributes collection of to-be-created span
+  - regardless of sampling decision: prepends `sampling.score` key-value pair
+    into tracestate of to-be-created span
+- Downstream service continues a trace but has different sampling rate (it's
+  also configured to use `ExternalScoreSampler`)
```
bogdandrutu, line 101: What happens if not configured to use `ExternalScoreSampler`? I think I already asked this question, sorry for duplication.

text/trace/0107-sampling-score.md
```diff
@@ -0,0 +1,217 @@
...
+score in order to interoperate with OpenTelemetry and should recommend
+customers to configure such sampler.
+
+It may be the case that after migration to OpenTelemetry is finalized, the need
+of `sampling.score` will decrease and customers can remove
+`ExternalScoreSampler` from configuration.
+
+## Prior art and alternatives
+
+Related discussions on [Probability sampler](https://github.com/open-telemetry/opentelemetry-specification/pull/570)
```
bogdandrutu, line 140: Some other related specs:
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#traceidratiobased

text/trace/0107-sampling-score.md
```diff
@@ -0,0 +1,217 @@
...
+
+### Sampling.score is NOT priority
+
+Priority is used by [OpenTracing](https://github.com/opentracing/specification/blob/master/semantic_conventions.md)
+as an implementation-specific hint for sampler to prioritize recording a span.
+
+[OpenTelemetry collector](https://github.com/open-telemetry/opentelemetry-collector/blob/60b03d0d2d503351501291b30836d2126487a741/processor/samplingprocessor/probabilisticsamplerprocessor/testdata/config.yaml#L10)
+uses `sampling.priority` to hint collector's sampler decision
+
+To avoid conflicts with existing implementations we fo not reuse priority term.
```
bogdandrutu, line 150: s/fo/do?

text/trace/0107-sampling-score.md
```diff
@@ -0,0 +1,217 @@
...
+    attributes collection of to-be-created span
+
+Here is a [proof of concept](https://github.com/lmolkova/opentelemetry-dotnet/pull/1)
+in .NET.
+
+### Specification Delta
+
+1. Add `SamplingResult.Tracestate` field: sampler should be able to assign a
+   new tracestate for to-be-created span
+2. Add convention for `sampling.score` attribute on span (TBC). Check out
```
bogdandrutu, line 116: s/TBC/TBD?

text/trace/0107-sampling-score.md
```diff
@@ -0,0 +1,217 @@
...
+as an implementation-specific hint for sampler to prioritize recording a span.
+
+[OpenTelemetry collector](https://github.com/open-telemetry/opentelemetry-collector/blob/60b03d0d2d503351501291b30836d2126487a741/processor/samplingprocessor/probabilisticsamplerprocessor/testdata/config.yaml#L10)
+uses `sampling.priority` to hint collector's sampler decision
+
+To avoid conflicts with existing implementations we fo not reuse priority term.
+
+## Open questions
+
+### Score calculation: can we use ProbabilitySampler?
```
bogdandrutu, line 154: No more probability sampler :) It is called TraceIdRatioBased.

text/trace/0107-sampling-score.md
```diff
@@ -0,0 +1,217 @@
...
+distribution.
+
+Assuming trace-ids are uniformly distributed, `ProbabilitySampler` can generate
+score, so the flow could look like this:
+
+`ExternalScoreSampler.ShouldSample`:
+
+- checks if `sampling.score` is available in the tracestate
+- if it's not there, invokes `ProbabilitySampler`, which calculates score
+  and populates it on the attributes
```
bogdandrutu, line 166: I like this.

text/trace/0107-sampling-score.md
```diff
@@ -0,0 +1,217 @@
...
+Fallback to `ProbabilitySampler` improves the case when `tracestate` is trimmed
+so there is a chance sampling could be consistent if same probability
+calculation algorithm was used.
+
+#### Cons
+
+- There is no requirement for trace-ids to be uniformly distributed
+- No clear boundary between `ProbabilitySampler` and `ExternalScoreSampler`.
+`ProbabilitySampler` needs to set score in attribute even if there is no
+`ExternalScoreSampler`.
```
bogdandrutu, line 180: Some of this may be added in the motivation as well to prove the current path does not achieve what you suggest.

text/trace/0107-sampling-score.md
lmolkova: I don't see why we'd need the sampler details on the tracestate. What scenario are you thinking about?

text/trace/0107-sampling-score.md
lmolkova: We don't always need to have a score. We need it when there are multiple possible sampling algorithms in the same app. 
Example when we need it:
```
Service A uses Azure Monitor SDK, uses hash1(trace-id), assigns score  -> Service B uses OTel with hash2(trace-id)
                                                                       -> Service C uses AzureMontior with hash1(trace-id)
```
                                                 
Services A and C use Azure Monitor SDK for years and it's unreasonable to ask them to upgrade to OTel. 
It is reasonable to ask them to update to the new version of Azure Monitor that supports score.

We don't need score when only OTel is used everywhere (or the same algorithm). 

So, we'll tell our customers to configure `ExternalScoreSampler` and fallback to OTel TraceIdRationBased algorithm if score is not in the tracestate.

Default case for OTel users does not change, this is opt-in behavior. 
Custom samplers that use  1 trace every 1 second are not compatible with score - they don't pursuit consistent sampling goal anyway and applying them together with `ExternalScoreSampler` seems to be a misconfiguration.

text/trace/0107-sampling-score.md
lmolkova: same as today - whatever sampler you configured would work according to the spec. Score (if present in tracestate) will be ignored and propagated futher

lmolkova:
thanks for the review @bogdandrutu. I believe I addressed your questions/comments in the spec.

Oberon00:
I think

> 1. Add `SamplingResult.Tracestate` field: sampler should be able to assign a
   new tracestate for to-be-created span

which you already created https://github.com/open-telemetry/opentelemetry-specification/issues/856 for should be done separately. That change is IMHO the most important one (breaking in many interface-based languages), it is also the simplest one and the other parts can be added later.

lmolkova:
> I think
> 
> > 1. Add `SamplingResult.Tracestate` field: sampler should be able to assign a
> >    new tracestate for to-be-created span
> 
> which you already created [open-telemetry/opentelemetry-specification#856](https://github.com/open-telemetry/opentelemetry-specification/issues/856) for should be done separately. That change is IMHO the most important one (breaking in many interface-based languages), it is also the simplest one and the other parts can be added later.

I agree that other parts could be added later implementation-wise.
At the same time, we are making some decisions now on how to support sampling interop between current Azure-specific algorithms and OpenTelemetry without breaking existing customers.  We want to have a consensus on this approach going forward so we can make such decisions today.  
The ultimate goal is to have this mechanism in vanilla OTel and avoid shipping Azure-specific packages (except exporters). It also seems to be generally useful for interop/sampling orchestration scenarios.
So I'd like to move the OTEP forward, I can mark it as post-GA if it helps. Certainly, implementation can wait till post-GA.

text/trace/0107-sampling-score.md
lmolkova: so I've tried to follow this approach in the new version. To make it clean, I suggest separating score generation from 
sampling.

Samplers can use score generation, and also attach attributes, change tracestate.
If we allow falling back to another sampler:
- we need to do score propagation back from one sampler to the ExtenralScore one
- coordinate possible multiple changes of tracestate and attributes

If we have a layer responsible for score calculation (random, deterministic of any sort), we can make it much simpler and cleaner.

lmolkova:
@specs-trace-approvers what could be the next steps to move this spec forward?

Oberon00:
I think we should definitely do open-telemetry/opentelemetry-specification#856 before GA but I suspect that most don't want to take the time to read and understand the whole OTEP before GA.
On the other hand there were some worries voiced in yesterday's SIG meeting that the Sampling SDK may not be in a GA-ready state yet, because no one has tried writing custom samplers, etc yet. So this OTEP might be a good chance to test the sampling SDK.

CC @open-telemetry/specs-approvers @open-telemetry/specs-trace-approvers

Oberon00:
@lmolkova You could maybe move this forward by making a spec PR for open-telemetry/opentelemetry-specification#856. Then after that is merged, you can make the OTEP a bit shorter 😃

bogdandrutu (approved):
Overall I am ok with the direction of this OTEP

text/trace/0107-sampling-score.md
```diff
@@ -0,0 +1,232 @@
...
+
+2. Sampling flag propagated from the head component/app is used by downstream
+   apps to sample in a given trace.
+   It requires to trust upstream decision and does not allow to have different
+   sampling rates across different components.
+
+## Explanation
+
+Sampling score is generated by the first service to make sampling
+decision. It's a random float (6-9 digits precision) number in [0, 1] range.
```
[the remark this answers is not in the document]
lmolkova: I could not find versions of IEEE 754 implemented in popular languages. They all refer to IEEE 754, I updated the spec.

text/trace/0107-sampling-score.md
```diff
@@ -0,0 +1,223 @@
...
+sampler's configuration, not specific to trace.
+
+Sampler can compare the *score* with the configured *probability* to make
+sampling decisions.
+
+Service that starts the trace calculates the score and adds it to the
+`tracestate` so downstream services can re-use it to make their sampling
+decisions *instead of* re-calculating score as a function of trace-id
+(or trace-flags). This allows to configure sampling algorithm on the first
+service ans avoid coordination of algorithms when multiple tracing tools are
```
MSNev, line 21: Typo: ans -> and

jmacd:
@ericmustin I remember the history of this thread, but I believe @lmolkova has moved on from OpenTelemetry.

The conclusion from #148 is that encoding inclusion probability (how we count sampled events) is different from how we ensure that traces are complete when sampled by some scheme. OTel's current specification includes only one scheme for ensuring that sampled traces are complete, the W3C "is-sampled" flag (part of TraceFlags, in SpanContext). 

See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spancontext

May we close this issue?

ericmustin:
@jmacd gotcha ty for the followup. makes sense, I’ll do my bike shedding(haha jk...unless?) over on #148 , looks like there’s some thoughtful work on inclusion probability there that would address the use case I had in mind.

oertl:
@jmacd I prefer to keep this issue open. Propagating the sample rate or _adjusted count_ from the root span, where the sampling decision was made, allows to set the _adjusted count_ for a span properly as already mentioned in https://github.com/open-telemetry/oteps/pull/148#discussion_r626404908. This would allow extrapolation of individual spans without having to retrieve this information from the root span.

jmacd:
@oertl OK, let's keep discussing. My understanding of this proposal in this PR is that it _does not propagate inclusion probability_. I am definitely in favor of solutions for propagating inclusion probability. The solution in this PR talks about the use of an explicit random number for use coordinating different-sampling schemes w/o addressing inclusion probability. I'm not familiar with the system @lmolkova describes here.

oertl:
@jmacd, you are right this is a different issue. However, it tackles a still unsolved problem with regard to sampling. 

The trace ID ratio based sampler, only works properly if the trace ID is uniformly distributed, which is AFAIK, not well specified. In order to get sampling right, if you do not know the distribution of the trace ID and if you just know that the trace ID is unique, you need to hash the trace ID. A high-quality hash function, is able to transform the trace ID into a uniformly distributed hash value which can then be used for ratio-based sampling. A list of fast, non-cryptographic, high-quality hash function can be found at https://github.com/rurban/smhasher#summary. Apart from the overhead, a further problem is that the hash function needs to be consistently implemented for all supported languages. 

To save hashing costs, the root span may calculate the hash value and propagate it with the trace ID, such that child spans do not need to hash the trace ID again and again. If the hash value is only calculated once at the root span and never again for child spans, because they are always using the same precalculated hash value from the parent, one could replace the hash value computed at the root span by any random number. This is exactly what is proposed here where the generated random value is called _sampling score_. It is essentially nothing else than a secondary trace ID that is uniformly distributed.

By the way, if ratio-based sampling is restricted to sample rates that are powers of one half as proposed (see https://github.com/open-telemetry/oteps/pull/148#issuecomment-803289438), it would be sufficient to propagate just the number of leading zeros of the _sampling score_, because this is the only information needed for the sampling decision in this case.

lmolkova:
I'm going to close this OTEP as I'm not working on sampling anymore and it doesn't look like there is much interest. If anyone is interested in it, feel free to take any parts of it, and happy to share any context.

jmacd:
This goals of this OTEP are, I believe, addressed in #168.

PR #136, opened 2020-09-09
Error flagging with status codes

tedsuo:
This proposal attempts to address error handling while making the least number of breaking changes to our current system. It retains SpanStatus. Other proposals we essentially reinventing it, only in a way that would introduce a number of breaking changes. It allows current instrumentation to continue to work as-is, while providing guidance to instrumentation authors. It includes a way to differentiate between the default errors provided by instrumentation and intentional error flagging provided by the end user.

If accepted, this proposal would replace OTEP #134, and addresses issues [#559](https://github.com/open-telemetry/opentelemetry-specification/issues/599) and [#706](https://github.com/open-telemetry/opentelemetry-specification/issues/706).


Oberon00 (commented):
All in all, looks sensible to me.

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+First of all, every tracing system to-date comes with a default set of errors. No system requires that end users start completely from scratch. So... be calm!! Have faith!!
+ 
+While flagging errors can be a subjective decision, it is true that many semantic conventions qualify as an error. By providing a default mapping of semantic conventions to errors, we ensure compatibility with existing analysis tools (e.g. Jaeger), and provide guidance to users and future implementers.
+ 
+Obviously, all systems are different, and users will want to adjust error reporting on a case by case basis. Unwanted errors may be suppressed, and additional errors may be added. The collector will provide a processor and a configuration language to make this a straightforward process. Working from a baseline of standard errors will provide a better experience than having to define a schema from scratch.
+ 
+Note that analysis tools are free to disregard Span Status, and do their own error analysis. For these systems, the only Status codes of import are `OK_OVERRIDE` and `ERROR_OVERRIDE`.
+ 
+Removing the need to explicitly set span status, and instead have it
```
iNikem, line 64: Incomplete sentence?

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+### Error Mapping Schema
+As part of the specification, OpenTelemetry provides a canonical mapping of semantic conventions to status codes. This removes any ambiguity as to what OpenTelemetry ships with out of the box.
+ 
+### Error Processor
+The collector will provide a processor and a configuration language to make adjustments to this error mapping schema. This provides the flexibility and customization needed for real world scenarios.
+ 
+### Semantic conventions
+`error.message` - A description of the error to be displayed in the analysis tool. This attribute may be set on both spans and events. This is optional, but useful when the error does not map to an exception, or an existing semantic convention. It is not necessary to set this for exceptions or our standard error mapping.
+ 
+### Convenience methods
```
iNikem, line 39: Can you elaborate on that, please? We had a lot of controversy about API changes before. Do you propose any new API methods? Or changes for the semantic of existing ones?

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+Analysis tools are free to disregard status codes, in favor of their own approach to error analysis. However it is strongly suggested that analysis tools handle `OK_OVERRIDE` and `ERROR_OVERRIDE`, as these are explicitly set by the end user and contain valuable information.
+ 
+ 
+### Error Mapping Schema
+As part of the specification, OpenTelemetry provides a canonical mapping of semantic conventions to status codes. This removes any ambiguity as to what OpenTelemetry ships with out of the box.
+ 
+### Error Processor
+The collector will provide a processor and a configuration language to make adjustments to this error mapping schema. This provides the flexibility and customization needed for real world scenarios.
+ 
+### Semantic conventions
```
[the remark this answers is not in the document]
tedsuo: Great, I'll remove this.

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+* `ERROR` Instrumentation has marked a span as an error. (This replaces `UNKNOWN`.)
+* `OK_OVERRIDE` The user has provided an override. The span should NOT be flagged as an error, regardless of other analysis.
+* `ERROR_OVERRIDE` The user has provided an override. The span SHOULD be flagged as an error, regardless of other analysis.
+ 
+(Note that our  current status codes include a long list of error types. We may choose to keep  them, change them, or drop them in favor of a single `ERROR` code. How many error types we have is not relevant to this proposal .)
+ 
+`OK_OVERRIDE` and `ERROR_OVERRIDE` are special status codes. These are explicit overrides provided by the end user, and should never be set by shared instrumentation. They should only be set by the application developer (via application code), or by the operator (via the collector).
+ 
+Analysis tools are free to disregard status codes, in favor of their own approach to error analysis. However it is strongly suggested that analysis tools handle `OK_OVERRIDE` and `ERROR_OVERRIDE`, as these are explicitly set by the end user and contain valuable information.
```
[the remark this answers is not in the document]
tedsuo: I was debating that. I think it depends on whether we keep the remaining status codes.

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+## Prior art and alternatives
+In OpenTracing, the lack of a Collector and error mapping schema proved to be unwieldy. It placed a burden on instrumentation plugin authors to set the flag correctly, and led to an explosion of non-standardized configuration options in every plugin just to adjust the default error flagging. This in turn placed a configuration burden on application developers.
+ 
+An alternative is the `error.hint` proposal, paired with the removal of status code. This would work, but essentially provides the same mechanism provided in this proposal, only with a large number of breaking changes. It also does not address the need for user overrides.
+ 
+## Future Work
+ 
+The inclusion of status codes and error mappings help the opentelemetry community speak the same language in terms of error reporting. It lifts the burden on future analysis tools, and (when respected) it allows users to employ multiple analysis tools without having to synchronize an important form of configuration across multiple tools.
+ 
+In the future, OpenTelemtry may add a control plane which allows dynamic configuration of the error mapping  schema.
```
reyang, line 89:
```suggestion
In the future, OpenTelemtry may add a control plane which allows dynamic configuration of the error mapping schema.
```

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+ 
+### Status Codes
+The following status codes are added to our current  schema.
+ 
+* `DEFAULT` No status has been set. Any errors must be detected by the analysis tool. (This replaces `OK`.)
+* `ERROR` Instrumentation has marked a span as an error. (This replaces `UNKNOWN`.)
+* `OK_OVERRIDE` The user has provided an override. The span should NOT be flagged as an error, regardless of other analysis.
+* `ERROR_OVERRIDE` The user has provided an override. The span SHOULD be flagged as an error, regardless of other analysis.
+ 
+(Note that our  current status codes include a long list of error types. We may choose to keep  them, change them, or drop them in favor of a single `ERROR` code. How many error types we have is not relevant to this proposal .)
```
reyang, line 23:
```suggestion
(Note that our current status codes include a long list of error types. We may choose to keep  them, change them, or drop them in favor of a single `ERROR` code. How many error types we have is not relevant to this proposal.)
```

text/trace/0000-error_flagging.md
tedsuo: I will do that. These convenience methods simply wire together multiple API calls. They should live in a helper package, and should not be directly added to existing API interfaces. Given how many semantic conventions we have, there will be a pile of them.

text/trace/0000-error_flagging.md
tedsuo: derp, should be deleted.

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+ 
+`OK_OVERRIDE` and `ERROR_OVERRIDE` are special status codes. These are explicit overrides provided by the end user, and should never be set by shared instrumentation. They should only be set by the application developer (via application code), or by the operator (via the collector).
```
reyang, line 25: How would an application developer set the status codes provided by shared instrumentation?

text/trace/0000-error_flagging.md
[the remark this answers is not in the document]
tedsuo: FYI that feature is marked as optional. We should ensure that it is actually implemented.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#status-creation

text/trace/0000-error_flagging.md
tedsuo: Possibly these overrides should be a separate field, so that we retain that data.

text/trace/0000-error_flagging.md
tedsuo: Updated.

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,86 @@
...
+However, there is confusion over the mapping of semantic conventions to status codes, and concern over the subjective nature of errors. Which network failures count as an error? Are 404s an error? The answer is dependent on the situation.
+ 
+There is one major exception. Both application developers and operators have a deep understanding of what constitutes an error in their system. OpenTelemetry must provide a way for these users to control error flagging, and explicitly indicate both when a span should and should not be counted as an error.
+A second exception is supporting analysis tools which require explicit error flagging in the data which they receive. In this case, an operator must be able to apply an error flagging schema at some point during the OTLP data processing pipeline.
+ 
+## Explanation
+The following changes add several missing features required for proper error reporting, and are completely backwards compatible with OpenTelemetry today.
+ 
+### Status Codes
+The following status codes are added to our current  schema.
```
flands, line 16:
```suggestion
The following status codes are added to our current schema.
```

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,86 @@
...
+ 
+* `DEFAULT` No status has been set. Any errors must be detected by the analysis tool. (This replaces `OK`.)
+* `ERROR` Instrumentation has marked a span as an error. (This replaces `UNKNOWN`.)
+* `OK_OVERRIDE` The user has provided an override. The span should NOT be flagged as an error, regardless of other analysis.
```
flands, line 20:
```suggestion
* `OK_OVERRIDE` The user has provided an override. The span SHOULD NOT be flagged as an error, regardless of other analysis.
```

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,86 @@
...
+There is one major exception. Both application developers and operators have a deep understanding of what constitutes an error in their system. OpenTelemetry must provide a way for these users to control error flagging, and explicitly indicate both when a span should and should not be counted as an error.
+A second exception is supporting analysis tools which require explicit error flagging in the data which they receive. In this case, an operator must be able to apply an error flagging schema at some point during the OTLP data processing pipeline.
+ 
+## Explanation
+The following changes add several missing features required for proper error reporting, and are completely backwards compatible with OpenTelemetry today.
+ 
+### Status Codes
+The following status codes are added to our current  schema.
+ 
+* `DEFAULT` No status has been set. Any errors must be detected by the analysis tool. (This replaces `OK`.)
```
flands, line 18: Naming is hard... what is the default? the default is unspecified. Perhaps UNSPECIFIED could be used here?

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,86 @@
...
+* `ERROR` Instrumentation has marked a span as an error. (This replaces `UNKNOWN`.)
+* `OK_OVERRIDE` The user has provided an override. The span should NOT be flagged as an error, regardless of other analysis.
+* `ERROR_OVERRIDE` The user has provided an override. The span SHOULD be flagged as an error, regardless of other analysis.
+ 
+(Note that our current status codes include a long list of error types. We may choose to keep  them, change them, or drop them in favor of a single `ERROR` code. How many error types we have is not relevant to this proposal.)
```
flands, line 23:
```suggestion
(Note that our current status codes include a long list of error types. We may choose to keep them, change them, or drop them in favor of a single `ERROR` code. How many error types we have is not relevant to this proposal.)
```

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,86 @@
...
+ 
+`OK_OVERRIDE` and `ERROR_OVERRIDE` are special status codes. These are explicit overrides provided by the end user, and should never be set by shared instrumentation. They should only be set by the application developer (via application code), or by the operator (via the collector).
```
flands, line 25:
```suggestion
`OK_OVERRIDE` and `ERROR_OVERRIDE` are special status codes. These are explicit overrides provided by the end-user, and SHOULD NOT be set by shared instrumentation. They SHOULD be set by the application developer (via application code), or by the operator (via the collector).
```

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,86 @@
...
+ 
+Analysis tools are free to disregard status codes, in favor of their own approach to error analysis. However it is strongly suggested that analysis tools handle `OK_OVERRIDE` and `ERROR_OVERRIDE`, as these are explicitly set by the end user and contain valuable information.
```
flands, line 27:
```suggestion
Analysis tools MAY disregard status codes, in favor of their own approach to error analysis. However, it is strongly suggested that analysis tools SHOULD handle `OK_OVERRIDE` and `ERROR_OVERRIDE`, as these are explicitly set by the end-user and contain valuable information.
```

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,86 @@
...
+ 
+ 
+## BUT ERRORS ARE SUBJECTIVE!! HOW CAN WE KNOW WHAT IS AN ERROR? WHO ARE WE TO DEFINE THIS?
+First of all, every tracing system to-date comes with a default set of errors. No system requires that end users start completely from scratch. So... be calm!! Have faith!!
+ 
+While flagging errors can be a subjective decision, it is true that many semantic conventions qualify as an error. By providing a default mapping of semantic conventions to errors, we ensure compatibility with existing analysis tools (e.g. Jaeger), and provide guidance to users and future implementers.
+ 
+Obviously, all systems are different, and users will want to adjust error reporting on a case by case basis. Unwanted errors may be suppressed, and additional errors may be added. The collector will provide a processor and a configuration language to make this a straightforward process. Working from a baseline of standard errors will provide a better experience than having to define a schema from scratch.
+ 
+Note that analysis tools are free to disregard Span Status, and do their own error analysis. For these systems, the only Status codes of import are `OK_OVERRIDE` and `ERROR_OVERRIDE`.
```
flands, line 61:
```suggestion
Note that analysis tools MAY disregard Span Status, and do their own error analysis. For these systems, the only Status codes of import are `OK_OVERRIDE` and `ERROR_OVERRIDE`.
```

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,86 @@
...
+A second exception is supporting analysis tools which require explicit error flagging in the data which they receive. In this case, an operator must be able to apply an error flagging schema at some point during the OTLP data processing pipeline.
+ 
+## Explanation
+The following changes add several missing features required for proper error reporting, and are completely backwards compatible with OpenTelemetry today.
+ 
+### Status Codes
+The following status codes are added to our current  schema.
+ 
+* `DEFAULT` No status has been set. Any errors must be detected by the analysis tool. (This replaces `OK`.)
+* `ERROR` Instrumentation has marked a span as an error. (This replaces `UNKNOWN`.)
```
flands, line 19: How does the instrumentation make this determination? It appears this is left to discretion but clearly needs to be consistent across client libraries. For example, should `http.status_code` in the range 500-599 automatically be marked as an error?

Oberon00: Note that ERROR != ERROR_OVERRIDE. I think this leaves the status quo of having  http.status_code in the range 500-599 being mapped to various status codes (none of which is UNKNOWN/ERROR): https://github.com/open-telemetry/opentelemetry-specification/blob/821118e5/specification/trace/semantic_conventions/http.md#status

text/trace/0000-error_flagging.md
Oberon00: I would prefer NOT_SET. Then hopefully no one's gonna ask "can you set it explicitly to NOT_SET?". I would also prefer to explicitly state that this status code SHOULD NOT be set (or if possible, make sure at compile time that it cannot be set, e.g. by only having this enumeration value in the spec and protocol, but not in actual APIs).

carlosalberto:
Overall looks good to me.

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,86 @@
...
+* `OK_OVERRIDE` The user has provided an override. The span should NOT be flagged as an error, regardless of other analysis.
+* `ERROR_OVERRIDE` The user has provided an override. The span SHOULD be flagged as an error, regardless of other analysis.
```
tigrannajaryan, line 21: If I understand correctly one of the differences between `ERROR` and `ERROR_OVERRIDE` is that `ERROR` is set by instrumentation and may be ignored by analysis tools, while `ERROR_OVERRIDE` is set by application and should be honored.

This seems to hint that we believe that the opinion of instrumentation developers is less reliable and may be ignored but the opinion of application developers is more important and should be honored.

Is this a correct reading of the proposal?

If so then the rationale is not completely clear. Why do we think (if we think so) the opinion of instrumentation developers is less reliable or less important than opinion of application developers?

I can see that we aim to resolve the conflict of 3 opinions here: 1) the instrumentation developer, 2) the application developer and 3) the analysis tool (and the last one may be a composite opinion influenced by the analysis tool developer and the analysis tool user if the tool is configurable).

I feel that the proposal would be stronger if it provided the rationale for choices made on how the conflict is resolved.

text/trace/0000-error_flagging.md
tigrannajaryan, line 23: I think we should drop the current long list. At the moment I do not see the value of keeping it. The list we have is an attempt to standardize all possible errors into a few buckets of errors that can ever happen but I am not convinced it is a valuable standardization.

I can see some possible ways to use the values from the long list in the backend: to filter or to aggregate. However in my opinion the values are not particularly useful such that filtering or aggregation can give me interesting insights that I can't already get by doing an equivalent filtering or aggregation on the original attribute that resulted in the status code (e.g. the http.status_code). The original attribute likely tells me a better and more precise story of what happened. (And if such original attribute is missing the question is why didn't the developer include it and how am I supposed to understand what the generic error codes mean in the particular situation).

I would like to hear some opposing opinions and anecdotes of people who did use the OpenCensus status code for successful troubleshooting using an analysis tool. I can be convinced in that case.

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,86 @@
...
+The collector will provide a processor and a configuration language to make adjustments to this error mapping schema. This provides the flexibility and customization needed for real world scenarios.
+ 
+### Convenience methods
+As a convenience, OpenTelemetry provides helper functions for adding semantic conventions and exceptions to a span. These helper functions will also set the correct status code. This simplifies the life of the instrumentation author, and helps ensure compliance and data quality.
+
+Note that these convenience methods simply wire together multiple API calls. They should live in a helper package, and should not be directly added to existing API interfaces. Given how many semantic conventions we have, there will be a pile of them.
+ 
+ 
+## Internal details
+Except for the renaming of two status codes, this proposal is backwards compatible with existing code, protocols, and the OpenTracing bridge.
```
tigrannajaryan, line 43: >this proposal is backwards compatible with existing code, protocols

For this to be true we need to keep all existing status codes and assign OK_OVERRIDE and ERROR_OVERRIDE numeric values that were previously unused (by numeric values I mean the values we use record them on the wire).

If we decide to remove existing (old) status codes as it is suggested as an option in the proposal (and which I think we should do) then it is not completely clear how we ensure the backward compatibility. Do we mark the previously used status codes and their corresponding numeric values on the wire as obsolete and disallow their usage in the generation side? Should the Collector continue to treat the obsolete codes seen on its input as errors?

A possible approach here which I think is reasonable is that we don't try to keep backward compatibility. We just break it and remove the old status codes from the API and don't care about the numeric values on the wire. I am not aware of existing production uses that may be affected by this and even if they are then perhaps it is one the things we break before the GA.

tigrannajaryan (commented):
Ted, thank you for the proposal. I think this nicely summarizes the Error WG discussion. I have a few questions/comments that would be good to clarify.

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,86 @@
...
+ 
+## Motivation
+Error reporting is a fundamental use case for distributed tracing. While we prefer that error flagging occurs within analysis tools, and not within instrumentation plugins, a number of currently supported analysis tools and protocols rely on the existence of an explicit error flag reported from instrumentation. In OpenTelmetry, the error flag is called "status codes."
+ 
+However, there is confusion over the mapping of semantic conventions to status codes, and concern over the subjective nature of errors. Which network failures count as an error? Are 404s an error? The answer is dependent on the situation.
+ 
+There is one major exception. Both application developers and operators have a deep understanding of what constitutes an error in their system. OpenTelemetry must provide a way for these users to control error flagging, and explicitly indicate both when a span should and should not be counted as an error.
+A second exception is supporting analysis tools which require explicit error flagging in the data which they receive. In this case, an operator must be able to apply an error flagging schema at some point during the OTLP data processing pipeline.
```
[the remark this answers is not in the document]
Oberon00: Maybe the reference to OTLP is misplaced but I understood this as "SDK exports as OTLP to collector, collector applies the error flagging schema, exports via Jaeger to backend or via OTLP to another collector."

text/trace/0000-error_flagging.md
justinfoote: I agree with NOT_SET. But I would expect someone to ask for an `unsetStatusCode` sort of operation.  I'm trying to think of what use case there would be to unset the status (but not to use an override status).

text/trace/0000-error_flagging.md
justinfoote: I initially agreed with the idea of splitting these into two fields. Then it occured to me that we might not have matching mental models.  There are two ways we could split these concepts into two fields:

```
  status: ERROR
  user_override: true
```
or
```
  status: ERROR
  user_status: OK
```

The first replaces the status (which is useful, so there's only one field to check in the back-end). The second introduces a new status field, which supercedes the other status field, but it also keeps both statuses, so a back-end could potentially see what the _instrumentation_ thought the status should be.

If one of our guiding principals is that user-provided status is unilaterally more important than instrumentation-provided status, I think the first option makes the most sense.

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,86 @@
...
+ 
+(Note that our current status codes include a long list of error types. We may choose to keep  them, change them, or drop them in favor of a single `ERROR` code. How many error types we have is not relevant to this proposal.)
+ 
+`OK_OVERRIDE` and `ERROR_OVERRIDE` are special status codes. These are explicit overrides provided by the end user, and should never be set by shared instrumentation. They should only be set by the application developer (via application code), or by the operator (via the collector).
+ 
+Analysis tools are free to disregard status codes, in favor of their own approach to error analysis. However it is strongly suggested that analysis tools handle `OK_OVERRIDE` and `ERROR_OVERRIDE`, as these are explicitly set by the end user and contain valuable information.
+ 
+ 
+### Error Mapping Schema
+As part of the specification, OpenTelemetry provides a canonical mapping of semantic conventions to status codes. This removes any ambiguity as to what OpenTelemetry ships with out of the box.
```
justinfoote, line 31: 👍 
This implies a new _Status Mapping_ section to all of the existing trace semantic conventions, yeah?

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,86 @@
...
+Note that these convenience methods simply wire together multiple API calls. They should live in a helper package, and should not be directly added to existing API interfaces. Given how many semantic conventions we have, there will be a pile of them.
+ 
+ 
+## Internal details
+Except for the renaming of two status codes, this proposal is backwards compatible with existing code, protocols, and the OpenTracing bridge.
+ 
+**OK is renamed to DEFAULT**
+ Using the term “OK” as the default implies more meaning than we intend. The span is not necessarily OK - it simply has not triggered our standard error mapping. The default status code should be renamed to `DEFAULT` to avoid confusion.
+ 
+ Note: I intentionally avoided terms like "unset" as it may imply to users that they are required to set the status code, which is not the intention.
```
justinfoote, line 48: If we don't want to use the term `UNSET`, because the _set_ verb implies that the user should be doing some setting, perhaps something like `NONE` or `NULL`.

justinfoote (commented):
In my opinion, this is a reasonable compromise and logical step forward.  Thanks for writing it up!

text/trace/0000-error_flagging.md
Oberon00: NO_OVERRIDE would make perfect sense to me

text/trace/0000-error_flagging.md
tedsuo: Yeah, I'm fine with the first one as well. This is a good improvement to this proposal, I'll make the change.

text/trace/0000-error_flagging.md
tedsuo: Heh, well given that it looks like we may move to a `user_override` flag instead of new statuses, the current default of `OK` starts to make sense again, as you can now tell the difference between the default and an explicit override from the user.

text/trace/0000-error_flagging.md
tedsuo: @flands the "Error Mapping" section tries to address this. We should provide a schema that explicitly describes how every semantic convention affects the status code. Basically, take the HTTP mapping example above, and expand it into a new page in the spec which contains all of the mappings.

text/trace/0000-error_flagging.md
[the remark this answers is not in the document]
tedsuo: Yeah I meant what @Oberon00 said. I mentioned  OTLP because it will be the only format which has these override features, but I can change it if it is confusing. I'm also realizing that we don't define our "data processing pipeline" anywhere, so I see how that term could be confusing, so I'll add an explanation.

text/trace/0000-error_flagging.md
tedsuo: Sorry, the intent is definitely not to disparage the instrumentation writers! The reason we need to differentiate between instrumentation and the end user is because most instrumentation is generic. Most instrumentation will only supply a standard mapping of semantics to status codes. On the other hand, statuses set by application developers and operators is much more service and deployment specific. Even if systems have their own schema, and ignore some or all statuses set by instrumentation, they will will still be interested in this information, as it is novel and very relevant. 

Does that sound reasonable?

text/trace/0000-error_flagging.md
tedsuo: Totally fine! I felt like making a decision on these additional status codes expands the scope of this OTEP unnecessarily. I haven't thought deeply about what the use cases for these codes might be, other than allowing for more nuance in error counting, without the overhead of scanning span attributes and events.

text/trace/0000-error_flagging.md
tedsuo: Since we're now moving this to a separate `user_override` flag, we will no longer need to change the status codes as part of this OTEP. I do agree that it will be tricky to remove some of the status codes. I'm not sure how bad the breakage would be.

text/trace/0000-error_flagging.md
tedsuo: Yes. But that is a good point, it should be called Status Mapping and not error mapping. Here is an example of where we are doing this already: https://github.com/open-telemetry/opentelemetry-specification/blob/821118e5/specification/trace/semantic_conventions/http.md#status

Note that if we end up removing all the error types from span status, this concept would be unnecessary.

text/trace/0000-error_flagging.md
mwlang: The way I see it, instrumentation developers make their best effort attempt at accurately reporting facts of what happened during instrumented calls.  The instrumentation developer must be far more strict about being subjective about what constitutes an error whereas the application developer has more freedom to be subjective.  An application developer, who is using an instrumentation has more context to make explicit decisions that may go against the generic (more context-less) indications an instrumentation is setting for error conditions.  In other words, what @tedsuo replied with is 100% reasonable and spot on to me.

tigrannajaryan: @tedsuo sounds good. If possible please add a couple words to the OTEP to make it clear.

tedsuo: Thanks. I updated several sections.

> There is one other missing piece, required for proper error flagging. Both application developers and operators have a deep understanding of what constitutes an error in their system. OpenTelemetry must provide a way for these users to control error flagging, and explicitly indicate that it is the end user setting the status code, and not instrumentation plugins. In these specific cases, the error flagging is known to be correct: the end user has decided the status of the span, and they do not want another interpretation. 

> Note that analysis tools MAY disregard Span Status, and do their own error analysis. There is no requirement that the status code is respected, even when `user_override` is set. However, it is strongly suggested that analysis tools SHOULD pay attention to the status code when `user_override` is set, as it represents a subjective decision made by either the operator or application developer.

tedsuo: Please let me know if it is clear enough now.

text/trace/0000-error_flagging.md
tedsuo: All gone now!

tedsuo:
FYI I have updated this proposal based on feedback. Rather than adding additional status codes to represent user overrides, we add a `user_override` boolean which indicated that the end user has confirmed that the status code is correct. The expected behavior is that the status code on the span should take precedence over any further analysis.

Please let me know if this solves our remaining issues with error reporting.

tedsuo:
Also, we will be discussing this proposal at the Errors Working Group tomorrow (thursday) at 8am Pacific. Please join if you can make it.

Details: https://docs.google.com/document/d/1HgUI69rBridFzCXxXuTjQrkG6jb-YcFQnZjPcyBcK1U/

iNikem:
I beg your pardon. Do you I read the current proposal correctly, that all it does is adding new `error_override` flag to Span? I don't see how this solves any problem we tried to address. 

It does not address open-telemetry/opentelemetry-specification#706. It does not address motivational part of #134 . It does not address part of its own motivation:
>However, there is confusion over the mapping of semantic conventions to status codes, and concern over the subjective nature of errors. Which network failures count as an error? Are 404s an error? The answer is often dependent on the situation, but without even a baseline of suggested status codes for each convention, the instrumentation author is placed under the heavy burden of making the decision. Worse, the decisions will not be in sync across different instrumentation packages.

I think the initial suggestion of this OTEP of reducing/renaming the enum of `Span.Status` made sense. Current form does not.

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,70 @@
...
+ 
+There is one other missing piece, required for proper error flagging. Both application developers and operators have a deep understanding of what constitutes an error in their system. OpenTelemetry must provide a way for these users to control error flagging, and explicitly indicate that it is the end user setting the status code, and not instrumentation plugins. In these specific cases, the error flagging is known to be correct: the end user has decided the status of the span, and they do not want another interpretation. 
+
+While generic instrumentation can only provide a generic schema, end users are capable of making subjective decisions about their systems. And, as the end user, they should get to have the final call in what consitutes an error. In order to accomplish this, there must be a way to differntiate between errors flagged by instrumentation, and errors flagged by the end user.
+ 
+## Explanation
+The following changes add several missing features required for proper error reporting, and are completely backwards compatible with OpenTelemetry today.
+ 
+### `span.user_override(boolean)`
```
arminru, line 16: Would this me a function/method on the span? The name `user_override` itself does not convey that this has to do anything with the span's status and is therefore very confusing IMHO. It sounds rather like "I'm taking control over this span which was created by auto instrumentation" or something like that.

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,70 @@
...
+The `user_override` method indicates that the end user has confirmed that the status code is correct. When using OTLP, this will set the `user_override` field. When setting status codes via the collector or application code, `user_override` can be set to ensure that the span status is not re-interpreted by further analysis.
```
arminru, line 17: This makes it sound like if there was already a field called `user_override` in OTLP that has some semantics and is already implemented but this is supposed to introduce a new field, right?

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,79 @@
...
+set, and an explicit OK status set by an end user.
+
+* `UNSET` is the default status code.
+* `ERROR` represents all error types.
+* `OK` represents a span which has been explicitly marked as being free of errors, and should not be counted against an error budget. Note that only end users should set this status. Instead, instrumentation should leave the status as `UNSET` for nominal operations.
+
+### `Status Source`
+The Status Source field identifies the origin of the status code on the span. This is important, as statuses set by application developers and operators have been confirmed by the end user to be correct to the particular situation. Statuses set by instrumentation, on the other hand, are only following a generic schema. These statuses When using OTLP, this will set the `user_override` field. When setting status codes via the collector or application code, `user_override` can be set to ensure that the span status is not re-interpreted by further analysis.
```
iNikem, line 25: This line seems to be corrupted at the end.

iNikem (approved):

tedsuo:
Ok, third draft is now available. This draft splits the difference between the first and second draft.

Changes:
* Rename the default status from OK to UNSET 
* Reduce Status codes to the following: 
   * UNSET (DEFAULT) 
   * OK 
   * ERROR 
* user_override boolean is expanded into an enum named Status Source, which adds clarity.
  * Status Source values: 
    * Instrumentation  (default)
    * Application 
    * Operator 
* Clarify that the Status Mapping  is required, but not a blocker for GA, and will grow over time.


Thank you all who have been contributing feedback! I am hopeful that this is the final iteration.

text/trace/0000-error_flagging.md
tedsuo: This is updated to "Status Source" for clarity.

text/trace/0000-error_flagging.md
tedsuo: Clarified, thank you.

text/trace/0000-error_flagging.md
tedsuo: fixed, thank you

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,79 @@
+# Error Flagging with Status Codes
+This proposal adds two status codes explicitly for use as overrides by the end user, and proposes a canonical mapping of semantic conventions to status codes. This clarifies how error reporting should work in OpenTelemetry.
+ 
+## Motivation
+Error reporting is a fundamental use case for distributed tracing. While we prefer that error flagging occurs within analysis tools, and not within instrumentation plugins, a number of currently supported analysis tools and protocols rely on the existence of an explicit error flag reported from instrumentation. In OpenTelemetry, the error flag is called "status codes."
```
reyang, line 5:
```suggestion
Error reporting is a fundamental use case for distributed tracing. While we prefer that error flagging occurs within analysis tools, and not within instrumentation plugins, a number of currently supported analysis tools and protocols rely on the existence of an explicit error flag reported from instrumentation. In OpenTelemetry, the error flag is called "status codes".
```

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,79 @@
...
+## Explanation
+The following changes add several missing features required for proper error reporting, and are completely backwards compatible with OpenTelemetry today.
+
+### Status Codes
+Currently, OpenTelemetry does not have a use case for differentiating between different types of errors. However, this use case may appear in the future. For now, we would like to reduce the number of status codes, and then add them back in as the need becomes clear. We would also like to differentiate between status codes which have not been
+set, and an explicit OK status set by an end user.
+
+* `UNSET` is the default status code.
+* `ERROR` represents all error types.
+* `OK` represents a span which has been explicitly marked as being free of errors, and should not be counted against an error budget. Note that only end users should set this status. Instead, instrumentation should leave the status as `UNSET` for nominal operations.
```
reyang, line 22: It is unclear what "nominal operations" mean.

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,79 @@
...
+
+Please note that semantic conventions, and thus status mapping from conventions, are still a work in progress and will continue to change after GA.
+ 
+### Status Processor
+The collector will provide a processor and a configuration language to make adjustments to this status mapping schema. This provides the flexibility and customization needed for real world scenarios.
+ 
+### Convenience methods
+As a convenience, OpenTelemetry provides helper functions for adding semantic conventions and exceptions to a span. These helper functions will also set the correct status code. This simplifies the life of the instrumentation author, and helps ensure compliance and data quality.
+
+Note that these convenience methods simply wire together multiple API calls. They should live in a helper package, and should not be directly added to existing API interfaces. Given how many semantic conventions we have, there will be a pile of them.
```
reyang, line 44: Are these helper functions part of API or SDK or neither? What's the compatibility guarantee?
Is this something that each language SDK will decide, or the spec will cover it (for consistency)?

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,79 @@
...
+
+### `Status Source`
+The Status Source field identifies the origin of the status code on the span. This is important, as statuses set by application developers and operators have been confirmed by the end user to be correct to the particular situation. Statuses set by instrumentation, on the other hand, are only following a generic schema. These statuses When using OTLP, this will set the `user_override` field. When setting status codes via the collector or application code, `user_override` can be set to ensure that the span status is not re-interpreted by further analysis.
+
+* `INSTRUMENTATION` is the default source. This is used for instrumentation contained within shared code, such as OSS libraries and frameworks. All instrumentation plugins shipped with OpenTelemetry use this status code.
+* `APPLICATION` identifies statuses set by the application developer, within their application code. 
+* `OPERATOR` identifies statuses which have been altered during data egress.
```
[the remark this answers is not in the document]
tedsuo: If we don't need a distinction, I can simply to `INSTRUMENTATION` and `USER`. That is probably better for now.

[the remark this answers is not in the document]
tigrannajaryan: INSTRUMENTATION and USER also sounds good to me. In the description of USER we can tell that it is either the application developer (who is the user of Otel library) or the Collector operator (who is the user of Collector). Sounds reasonable to me.

text/trace/0000-error_flagging.md
tedsuo: These helper functions should be higher order functions which make a series of API calls on your behalf. They could go in their own package, but either way they should not be added directly to the Span API.

text/trace/0000-error_flagging.md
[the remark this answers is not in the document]
tedsuo: consolidated.

text/trace/0000-error_flagging.md
tedsuo: How about "normal"?

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,78 @@
...
+## Explanation
+The following changes add several missing features required for proper error reporting, and are completely backwards compatible with OpenTelemetry today.
+
+### Status Codes
+Currently, OpenTelemetry does not have a use case for differentiating between different types of errors. However, this use case may appear in the future. For now, we would like to reduce the number of status codes, and then add them back in as the need becomes clear. We would also like to differentiate between status codes which have not been
+set, and an explicit OK status set by an end user.
+
+* `UNSET` is the default status code.
+* `ERROR` represents all error types.
+* `OK` represents a span which has been explicitly marked as being free of errors, and should not be counted against an error budget. Note that only end users should set this status. Instead, instrumentation should leave the status as `UNSET` for normal operations.
```
yurishkuro, line 22: Clarification - do you mean that ERROR can be set by instrumentation, but OK can only be set by user?

tedsuo: Yes correct. Instrumentation should not set OK on every span which does not have an error, it should leave the status code unset. OK is for end users to clearly communicate that they would like to suppress any errors which would otherwise be triggered by the span.

text/trace/0000-error_flagging.md
iNikem: This still begets interpretation. Maybe remove this completely: 
> Instead, instrumentation SHOULD leave the status as `UNSET`.

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,78 @@
...
+
+### `Status Source`
+A new Status Source field identifies the origin of the status code on the span. This is important, as statuses set by application developers and operators have been confirmed by the end user to be correct to the particular situation. Statuses set by instrumentation, on the other hand, are only following a generic schema.
```
Oberon00, line 25: Rather than talking about fields, we should talk about APIs. It seems to me that on an API level, you would sometimes get both an instrumentation status and an user status. It would semantically make sense for an implementation to record both.

text/trace/0000-error_flagging.md
Oberon00: Should we also move RecordException out of the core span interface API? I guess this is a bit of a special case because vendors could have large differences in what data they record here.

text/trace/0000-error_flagging.md
```diff
@@ -0,0 +1,78 @@
...
+The collector will provide a processor and a configuration language to make adjustments to this status mapping schema. This provides the flexibility and customization needed for real world scenarios.
+ 
+### Convenience methods
+As a convenience, OpenTelemetry provides helper functions for adding semantic conventions and exceptions to a span. These helper functions will also set the correct status code. This simplifies the life of the instrumentation author, and helps ensure compliance and data quality.
+
+Note that these convenience methods simply wire together multiple API calls. They should live in a helper package, and should not be directly added to existing API interfaces. Given how many semantic conventions we have, there will be a pile of them.
+ 
+ 
+## Internal details
+This proposal is mostly backwards compatible with existing code, protocols, and the OpenTracing bridge. The only potential exception is the removal of status codes enums from the current OTLP protocol, and the rewriting of the small number of instrumentation plugins that were making use of them.
```
Oberon00, line 47: So by default the status will be Instrumentation?
Because of thread safery alone, we must record source and status in the same API call.

Oberon00 (approved):
LGTM. Details can be resolved when writing the spec.

text/trace/0000-error_flagging.md
carlosalberto: I think we should keep that as it simply creates an `Event` (and should not affect directly the final `Status` value).

Oberon00: But doensn't this OTEP want to change that?

> As a convenience, OpenTelemetry provides helper functions for adding semantic conventions and exceptions to a span. These helper functions will also set the correct status code.

Oberon00: Although I concur that exceptions should not affect the status by default (unless they are escaped #784)

carlosalberto: I think `RecordException()` is low-level enough to keep it IMHO (unless @tedsuo thinks otherwise).

tigrannajaryan (approved):
The proposal is clear. There are still details that need to be fleshed out, i.e. how exactly this is going to look like in the API.

I agree with the spirit of this proposal and for the sake of making progress I am approving this, with the understanding that there is a additional work needed when applying this OTEP to the spec and proto.

I believe it is important at this stage to gain a reasonable consensus on the suggested approach first. If we manage to do that we can then clarify how exactly this proposal should be applied to the spec and what the corresponding API and protocol field changes will look like.

I would also like @iNikem and others who work on auto-instrumentation to vet this proposal from their perspective.

tigrannajaryan:
@tedsuo please fix markdownlint errors.

text/trace/0000-error_flagging.md
tedsuo: Yes, I agree Instrumentation should be the default.

text/trace/0000-error_flagging.md
tedsuo: Either way; I'm fine with keeping RecordException as low level, for the reasons @Oberon00 mentions. I imagine that instrumentation authors will have way more use of these functions than application developers, so keeping RecordException as part of the Span API would be helpful as it is very important.

Rather than add yet-another-package, these functions could be added to the same package that contains the constants for semantic conventions. That may be the best place for them.

carlosalberto (approved):

text/trace/0136-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Remind me why we need status codes again?
+
+Status codes provide a low overhead mechanism for checking if a span counts against an error budget, without having to scan every attribute and event. It is an inexpensive and low cardinality approach to track multiple types of error budgets. This reduces overhead and could be a benefit for many systems.
+
+However, adding in an existing set of error types without first clearly defining their use and how they might be set has caused confusion. If the status codes are not set consistently and correctly, then the resulting error budgeting will not be useful. So we are consolidating all error types into a single ERROR type, to avoid this situation. We may add more error types back in if we can agree on their use cases and a method for applying them consistently.
```
jmacd, line 71: > We may add more **error types** back in if

(emphasis mine) be careful to refer to these as **status codes** to signal that they are (theorized to be) useful for reporting status types that are not errors, for example "canceled".

jmacd (approved):

reyang (approved):
LGTM.

text/trace/0136-error_flagging.md
tedsuo: Thanks, I keep catching myself doing that. :)

text/trace/0136-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+
+The following changes add several missing features required for proper error reporting, and are completely backwards compatible with OpenTelemetry today.
+
+### Status Codes
+
+Currently, OpenTelemetry does not have a use case for differentiating between different types of errors. However, this use case may appear in the future. For now, we would like to reduce the number of status codes, and then add them back in as the need becomes clear. We would also like to differentiate between status codes which have not been
+set, and an explicit OK status set by an end user.
+
+* `UNSET` is the default status code.
+* `ERROR` represents all error types.
```
bogdandrutu, line 25: Can this be named "UnknownError":
1. Consistent with status codes;
2. In the future if other errors are added this means not other more specific error was identified;

text/trace/0136-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+* `OK` represents a span which has been explicitly marked as being free of errors, and should not be counted against an error budget. Note that only end users should set this status. Instead, instrumentation should leave the status as `UNSET` for normal operations.
+
+### `Status Source`
```
bogdandrutu, line 28: Can this be added later, not sure we thought about this enough and feels like a bit rushing. Is there any downside if we do that?

cijothomas (approved):

text/trace/0136-error_flagging.md
Oberon00: I have mixed feelings about splitting the boolean from the main status code enum (see earlier versions of this PR), but I think this OTEP does not make sense without having the user-supplied-info/instrumentation-derived-heuristic split.

text/trace/0136-error_flagging.md
Oberon00: In the future if other errors are added, I would rather add another enum ErrorCategory or something like that (maybe as semantic convention). That would also be backwards compatible with code that would then check `if (status == ERROR) { assume error } else { assume no error }`.

tedsuo: I think it should map to `UNKOWN`, but the name `UNKOWN` doesn't make sense when we only have one kind of error. When we add more codes, `ERROR` will then mean "generic error, no details" which is more or less how we treat `UNKOWN` today.

text/trace/0136-error_flagging.md
tedsuo: For backends which want to ignore errors from instrumentation and instead use their own heuristic, we need a way to inform them that the status code is an override which they should respect. We could have `ERROR_INSTRUMENTATION `and `ERROR_USER`, but that would lead to a duplication of all the different error codes as we add them back in. People didn't like that part - this felt clearer. Having them all be status codes would be atomic, which is nice.

tedsuo: (But, to be clear, I think this is a fine compromise.)

text/trace/0136-error_flagging.md
```diff
@@ -0,0 +1,89 @@
+# Error Flagging with Status Codes
+
+This proposal adds two status codes explicitly for use as overrides by the end user, and proposes a canonical mapping of semantic conventions to status codes. This clarifies how error reporting should work in OpenTelemetry.
```
arminru, line 3: It's actually three status codes if "`UNSET` is the default status code" is included. It's also not only intended for overriding by the end user but also to be set by instrumentations.

arminru, line 3: What is meant by the _canonical mapping of semantic conventions to status codes_?

text/trace/0136-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Motivation
+
+Error reporting is a fundamental use case for distributed tracing. While we prefer that error flagging occurs within analysis tools, and not within instrumentation plugins, a number of currently supported analysis tools and protocols rely on the existence of an explicit error flag reported from instrumentation. In OpenTelemetry, the error flag is called "status codes".
```
arminru, line 7:
```suggestion
Error reporting is a fundamental use case for distributed tracing. While we prefer that error flagging occurs within analysis tools, and not within instrumentations, a number of currently supported analysis tools and protocols rely on the existence of an explicit error flag reported from instrumentation. In OpenTelemetry, the error flag is called "status codes".
```

This would introduce a new term, usually it's simply called instrumentation in the spec.

text/trace/0136-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+
+However, there is confusion over the mapping of semantic conventions to status codes, and concern over the subjective nature of errors. Which network failures count as an error? Are 404s an error? The answer is often dependent on the situation, but without even a baseline of suggested status codes for each convention, the instrumentation author is placed under the heavy burden of making the decision. Worse, the decisions will not be in sync across different instrumentation packages.
```
arminru, line 9:
```suggestion
However, there is confusion over the mapping of semantic conventions to status codes, and concern over the subjective nature of errors. Which network failures count as an error? Are 404s an error? The answer is often dependent on the situation, but without even a baseline of suggested status codes for each convention, the instrumentation author is placed under the heavy burden of making the decision. Worse, the decisions will not be in sync across different instrumentations.
```

Yet another new term ;-)

text/trace/0136-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+
+There is one other missing piece, required for proper error flagging. Both application developers and operators have a deep understanding of what constitutes an error in their system. OpenTelemetry must provide a way for these users to control error flagging, and explicitly indicate that it is the end user setting the status code, and not instrumentation plugins. In these specific cases, the error flagging is known to be correct: the end user has decided the status of the span, and they do not want another interpretation.
```
arminru, line 11:
```suggestion
There is one other missing piece, required for proper error flagging. Both application developers and operators have a deep understanding of what constitutes an error in their system. OpenTelemetry must provide a way for these users to control error flagging, and explicitly indicate that it is the end user setting the status code, and not an instrumentation. In these specific cases, the error flagging is known to be correct: the end user has decided the status of the span, and they do not want another interpretation.
```

arminru, line 11: End user would be the person _using_ the application (e.g., the person shopping on the instrumented web shop) rather than the developer instrumenting the application, right? Maybe we can call it application or instrumentation developer? I think end user only makes sense from a back end vendor's POV (like we both are) but not for a general audience.

text/trace/0136-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+
+While generic instrumentation can only provide a generic schema, end users are capable of making subjective decisions about their systems. And, as the end user, they should get to have the final call in what constitutes an error. In order to accomplish this, there must be a way to differentiate between errors flagged by instrumentation, and errors flagged by the end user.
```
arminru, line 13: Same as above: I don't think end user is the right term to use here.

text/trace/0136-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+The following changes add several missing features required for proper error reporting, and are completely backwards compatible with OpenTelemetry today.
+
+### Status Codes
+
+Currently, OpenTelemetry does not have a use case for differentiating between different types of errors. However, this use case may appear in the future. For now, we would like to reduce the number of status codes, and then add them back in as the need becomes clear. We would also like to differentiate between status codes which have not been
+set, and an explicit OK status set by an end user.
+
+* `UNSET` is the default status code.
+* `ERROR` represents all error types.
+* `OK` represents a span which has been explicitly marked as being free of errors, and should not be counted against an error budget. Note that only end users should set this status. Instead, instrumentation should leave the status as `UNSET` for normal operations.
```
arminru, line 26:
```suggestion
* `OK` represents a span which has been explicitly marked as being free of errors, and should not be counted against an error budget. Note that only end users should set this status. Instead, instrumentations should leave the status as `UNSET` for normal operations.
```

arminru, line 26: What are _normal operations_?

text/trace/0136-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+
+### `Status Source`
+
+A new Status Source field identifies the origin of the status code on the span. This is important, as statuses set by application developers and operators have been confirmed by the end user to be correct to the particular situation. Statuses set by instrumentation, on the other hand, are only following a generic schema.
```
arminru, line 30: Is the _end user_ mentioned here different from _application developers and operators_?

text/trace/0136-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+
+* `INSTRUMENTATION` is the default source. This is used for instrumentation contained within shared code, such as OSS libraries and frameworks. All instrumentation plugins shipped with OpenTelemetry use this status code.
+* `USER` identifies statuses set by the end user, either in application code or the collector.
```
arminru, line 33:
```suggestion
* `USER` identifies statuses set by application developers or operators, either in application code or the collector.
```

text/trace/0136-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+
+Analysis tools MAY disregard status codes, in favor of their own approach to error analysis. However, it is strongly suggested that analysis tools SHOULD pay attention to the status codes when set by `USER`, as it is a communication from the end-user and contains valuable information.
```
arminru, line 35:
```suggestion
Analysis tools MAY disregard status codes, in favor of their own approach to error analysis. However, it is strongly suggested that analysis tools SHOULD pay attention to the status codes when set by `USER`, as it is a communication from the application developer or operator and contains valuable information.
```
The data an instrumentation provides is usually also valuable 😉

text/trace/0136-error_flagging.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Status Mapping Schema
+
+As part of the specification, OpenTelemetry provides a canonical mapping of semantic conventions to status codes. This removes any ambiguity as to what OpenTelemetry ships with out of the box.
```
arminru, line 39: I don't really understand this sentence.
Does this mean the semantic conventions will be expanded to provide a mapping of a library's/framework's return status to whether it should be `ERROR` or `OK` or something else? Can we provide a reasonable mapping anywhere other than for HTTP? Our semantic conventions target generic _concepts_ like DB calls, RPC or messaging but usually not concrete libraries (e.g. MongoDB return code X). Will we add something like that?

arminru (approved):
Mostly editorial suggestions and some statements that need to be clarified but nothing to object. I'm satisfied with the general direction and agree with @tigrannajaryan that, given the initial consensus we have here, we'll be able to work out the details in the spec (and proto) PRs.

text/trace/0136-error_flagging.md
tedsuo: operations which do not generate an error.

text/trace/0136-error_flagging.md
tedsuo: no, it is the same.

text/trace/0136-error_flagging.md
justinfoote: Sorry to nitpick your nitpick, but this shouldn't be pluralized here.
```suggestion
Error reporting is a fundamental use case for distributed tracing. While we prefer that error flagging occurs within analysis tools, and not within instrumentation, a number of currently supported analysis tools and protocols rely on the existence of an explicit error flag reported from instrumentation. In OpenTelemetry, the error flag is called "status codes".
```

text/trace/0136-error_flagging.md
justinfoote: Same
```suggestion
However, there is confusion over the mapping of semantic conventions to status codes, and concern over the subjective nature of errors. Which network failures count as an error? Are 404s an error? The answer is often dependent on the situation, but without even a baseline of suggested status codes for each convention, the instrumentation author is placed under the heavy burden of making the decision. Worse, the decisions will not be in sync across different instrumentation.
```

justinfoote: Though, actually, I don't like how this reads like this. Perhaps _instrumentation providers_? _Sources of instrumentation_?
Now that I'm struggling with wording here, I wonder if it's really that important.

text/trace/0136-error_flagging.md
tedsuo: Yes, I would like to improve these conventions over time, to ensure that instrumentation is normalized across languages and. For example, we do not want to label a MongoDB code an error in one client, but not an error in another. We should do our best to record in the spec what counts as an error in the instrumentation we provide, when those errors relate to something cross-langauge, like a database protocol. 

It's also not clear to me if "generic" concepts like DB will be sufficient in the long run. SQL, Mongo, and Redis are all very different protocols. But even if the DB conventions remains generic, we still need to describe how to map each protocol to this generic convention, to ensure the data that we emit is consistent. I would like to bring in experts from these different databases to help with this.

text/trace/0136-error_flagging.md
justinfoote:
```suggestion
There is one other missing piece, required for proper error flagging. Both application developers and operators have a deep understanding of what constitutes an error in their system. OpenTelemetry must provide a way for these users to control error flagging, and explicitly indicate that it is the end user setting the status code, and not instrumentation. In these specific cases, the error flagging is known to be correct: the end user has decided the status of the span, and they do not want another interpretation.
```

text/trace/0136-error_flagging.md
justinfoote:
```suggestion
* `OK` represents a span which has been explicitly marked as being free of errors, and should not be counted against an error budget. Note that only end users should set this status. Instead, instrumentation should leave the status as `UNSET` for operations that do not generate an error.
```

text/trace/0136-error_flagging.md
justinfoote: I think we can provide reasonable mappings for databases, though the conventions will need to include special cases (but really, have you seen the db semantic conventions? It feels like it's mostly special cases 😄). We can probably provide something pretty good for messaging RPC too.  

Also, I'd bet we can make a pretty good go at a general convention as well.  If the operation results in an exception or a panic, it's probably an error (is that controversial?).

text/trace/0000-error_flagging.md
tedsuo: While I agree that the status code could be switched multiple times, I think recording a list of statuses and status sources - rather than a single enum for each - is too much. It adds complexity, and it is not clear how this data would be helpful. Only the final status matters.

text/trace/0136-error_flagging.md
tedsuo: fixed

tedsuo: removed

text/trace/0136-error_flagging.md
tedsuo: Fixed. Officially we call them [instrumentation libraries](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/glossary.md#instrumentation-library) in the spec. I added a clarification at the beginning for the terms `end user` and `instrumentation`.

text/trace/0136-error_flagging.md
tedsuo: We officially call them [instrumentation libraries](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/glossary.md#instrumentation-library). I included a note at the top to clarify that this is what the term "instrumentation" refers to in this OTEP.

text/trace/0136-error_flagging.md
tedsuo: I added a the top to clarify this.

text/trace/0136-error_flagging.md
tedsuo: Great minds think alike :)

I added a clarification to this section.

tedsuo:
Thanks @arminru. I adjusted the language to clarify the points you raised. Is this good enough to merge?

text/trace/0136-error_flagging.md
```diff
@@ -1,14 +1,16 @@
 # Error Flagging with Status Codes
 
-This proposal adds two status codes explicitly for use as overrides by the end user, and proposes a canonical mapping of semantic conventions to status codes. This clarifies how error reporting should work in OpenTelemetry.
+This proposal reduces the number of status codes to three, adds a new field to identify status codes set by application developers and operators, and adds a mapping of semantic conventions to status codes. This clarifies how error reporting should work in OpenTelemetry.
+
+Note: the term **end user** is defined as the application developers and operators of the system running opentelemetry. The term **instrumentation** refers to [instrumentation libraries](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/glossary.md#instrumentation-library) for common code shared between different systems, such as web frameworks and database clients.
```
arminru, line 5:
```suggestion
Note: The term **end user** in this document is defined as the application developers and operators of the system running OpenTelemetry. The term **instrumentation** refers to [instrumentation libraries](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/glossary.md#instrumentation-library) for common code shared between different systems, such as web frameworks and database clients.
```

text/trace/0136-error_flagging.md
arminru: I see! That sounds reasonable to me.

@tedsuo I think you forgot to push the clarification but once that's done we're good to go :shipit:

arminru (approved):
LGTM! The details will be sorted out when this is poured into a spec PR. Thanks Ted!

text/trace/0136-error_flagging.md
tedsuo: @arminru derp! Pushed. :)

carlosalberto:
Merging this as we have enough votes and don't want to decrease the velocity - and as @arminru said, we will iron out the actual details in the Specification.

cc @bogdandrutu

bogdandrutu:
@carlosalberto not that I am totally against this (I would probably request changes in that case) but as mentioned multiple times, we cannot use the velocity argument to merge changes before GA because that may cost us a lot in the future if we merge half baked changes now to rush things.

carlosalberto:
@bogdandrutu Oh, definitely. I merged cause this is 'merely' an OTEP, but let's definitely find full agreement when the actual changes hit the Specification (as long as they take).

tedsuo:
@bogdandrutu to be clear: I agree, but to be clear, this was merged because it had more than enough votes, was open for an acceptable amount of time, and there were no further objections.

PR #140, opened 2020-10-27
Fix link on Auto Resource Detection OTEP

mx-psi:
This link was broken on a repo restructuring. The name of the spec was also changed from Correlation-Context to Baggage (see w3c/baggage#27)


arminru (approved):
Thanks!

Issue #112, opened 2020-06-03
Proposal: Exemplars


cnnradams:
This proposal defines exemplars within OpenTelemetry and specifies behaviour for exemplars for the default set of aggregations: [Proposal doc](https://docs.google.com/document/d/1XpSCxzt5Q7pAIv6EeNt5x4jbqYO6pERVBO02sJvvMws/edit?ts=5ed6bf8e)

MrAlias:
From Gitter: the plan is to replace this with a PR containing the proposal.

PR #146, opened 2021-02-18
Metrics prototype scenario

reyang:
Follow up on the [02/11/2021 #108 Metrics API/SDK SIG Mtg](https://docs.google.com/document/d/1-bCYkN-DWJq4jw1ybaDZYYmx-WAe6HnwfWbkm8d57v8/edit#heading=h.hwga7jdklmjd), I've created this OTEP which has described two scenarios for the metrics prototyping work.

The actual prototype will be submitted as PR(s) to the language client repo, for example:
* https://github.com/open-telemetry/opentelemetry-dotnet/pull/1831
* https://github.com/open-telemetry/opentelemetry-java/pull/2818


text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,214 @@
...
+
+* By end of 10/2021, we should mark the metrics API/SDK specification as
+  [`Feature-freeze`](https://github.com/open-telemetry/opentelemetry-specification/blob/1afab39e5658f807315abf2f3256809293bfd421/specification/document-status.md#feature-freeze),
+  and focusing on bug fixing or editorial changes.
+
+* By end of 2021, we want to have a stable release of metrics API/SDK
+  specification, with multiple language SIGs providing RC (release candidate) or
+  GA (general available) clients.
+
+In this document, we will focus on two scenarios that we use for prototypeing
```
cijothomas, line 24:
```suggestion
In this document, we will focus on two scenarios that we use for prototyping
```

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,214 @@
...
+  from 7/1/2021 the specification should not have major surprise or big changes
+  which would thrust the language client maintainers.
+
+* By end of 10/2021, we should mark the metrics API/SDK specification as
+  [`Feature-freeze`](https://github.com/open-telemetry/opentelemetry-specification/blob/1afab39e5658f807315abf2f3256809293bfd421/specification/document-status.md#feature-freeze),
+  and focusing on bug fixing or editorial changes.
+
+* By end of 2021, we want to have a stable release of metrics API/SDK
+  specification, with multiple language SIGs providing RC (release candidate) or
+  GA (general available) clients.
```
cijothomas, line 22:
```suggestion
  GA (generally available) clients.
```

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,214 @@
...
+
+The _HTTP Server_ scenario covers how a library developer X could use metrics
+API to instrument a library, and how the application developer Y can configure
+the library to use OpenTelemetry SDK in a final application. X and Y are working
+for different companies and they don't communicate. The demo has two parts - the
+library (HTTP lib owned by X) and the server app (owned by Y):
+
+* How developer X could instrument the library code in a vendor agnostic way
+  * Performance is critical for X
+  * X doesn't know which metrics and which dimension will Y pick
```
cijothomas, line 118:
```suggestion
  * X doesn't know which metrics and which dimensions will Y pick
```

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,214 @@
...
+* [System CPU Usage](#system-cpu-usage) reported every 5 seconds
+* [Server Room Temperature](#server-room-temperature) reported every minute
+* [HTTP Server Duration](#http-server-duration), reported every 5 seconds, with
+  a subset of the dimensions:
+  * Host Name
+  * HTTP Method
+  * HTTP Host
+  * HTTP Status Code
+  * Client Type
+  * 90%, 95%, 99% and 99.9% latency
```
cijothomas, line 203:
```suggestion
  * 90%, 95%, 99% and 99.9% server duration
```

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,214 @@
...
+
+```python
+store = GroceryStore("Portland")
+store.process_order("customerA", {"potato": 2, "tomato": 3})
+store.process_order("customerB", {"tomato": 10})
+store.process_order("customerC", {"potato": 2})
+store.process_order("customerA", {"tomato": 1})
+```
+
+When the store is closed, we will report the following metrics:
```
cijothomas, line 89: It'd be good to mention the time window. For example, the store could be open from 9 am to 5pm, so is the following metrics the total number between these times? Or the SDK could export metrics every minute, and the following is the final numbers stored in some backend? If SDK is exporting every minute, does it export the delta since the last reported time, or it reports the cumulative since process start.

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,214 @@
...
+* HTTP request counters, reported every 5 seconds:
+  * Total number of received HTTP requests
+  * Total number of finished HTTP requests
+  * Number of currently-in-flight HTTP requests (concurrent HTTP requests)
+
+  | Host Name | Process ID | HTTP Host | Received Requests | Finished Requests | Concurrent Requests |
+  | --------- | ---------- | --------- | ----------------- | ----------------- | ------------------- |
+  | MachineA  | 1234       | otel.org  | 630               | 601               | 29                  |
+  | MachineA  | 5678       | otel.org  | 1005              | 1001              | 4                   |
+* Exception samples (examplar) - in case HTTP 5xx happened, developer Y would
```
cijothomas, line 213: an example of how this exemplar looks?

cijothomas, line 213:
```suggestion
* Exception samples (exemplar) - in case HTTP 5xx happened, developer Y would
```

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,214 @@
...
+Here are the languages we've agreed to use during the prototyping:
+
+* C#
+* Java
+* Python
+
+Instead of boiling the ocean, we will need to divide the work into multiple
+stages:
+
+1. Do the end-to-end prototype to get the overall understanding of the problem
```
cijothomas, line 39: Does end-to-end mean exporting the metrics to some test exporter (which just prints on console), or an actual exporter to well known system like Prometheus?

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,214 @@
...
+| Host Name | Process ID | Client Type | HTTP Method | HTTP Host | HTTP Status Code | HTTP Flavor | Peer IP   | Peer Port | Host IP   | Host Port | Duration (ms) |
+| --------- | ---------- | ----------- | ----------- | --------- | ---------------- | ----------- | --------- | --------- | --------- | --------- | ------------- |
+| MachineA  | 1234       | Android     | GET         | otel.org  | 200              | 1.1         | 127.0.0.1 | 51327     | 127.0.0.1 | 80        | 8.5           |
+| MachineA  | 1234       | Android     | POST        | otel.org  | 304              | 1.1         | 127.0.0.1 | 51328     | 127.0.0.1 | 80        | 100.0         |
+
+### Application Requirements
+
+The application owner (developer Y) would only want the following metrics:
+
+* [System CPU Usage](#system-cpu-usage) reported every 5 seconds
```
cijothomas, line 194: Not sure if this is something we want in initial stage, but i'd like to add a requirement of *same* metric being reported with different interval, potentially with diff. dimensions.
For example, the app owner wants to see HTTP Server Duration metric exported every 1 second with only HttpStatusCode dimension, and HTTP Server Duration metric exported every 30 seconds with dimensions {hostname, HTTP Method, Host, Status Code, Client Type}. The former is typically used for near-real-time dashboards, and the latter for more permanent storage.

text/metrics/0146-metrics-prototype-scenarios.md
reyang: I'm struggling with this every time - perhaps we should use "General Availability"?  https://en.wikipedia.org/wiki/Software_release_life_cycle#General_availability_(GA).

text/metrics/0146-metrics-prototype-scenarios.md
reyang: Good question, "end-to-end" is a subjective term. I think just one console/dummy exporter should be good enough for the API/SDK prototype scope.

The "real end-to-end" might include dashboard, alerting, auto deployment/rollback which are definitely out of scope.

text/metrics/0146-metrics-prototype-scenarios.md
reyang: Mitigated the problem by linking to the OpenTelemetry Stable definition.

text/metrics/0146-metrics-prototype-scenarios.md
reyang: I added one, please review.

text/metrics/0146-metrics-prototype-scenarios.md
reyang: In your example - I guess normally people would only report the 1 second one from the SDK pre-aggregation, and rely on the metrics backend to aggregate the 30 seconds one (and daily/weekly/monthly summary if there is a need).

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+* How to instrument the code in a vendor agnostic way
+* How to configure the SDK and exporter
+
+Considering there might be multiple grocery stores, the metrics we collect will
+have the store name as a dimension - which is fairly static (not changing while
+the store is running).
+
+The store has plenty supply of potato and tomato, with the following price:
+
+* Potato: $1.0 / ea
```
victlu, line 76: I suggest a double/float price to show int vs double.  Maybe Potato should be $1.15 /ea

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+
+```python
+store = GroceryStore("Portland")
+store.process_order("customerA", {"potato": 2, "tomato": 3})
+store.process_order("customerB", {"tomato": 10})
+store.process_order("customerC", {"potato": 2})
+store.process_order("customerA", {"tomato": 1})
+```
+
+When the store is closed, we will report the following metrics:
```
victlu, line 90: I will assume you mean we instrument enough data so that a backend system can generate the report per day when the store is closed?  I don't think you mean we only send metric ONCE at end of day.

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+collected based on the ask from consumer(s). If there is no ask from the
+consumer, the value will not be reported at all (e.g. there is no API call to
+fetch the room temperature unless someone is asking for the room temperature).
+
+#### Process CPU Usage
+
+Note: the **Host Name** should leverage [`OpenTelemetry
+Resource`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md),
+so it should be covered by the metrics SDK rather than API, and strictly
+speaking it is not considered as a "dimension" from the SDK perspective.
```
victlu, line 144: I will ask about how vendors can "enrich" a data point.  I assume we have some "Processors" or extension points for vendors to enrich or alter data points (what is allowed/disallowed is also a discussion).

Separately, it would be good to clearly state who is responsible for providing the "Resource" labels?  Is it an Exporter or a Processor, or ???

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+| Host Name | Process ID | Client Type | HTTP Method | HTTP Host | HTTP Status Code | HTTP Flavor | Peer IP   | Peer Port | Host IP   | Host Port | Duration (ms) |
+| --------- | ---------- | ----------- | ----------- | --------- | ---------------- | ----------- | --------- | --------- | --------- | --------- | ------------- |
+| MachineA  | 1234       | Android     | GET         | otel.org  | 200              | 1.1         | 127.0.0.1 | 51327     | 127.0.0.1 | 80        | 8.5           |
+| MachineA  | 1234       | Android     | POST        | otel.org  | 304              | 1.1         | 127.0.0.1 | 51328     | 127.0.0.1 | 80        | 100.0         |
+
+### Application Requirements
+
+The application owner (developer Y) would only want the following metrics:
+
+* [System CPU Usage](#system-cpu-usage) reported every 5 seconds
```
victlu, line 195: I assume we need to create multiple "pipelines" so that each pipeline can be configured individually.  This implies IMHO that measurements reported from API needs to be routed to each individual pipelines.

Currently, I think this is an issue with our use of a "Default" provider in Library X, if Library X does not take a dependency on the SDK.

text/metrics/0146-metrics-prototype-scenarios.md
reyang: > I suggest a double/float price to show int vs double. Maybe Potato should be $1.15 /ea. Also, I would assume the dev will treat this as a currency, so it's $3.00 (two "zeros").

Good point, let me make the change.
 
> Thus, a follow up question. Do we need market specific types like Currency? Or rather, a way to represent a unit like "USD" (U.S. Dollar).

I don't know 😄
Unit to me is very similar to semantic convention - which is hard to align.

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+community is willing to spend more energy on metrics API/SDK. The goal is to get
+the metrics API/SDK specification to
+[`Experimental`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#experimental)
+state by end of 5/2021, and make it
+[`Stable`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#stable)
+before end of 2021:
+
+* By end of 5/2021, we should have a good confidence that we can recommend
+  language client owners to work on metrics preview release. This means starting
+  from 7/1/2021 the specification should not have major surprise or big changes
```
jsuereth, line 13: nit: s/surprise/surprises/


Also should we make any statement about avoiding breaking changes as much as possible, without guaranteeing there won't be any breaking change?

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+stages:
+
+1. Do the end-to-end prototype to get the overall understanding of the problem
+   domain. We should also clarify the scope and be able to articulate it
+   precisely during this stage, here are some examples:
+
+    * Why do we want to introduce brand new metrics APIs versus taking a well
+      established API (e.g. Prometheus and Micrometer), what makes OpenTelemetry
+      metrics API different (e.g. Baggage)?
+    * Do we need to consider OpenCensus Stats API shim, or this is out of scope?
```
jsuereth, line 47: FYI - If we need to spring up a separate working-group, we can do that.  

Even today, it's likely OpenCensus will have to do an SDK Export => SDK Export bridge rather than API => API (just given the nature of how it works).  I'd rather aggressively get a stable OTEL API, and we can determine shims after that.

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+* Python
+
+Instead of boiling the ocean, we will need to divide the work into multiple
+stages:
+
+1. Do the end-to-end prototype to get the overall understanding of the problem
+   domain. We should also clarify the scope and be able to articulate it
+   precisely during this stage, here are some examples:
+
+    * Why do we want to introduce brand new metrics APIs versus taking a well
```
jsuereth, line 44: This point (and the next point) beg the question:

Should we make ONE of the use cases be "hide OTEL behind another library to help users take advantage of OTEL telemetry-unfiication concepts, like Baggage + Resource".   This could be Micrometer, OpenCensus, whatever.

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+  * How should Y hook up the metrics SDK with the library
+  * How should Y configure the time window(s) and destination(s)
+  * How should Y pick the metrics and the dimensions
+
+### Library Requirements
+
+The library developer (developer X) will expose the following metrics out of
+box:
+
+### Pull Metrics
```
jsuereth, line 132: I'm having trouble processing the context here.  Would you mind adjusting the Header-values so it's more clear if this is in "Library Requirements".

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+
+These are pull metrics - the value is always available, and is only reported and
+collected based on the ask from consumer(s). If there is no ask from the
+consumer, the value will not be reported at all (e.g. there is no API call to
+fetch the room temperature unless someone is asking for the room temperature).
+
+#### Process CPU Usage
```
jsuereth, line 139: It may be useful in your tables to show "resource" with sub-tables for the components coming from resource (host name, process_id I assume)

text/metrics/0146-metrics-prototype-scenarios.md
jsuereth: IIUC - This is done in Exporters for Trace.  If a vendor isn't directly supporting `Resouce` (or where their notion of `Resource` doesn't fully align with OTLP), the vendor can lift `Resource` labels onto the trace in the export method.  We actually plan to do this for instrumentation library (although I don't think we do it today).

text/metrics/0146-metrics-prototype-scenarios.md
jsuereth, line 139: Is "Process CPU Usage" really an HTTP server concern?   I feel like this may be a general instrumentation concern, and the HTTP metrics should really be focused only on things HTTP libraries do.  CPU consumption is more of a process-wide concern.

I'd suggest using `Active HTTP Connections` as the pull metric from the HTTP library.


For Server Room Temperature, I love what it's trying to do, but I'm having trouble buying it's part of an http library.  Maybe move it into its own instrumentation component?

jsuereth (changes requested):

text/metrics/0146-metrics-prototype-scenarios.md
reyang: I've fixed the s/surprise/surprises nit.

Regarding breaking changes, I think we only have the "no breaking change" guarantee after we announce spec `Stable`. Before stable it will be `Experimental` so we can make breaking changes, we should always keep a high bar though.

text/metrics/0146-metrics-prototype-scenarios.md
reyang: We have discussed this during the [02/18/2021 #112 Metrics API/SDK SIG Mtg](https://docs.google.com/document/d/1-bCYkN-DWJq4jw1ybaDZYYmx-WAe6HnwfWbkm8d57v8/edit#heading=h.jirzlx3f0p86) and the consensus is to provide the semantic support but not try to do API <-> API mapping.

text/metrics/0146-metrics-prototype-scenarios.md
reyang: I've updated the ToC.

text/metrics/0146-metrics-prototype-scenarios.md
reyang: I'm going to resolve this comment and leave the unit question as a Data Model problem (rather than API/SDK problem).

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+
+The _HTTP Server_ scenario covers how a library developer X could use metrics
+API to instrument a library, and how the application developer Y can configure
+the library to use OpenTelemetry SDK in a final application. X and Y are working
+for different companies and they don't communicate. The demo has two parts - the
+library (HTTP lib owned by X) and the server app (owned by Y):
+
+* How developer X could instrument the library code in a vendor agnostic way
+  * Performance is critical for X
+  * X doesn't know which metrics and which dimensions will Y pick
```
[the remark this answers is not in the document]
jmacd: Yes, adding to @noahfalk's observation, the OTel metrics discussion has at times gone back and forth on the topic of "recommended dimensions" (https://github.com/open-telemetry/opentelemetry-specification/issues/463), more recently using the terms "Descriptive" and "Identifying" to distinguish optional and non-optional attributes (https://github.com/open-telemetry/opentelemetry-specification/issues/1298#issuecomment-775749864).

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+  * Host Name
+  * HTTP Method
+  * HTTP Host
+  * HTTP Status Code
+  * Client Type
+  * 90%, 95%, 99% and 99.9% server duration
+* HTTP request counters, reported every 5 seconds:
+  * Total number of received HTTP requests
+  * Total number of finished HTTP requests
+  * Number of currently-in-flight HTTP requests (concurrent HTTP requests)
```
jmacd, line 208: I like how this example asks for three counters, because it seems possible to achieve with two instruments: a count of received requests and a histogram of response durations (i.e., seems to call for either a view or a 3rd instrument).

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+
+```python
+store = GroceryStore("Portland")
+store.process_order("customerA", {"potato": 2, "tomato": 3})
+store.process_order("customerB", {"tomato": 10})
+store.process_order("customerC", {"potato": 2})
+store.process_order("customerA", {"tomato": 1})
+```
+
+When the store is closed, we will report the following metrics:
```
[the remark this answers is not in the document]
jmacd: I see (1) and (2) as good use-cases for monitoring using a metrics API, but maybe not (3). Although the example feels like it fell out of a textbook, you could re-imagine the store as a Message-Queue consumer processing orders in a horizontally scalable store. Can we ask another form of query: "how many stores were in operation at a given time?"

jmacd (approved):

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+the metrics API/SDK specification to
+[`Experimental`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#experimental)
+state by end of 5/2021, and make it
+[`Stable`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#stable)
+before end of 2021:
+
+* By end of 5/31/2021, we should have a good confidence that we can recommend
+  language client owners to work on metrics preview release. This means starting
+  from 6/1/2021 the specification should not have major surprises or big changes
+  which would thrust the language client maintainers, and we start to recommend
```
lzchen, line 14:
```suggestion
, and we start to recommend
```

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+  client maintainers to implement it. We might introduce additional features but
+  there should be a high bar.
+
+* By end of 9/30/2021, we should mark the metrics API/SDK specification as
+  [`Feature-freeze`](https://github.com/open-telemetry/opentelemetry-specification/blob/1afab39e5658f807315abf2f3256809293bfd421/specification/document-status.md#feature-freeze),
+  and focusing on bug fixing or editorial changes.
```
lzchen, line 20:
```suggestion
  and focus on bug fixing or editorial changes.
```

text/metrics/0146-metrics-prototype-scenarios.md
lzchen, line 14:
```suggestion
  . We will then start recommending
```

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+
+* By end of 2021, we want to have a stable release of metrics API/SDK
+  specification, with multiple language SIGs providing RC (release candidate) or
+  [stable](https://github.com/open-telemetry/opentelemetry-specification/blob/9047c91412d3d4b7f28b0f7346d8c5034b509849/specification/versioning-and-stability.md#stable)
+  clients.
+
+In this document, we will focus on two scenarios that we use for prototyping
+metrics API/SDK. The goal is to have two scenarios which clearly capture the
+major requirements, so we can work with language client SIGs to prototype,
+gather the learnings, determine the scopes and stages. Later the scenarios can
```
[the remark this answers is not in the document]
lzchen: Could you provide an example of one of these?

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+be used as examples and test cases for all the language clients.
+
+Here are the languages we've agreed to use during the prototyping:
+
+* C#
+* Java
+* Python
+
+Instead of boiling the ocean, we will need to divide the work into multiple
```
lzchen, line 39:
```suggestion
In order to not undertake such an enormous task at once, we will need to have an incremental approach and divide the work into multiple
```

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+   might be very similar (e.g. if we know how to record an integer, we don't
+   have to work on float/double as we can add them later by replicating what
+   we've done for integer).
+
+3. Focus on a core subset of SDK. This would help us to get the end-to-end
+   application.
+
+4. Replicate stage 2 to cover the complete set of APIs.
+
+5. Replicate stage 4 to cover the complete set of SDKs.
```
lzchen, line 62:
```suggestion
5. Replicate stage 3 to cover the complete set of SDKs.
```

lzchen (approved):

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+| --------- | ---------- | ----------- | ----------- | --------- | ---------------- | ----------- | --------- | --------- | --------- | --------- | ------------- |
+| MachineA  | 1234       | Android     | GET         | otel.org  | 200              | 1.1         | 127.0.0.1 | 51327     | 127.0.0.1 | 80        | 8.5           |
+| MachineA  | 1234       | Android     | POST        | otel.org  | 304              | 1.1         | 127.0.0.1 | 51328     | 127.0.0.1 | 80        | 100.0         |
+
+### Application Requirements
+
+The application owner (developer Y) would only want the following metrics:
+
+* Server temperature - reported every 5 seconds
+* Server humidity - reported every minute
```
victlu, line 196: AFAIK, we need to have multiple "pipelines" for different configurations.  This may include...
- Reporting period which is related to Collection rate, Export rate, etc...
- Selection / grouping of desired metrics for this pipeline

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,220 @@
...
+* HTTP request counters, reported every 5 seconds:
+  * Total number of received HTTP requests
+  * Total number of finished HTTP requests
+  * Number of currently-in-flight HTTP requests (concurrent HTTP requests)
+
+  | Host Name | Process ID | HTTP Host | Received Requests | Finished Requests | Concurrent Requests |
+  | --------- | ---------- | --------- | ----------------- | ----------------- | ------------------- |
+  | MachineA  | 1234       | otel.org  | 630               | 601               | 29                  |
+  | MachineA  | 5678       | otel.org  | 1005              | 1001              | 4                   |
+* Exception samples (exemplar) - in case HTTP 5xx happened, developer Y would
```
victlu, line 214: Would this still be Metrics?  It seems like this is more logging.

reyang: Yes.

Exemplars (correlating metrics and traces) is a top goal for OpenTelemetry.
Prometheus will also add this very soon https://grafana.com/blog/2020/07/23/the-future-of-prometheus-and-its-ecosystem/.

text/metrics/0146-metrics-prototype-scenarios.md
reyang: Updated. Thanks!

text/metrics/0146-metrics-prototype-scenarios.md
reyang: Good catch. Fixed.

text/metrics/0146-metrics-prototype-scenarios.md
reyang: I've removed CPU since itself is a complex topic and not the focus on this OTEP/prototype.
I've extracted temperature/humidity to a separate lib to make the scenario more realistic/reasonable.

For Active HTTP Connections, I don't know if we want to do it by reporting "total received - total processed" or doing it differently (see the discussion [here](https://github.com/open-telemetry/opentelemetry-specification/pull/1378)). Given this discussion might take extra time, probably leave it out for now?

jsuereth (approved):
My main comment, thinking deeper on this, is whether or not "external annotation" needs to be called out specifically as a non-goal.

Specifically, should we call out a use case where an DevOps wants to add metric labels *POST DEVELOPMENT* to the API.   I think this is an SDK concern and should be called out there (e.g. how you can add annotations to Resource via ENV, which would then influence metrics), but I think the scenarios listed here are sufficient for API discussions.

text/metrics/0146-metrics-prototype-scenarios.md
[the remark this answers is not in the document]
reyang: +1, besides "recommended dimensions", the library might need to give recommendation on the buckets (consider the histogram scenario, how would the user know what's the best practice?)

text/metrics/0146-metrics-prototype-scenarios.md
reyang: And it might affect the semantic convention https://github.com/open-telemetry/opentelemetry-specification/pull/1378#discussion_r578093122.

text/metrics/0146-metrics-prototype-scenarios.md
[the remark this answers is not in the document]
cijothomas: There are 6 instrumented in current [API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#the-instruments). Anything from that list which is not used for the above 2 scenarios.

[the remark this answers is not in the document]
noahfalk: Another example I am anticipating are the batching APIs. My cursory review of some other pre-existing metric APIs suggested that atomic modification of multiple metrics wasn't a common feature and nothing seemed apparent to me in these example scenarios that required batching to exist.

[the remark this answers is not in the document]
jmacd: The draft OTel APi included a number of things that were supposedly demanded for performance. One of those was a batch API for recording metrics, but while we agreed on that there were many unanswered questions about the bigger picture--particularly because there is a conflicting request for per-instrument collection intervals. Nothing that can't be solved, just growing scope and complexity.

> Could you provide an example of one of these?

There are no Gauge use-cases in these scenarios. Although the 6 instruments in the draft API cover most scenarios, I've become convinced that users will continue to ask for a synchronous form of Gauge, which makes at least 7 use-cases, but even if we are convinced that all valid use-cases can be mapped into 7 fundamental instruments, it doesn't mean users want those APIs.

For the current scenario 2, there's a relatively heavy-weight instrumentation called for in a high-value piece of instrumentation. You are likely to invest in more than one line of code for a timing measurement in that case.

OTOH, in less high-value code, I'm likely to use a one-liner for measuring a duration metric. Here's an example OTel-Go `ValueRecorder` wrapper that emits a timing measurement using a one-line expression like

```
func myConnect() (retErr error) {
  defer connectDuration.Start(ctx).Stop(&retErr)
  // ... 
}
```

It's not clear that 7 scenarios is enough, but I think starting with 2 is great and avoiding Gauges to start is :100:.

text/metrics/0146-metrics-prototype-scenarios.md
jmacd: (Reply: https://github.com/open-telemetry/opentelemetry-specification/pull/1378#issuecomment-784910215)

text/metrics/0146-metrics-prototype-scenarios.md
jmacd: These are all SDK configuration topics, to me. The use-cases in this document are about how code is instrumented, I think, and what the API looks like.

OpenCensus had a programmatic API for configuring the kinds of variables you mentioned. I'm not sure if a programmatic API or a configuration file is what user's want, but I'd argue to keep this setup out of the instrumentation API.

text/metrics/0146-metrics-prototype-scenarios.md
jmacd: The thing about CPU is that it's a perfect example of an asynchronous instrument in the draft API. We want these to be recorded though callbacks, because it's expensive. This means the value is returned in cumulative form, not in delta form.

lalitb (approved):

text/metrics/0146-metrics-prototype-scenarios.md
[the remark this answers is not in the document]
reyang: I've added one bullet point here to address the feedback.

text/metrics/0146-metrics-prototype-scenarios.md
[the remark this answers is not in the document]
reyang: > It's not clear that 7 scenarios is enough, but I think starting with 2 is great and avoiding Gauges to start is 💯.

+100 💯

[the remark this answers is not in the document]
reyang: We will need to take the step by step approach.

For things that are currently in the draft spec, but not covered/used by the scenarios here, we should table them for now and only revisit them after we covered the basics.

Please refer to the next section - "we will need to have an incremental approach and divide the work into multiple stages", it should clarify the approach we're taking.

text/metrics/0146-metrics-prototype-scenarios.md
reyang: Makes sense, let's change it to every minute.

text/metrics/0146-metrics-prototype-scenarios.md
reyang: Yep, this is similar to what @cijothomas mentioned [here](https://github.com/open-telemetry/oteps/pull/146/files#r578138526), let me change it to a per minute report.

text/metrics/0146-metrics-prototype-scenarios.md
reyang: These are more like the SDK design question and implementation question. Probably not putting too much info in this doc since we're trying to cover the scenario/scope here?

text/metrics/0146-metrics-prototype-scenarios.md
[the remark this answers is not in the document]
noahfalk: Sounds great, thanks for the clarification : )

noahfalk (approved):
Looks great, thanks @reyang!

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,222 @@
...
+| Portland | customerB | tomato | 10    |
+| Portland | customerC | potato | 2     |
+
+## Scenario 2: HTTP Server
+
+The _HTTP Server_ scenario covers how a library developer X could use metrics
+API to instrument a library, and how the application developer Y can configure
+the library to use OpenTelemetry SDK in a final application. X and Y are working
+for different companies and they don't communicate. The demo has two parts - the
+library (HTTP lib owned by X) and the server app (owned by Y):
```
cijothomas, line 116:
```suggestion
library (HTTP lib and ClimateControl lib owned by X) and the server app (owned by Y):
```

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -0,0 +1,222 @@
...
+
+The library developer (developer X) will provide two libraries:
+
+* Server climate control library - a library which monitors and controls the
+  temperature and humidity of the server.
+* HTTP server library - a library which provides HTTP service.
+
+Both libraries will provide out-of-box metrics, the metrics have two categories:
+
+* Push metrics - the value is reported (via the API) when it is available, and
```
cijothomas, line 141: nit: Do you want to clarify which metric in the example is Push/Pull? (I assume climate one only has Pull ones?)

cijothomas (approved):

text/metrics/0146-metrics-prototype-scenarios.md
reyang: I think push/pull metrics can be implementation details. For example, the HTTP server duration can be implemented as either push or pull.

reyang: If later we feel there is a need to make it concrete, we can have a small follow up PR.

jmacd:
I reviewed the open comments and believe that @reyang has addressed them all. Thank you!

PR #147, opened 2021-03-05
Upgrade procedures for OpenTelemetry

tedsuo:
This OTEP clarifies the stability requirements in the [Versioning and Support document](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md).

* Clarify that the API interfaces are backwards compatible for callers which use them, not SDKs which implement them.
* Clarify that the SDK plugin interfaces are backwards compatible for implementors.
* Explanation as to why we need these requirements, and the problems that they solve.
* Explanation as to how this design avoids the need to backport patches to older versions of the SDK, or maintain multiple versions of instrumentation.

A google doc version of this OTEP is available here, in case you find that format easier to comment on:
https://docs.google.com/document/d/1p_KDgfBo6_7kO7cvio4ZrcjsJqZqr3UyT00zimUI7qY/edit#


text/0000-upgrade-procedures
```diff
@@ -0,0 +1,56 @@
...
+
+The API refers to the set of software packages that contain all of the interfaces and constants needed to write OpenTelemetry instrumentation. An implementation of the API may be registered during application startup. If no other implementation is registered, the API registers a no-op implementation by default.
+
+The SDK refers to a framework which implements the API, provided by the OpenTelemetry project. While alternative API implementations may be written to handle special cases, we expect most users to install the SDK when running OpenTelemetry.
+
+Plugin Interfaces refer to extension points provided by the SDK. These include interfaces for controlling sampling, exporting data, and various other lifecycle hooks. Note that these interfaces are not part of the API. They are part of the SDK.
+
+Instrumentation refers to any code which calls the API. This includes the instrumentation provided by the OpenTelemetry project, third party instrumentation, plus application code and libraries which instrument themselves natively.
```
jsuereth, line 17: nit: "Instrumentation refers to any code which calls the API to produce instrumentation signals".  (albeit non-instrumentation usage of the API is limited to SDK or SDK replacements)

text/0000-upgrade-procedures
```diff
@@ -0,0 +1,56 @@
...
+
+## The OpenTelemetry upgrade path
+
+Before we get into the design requirements, here’s how upgrading actually works. Note that all OpenTelemetry components (API, SDK, plugins, instrumentation) have separate version numbers.
+
+### API changes
+
+When new functionality is added to the OpenTelemetry API, and a new minor version of the API is released. These API changes are always additive and backwards compatible from the perspective of existing Instrumentation packages which import and call prior versions. Instrumentation written against all prior minor versions of the API continues to work, and may be composed together into the same application without creating a dependency conflict.
+
+API implementations are expected to always target the latest version of the API. When a new version of the API is released, a version of the SDK which supports the API is released in tandem. New versions of the API are not expected to support older versions of the SDK.
```
jsuereth, line 31: Maybe bold/italicize the "new versions of the API are not expected to support older version of the SDK"

text/0000-upgrade-procedures
```diff
@@ -0,0 +1,56 @@
...
+
+This would be an enormous maintenance effort at best. But, as the OpenTelemetry project only controls a portion of that ecosystem, it is also unfeasible. OpenTelemetry cannot require that libraries with native instrumentation support multiple versions of the API. Ensuring that application owners and operators can upgrade to the latest version of the SDK resolves this issue.
+
+The primary blocker to upgrading the SDK is out of date Plugins. If a new version of the SDK were to break existing Plugin Interfaces, no user would be able to upgrade their SDK until the Plugins they depend on have been upgraded. Users could be caught between instrumentation they depend on requiring a version of the API which is not compatible with the version of the SDK which supports their Plugins.
+
+By following a deprecation pattern with Plugin Interfaces, we create a one year window in which the Plugin ecosystem can upgrade after the release of a new SDK. We believe this is sufficient time for any Plugin which is actively maintained to make an upgrade, and for defunct Plugins to be identified and replaced.
```
jsuereth, line 54: I've been thinking a  lot about this here, and I want to through out an idea i've been noodling on.

1. Keep this 1-year plugin interface deprecation policy for most SDK plugin interfaces.
2. To alleviate concerns around 1-year timeline we provide a longer timeline on the Collector
  - We keep a 3-year compatibility guarantee on OTLP exporter to collector
  - We provide longer stability guarantees of "plugins API"s through the Collector

This is a non blocking comment, and just a thought I've had when reading this.  I think what you've written represents a balance of "what we can do" and "what users will want".   However, I know from "build plugin" space that 1-year can actually be "fast" for some old and highly used (but unchanged) plugins.   I'm not sure that'll be true in our space but want to call out the parallel.   Build plugins *likely* are more akin to instrumentation, but I still worry 1 year will be too short for the edge case scenarios we'll see people solve.

jsuereth (approved):

text/0000-upgrade-procedures
carlosalberto: Maybe add a note on how important it will be to release the SDK in tandem, i.e. "a version of the SDK which supports the API MUST be released in tandem"

text/0000-upgrade-procedures
```diff
@@ -0,0 +1,56 @@
...
+
+API implementations are expected to always target the latest version of the API. When a new version of the API is released, a version of the SDK which supports the API is released in tandem. New versions of the API are not expected to support older versions of the SDK.
+
+### SDK changes
+
+Bugs fixes, security patches, and performance improvements are released as patch versions of the SDK. Support for new versions of the API are released as minor versions. New Plugin Interfaces and configuration options are also released as minor versions.
+
+Breaking changes to Plugin Interfaces are handled through deprecation. Instead of breaking a Plugin Interface, a new interface is created and the existing interface is marked as deprecated. Plugins which target the deprecated interface continue to work, and the SDK provides default implementations of any missing functionality. After one year, the deprecated Plugin Interface is removed as a major version release of the SDK.
```
carlosalberto, line 37: For some languages adding the new element to the existing interface **along with** a default implementation will _probably_ be a better fit. cc @jkwatson

text/0000-upgrade-procedures
```diff
@@ -0,0 +1,56 @@
...
+
+## Design requirements and explanations
+
+This approach to upgrading solves two critical design requirements, while minimizing the maintenance overhead associated with legacy code.
+
+* Callers of the API are never broken.
+* Users of the SDK can easily upgrade to the latest version.
+
+Indefinite support for existing Instrumentation is a critical feature of OpenTelemetry. Millions of lines of code are expected to be written against the API. This includes shared libraries which ship with integrated OpenTelemetry instrumentation. These libraries must be able to compose together into applications without OpenTelemetry creating a dependency conflict. While some Instrumentation will be updated to the latest version of the API, such as that provided by the OpenTelemetry project, other Instrumentation will never be updated.
```
carlosalberto, line 46: +1

carlosalberto (approved):

text/0000-upgrade-procedures
tedsuo: So, in this doc I want to define the term Instrumentation to specify _any_ caller of the API package, vs any implementor of the package interfaces. In other words, focus on the directionality of the dependency and not the use case. The relevant information for the reader is to understand that directionality, not the use case. I will try to make it super clear that the term "instrumentation" cover other random use cases as well, as far as this document is concerned.

text/0000-upgrade-procedures
tedsuo: Sounds good. I'm not using MUST/SHOULD/etc terminology in the OTEP just yet but I will make sure to add it before we commit it to the specification.

text/0000-upgrade-procedures
tedsuo: I think that would be @tigrannajaryan/@bogdandrutu/collector-maintainers decision.

In general, I think we can support most deprecated plugin interfaces indefinitely, and the window is actually much wider. But I' prefer not to officially make such a long timeline a rule, as there may be edge cases where a Plugin Interface is being changed in order to support a change in the underlying framework architecture, and maintaining the old plugin interface is actually very burdensome.

text/0000-upgrade-procedures
jkwatson: If it's truly a breaking change, then a new interface would need to be provided. But, the old interface should still be supported.

text/0000-upgrade-procedures
```diff
@@ -0,0 +1,56 @@
...
+
+Consuming new Instrumentation will require users to upgrade to the latest version of the SDK. If it was not easy to perform this upgrade, the OpenTelemetry project would be forced to support older versions of the SDK, as well as older versions of the entire Instrumentation ecosystem.
```
jkwatson, line 48:
```suggestion
Consuming new Instrumentation might require users to upgrade to the latest version of the SDK. If it was not easy to perform this upgrade, the OpenTelemetry project would be forced to support older versions of the SDK, as well as older versions of the entire Instrumentation ecosystem. 
```

jkwatson:
Minor nit...if this were a `.md` file, then github would let us read the formatted version, which would be very much preferable to having to scroll to the right in the window so much.

text/0000-upgrade-procedures
Aneurysm9: I would view this as a floor applied across the entire project.  If individual SIGs wanted to provide longer compatibility guarantees then I wouldn't see this as necessarily precluding them from doing so.

text/0000-upgrade-procedures
```diff
@@ -0,0 +1,56 @@
...
+
+There is an important distinction between Plugins and Instrumentation. Plugins implement the Plugin Interfaces. Instrumentation calls the API. This difference is relevant to OpenTelemetry’s approach to upgrading.
+
+## The OpenTelemetry upgrade path
+
+Before we get into the design requirements, here’s how upgrading actually works. Note that all OpenTelemetry components (API, SDK, plugins, instrumentation) have separate version numbers.
+
+### API changes
+
+When new functionality is added to the OpenTelemetry API, and a new minor version of the API is released. These API changes are always additive and backwards compatible from the perspective of existing Instrumentation packages which import and call prior versions. Instrumentation written against all prior minor versions of the API continues to work, and may be composed together into the same application without creating a dependency conflict.
```
Aneurysm9, line 29:
```suggestion
When new functionality is added to the OpenTelemetry API, a new minor version of the API is released. These API changes are always additive and backwards compatible from the perspective of existing Instrumentation packages which import and call prior versions. Instrumentation written against all prior minor versions of the API continues to work, and may be composed together into the same application without creating a dependency conflict.
```

Aneurysm9 (approved):

text/0000-upgrade-procedures
punya: I thought this document was exclusively about language APIs and SDKs. Do we need a separate doc, or at least a section, for the collector?

I think of the collector as having *only* plugin interfaces, no APIs. Also, since the collector is written in a single language (Go), we can be a lot more specific about the meaning of compatibility.

text/0000-upgrade-procedures
reyang: I guess @jsuereth was saying "SDK is not an instrumentation" (although SDK calls the API)?

mtwo (approved):

text/0000-upgrade-procedures
tedsuo: @jsuereth want to follow up on this when I add it to the spec?

text/0000-upgrade-procedures
tedsuo: @punya correct this document is just about the OpenTelemetry clients. In my opinion the Collector should have the same minimal guarantee (One year support for deprecated plugin interfaces), but I agree that there should be a stability and support document written which is specific to the Collector.

text/0000-upgrade-procedures
tedsuo: Totally. Some languages are more flexible than others, so its not the case that every language will experience deprecation at the same rate. The intent is to only apply these rules for situations where a deprecation would lead to a breaking change to plugin implementations.

bogdandrutu (approved):

carlosalberto:
Merging as we have 4 reviews (and no changes requested). @tedsuo let's follow an issue in the Specification to get this integrated there, etc.

PR #148, opened 2021-03-06
Probability sampling basics for telemetry events

jmacd:
This OTEP defines a foundation for probability sampling in OpenTelemetry.

This drafts a specification for how to encode the sampling probabilities in a span, to enable statistical summarization from sampled traces using the `sampler.name` and `sampler.adjusted_count` attributes.


text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+algorithms for calculating statistics about the population using just
+the sample data.  For the broad class of telemetry sampling
+application considered here, we need an estimator for the population
+total represented by each individual event.
+
+### Sample count adjustment
+
+The estimated population total divided by the individual event value
+equals the event's _adjusted count_.  The adjusted count is named
+`sample_count` where it appears as a field in a telemetry data type.
```
yurishkuro, line 41: Why not `sample_weight`? Count (a) sounds like an integer, and (b) sounds like we literally received N identical events and compressed into one, while in practice we didn't even look at N-1 events and just assumed that this event represents N because of how it was selected.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+The sampling approach covered here dictates how to select root spans
+in a probabilistic way.  When recording root spans, the `sample_count`
+field should be set as described above.  The adjusted `sample_count`
+of the root span applies the trace, meaning the trace should be
+considered as representative of `sample_count` traces in the
+population.
+
+When non-root spans are recorded because they are part of an ongoing
+trace, they are considered non-probabilistic exemplars.  Non-root
+spans should have `sample_count` set to zero.
```
yurishkuro, line 180: This sounds like it's saying that the cumulative statistics (e.g. throughput or error rate) cannot be inferred for a service that is not a root?

jmacd: @akehlenbeck Agree! I left too much of a gap in this document by talking about what we record in the span event as opposed to what a tracing system can do to sample and re-sample spans at various levels in a trace. 

I remember the Dapper details you describe--unfortunately it wasn't discussed in the Dapper paper. I heard that @bogdandrutu may recall removing the `inverse_probability` field from trace propagation in the present-day Dapper system. (I hear other means are now used.)

The conversation about how to create new roots at interior nodes of a trace is important (and was to Dapper) when a high-throughput service occasionally calls a low-throughput service. Uniform sampling of the high-throughput service rarely traces the low-throughput service; starting new roots makes low-throughput interior nodes visible. 

In some ways, you could use all "true" roots in your trace collection to analyze the behavior of a certain class of interior span. This would be a lot like analyzing a stratified sample. You could also compute an independent sample for traces rooted at the same class of interior span.  You could use both data sets to analyze the behavior of that class of interior span, as they support estimatation of the same quantities.

Dapper chose a less expennsive alternative to what I wrote above,which allows spans sampled as parts of rooted traces to be mixed with spans sampled as parts of "new-roots".  I believe what Dapper did and you describe is probabilistically correct, and I believe the proposed `sample_count` field can be used to carry the intended meaning if this approach is taken.  The proof would be to show how the sume of `sample_count` over a stream of spans is expected to equal the true span count.

I think a new OTEP to discuss propagating "inclusion probability" (or its inverse) for recording in sample spans is called for.

oertl: @yurishkuro Reproducing the three possible cases in your example:
1. X and Y are sampled in 10% of all cases. The `sample_count` for Y would be 1.
2. X is not sampled while Y is sampled in 0.1 * 0.9 = 9% of all cases. The `sample_count` of Y would be 1 / (0.9 * 0.1).
3. X and Y are both not sampled in 81% of all cases. Y is not recorded.

The expectation of the sample count for Y would therefore be

10% * 1 + 9% * (1 / (0.9 * 0.1)) = 1.1

showing that there is something wrong. Unbiased sampling would give an expected sample count of 1.

This quite simple example already shows that it's not that easy to get subtree sampling right.

akehlenbeck: > I heard that @bogdandrutu may recall removing the `inverse_probability` field from trace propagation in the present-day Dapper system. (I hear other means are now used.)

Is anyone able to share details about these other means?

jmacd:
I had an off-line meeting with @oertl, who mentioned that Dynatrace has a field named `multiplicity` used to convey sampling information. This is a term I had not considered, and one that I like.

text/0148-sampling-probability.md
paulosman: > So perhaps in case 1 the right thing to record is neither `p=0` like Josh wrote nor `p=1` like I wrote, but just `p=0.1 and Y isn't a root`, where the "Y isn't a root" part indicates that we're in case 1 and we shouldn't combine with other cases.

That makes a lot of sense to me and seems sufficient for tracing systems that don't support promoting child spans with missing parents. I'd also be curious if there's a way to combine this with case 2, but I don't have any ideas myself :-)

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+### Applicability for tracing
+
+The sampling approach covered here dictates how to select root spans
+in a probabilistic way.  When recording root spans, the `sample_count`
+field should be set as described above.  The adjusted `sample_count`
+of the root span applies the trace, meaning the trace should be
+considered as representative of `sample_count` traces in the
+population.
+
+When non-root spans are recorded because they are part of an ongoing
```
paulosman, line 178: @jmacd It seems like most use cases discussed below support having non-root spans carry the same `sample_count` field as the trace. What do you think of just dropping the paragraph from this section?

paulosman (approved):
I really like this proposal, thank you!

The described approach solves two particular use cases that I think come up quite often:

1. Allowing a tracing system to easily estimate the size of a population (in visualizations, counts, etc)
2. Allowing clients (probably through tail sampling) to make dynamic sampling decisions

I've made some suggested changes, one that should make the linters happy and a couple of minor edits.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+
+The specified sampling adjustment sets the `sample_count` of each
+sampled event to the inverse of the event's inclusion probability.
+Conveying the inverse of the inclusion probability is convenient for
+several reasons:
+
+- A `sample_count` of 1 indicates no sampling was performed
+- A `sample_count` of 0 indicates an unrepresentative event
+- Larger `sample_count` indicates greater representivity
+- Smaller `sample_count` indicates smaller representivity
```
seh, line 58:
```suggestion
- A `sample_count` of 1 indicates no sampling was performed.
- A `sample_count` of 0 indicates an unrepresentative event.
- A larger `sample_count` indicates greater representativity.
- A smaller `sample_count` indicates smaller representativity.
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+- The sum of `sample_count` in a sample equals the expected 
+  population size.
+  
+The zero `sample_count` value supports collecting values that were
+rejected from a sample to use as exemplars, which supports encoding of
```
seh, line 63: Does "rejected from a sample to use as exemplars" mean that the event is _not_ exemplary, or that it's an example of something being elided? The list above says that a `sample_count` of zero indicates the former, using different language.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+values that should not be counted in an estimate of the population
+total.  This has applications to reservoir sampling designs, where
+events may be selected for a sample only to be rejected before the end
+of the frame.
+
+### Sampling with attributes
+
+Sampling is a powerful approach when used with event data that has
+been annotated with key-value attributes.  It is possible to select
+arbitrary subsets of the sampled data and use it to estimate the count
```
seh, line 73:
```suggestion
arbitrary subsets of the sampled data and use them to estimate the count
```
That is, assuming that "it" referred to the subsets. Alternately:
```suggestion
arbitrary subsets of the sampled data and use each to estimate the count
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+Applied correctly, this approach provides accurate estimates for
+population counts and distributions with support for ad-hoc queries
+over the data.
+
+### Changes proposed
+
+This proposal leads to three change requests that will be carried out in
+separate places in the OpenTelemetry specification.  These are:
+
+1. For tracing, the SpanData message type should be extended with
```
seh, line 93:
```suggestion
1. _For tracing:_ The SpanData message type should be extended with 
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+over the data.
+
+### Changes proposed
+
+This proposal leads to three change requests that will be carried out in
+separate places in the OpenTelemetry specification.  These are:
+
+1. For tracing, the SpanData message type should be extended with 
+   the `sample_count` field defined above.
+2. For metrics aggregate data: Count information aggregated from
```
seh, line 95:
```suggestion
2. _For metrics aggregate data:_ Count information aggregated from
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+   sample metric events will have floating point values in general.
+   Histogram and Counter data must be able to support floating point
+   values.
+3. For metrics raw events: Exemplars should be extended with the
```
seh, line 99:
```suggestion
3. _For metrics raw events:_ Exemplars should be extended with the 
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+
+### Example: Dapper tracing
+
+Google's [Dapper](https://research.google/pubs/pub36356/) tracing
+system describes the use of sampling to control the cost of trace
+collection at scale.
+
+### Example: Statsd metrics
+
+A Statsd counter event appears as a line of text, for example a metric
```
seh, line 110:
```suggestion
A Statsd counter event appears as a line of text.

For example, a metric
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+Probability 0.1 leads to a `sample_count` of 10.  Assuming the sample
+was selected using an unbiased algorithm (as by a "fair coin"), we can
+interpret this event as probabilistically equal to a count of `100/0.1
+= 1000`.
+
+## Internal details
+
+The statistical foundation of this technique is known as the
+Horvitz-Thompson estimator ("A Generalization of Sampling Without
+Replacement From a Finite Universe", JSTOR 1952).  The
```
seh, line 134:
```suggestion
Replacement From a Finite Universe," JSTOR 1952).  The
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+true population totals.  This means that statistics derived from
+sample data in this way are always accurate, and that more data will
+always improve precision.
+
+### Non-probabilistic rate-limiters
+
+Rate-limiting stages in a telemetry collection pipeline interfere with
+sampling schemes when they operate in non-probabilistic ways.  When
+implementing a non-probabilistic form of rate-limiting, processors
+MUST set the `sample_count` to zero.
```
seh, line 161: It would help to present a scenario here where _not_ setting `sample_count` to zero predicts the wrong population value. It's not immediately obvious to the reader why rate limiting invalidates the estimate.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+
+### No Sampler configured
+
+When no Sampler is in place and all telemetry events pass to the
+output, the `sample_count` field SHOULD be set to 1 to indicate
+perfect representivity, indicating that no sampling was performed.
```
seh, line 167:
```suggestion
perfect representativity, indicating that no sampling was performed and that all events were preserved.
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+
+### Applicability for tracing
+
+The sampling approach covered here dictates how to select root spans
+in a probabilistic way.  When recording root spans, the `sample_count`
+field should be set as described above.  The adjusted `sample_count`
+of the root span applies the trace, meaning the trace should be
```
seh, line 174:
```suggestion
of the root span applies to the entire trace, meaning the trace should be
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+timeseries, it is possible to defer decisions about cardinality
+reduction to the server, without unreasonable cost to the client.
+
+## Prior art and alternatives
+
+The name `sample_count` is proposed because the resulting value is
+effectively a count and may be used in place of the exact count.
+
+Statsd conveys inclusion probability instead of `sample_count`, where
+it is often called "sample rate".
```
seh, line 197:
```suggestion
it is often called "sample rate."
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+
+Another name for the proposed `sample_count` field is
+`inverse_probability`, which is considered less suggestive of the
+field's purpose.
```
seh, line 201: I find the latter to be clearer within the topic of sampling, though not fit for purpose here. I would prefer `sampling_factor` or `compression_factor`, because it's more general than "count"—in that it need not be integral—and it suggests multiplying by it to find another value.

oertl:
Maybe we should also consider allowing only discrete sampling rates. I'm thinking in particular of powers of 1/2 (1/2^0, 1/2^1, 1/2^2,1/2^3,...). In my opinion, this is not a big limitation in practice, but would have some really nice benefits:
1) A single byte for the exponent would be sufficient to represent the sampling rate.
2) The extrapolation factor (`sample_count`) would always be an integer. When extrapolating integer quantities such as counts, floating-point operations could be completely avoided and the estimate would be an integer as well.
3) Trace ID ratio based sampling would be much simpler. The sampling decision could be made simply based on the number of leading zeros (NLZ) of the trace ID. If the NLZ is equal to or greater than the exponent of the sampling rate, the span would be sampled. Only a few cheap CPU operations would be necessary for a sampling decision. Furthermore, by avoiding floating-point operations, sampling decisions would become more consistent across different platforms.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+application considered here, we need an estimator for the population
+total represented by each individual event.
+
+### Sample count adjustment
+
+The estimated population total divided by the individual event value
+equals the event's _adjusted count_.  The adjusted count is named
+`sample_count` where it appears as a field in a telemetry data type.
+The field indicates that the individual sample event is estimated to
+represent `sample_count` number of identical events in the population.
```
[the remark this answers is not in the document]
jmacd: As for whether "count" implies an integer or not, I'm less sure. I get that idea, but I also think in terms of fractional counts quite easily. My odometer reads in 1/10ths of a kilometer. If we restrict this field to an integer, conceptually, we force the user to introduce bias or limit the kinds of sampling algorithm they can use. 

> Do we really intend to say that a positive sample_count value implies that there are similar-as-can-be duplicates that were elided, or just that this sample somehow represents a larger population?

Yes. Yes! It may sound like magic, but the statistical principles referenced here (in my not-a-real-mathematician-or-statistician kind of way) justify this interpretation.

I can see how a feeling of uneasiness creeps in. In your example, there were 35 events and we recorded only four of them. Suppose we had used simple 1-in-7 sampling but only 4 events were selected. The expected number of items in the sample was 5, but we only turned up 4. In this case we would estimate a total count of 28 for that period, which is off by 7. This is a form of variance: also known as sampling error.

There's a silver lining to this. When sampling is unbiased, sampling errors cancel out when estimating the mean. When we combine enough data, all the under-counting and over-counting averages out. For example, although in one period your 1-in-7 sampling scheme turns up 4 items, other periods will turn up 5 and 6 items, and over a long enough period of time, the sum of sample_count is expected to equal the sum of the population count. The same can be said when examining subsets of the data (see the [paper](https://dl.acm.org/doi/abs/10.1145/1314690.1314696) referenced above).

text/0148-sampling-probability.md
jmacd: I don't see these two (not exemplary and elided) as any different. The point I was trying to establish is that when sampling is in effect, and when the `sample_count` field is being interpreted from a stream of events, the value of zero is meaningful and necessary to ensure that the sum of sample_count equals the expected population count if we are to record a span that was not sampled.

A span with a count of zero should be considered a span that happened once but is not counted. You might use that span to complete another trace (rooted somewhere else), or you might use that span as an "auxiliary variable" (a term from Survey Analysis).

The specific use-case I have in mind when writing this stems from the weighted sampling algorithms mentioned here and above. If we want a sampling algorithm to compute a fixed size sample, when the number of events that will happen during an interval is unknown, there will be a chance that some item is first selected for a sample and later rejected from a sample (see [here](https://dl.acm.org/doi/10.1145/3147.3165)).

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+
+Consider a hypothetical telemetry signal in which an API event
+produces a unit of data that has one or more associated numbers.
+Using the OpenTelemetry Metrics data model terminolgy, we have two
+scenarios in which sampling is common.
+
+1. Counter events: Each event represents a count, signifying the change in a sum
+2. Histogram events: Each event represents an individual variable, signifying new membership in a distribution
+
+A Tracing Span event qualifies as both of these cases simultaneously.  It is
```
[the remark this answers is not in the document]
jmacd: I added an example using request bytes.

text/0148-sampling-probability.md
jmacd: Considering all the feedback on this OTEP, I removed the recommendation of a specifically named field ("sample_count") in favor of a specific term used to describe the concept, "adjusted count". I reason that this is a count, not a weight, because it tells you how many times to count the event.

text/0148-sampling-probability.md
jmacd: Added a section on the use of zero adjusted count.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,209 @@
...
+
+### No Sampler configured
+
+When no Sampler is in place and all telemetry events pass to the
+output, the `sample_count` field SHOULD be set to 1 to indicate
+perfect representivity, indicating that no sampling was performed.
+
+### Applicability for tracing
+
+The sampling approach covered here dictates how to select root spans
```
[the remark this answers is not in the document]
jmacd: I removed this section "applicability for tracing". I've made an example above for one way to use "adjusted count" or inverse inclusion probability in a tracing system, but this OTEP has exposed that there are many reasonable ways to do that. I believe another OTEP is needed specifically to address the use of sampling in a tracing system. This OTEP goes far enough to cover the metrics use-cases that I wanted covered.

text/0148-sampling-probability.md
jmacd: I dropped it!

text/0148-sampling-probability.md
jmacd: This thread led me to remove a bunch of text about tracing specifics. I understand now that there are several ways to convey sampling information when collecting traces. I am now giving one example of how this can be used in a tracing system where only roots are sampled. I've also expanded the text about Dapper by a bit to explain this thread, somewhat.

text/0148-sampling-probability.md
jmacd: I've removed this name from my proposal. I'm using the term "adjusted count" to explain the concept, and would like you to review and see how this reads now.

jmacd:
Please take another look. @oertl I included your recipe for encoding inclusion probability as a good option. This OTEP no longer makes a specific recommendation about how to encode this information, only that when this information is encoded we know what it means.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,291 @@
...
+- sum the adjusted counts of the subset to estimate the sub-population total.
+
+Applied correctly, this approach provides accurate estimates for
+population counts and distributions with support for ad-hoc queries
+over the data.
+
+### Changes proposed
+
+This OTEP proposes no formal changes in the OpenTelemetry
+specitication.  It is meant to lay a foundation for importing sampled
```
pyohannes, line 189:
```suggestion
specification.  It is meant to lay a foundation for importing sampled
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,291 @@
...
+Consider a hypothetical telemetry signal in which an API event
+produces a unit of data that has one or more associated numbers.
+Using the OpenTelemetry Metrics data model terminology, we have two
+scenarios in which sampling is common.
+
+1. _Counter events:_ Each event represents a count, signifying the change in a sum.
+2. _Histogram events:_ Each event represents an individual variable, signifying new membership in a distribution.
+
+A Tracing Span event qualifies as both of these cases simultaneously.
+It is a Counter event (of 1 span) and at lesat one Histogram event
```
oertl, line 27: typo "lesat"

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,291 @@
...
+### Encoding inclusion probability
+
+Some possibilities for encoding the inclusion probability, depending
+on the circumstances and the protocol, briefly discussed:
+
+1. Encode the adjusted count directly as a floating point or integer number in the range [0, +Inf).  This is a conceptually easy way to understand sampling because larger numbers mean greater representivity.
+2. Encode the inclusion probability directly as a floating point number in the range [0, 1).  This is typical of the Statsd format, where each line includes an optional probability.  In this context, the probability is commonly referred to as a "sampling rate".  In this case, smaller numbers mean greater representivity.
+3. Encode the negative of the base-2 logarithm of inclusion probability.  This restricts inclusion probabilities to powers of two and allows the use of small non-negative integers to encode power-of-two adjusted counts.
+4. Fold the adjusted count into the data.  This is appropriate when the data itself carries counts, such as for OTLP Metrics Sum and Histogram points encoded using delta aggregation temporality.  This may lead to rounding errors, when adjusted counts are not integer valued.
+
```
oertl, line 122: Maybe mention that the 4th possibility is different from the first 3 as you loose information about the variance. Without additional information, you no longer know if the count is exact or represents an estimate, and if the sampling rate was the same for all increments or not.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,291 @@
...
+Consider a span sampling design that applies a sampling decision only
+to the roots of a trace.  Non-root spans must be recorded when their
+parent span is selected for a sample.  For a class of spans that
+sometimes are and sometimes are not the root of a trace, we have three
+outcomes:
+
+1. Span is part of another trace: adjusted count is zero
+2. Span is a root, was selected for the sample: adjusted count is non-zero
+3. Span is a root, was not selected for the sample: not recorded.
+
```
oertl, line 152: What I would expect, when aggregating the _adjusted counts_ of all spans of one particular service, is an unbiased estimate of the true number of requests to that service. Unfortunately this is not the case, if there are non-root spans for that particular service, because the sampling decision is made at root spans, and the corresponding sample rates and hence _adjusted counts_ are not known for non-root spans. Consequently, a zero _adjusted count_ would just indicate that the _adjusted count_ is not known and needs to be retrieved from the root span. However, it does not necessarily mean that the span does not contribute to the total count at all.

Therefore, I would prefer to propagate the sample rate that was used for some root span to its descendants, such that they know the sample rate and are able to set the _adjusted count_ properly.

text/0148-sampling-probability.md
jmacd: Done.

text/0148-sampling-probability.md
jmacd: I was trying to illustrate an example, not to suggest a real tracing Sampler, however I can see that no one is interested in this case, so it's not a good example.

I removed these paragraphs. Instead, I added more depth to the description of the Dapper system, and how it propagated the parent's inclusion probability, particularly when it made a negative sampling decision, so the child could calculate the conditional probability needed to ensure an overall rate of its own traces. I sense there is more interested in this approach, now that tracing has reached GA.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,376 @@
...
+In tracing, metrics, and logs, there are widely known techniques for
+sampling a stream of events that, when performed correctly, enable
+collecting a fraction of the complete data while maintaining
+substantial visibility into the whole population of events.
+
+These techniques are all forms of approximate counting.  Estimates
+calculated by the forms of sampling outlined here are considered
+accurate, in that they are random variables with an expected value
+equal to the true count.  With enough observations,
```
pirgeo, line 15:
```suggestion
equal to the true count.
```

This seems to be dangling here, is it missing somewhere down below?

jmacd: Removed. Thanks.

jmacd:
@paulosman please take another look?

carlosalberto:
Approving after reading these PR a pair of times (plus reading related documents, etc).

carlosalberto (approved):

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+Google's [Dapper](https://research.google/pubs/pub36356/) tracing
+system describes the use of sampling to control the cost of trace
+collection at scale.  Dapper's early Sampler algorithm, referred to as
+an "inflationary" approach (although not published in the paper), is
+reproduced here.
+
+This kind of Sampler allows non-root spans in a trace to raise the
+probability of tracing, using a conditional probability formula shown
+below.  Traces produced in this way are complete sub-trees, not
+necessarily complete.  This technique is succesful especially in
```
paulosman, line 485:
```suggestion
necessarily complete.  This technique is successful especially in
```

paulosman (approved):
Made comments fixing small typos that I caught while reading and added a question about floating-point adjusted count values, but overall, this looks great!

From a vendor perspective, this would allow us to support deterministic samplers that we did not write ourselves - at the moment we have to provide sampler plugins that include "Sample Rate" (basically adjusted count) as an attribute in the returned `SamplingResult`.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,726 @@
...
+There are several ways of encoding this information:
+
+- as a dedicated field in an OTLP protobuf message
+- as a non-descriptive Attribute in an OTLP Span, Metric, or Log
+- without any dedicated field.
+
+#### Encoding adjusted count
+
+We can encode the adjusted count directly as a floating point or
+integer number in the range [0, +Inf).  This is a conceptually easy
```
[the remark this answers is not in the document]
jmacd: A uniform sample of 40% has adjusted count of 2.5. To me this, makes sense, but I realize that it can present problems in some systems. This is mentioned in the example about computing Histogram metrics from spans, because OTLP Histogram points do not have a way to represent non-integer counts.  One of the reasons I like using an attribute to reflect adjusted count, not multiplying these values into the data itself, is that it allows the use of floating-point counts even where values are integers; OTOH, it punts the problem for the next consumer. I am happy to recommend the use of integer-reciprocal sampling rates in situations where this is a real problem.

jmacd:
Reviewers please take a look at this PR if you are interested in probability sampling and/or counting spans. 
Particularly, @yurishkuro I am interested in your take on the problems mentioned with trace sampling. If we use TraceIDRatio sampling, we do not know when traces are complete. If we use Parent sampling, we can know traces are comp,lete, but we have to propagate inclusion probability in order to count spans (if we are to count spans before assembling complete traces). This document proposes that we propagate inclusion probability, which still doesn't help with the TraceIDRatio completeness problem.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+(where the adjusted count is zero), therefore we prefer to use the
+inclusion probability and not the adjusted count when propagating the
+sampling rate via trace context.  The inclusion probability of a
+context is referred to as its `head inclusion probability` for this
+reason.
+
+In addition to propagating head inclusion probability, to count
+Parent-sampled spans, each span must directly encode its adjusted
+count in the corresponding `SpanData`.  This may use a non-descriptive
+Resource or Span attribute named `sampling.parent.adjusted_count`, for
```
yurishkuro, line 439:
```suggestion
Span attribute named `sampling.parent.adjusted_count`, for
```
Resource is supposed to be immutable

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+events when they provide useful information despite their effective
+count; we can use this to combine multiple sampling schemes in a
+single stream.
+
+For example, consider collecting two samples from a stream of spans,
+the first sample from those spans with attribute `A=1` and the second
+sample from those spans with attribute `B=2`.  Any span that has both
+of these properties is eligible to be selected for both samples, in
+which case one event could have two non-zero adjusted counts (e.g.,
+`sampling.by_a.adjusted_count` and `sampling.by_b.adjusted_count`).
```
yurishkuro, line 613: Another situation that gives rise to multiple weights is when different policies compete for the same trace. Eg the root service may have multiple sampling policies attached that have different purposes, with different sampling probabilities (and potentially different verbosity of perf data collection). When more than one policy "fires" they "share the body" of a single trace.

Anyway, it's not clear what is suggested here. If two weights are recorded, what happens to the suggestion of propagating the weight with the context downstream?

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+
+## Proposed Tracing specification
+
+For the standard OpenTelemetry Span Sampler implementations to support
+a range of probability sampling schemes, this document recommends the
+use of a Span attribute named `sampling.sampler_name.adjusted_count`
+to encode an unbiased adjusted count computed by a Sampler.
```
yurishkuro, line 620: What about situations where sampling is either biased or is subject to after-sampling rate limiting?

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+Samplers](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#built-in-samplers),
+a new section will introduce the term "adjusted count" and relate it
+to inclusion probability.  For example:
+
+```markdown
+### Probability sampling
+
+Probability Samplers are Samplers that output statistically unbiased
+inclusion probability.  Inclusion probability in the context of
+tracing is the *effective* probability of the Sampler returning
```
yurishkuro, line 641: The meaning of "Effective" is unclear

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+probability sampling scheme SHOULD use a Span attribute name with
+prefix `sampling.` and with suffix `.adjusted_count` (e.g.,
+`sampling.sampler_name.adjusted_count`).  Adjusted count attributes
+MAY be integer or floating-point values.
+
+#### Inclusion probability tracestate value
+
+The recommended way to convey sampling probability *for contexts* in
+OpenTelemetry is through the W3C Trace Context tracestate using the
+key `head_probability`.
```
yurishkuro, line 672: I think this is a common misunderstanding of tracestate header: it is not meant for arbitrary keys that have shared semantics. It is meant for vendor data.

What would be a valid use of tracestate is something like otel=headprob:0.123

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+
+For each span processed, locate all span attributes with prefix
+`sampling.` and suffix `.adjusted_count`.  If there are no adjusted
+count attributes on the span, count a single event.  If there is
+exactly one adjusted count attribute, count that many identical span
+events.
+
+If there are more than one adjusted count attribute on the span, the
+processor SHOULD make a consistent choice for spans of a given
+resource.  By default, the SDK's first preference MUST the builtin
```
yurishkuro, line 700: a) why this specific priority?
b) this section describes downstream processor, why the reference to SDK?

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+of adjusted counts in the sample equals the true count of spans in the
+population.
+
+Attributes used to express the adjusted count in an unbiased
+probability sampling scheme SHOULD use a Span attribute name with
+prefix `sampling.` and with suffix `.adjusted_count` (e.g.,
+`sampling.sampler_name.adjusted_count`).  Adjusted count attributes
+MAY be integer or floating-point values.
+
+#### Inclusion probability tracestate value
```
yurishkuro, line 668: I think this should be optional

yurishkuro (commented):
Very good read. One part that is missing is a discussion/rationalization of using root span's weight for estimation of downstream services stats. Intuitively it still works, but is there any theory to back it up? And there is a case where this estimation does not work for downstream service if it's getting traffic that does not follow the sampling rules outlined here (eg from uninstrumented service or some biased form of sampling).

oertl:
Knowing the adjusted count for each span allows you to at least extrapolate individual spans. Things get more complicated, however, if you want to count the number of traces that called a service A AND a service B, as an example. Let's assume that A and B are in different branches of the same trace, so that neither one of them is the parent of the other. Depending on the sampling strategies chosen, up to 4 cases are possible.

1. Neither A nor B are sampled.
2. Only A is sampled.
3. Only B is sampled.
4. A and B are both sampled.

Obviously, only the last case is able to contribute to the estimate, because only in this case we can be sure that both services have been called. The question now is what is the adjusted number for this trace so that we get an unbiased estimate?

Assume that w_A and w_B denote the adjusted counts for spans of A and B, respectively. Depending on the samplers, the sampling decision of A and B may or may not be correlated. If the sampling decision is independent, the adjusted count of the trace would be w_A*w_B, and if spans A and B are sampled consistently using trace ID ratio based sampling, the correct adjusted count of the trace would be max(w_A, w_B). 

This example shows that it is not sufficient to know only the adjusted counts of the spans. It is important that the sampling process is also either known or well defined. In my opinion, the simplest solution is to allow only trace ID ratio based sampling and nothing else. Head-based sampling can actually be realized by propagating the sample rate (instead of a flag indicating the sampling decision) so that child spans can use the same sample rate as the parent. If the same sample rate is used for all spans, trace ID ratio based sampling will ensure that either all spans are sampled or nothing is sampled, as with head-based sampling.

One could think about redefining the sampler interface. Instead of returning the sampling decision, samplers should only be responsible for returning a sample rate. The sampling decision should then always be made by comparing the sample rate to the trace ID (as trace ID ratio-based sampling does) or alternatively to a sampling score (if the trace ID is not uniformly distributed, compare https://github.com/open-telemetry/oteps/pull/135). In this way, samplers are forced to provide a sample rate, and thus the adjusted counts are always known for the spans as needed for extrapolation. In addition, the sampling process would always be well-defined, making it possible to extrapolate traces rather than just spans.

text/0148-sampling-probability.md
ericmustin: Can you clarify @yurishkuro ? It seems essential to me to convey the probability in _some_ form when exporting   to consumers of span data (like the collector  and it's various exporters)? Otherwise how can those exporters know how to upscale any statistics or metrics?

yurishkuro: "essential to me" is the key part. This whole process (such as generating metrics on the fly from spans) may not be essential to some users, so the behavior should be opt-in.

ericmustin: > "some". 

I think _a majority_ of users would find this useful. It's not _just_ on the fly metric generation. Including the sampling rate _somewhere_ in the exported payload means providing users a more complete picture of the sampling decision that occurred at the time in the service/application/etc. Not doing that means we're giving them an incomplete picture. Doing on the fly metric generation is just one way that can then be used down the line. 

Could you give an example of a user who would prefer the tradeoff off off-by-default incomplete/missing sampling information for the benefit of not having to include/compute the extra key/value pair in their span data? I'm genuinely not trying to be snarky here I just am not sure I understand the "user story", so to speak. The use case I can think of (web-scale/perf costs/reduce payload size/etc) feels better served by `on-by-default, optional opt-out`.

Obviously I don't have the depth of knowledge you do here, but I think of sampling as a 1st class citizen in _tracing_ and what separates it from just structured logging, and so we should aim for as rich a sampling picture as is reasonable.

Anyway, hope that helps clarify what "essential to me" means.

yurishkuro: Jaeger, Zipkin, OpenCensus SDKs do not propagate sampling weight. W3C Trace Context does not include the weight as a required part of the context. I think this provides enough evidence that this feature is not a requirement for a functioning tracing system.

I am completely on board with capturing the weight as a span attribute - all Jaeger SDKs do that, and any post-collection processing that hopes to build any aggregate data from traces would need that. But in-band propagation is not required for most use cases.

ericmustin: yup, per usual i'm an idiot and misread the previous section. plain ole span attribute works just fine for me, in-band (traceState) far less important and i don't have strong opinions here. txs for the patience.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+For every span it receives, the example processor will synthesize
+metric data as though a Histogram instrument named `S.duration` for
+span named `S` had been observed once per span at the original `End()`
+call site.
+
+The OpenTelemetry Metric data model does not support histogram buckets
+with non-integer counts, which forces the use of integer adjusted
+counts here (i.e., 1-in-N sampling rates where N is an integer).
+
+Logically spaking, this processor will observe the span's duration its
```
richardstartin, line 110: Typo

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+attribute named like `sampling.sampler_name.adjusted_count`.
+Consumers of spans, metrics, and logs annotated with adjusted counts
+are able to calculate accurate statistics about the population of
+events, at a basic level, without knowing details about the sampling
+configuration.
+
+### Span sampling
+
+<details> 
+<summary> Example use-cases for probability sampling of spans
```
reyang, line 85:
```suggestion
<summary>Example use-cases for probability sampling of spans
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+generally involve generating metrics from spans.</summary>
+
+#### Sample spans to Counter Metric
+
+For every span it receives, the example processor will synthesize
+metric data as though a Counter instrument named `S.count` for span
+named `S` had been incremented once per span at the original `Start()`
```
reyang, line 92: One potential challenge here is that the span `Start()` will not have access to all the attributes (e.g. some of the attributes might be added later).

text/0148-sampling-probability.md
reyang:
```suggestion
Logically speaking, this processor will observe the span's duration its
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+
+Logically spaking, this processor will observe the span's duration its
+adjusted count number of times for every span it receives, at the end
+time of the span.
+
+#### Sample span rate limiting
+
+A collector processor will introduce a slight delay in order to ensure
+it has received a complete frame of data, during which time it
+maintains a fixed-size buffer of input spans.  If the number of spans
```
reyang, line 118: Will this buffer only contain finished spans, or it can have spans that are started but not finished?

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+Events in the example have with 0.1 inclusion probability have
+adjusted count of 10.  Assuming the sample was selected using an
+unbiased algorithm, we can interpret this event as having an expected
+count of `100/0.1 = 1000`.
+
+#### Metric exemplars with adjusted counts
+
+The OTLP protocol for metrics includes a repeated exemplars field in
+every data point.  This is a place where histogram implementations are
+able to provide example context to corrlate metrics with traces.
```
reyang, line 167:
```suggestion
able to provide example context to correlate metrics with traces.
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+
+Consider a hypothetical telemetry signal in which an API event
+produces a unit of data that has one or more associated numbers.
+Using the OpenTelemetry Metrics data model terminology, we have two
+scenarios in which sampling is common.
+
+1. _Counter events:_ Each event represents a count, signifying the change in a sum.
+2. _Histogram events:_ Each event represents an individual variable, signifying membership in a distribution.
+
+A Tracing Span event qualifies as both of these cases simultaneously.
```
reyang, line 208: This is awesome! 💯

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+
+After executing a sampling design over a frame, each item selected in
+the sample will have known _inclusion probability_, that determines
+how likely the item was to being selected.  Implicitly, all the items
+that were not selected for the sample have zero inclusion probability.
+
+Descriptive words that are often used to describe sampling designs:
+
+- *Fixed*: the sampling design is the same from one frame to the next
+- *Adaptive*: the sampling design changes from one frame to the next
```
reyang, line 241: I understand that these are hard to define.
Just curious if we want to consider something like "Adaptive: the sampling design changes from one frame to the next based on the input values". And here goes the reason:

If the sampling design changes dynamically based on the configuration from the operator, it sounds like it is "not fixed" but also "not adaptive".

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+together across time.  For example, although the estimates drawn from
+a one-minute sample may have high variance, combining an hour of
+one-minute sample frames into an aggregate data set is guaranteed to
+lower variance.  It must, because the data remains unbiased.
+
+### Conveying the sampling probability
+
+Some possibilities for encoding the adjusted count or inclusion
+probability are discussed below, depending on the circumstances and
+the protocol.
```
reyang, line 330: Curious, is the protocol here specific to the exporting protocol, or it covers the RPC protocol where service A propagates the sampling decision+design to service B, so service B could provide consistent sampling decision instead of making its individual sampling decision?

It looks like we're only talking about the exporting protocol here.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+representivity.
+
+#### Multiply the adjusted count into the data
+
+When the data itself carries counts, such as for the Metrics Sum and
+Histogram points.
+
+This technique is less desirable because while it preserves the
+expected value of the count or sum, the data loses information about
+variance.  This may also lead to rounding errors, when adjusted counts
```
reyang, line 368: +1, all the folks that I know + who made such choice, regret for it. This is especially painful when the metrics are used for alerting, the information loss about variance makes it hard to tell the confidence level of the alert.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,744 @@
...
+resource.  By default, the SDK's first preference MUST the builtin
+`TraceIDRatio` sampler, and its second preference MUST be the builtin
+`Parent` sampler.
+```
+
+For the `TraceIDRatio` sampler, include the following additional text:
+
+```md
+When returning a `RECORD_AND_SAMPLE` decision, the TraceIDRatio
+Sampler MUST include the attribute
```
reyang, line 709: Would this be a breaking change for the spec (e.g. we will have to bump the spec version to 2.x)?

carlosalberto: If it's because of the addition of `sampling.traceidratio.adjusted_count` attribute, I'd say we are fine as that's more related to semantic conventions (and we already 'handle' that part).

text/0148-sampling-probability.md
jmacd: I see now the consequences of moving the examples above the introductory text, some missing bits. I've added "complete" here to avoid the ambiguity. I'm trying to stick the OTel tracing data model, where spans encoded in OTLP are already finished.

text/0148-sampling-probability.md
jmacd: 😀

text/0148-sampling-probability.md
jmacd: I added "based on the observed data".

text/0148-sampling-probability.md
jmacd: I added ```Here, the focus is on how to count sampled telemetry
+events in general, not a specific kind of event.  As we shall see in
+the following section, tracing comes with its addional complications.```

text/0148-sampling-probability.md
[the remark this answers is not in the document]
jmacd: I added a paragraph to state the above. ^^^

text/0148-sampling-probability.md
jmacd: :+1:

text/0148-sampling-probability.md
jmacd: I added "finished" to describe these spans.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,805 @@
...
+
+Events in the example have with 0.1 inclusion probability have
+adjusted count of 10.  Assuming the sample was selected using an
+unbiased algorithm, we can interpret this event as having an expected
+count of `100/0.1 = 1000`.
+
+#### Metric exemplars with adjusted counts
+
+The OTLP protocol for metrics includes a repeated exemplars field in
+every data point.  This is a place where histogram implementations are
```
yurishkuro, line 166: Do exemplars only apply to histograms?

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,805 @@
...
+
+Considering data points received during the interval, when the number
+of points exceeds K, select a probability proportional to size sample
+of points, output every point with an additional (non-descriptive)
+`sampling.cardinality_limit.adjusted_count` attribute.
+</details>
+
+## Explanation
+
+Consider a hypothetical telemetry signal in which an API event
```
yurishkuro, line 200: What is an API event?

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,805 @@
...
+### Model and terminology
+
+This model is meant to apply in telemetry collection situations where
+individual events at an API boundary are sampled for collection.  Once
+the process of sampling individual API-level events is understood, we
+will learn to apply these techniques for sampling aggregated data.
+
+In sampling, the term _sampling design_ refers to how sampling
+probability is decided and the term _sample frame_ refers to how
+events are organized into discrete populations.
```
yurishkuro, line 228: Unclear what sample frame is. Is it a time interval? That's how it seems to be used in examples (eg "over a frame"), except for the next paragraph.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,805 @@
...
+
+The use of unbiased sampling outlined above makes it possible to
+estimate the population total for arbitrary subsets of the sample, as
+every individual sample has been independently assigned an adjusted
+count.
+
+There is a natural relationship between statistical bias and variance.
+Approximate counting comes with variance, a matter of fact which can
+be controlled for by the sample size.  Variance is unavoidable in an
+unbiased sample, but it vanishes when you have enough data.
```
yurishkuro, line 314: This sounds like there's some magic threshold for sample size. Maybe rephrase that variance diminishes with growing sample size.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,805 @@
...
+
+Although this makes it sound like small sample sizes are a problem,
+due to expected high variance, this is just a limitation of the
+technique.  When variance is high, use a larger sample size.
+
+An easy approach for lowering variance is to aggregate sample frames
```
yurishkuro, line 320: Not following this. The way to decrease variance is to increase the ratio of sample size vs population size. Aggregating 1m to 1hr does not change that ratio.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,805 @@
...
+Tracer overhead.  Tracer overhead is reduced by not recording spans
+for unsampled traces and requires making the sampling decision for a
+trace before all of its attributes are known.
+
+Traces are expected to be complete, meaning that a tree or sub-tree of
+spans branching from a certain root are expected to be fully
+collected.  When sampling is applied to reduce Tracer overhead, there
+is generally an expectation that complete traces will still be
+produced.  Sampling techniques that lower Tracer overhead and produce
+complete traces are known as _Head trace sampling_ techniques.
```
yurishkuro, line 394:
```suggestion
complete traces are known as _Head-based trace sampling_ techniques.
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,805 @@
...
+the root span starts, to avoid incomplete traces.  Then, assuming
+complete traces can be collected, the adjusted count of the root span
+determines an adjusted count for every span in the trace.
+
+#### Counting child spans using root span adjusted counts
+
+The adjusted count of a root span determines the adjusted count of
+each of its children based on the following logic:
+
+- The root span is considered representative of _adjusted count_ many
```
yurishkuro, line 406: The phrasing tripped me a bit, I would use adjusted_count (fixed width font) to indicate that we're referring to a single variable, which space separated italic does not convey.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,805 @@
...
+sample from those spans with attribute `B=2`.  Any span that has both
+of these properties is eligible to be selected for both samples, in
+which case one event could have two non-zero adjusted counts (e.g.,
+`sampling.by_a.adjusted_count` and `sampling.by_b.adjusted_count`).
+
+## Proposed Tracing specification
+
+For the standard OpenTelemetry Span Sampler implementations to support
+a range of probability sampling schemes, this document recommends the
+use of a Span attribute named `sampling.sampler_name.adjusted_count`
```
yurishkuro, line 680: I would prefer a different naming scheme. Trace processing pipeline does not know which sampler is used for each individual span, and in most cases it doesn't care either as long as the adjusted count is accurate. But since the attribute name includes the sampler name, the processor has no way of retrieving the count via O(1) lookup, it must iterate through all attributes with prefix match.

text/0148-sampling-probability.md
jmacd: I threw this paragraph away and have removed all references to multiple adjusted counts.

I was trying to solve a problem when more than one framing is being used, but I see this was a mistake.

The key property is that if we expect to count all spans, we want to ensure the framing refers to all spans, so I've written that `sampling.adjusted_count` SHOULD refer to the whole population of spans.

text/0148-sampling-probability.md
jmacd: Having laid out the argument for propagating head inclusion probability, I can agree we should all be on the fence about it.

If the Parent sampler will be used and head inclusion probability is not propagated when the is-sampled bit is set, we will have a difficult time computing metrics from spans. However, the probability is only needed when is-sampled is true, so by turning on sampling the cost of propagating the sampling rate is automatically lowered.

Only the Dapper "inflationary" approach requires propagating the inclusion probability for negative sampling decisions, and that approach is not one I'm trying to defend. Unfortunately, we are left with few options that meet the three requirements stated in this document, which are:

1. Reduces Tracer overhead
2. Produces complete traces
3. Spans are countable.

OpenCensus was trying to switch to the use of TraceIDRatio sampling by default, which satisfies 1 and 3 easily. OpenTelemetry has a problem with 2, but I'm not sure it has to be solved. Completeness is less important than spans-to-metrics support, from our perspective as a vendor.

jmacd:
@oertl 

> This example shows that it is not sufficient to know only the adjusted counts of the spans. 

If a common ancestor of A and B was sampled (whether using a consistent scheme, or if W3C is-sampled is used and sampling probability is propagated), then we have information about traces containing both A and B rooted at a particular ancestor. The adjusted count of any common ancestor can be used. I don't see a gap, I think I can use the adjusted count of any common ancestor regardless of how the sample was computed as long as it was unbiased.

jmacd:
I've simplified this based on @yurishkuro's feedback to only discuss how to standardize a single adjusted count per event, being the one understood to apply to the "whole population".

oertl:
@jmacd 
> If a common ancestor of A and B was sampled (whether using a consistent scheme, or if W3C is-sampled is used and sampling probability is propagated), then we have information about traces containing both A and B rooted at a particular ancestor. The adjusted count of any common ancestor can be used. I don't see a gap, I think I can use the adjusted count of any common ancestor regardless of how the sample was computed as long as it was unbiased.

If a common ancestor of A and B was not sampled and the two branches containing A and B are not connected, we can still infer that the trace called service A and service B based on the common trace ID. Unfortunately, this information cannot be used for estimation because we do not know how the different branches were sampled (in a correlated or uncorrelated way). If only traces, that called service A and service B and for which a common ancestor was also sampled, are used for estimation, we do not try to get the maximum information out of the sampled spans.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,796 @@
...
+which is the reciprocal (i.e., mathematical inverse function) of
+inclusion probability.
+
+The implied goal of probability sampling is to support estimating the
+count of spans in the population using the spans that were sampled.
+The adjusted count associated with a Span is the expected value of the
+number of identical Spans within the population that each Span
+represents.  Probability samplers SHOULD ensure the samples they
+compute are unbiased, which implies that the expected value of the sum
+of adjusted counts in the sample equals the true count of spans in the
```
yurishkuro, line 714:
```suggestion
compute are unbiased, which implies that the sum
of adjusted counts in the sample approximates the true count of spans in the
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,796 @@
...
+have different inclusion probabilities.
+
+#### Counting probabilistically sampled spans
+
+Consumers of a stream of span data that may or may not have been
+sampled can follow these steps to count or approximately count the
+total number of spans in the population.
+
+For each span processed, locate the `sampling.adjusted_count` attribute. 
+If there is none, count a single span event.  If the attribute is set,
```
yurishkuro, line 747: > If there is none, count a single span event.

I don't think this is correct, because spans inherit adjusted count from their parent

text/0148-sampling-probability.md
yurishkuro: still have the same two questions

text/0148-sampling-probability.md
yurishkuro: I would be reluctant to bump the version of traceparent just to capture this value, since this can be done without changing w3c spec via `tracestate: otel=headprob:0.123`.

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,796 @@
...
+when it represents the whole population of events.  Adjusted count 
+attributes MAY be integer or floating-point values.
+
+#### Inclusion probability tracestate value
+
+The recommended way to convey sampling probability *for contexts* in
+OpenTelemetry is through the W3C Trace Context tracestate using the
+key `head_probability`.
+
+Probability Samplers SHOULD encode the effective sampling inclusion
```
yurishkuro, line 728:
```suggestion
Probability Samplers MAY encode the effective sampling inclusion
```

text/0148-sampling-probability.md
jmacd: I see, thanks. :+1: Can we say "using the key `otel` and sub-key `headprob`" or something similar?

text/0148-sampling-probability.md
jmacd: There are several ways to interpret a count. We can count spans by the adjusted count of all the trace roots, or we can count spans by the adjusted count of all the spans independently. I aimed to talk about counting spans in their own right in this section, without referring to trace roots.

Spans may be counted using the adjusted count of the parent, in which case once in a while the parent's adjusted count comes through along with `is-sampled`. This also implies that much of the time `is-sampled` is not set and a zero adjusted count is assumed.

Spans may be counted by sampling at the call site independently, in which case if there is no sampling being applied, there will be no adjusted count set and the correct count is 1.

You are correct, I have failed to call out an error scenario, when `is-sampled` is set and the `otel=headprob:` value is unknown. I've written that the Parent Sampler MUST set `adjusted_count=0`, to signal that they must be counted using trace roots, not using span adjusted counts.

That way, when there is no sampling taking place, 100% of spans will be recorded and counted as one span each.

text/0148-sampling-probability.md
jmacd: I meant to remove this section, it's now removed.

text/0148-sampling-probability.md
jmacd: I replaced the MUST with MAY to signify this is optional support.

jmacd:
@oertl 

> we do not try to get the maximum information out of the sampled spans.

I take it we would need to record information about the Sampler in order to gain this information. I am hoping we can somehow solve the traces-may-be-incomplete problem when sampling before trying to solve this one, if I understand it correctly.

jmacd:
Lightstep has open-sourced its implementation of the VarOpt sampling algorithm, one of the algorithms listed in this OTEP. 

https://github.com/lightstep/varopt

VarOpt is a useful tool for building second-stage Tail samplers because they (1) accept unequal adjusted-count inputs and (2) produce fixed-size output. We wish to promote this as an alternative to the "leaky bucket" rate-limited sampler discussed in https://github.com/open-telemetry/opentelemetry-specification/issues/1769. I believe VarOpt and the Leaky-bucket samplers will have similar complexity, but VarOpt is able to output unbiased adjusted counts, making it possible to probabilistically count spans that make it through the Sampler.

oertl:
Dynatrace has published a paper on partial trace sampling with a focus on the unbiased estimation from incomplete trace data https://arxiv.org/abs/2107.07703. It provides arguments for limiting the sampling rate to powers of 1/2 (see section "2.8 Practical Considerations").

jmacd:
@yurishkuro Do you feel that we can merge this OTEP? I believe I addressed your concerns, and any remaining concerns or matters of opinion can be ironed out as we move on to update the specification. 

The recent sampling SIG meeting, summarized in https://github.com/open-telemetry/opentelemetry-specification/issues/1819, found little objection over the contents of this OTEP. We seem to have reached a consensus about the use of TraceIDRatio sampling.

jmacd:
@oertl Thank you for posting your research. Partial Sampling is a fantastic addition to the state of the art, and now I understand why you've been proposing power-of-two sampling rates. The "negative base-2 logarithm" topic is mentioned, and I can add more current information if that will help.

I take it you would like to see the specification support a directly-encoded span probability, using this approach. It would be an unsigned integer field in the protocol containing the base-2 logarithm of the adjusted count (i.e., the negative base-2 logarithm of the inclusion probability):

0: The span's adjusted count is 1
1: The span counts for 2
2: The span counts for 4
3: The span counts for 8

So, perhaps the `Span` field should be named `log2_adjusted_count`? I will support that as an optimization, if there is a general agreement. However, this field will not support tail sampling at arbitrary rates, which is an application with known potential and existing uses. In that sense, the proposed use of a span attribute ending in `adjusted_count` seems less contentious and more flexible (e.g., supports integer and non-integer adjusted counts).

text/0148-sampling-probability.md
yurishkuro: > Can we say "using the key otel and sub-key headprob" or something similar?

yes

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,794 @@
...
+
+#### Inclusion probability tracestate value
+
+The recommended way to convey sampling probability *for contexts* in
+OpenTelemetry is through the W3C Trace Context tracestate using the
+key `head_probability`.
+
+Probability Samplers MAY encode the effective sampling inclusion
+probability using tracestate, for the context that was in effect when
+the W3C is-sampled bit was set.  The tracestate field SHOULD be set in
```
yurishkuro, line 730: > The tracestate field SHOULD be set in

This needs to be qualified, so that SHOULD is interpreted conditionally, e.g.

> **If the inclusion probability is propagated**, it SHOULD be represented with subkey `headprob` within the field `otel`... For example:
```
tracestate: otel=headprob:0.1234
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,794 @@
...
+For the `Parent` sampler, include the following additional text:
+
+```md
+When returning a `RECORD_AND_SAMPLE` decision, the Parent Sampler MAY
+include the attribute `sampling.adjusted_count=C`, where `C` is the
+reciprocal of the parent trace context's head inclusion probability.
+
+The tracestate `otel` key with the sub-key `headprob` is used to lookup
+and propagate the configured  trace ID ratio.  When the `otel` key with 
+sub-key `headprob` is not located and `is-sampled` is set, the Sampler
```
yurishkuro, line 772: the tracestate seems to be described multiple times, why not do it once only?

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,794 @@
...
+
+The returned tracestate used for the child context MAY have the
+tracestate `otel` key with the sub-key `headprob` set to the configured 
+trace ID ratio.
+```
+
+For the `Parent` sampler, include the following additional text:
+
+```md
+When returning a `RECORD_AND_SAMPLE` decision, the Parent Sampler MAY
```
yurishkuro, line 766: a few things here:
  1. parent-present and parent-absent must be treated differently, the attribute should only be set when parent is present, otherwise defer to the fallback sampler
  2. MAY keyword for recording the attribute seems unnecessary weak, considering that ratio sampler has MUST for the same
  3. on the other hand, the parent probability propagation is optional, so it could be MUST but with a qualification

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,794 @@
...
+OpenTelemetry is through the W3C Trace Context tracestate using the
+key `head_probability`.
+
+Probability Samplers MAY encode the effective sampling inclusion
+probability using tracestate, for the context that was in effect when
+the W3C is-sampled bit was set.  The tracestate field SHOULD be set in
+both sampled and unsampled cases, to convey the inclusion probability
+even for unsampled contexts.
+
+The `head_probability` tracestate key is set to a floating-point
```
yurishkuro, line 734:
```suggestion
The `headprob` subkey is set to a floating-point
```

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,794 @@
...
+compute are unbiased, which implies that the sum
+of adjusted counts in the sample approximates the true count of spans in the
+population.
+
+The adjusted count in an unbiased probability sampling scheme SHOULD
+be expressed using a Span attribute named `sampling.adjusted_count`
+when it represents the whole population of events.  Adjusted count 
+attributes MAY be integer or floating-point values.
+
+#### Inclusion probability tracestate value
```
yurishkuro, line 722: I am not sure it is clear how the SDKs should behave given that the propagation is optional. Propagating probability is a clear additional overhead that some sites may not want to pay, especially if they don't use any inflationary or partial sampling. So how does an SDK know if it should or should not include the probability in the tracestate? Should it be a configuration of the SDK (tricky because guaranteeing identical configuration throughout all microservices is quite difficult)?

text/0148-sampling-probability.md
```diff
@@ -0,0 +1,794 @@
...
+
+```md
+When returning a `RECORD_AND_SAMPLE` decision, the Parent Sampler MAY
+include the attribute `sampling.adjusted_count=C`, where `C` is the
+reciprocal of the parent trace context's head inclusion probability.
+
+The tracestate `otel` key with the sub-key `headprob` is used to lookup
+and propagate the configured  trace ID ratio.  When the `otel` key with 
+sub-key `headprob` is not located and `is-sampled` is set, the Sampler 
+MUST set `sampling.adjusted_count=0` to signal that spans cannot be
```
yurishkuro, line 773: How would setting it to 0 be different from not setting at all?

If we look at the existing trace context formats like Zipkin's B3 or Jaeger's, they do not propagate probability and they use classic head-based sampling with sampling decision propagated. Extrapolation from individual spans is not possible with those, since there's no way of knowing probability or count of children spans. Extrapolation from re-assembled trace is possible as long as the count is captured in the root. I think whichever behavior we propose, it should degrade nicely into this legacy (and very widespread) mode.

yurishkuro (commented):
> @yurishkuro Do you feel that we can merge this OTEP?

So I have two concerns:
  1. The propagation story is not completely clear, see comments inline.
  2. There is only one approval so far. I think this topic needs a lot more attention, especially from different vendors who already had to deal with sampling.

jmacd:
@yurishkuro Thank you. I will address your questions with one more round of work on this PR, however I want to highlight that the result of this OTEP is largely negative on propagating probability, and while I have given a proposed/draft specification for it I do not expect we will move forward with a specification that involves propagating anything. We are left with a desire to complete the TraceIDRatio sampler (which is to achieve consistent pseudo-randomness) and (independently) to know when traces are complete.

I believe this OTEP has met the standard for an OTEP, discussed [here](https://github.com/open-telemetry/oteps#submitting-the-otep). This OTEP is now too long to review and another PR is needed. The next step will be, I think, another PR to revise this document, to focus specifically on the questions you've raised. Another piece of text is needed to fully document how to count spans based on the adjusted count that is recorded in the span.

> Extrapolation from re-assembled trace is possible as long as the count is captured in the root. 
> should degrade nicely into this legacy mode

I don't see any part of this proposal that would prevent legacy counting strategies. If you agree that this proposal needs more editing, please approve so we can merge it and open a new PR. Thanks!

yurishkuro:
> I believe this OTEP has met the standard for an OTEP, discussed [here](https://github.com/open-telemetry/oteps#submitting-the-otep).

Not sure I agree with that. The process we agreed for OTEPs is that we do not capture OTEP status in the document, but use GitHub status as a proxy. I.e. approved & merged PR means approved OTEP, which in turn means it is the official position for the project and only pending mechanical translation into Spec changes. The link you provided does not say that a PR can be merged with intention for revisions via another PR, it actually says the opposite that the old PR should be closed if revisions are needed. Otherwise we are left with officially looking document in OTEPs that does not reflect the agreements.

Concretely, if you do not believe that we should proceed with implementing probability propagation, then I would move that text into a Discussion area with pros/cons, and not include it in the normative portion that recommends the actual changes to the spec (which, incidentally, most of my comments are about since the proposed changes are inconsistent).

jmacd:
If I remove the entire proposed specification section of this document, would you approve? The goal was to outline our options, and I included the specification text as an example. It states ("For exampe:") before that section of text. It's the least interesting part of this document to me, what it really did was show us how we do not want to propagate probability. I'm interested in updating this OTEP with a minimal summary of what we concluded and merging it, not continuing to address minor points in what is ultimately not a specification document.

yurishkuro:
If you want to remove the proposed normative changes section, then it ceases to be an enhancement proposal (OT*EP*) imo. But I wouldn't mind merging it as it's a great read that can be referenced from other places.

jmacd:
I think I will prefer to close and re-open a new PR. I will leave this open until then.

jmacd:
@yurishkuro I wonder if we can salvage the bulk of this PR by removing the normative text that I had and specifying much less. My original goal with this OTEP was to specify a foundation, which is the basic idea that we can record an adjusted count when sampling to convey probability, and that adjusted count is a good way to convey that because users intuitively understand how to count but do not intuitively understand how to count inverses.

Thus, I've replaced all the text with a proposal for a new `trace/semantic_conventions/sampling.md` file with two attributes, `sampler.adjusted_count` and `sampler.name`. I think this is something we could all agree on, but it leaves a lot to be specified. It doesn't tell anyone how to modify the Samplers, but at least it would let users of OTLP who are not yet using OTel SDKs convey sampling information.

Originally I had proposed only an `adjusted_count` attribute, not the `name` attribute, but @oertl has shown why knowing which sampler computed the probability matters in some cases. Moreover, knowing the sampler name addresses an ambiguity that was discussed above: If the probability is not propagated then we cannot know the inclusion probability of a span written by the Parent sampler. We also should not presume the adjusted count is 1 in this case, so we have several possibilities:

Parent sampling, unknown probability:

```
sampler.name=Parent
```

Parent sampling, known probability (for 0.1 probability)

```
sampler.name=Parent
sampler.adjusted_count=10
```

TraceIDRatio sampling (for 0.1 probability)

```
sampler.name=TraceIDRatio
sampler.adjusted_count=1
```

And for no sampler, no attributes are needed. Spans carrying these attributes can be correctly counted, the only problem is for Parent sampling with unknown probability. At least now we have clearly identified when the adjusted count is missing. See the replacement text here: https://github.com/open-telemetry/oteps/pull/148/commits/c4c06cd58ceae3f01b4f46d9edadf099cd5ed395

I will take all the discussion about how to modify Samplers to produce these attributes as well as how to optionally propagate inclusion probability into another PR. Thanks.

jmacd:
> Why not always record sampler name?

The AlwaysOn sampler is (afaik) equivalent to no sampler, so I don't see any use for the name in that case. The adjusted count is 1 with or without the AlwaysOn sampler.

> avoid inventing new naming scheme here

Yes, I see this point and feel ambivalent about it.  I do not see any viable uses for the existing Description other than to configure a composite sampler in the SDK. It has the appearance of something that can be logged and parsed, but I wouldn't use it.

If the sampler description is "jaeger_remote", for example, it tells me nothing useful. The piece of information that is needed is not the composite policy that was configured as a Sampler, it is the effective "leaf" Sampler that was selected. If the Jaeger remote sampler selects a TraceIDRatio policy, that's what I want to know.

For the TraceIDRatio description, it encodes a floating point probability which is also (IMO) not a great representation for conveying adjusted count. The specification talks about how much precision should be logged for example, which only adds to the confusion. If I am logging 1 in 1024 spans, which @oertl would prefer we represent using the number 10 (i.e., 1 byte to say sampling probability is 1/2^10), the TraceIDRatio description will read "0.000977" (i.e., 8 bytes to encode a number with the addition of a 0.05% error, since 1/0.000977 = 1023.5).

As a vendor, we aren't actually concerned with knowing the sampler name. As this OTEP hopes to convince us, we can count spans without anything more than an adjusted count. I had two reasons to include it here:
1. @oertl gave a convincing reason why knowing that TraceIDRatio was used for example, as opposed to a tail-based sampling scheme
2. We had identified a gap for the Parent sampler, where we may or may not know its probability. Knowing the sampler avoids any ambiguity here and leaves the door open for specifying how to propagate the Parent probability.

text/0148-sampling-probability.md
jmacd: The text is outdated, resolving.

text/0148-sampling-probability.md
jmacd: The text is outdated, resolving. I am now proposing non-optional support. See https://github.com/open-telemetry/oteps/pull/168

text/0148-sampling-probability.md
jmacd: Thanks for your help. I tossed most of this text and came up with https://github.com/open-telemetry/oteps/pull/168

text/0148-sampling-probability.md
jmacd: I realized that my employer is definitely not interested in optional support, so I have returned with a mandatory but limited proposal here: https://github.com/open-telemetry/oteps/pull/168

jmacd:
See the related OTEP https://github.com/open-telemetry/oteps/pull/168

reyang (approved):
LGTM.

jmacd:
I will re-open this OTEP with a fresh PR. This is too long to review.

jmacd:
This is replaced by #170.

PR #149, opened 2021-03-08
Add exponential bucketing to histogram protobuf

yzhuge:
Follow up of https://github.com/open-telemetry/opentelemetry-proto/pull/226


jmacd:
@oertl please note.

text/0149-exponential-histogram.md
```diff
@@ -0,0 +1,162 @@
...
+    double base = 1;
+    double zero_count = 2; // Count of values exactly at zero.
+    ExponentialBucketCounts positive_value_counts = 3;
+    ExponentialBucketCounts negative_value_counts = 4; // Negative values are bucketed with their absolute values
+}
+
+// "repeated double bucket_counts" represents an array of N numbers from bucket_counts[0] to bucket_counts[N-1].
+// With index i starting at 0, ending at N-1, ExponentialBucketCounts defines N buckets, where
+// bucket[i].start_bound = base ^ (i + exponent_offset)   // inclusive bound
+// bucket[i].end_bound = base ^ (i + 1 + exponent_offset) // exclusive bound
```
CharlesMasson, line 38: As a note: I believe that removing the bound inclusiveness constraint would in some cases allow or make it easier to optimize the sketch insertion speed at no relative accuracy cost (see [this](https://github.com/DataDog/sketches-java/blob/master/src/main/java/com/datadoghq/sketch/ddsketch/mapping/LogLikeIndexMapping.java#L80) in DDSketch for instance). It may come with some disadvantages (e.g., non-determinism or added complexity in the bucket definition), which we could mitigate maybe, but given the significant performance gain it may allow, I thought it was worth mentioning.

oertl: In my opinion, the inclusiveness/exclusiveness is irrelavant in practice. It is almost impossible to get consistent boundaries on different platforms anyway. The reason is that the result of the power function is platform/implementation dependent, if the base is not a power of two. A base that is a power of two would imply relative errors in the order of 100% which is rarely useful. If boundaries are implicitly defined, I think there is no other way than to precisely define all numerical operations required to compute the boundaries or - more reasonably - to simply be tolerant with respect to mapping values to buckets.

yzhuge: I was on the fence on this one. So unless there is strong objection, I will remove the inclusive/exclusive comment in the next update.

jmacd: Interesting -- I would be happy to see this left unspecified, in order to allow implementations to decide how to treat values that lie on the boundary (subject to precision, ...).

jmacd (approved):
Very well done! I look forward to another take at https://github.com/open-telemetry/opentelemetry-proto/pull/226. Thank you @yzhuge

jmacd:
@open-telemetry/specs-approvers

reyang (approved):
LGTM.

hdost (commented):
The one thing that I am left with on this is "what is the benefit of doing this?"
Maybe I am missing something here. 
Is it that they are "Mergeable" but not universally mergeble?

Or is it mainly:
> distributions, which are common in the OTEL target application like response time measurement. Exponential buckets (ie. log scale buckets) need far fewer buckets than linear scale buckets to cover the wide range of a long tail distribution.

I know that there's the Open Questions section mentioning it and the options. 

Edit: I guess even without that based on your referenced implementation it would be possible.  I was attempting to think about it from the use case perspective of if I have different servers getting calls of different latencies such that they don't produce the same buckets as output. How does that look in this. I'd assume based on proto that it would work fine. Or should we in fact still need to require the implementation to define these bounds?

I apologize if this is not germane to this particular topic. I want to make sure I'm understanding the intent/implications.

yzhuge:
@hdost The intention is to have OTEL support for log scale histograms natively and efficiently. Log scale is commonly used for high dynamic range data such as response time with long tail. 
In this proposal, I basically left the mergeability issue open. The good news is that we have a way forward for "universally mergeable" histograms. But as the initial step, we just do a generic one  (any base is allowed).

text/0149-exponential-histogram.md
```diff
@@ -0,0 +1,162 @@
...
+3. Answering question like "what percentage of values fall below 100". When the threshold is on 10, 100, 1000, etc, base10 histograms do give exact answer. But even in the decimal world, power of 10 numbers is a small population. For thresholds like 200, 500, etc. base10 histograms have no advantage over base2. If an exact answer is required in these cases, a user should create explicit buckets at these bounds, instead of using exponential buckets.
+
+So the "human friendliness" of base10 exponential histograms is largely an illusion. To some extent, the base2 vs. base10 question has been answered long ago: computers convert input from base10 to base2, do processing in base2, then convert final output from base2 to base10. In the histogram case, we take input in "double", which is already in [binary float point format](https://en.wikipedia.org/wiki/Double-precision_floating-point_format). Bucketing these numbers in base10 bounds effectively switches base during processing. It just adds complexity and computational cost.
```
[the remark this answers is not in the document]
HeinrichHartmann: I agree with the conclusion in the context of this discussion: For log-histograms there is little point in using base-10.

I see a real usability advantage of using base-10 when using log-linear bins (...,100,110,...200,210,...,1000,1100,..) since these cover all the interesting cases for humans with regards to question 3. ("what percentage of values fall below 100"). I think this is the point that others have been making in related discussions. For the purpose of the current scope (log-bins) the benefits with regards to (3) are marginal.

text/0149-exponential-histogram.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Trade-offs and mitigations
+
+Simplicity is a main design goal. The format targets the most common scenarios. For now, histograms not conforming to ExponentialBuckets may be encoded as explicit buckets. If a histogram type is common enough, a new bucket type may be added in the future.
+
+The followings are restrictions of ExponentialBuckets:
+
+* Buckets for positive and negative values must have the same "base".
+* Buckets must cover the full value range. In the future, ExponentialBucketCounts might add an overflow_count and an underflow_count for counts above the highest bucket and below the lowest buckets, respectively. However, overflow or underflow bucket breaks the "pure log scale" property. "Rescale" is preferred when reducing memory cost. See "merge" section below.
+* ExponentialBucketCounts is designed for dense buckets. Between the lowest and the highest bucket, most buckets are expected to be non-empty. This is the common case in telemetry applications.
```
HeinrichHartmann, line 60: > ExponentialBucketCounts is designed for dense buckets. Between the lowest and the highest bucket, most buckets are expected to be non-empty. This is the common case in telemetry applications.

My experience is somewhat different here:

a) Latency metrics have a long tail, if your bucketing scheme has 1% accuracy, there might be many empty buckets before you reach the latency outliers (>10sec).
b) In order to avoid up-front choices it's very convenient to have a lot's of headroom at the top and bottom end of your scale, e.g. covering 10^+/-100, in this case you will have lot's of empty buckets as well. The problem is very similar to floating point arithmetic: When you declare a float you don't want to provide min-max exponents.

The conclusion that we (back at Circonus [1]) and now also Prometheus [2] reached is, that sparse encoding is the better alternative.

[1] [Circllhist](https://arxiv.org/abs/2001.06561)
[2] [Sparse high-resolution histograms for Prometheus](https://docs.google.com/document/d/1cLNv3aufPZb3fNfaJgdaRBZsInZKKIHo9E6HinJVbpM/edit#)

text/0149-exponential-histogram.md
```diff
@@ -0,0 +1,162 @@
...
+
+where base is a parameter of a bound series, and exponent is an integer. Note that exponent may be positive, negative, or 0. Such bounds are also commonly known as "log scale bounds".
+
+## Internal details
+
+Proposed message type to be added:
+
+```
+message ExponentialBuckets {
+    double base = 1;
```
HeinrichHartmann, line 29: typo? `base=1` will result in `bound = 1^ exponent`, which is always `==1`

text/0149-exponential-histogram.md
```diff
@@ -0,0 +1,162 @@
...
+double base = 1;
+
+// Future. Option 2. Changing a single value into a member of a new oneof is safe and binary compatible
+oneof base_spec {
+   double base = 1;
+   sint32 base_scale = 99;  // base = 2 ^ (2^base_scale). base_scale may be negative.
+}
+```
+
+**Or should we just do option 2 right now?**
```
HeinrichHartmann, line 155: +1

Leaving `referenceBase` configurable opens up a lot of possibilities to end-up with non-mergable histograms, by using different base values. You already explained how the choice of base-2 or base-10 is largely inconsequential. So if there is no clear advantage of leaving this configurable, I would opt for hard coding this value. Since floats are alreday base-2, `referenceBase=2` seems like the best option.

HeinrichHartmann:
My 2-cents from the side-lines:

* I like this proposal a lot in general. Log-histograms are a pragmatic and effective way forward. Theoretical backing from the DDSketch paper, and there is real-world experience from the DataDog implementation of this method.

* Configuring histograms can be a high barrier of entry, since most users will not be familiar with the subtleties of the involved bin layout, and will not be able to predict value ranges up-front very well. This is constant struggle, e.g. with Prom-Histograms, HDR-Histograms and also DDSKetch to a certain degree. We should strive to avoid configuration where possible.
* 
* If different configurations result in un-mergable histograms, this is a tragedy. AFAIK, this is one of the main regrets of Google's initial Histogram implementation in Borgmon. Universal Mergeability is a very nice insight, that can be used to address this problem.

Fortunately, with log histograms you can get-away with universal choices, that work good enough for all operational use cases I have seen. Those choices are:

a) Sparse bin-layout with bin range covering a very large value range of 10^{+/- 100}
b) referenceBase=2 (or 10 does not matter, just pick one)
c) referenceScale=-3, resulting in theoretical max relative error of 5% (in practice this is much lower => DDSketch paper).

There is a case to be made for leaving referenceScale configurable, that I can get behind, but setting the default to -3 with a strong recommendation to leave it as-is seems appropriate.

text/0149-exponential-histogram.md
jsuereth: this is a protocol buffer definition, 1 = the encoded field number for serialization no the value of the base.

HeinrichHartmann: whops! Thx for the clarification.

text/0149-exponential-histogram.md
yzhuge: There are many known methods for sparse array encoding. I'd like to start with a simple array and extend to fancier forms if necessary. 
Sparse vs. dense may have to do with config. In New Relic "Distribution" metric, the number of buckets is limited to 320. A typical histogram aggregates over thousands to millions of values, which are likely to make the buckets dense.  
More discussion in the "config" comment below. 
For "b)", many implementation, including DDSketch and New Relic (internal use) dynamically shift the array to cover only the dense part, avoiding empty buckets at the bottom and top.

yzhuge:
So far many open source histogram libs (HDR, DDSketch, etc.) have fixed/config'ed accuracy, variable memory footprint. This makes config harder and risks "out of memory" error, which often kills the app. 
What I envision is fixed/config'ed memory footprint, variable accuracy. For example, New Relic Distribution metric defaults to max 320 buckets, which usually cost just about 1KB (dense array, 4 byte counter). This gives about 3% relative error for a data max/min contrast up to 1M (example, 1ms to 1 million ms). When contrast exceeds the limit, it auto reduces accuracy to keep memory footprint constant. The rationale is "less accurate histogram is better than breaking monitored app". 
Of course, such histograms requires built-in auto rescaling. New Relic uses the 2 to 1 bucket merge, which is also proposed in UDDSketch paper, and in Google internal use. 
Look at this another way: Regardless of the value range of the data, the histogram provides a constant number of "steps" (ie. buckets) for understanding the distribution. 320 buckets is enough to visualize the "shape" of the distribution. 
New Relic does have plan to open source its histogram lib. No promise on timeline is given though.

text/0149-exponential-histogram.md
HeinrichHartmann: This makes sense. I can see a dynamically allocated (!) dense-region working in practice.

HeinrichHartmann:
> many open source histogram have fixed/config'ed accuracy, variable memory footprint. This makes config harder and risks "out of memory" error, which often kills the app.
> What I envision is fixed/config'ed memory footprint, variable accuracy.

I clearly see the benefits of the "fixed memory / variable accuracy" design. Seems like a good fit here. If we can realize "fixed memory / variable accuracy" histograms with near-zero configuration, and full mergeability this is a clear win.

Let me just remark, that my experience with "fixed accuracy / variable memory" has been different (a) Circllhist has zero configuration, and (b) I *never* have seen them blowing up or "killing an app" in practice. So the "fixed accuracy / variable memory" should be a viable as well.

hdost (approved):

jsuereth (approved):
Sorry I didn't approve this before.  I'm totally on board with this OTEP.

I think Heinrich raised some good questions that I'd like to see addressed when we formalize this specification and model, but I think this can be done as part of the actual submission into the specification.

jmacd (approved):

jmacd:
All: we plan to merge this OTEP but need one more approval (@bogdandrutu?). The consensus is to go forward with a fixed referenceBase of 2.

jmacd:
@oertl, @CharlesMasson do you approve the use of a fixed referenceBase value of 2 as discussed at the end of this review? If we have consensus on this, there is a very clear path forward. Thanks.

oertl:
@jmacd I also prefer a fixed referenceBase value of 2. Just a minor thing that could be improved in my opinion. I would prefer to have `base = 2 ^ (2 ^ (-baseScale))` instead of `base = 2 ^ (2 ^ baseScale)`, such that `baseScale` is always positive for relevant configurations. Then baseScale could be interpreted as a kind of precision parameter. The larger baseScale, the more precise the histogram.

githomin:
@jmacd a referenceBase value of 2 works, and I also like @oertl 's suggestion.

jmacd:
> that baseScale is always positive for relevant configurations
sounds good to me

jmacd:
@open-telemetry/specs-metrics-approvers please :)

oertl:
As a proof of concept, I implemented the proposed exponential bucketing scheme with fixed referenceBase 2 for DynaHist (see [here](https://github.com/dynatrace-oss/dynahist/blob/8d49f4be36d33df113a2f4a00d77029370253294/src/main/java/com/dynatrace/dynahist/layout/OpenTelemetryExponentialBucketsLayout.java#L145)). Mapping double-precision floating-point numbers to bucket indices can be implemented without any floating-point operations and with only a couple of array lookups, achieving recording times well below 10ns. This shows that the proposed bucketing scheme can also be used if performance is a concern.

To make the mapping of double-precision floating-point numbers to buckets platform-independent, it would be sufficient to specify the mantissa of the double-precision floating-point representation of  `2^(i*2^(-p))` for all `i` in `{1,2,...,(2^p)-1}`. If platform-independence is important, we could think of adding those constant values to the specification.

yzhuge:
Computers like binary numbers. Binary histogram reduces bucketing problem to a single 1 to 2 range. At a given base scale, there is a finite number of buckets in the range. So precomputed lookup tables can be used to avoid standard log() function, as shown in @oertl 's comment. So looks like we have a happy ending on this histogram journey. Scaled binary histogram is:
* Memory optimal (compared to log approximation like log linear, quadratic, or cubic)
* Fast to compute, when using lookup table
* Consistent across platforms, when a published lookup table is used, as suggested by @oertl 
BTW, I like the idea of flipping baseScale to positive range for commonly used scales (those with relative error of a few percent).

oertl:
The maximum number of buckets needed for recording all positive double-precision floating-point numbers (excluding 0, +inf, NaN) dependent on `p` is given below. This can be used to estimate the worst case space requirements.
```
 p    num. postive buckets    relative bucket width
---------------------------------------------------
 0                    2098                100.000 %
 1                    4195                 41.421 %
 2                    8387                 18.921 %
 3                   16767                  9.051 %
 4                   33518                  4.427 %
 5                   67005                  2.190 %
 6                  133946                  1.089 %
 7                  267764                  0.543 %
 8                  535273                  0.271 %
 9                 1070035                  0.135 %
10                 2139047                  0.068 %
```


@yzhuge 
> Memory optimal (compared to log approximation like log linear, quadratic, or cubic)

This depends on how you define optimality. If you need a maximum relative bin width of 1%, you will be forced to use `p=7` with a relative bin width of 0.543%, which is clearly not optimal.

lizthegrey:
hearing from @postwait there's the possibility of an apache 2 code contribution here to speed this process along...

postwait:
OpenHistogram has indeed been relicensed Apache-2

jmacd:
Responding to @lizthegrey's question above, about the possibility of adopting the OpenHistogram libraries and protocol as a short-cut for OpenTelemetry's metrics protocol.

OpenHistogram uses a decimal base with linear sub-buckets, which this OTEP argues is additional complexity compared with a simple exponential histogram.

This OTEP argues in favor of base-2 histogram with variable scale, with support for lower-resolution histograms and something we may call "perfect subsetting"--the idea that we can automatically lower resolution without introducing error by a change of scale.  OpenHistogram would have to be extended to support change of scale, where resolutions like 3, 9, and 18 (30x, 10x, and 5x reductions) make sense.

We have at least three vendors and platform providers willing to contribute client code for computing these histograms who have spoken in this review thread. The decision facing OpenTelemetry and OpenMetrics communities -- which histogram to adopt? -- will impact development in the OpenTelemetry collector and Prometheus server far more than in client SDKs, because we will be handling both our old explicit-boundary histogram and the new exponential histogram. This is why I think adopting OpenHistogram probably would not speed our overall process.

We are missing an approval on this OTEP, @open-telemetry/specs-approvers please consider reviewing. I want to strongly encourage that we merge this PR as a signal that we found an agreement, then I would propose that @yzhuge follow up with a PR in the protocol repository. If there are loose ends, such as about overflow buckets and whether the zero bucket has finite width (these points have been raised to me, privately), then I suggest we discuss them in the following PR. 

Thank you @yzhuge.

MrAlias (approved):

SergeyKanzhelev (approved):

SergeyKanzhelev:
I think this otep has reached critical mass of approvals. Should we give another day to raise concerns. @lizthegrey do you agree with @jmacd reasoning?

jmacd:
Thank you @SergeyKanzhelev. I agree with leaving this open for comment a little longer, and I think we should merge it either way because it is the result of a valuable engineering discovery process. We could still decide to adopt a base-10 solution following this OTEP, but I don't see enough support for it.

For me, the final signal that base-2 is our best choice  came from @beorn7, who has been (quietly) changing the [Prometheus spare histograms prototype](https://docs.google.com/document/d/1cLNv3aufPZb3fNfaJgdaRBZsInZKKIHo9E6HinJVbpM/edit#) being built in https://github.com/prometheus/client_golang/tree/sparsehistogram to use a reference base of 2 and power-of-2 scale, i.e., we have reached agreement with a key figure in the OpenMetrics community.

There are a number of contributors working on this, every one of us representing a vendor, so it's difficult to separate purely technical considerations from business-motivated reasoning in this arena. Let us know what you think, @lizthegrey, thanks!

postwait:
I realize I'm only one voice in the choir here, but after dealing with a
lot of end user consumers of histograms we switched from base 2 to base 10
for two reasons (both stakeholder driven).

The first was that base 10 SAS easier to deal with on systems without
floating point not because of the log based part but because the second
stage linear binning lined up well.

The second reason wss stronger. End users including business people,
analysts, operations people and software engineers were able to reason
about the formed histograms more intuitively. While cs people think they
think in base 2, or research showed that *everyone* reasoned more
competently in base 10.

I strongly believe that base 10 is a better solution for these reasons, but
will freely admit that it seems that base 2 has one the popularity contest.
I suspect that important stakeholders (the ultimate end users of this data)
haven't been well represented in this decision as they had in our research.

Best regards,

Theo

text/0149-exponential-histogram.md
jmacd: This has been accepted.

jmacd:
Thank you @yzhuge.

All, please see https://github.com/open-telemetry/opentelemetry-specification/issues/1776.

giltene:
I realize this may seem like some late in the game stuff, but I'd suggest looking at HdrHistogram's auto-ranging, auto-resizing [DoubleHistogram](https://github.com/HdrHistogram/HdrHistogram/blob/master/src/main/java/org/HdrHistogram/DoubleHistogram.java) for some mature conceptual work around a close-to-zero-config-as-possible histogram for positive floating point numbers. A DoubleHistogram can accommodate any value in the positive double precision floating point range (of 0...2^1023), and only limits the *range* of the actually recorded values in a single histogram to not exceed a ratio of e.g. 1 : 2^55 for 2 decimal point precision (+/- 1% value error worst case) and e.g. 1 : 2^53 for 3 for 3 decimal point precision (+/- 0.1% value error worst case). Those ranges are high enough that e.g. when measurement units end up being as a fine grained to be in nanoseconds, values as high as Billions of years can be tracked in the same histogram as a nanosecond, all while maintaining the required +/- % error rate.

DoubleHistogram was borne out of real world needs, and seems to have served some of those well. While configuration options do exist when one actually wants to control or contain in-memory data structures size such that they will error rather than auto-resize to accommodate recorded values, the only *required* configuration parameter in creating a DoubleHistogram is the precision (stated as number of decimal points) (see [Java constructor example](https://www.javadoc.io/static/org.hdrhistogram/HdrHistogram/2.1.12/org/HdrHistogram/DoubleHistogram.html#DoubleHistogram-int-)). And as the most commonly used precision is "2" decimal points (i.e. relative error for values is kept to below 1% across the entire covered range), a default of 2 can be used in e.g. specification if one wants to reach true zero-config.  The wire form for HdrHistogram (and DoubleHistogram) includes the actual configuration of the recorded histogram, allowing stored histograms of varying configurations to be read and manipulated, merged and added etc. into histograms with different precisions and experienced value ranges, all without having the source configuration dictate or dominate the outcome.

There are probably many possible ways to shave this Yak, but HdrHistogram has been at it for almost a decade, with some pretty wide use and exposures, so it may be a good starting point and a good way to get a head start on some of the learnings that often come from attempts to set one-scheme-fits-the-all solutions.

lizthegrey:
I've been on PTO, please proceed without me.

jmacd:
I meant to merge this PR, not to close it!

PR #152, opened 2021-04-12
Introduce Telemetry Schemas

tigrannajaryan:
Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/1324

I believe changes to semantic conventions and the shape of emitted telemetry
are likely to occur during the lifetime of an instrumentation library.
I do not think we should aim to lock the telemetry schema and disallow changes to it.
Such locking would place a huge limitation on how instrumentation can evolve and
would make it nearly impossible to fix mistakes in the semantic conventions, in the
schema or  in the implementation of the instrumentation (which will inevitably happen
sooner or later).

This OTEP introduces the concept of telemetry schemas that allows semantic conventions
and instrumentations to evolve over time without breaking consumers of telemetry.


text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1124 @@
...
+
+If schema_url field is non-empty both in Resource* message and in the contained
+InstrumentationLibrary* message then the value in InstrumentationLibrary*
+message takes the precedence.
+
+## API and SDK Changes
+
+OpenTelemetry API and SDK require the following changes:
+
+- Add method SetSchema(schema_url) to Tracer. After this call all telemetry
```
yurishkuro, line 744: I think it would be better to only allow this to be called once, e.g. by making it an optional argument to the Provider.

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1124 @@
...
+  possible transformation types that can handle all possible changes to schemas
+  that we can imagine. That would be too complicated and very likely
+  superfluous. Any new transformation types should be proposed and added in the
+  future to the schema file format when there is an evidence that they are
+  necessary for the evolution of OpenTelemetry.
+
+## Use Cases
+
+This section shows a couple interesting use-cases for the telemetry schemas
+(other uses-cases are also possible, this is not an exhaustive list).
```
pyohannes, line 164: Regarding the use case where the backend is not aware of schemas and the collector is not used: should there be any SDK functionality for this use case, or is it left to each individual exporter to implement schema conversions?

text/0152-telemetry-schemas.md
tigrannajaryan: This is good point and I was thinking about it, but I was not sure we want to alter the signature of an API that is already declared stable (GetTracer). If I am not wrong it would be a backward compatible API change but it will likely be a breaking ABI change (I am forgetting now if we require ABI compatibility for languages like C++).

Alternatively for languages which allow overloading we could introduce a new GetMeter/GetTracer/GetLogEmitter overload which takes 3 parameters (the 3rd parameter being schema_url).

I am open for suggestions.

text/0152-telemetry-schemas.md
tigrannajaryan: It is a possibility. We will need to decide if we want to put that burden on each language SDK vs implementing schema conversion once in the Collector.

text/0152-telemetry-schemas.md
yurishkuro: We need to have a design principle that certain API methods are extensible, and use language-specific constructs to achieve that. E.g. in Java the tracer shouldn't be accessed via `provider.getTracer`, but using a builder. In Python it can be getTracer method with kw args.

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1124 @@
...
+  they emit. This may be because for example the semantic conventions evolved
+  and we want to make our telemetry match the newly introduced conventions.
+
+- In an observability system there may simultaneously exist telemetry sources
+  that produce data that confirms to different telemetry schemas because
+  different sources evolve at a different pace and are implemented and
+  controlled by different entities.
+
+- Telemetry consumers have a need to understand what schema a particular piece
+  of received telemetry confirms to. The consumers also need a way to be able to
```
jkwatson, line 86:
```suggestion
  of received telemetry conforms to. The consumers also need a way to be able to
```

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1124 @@
...
+"Incompatible" means that transformations between the new version and the
+previous version are not possible.
+
+Note: version numbers use the
+[ordering rules](https://semver.org/#spec-item-11) defined by semver 2.0
+specification.
+
+## Schema File
+
+Schema File is a YAML file that describes the schema of a particular version. It
```
jkwatson, line 262:
```suggestion
A Schema File is a YAML file that describes the schema of a particular version. It
```

text/0152-telemetry-schemas.md
jkwatson: I agree that this should be set only once, preferably at creation time, and once the tracer has been used at all, that it should not be able to have it's schema value changed. 

For instance, I think in Java, we'd probably just add an overload to the TracerProvider.get() method to allow the specification of a schema URL, and under the hood, we'd default it to whatever schema version the semantic convention library aligned with the SDK version used (or something functionally equivalent to this... we might choose to introduce an optional builder for tracers but it would effectively do the same thing).

tigrannajaryan: I rewrote "API and SDK Changes" section to reflect this discussion. Please check.

text/0152-telemetry-schemas.md
tigrannajaryan: Fixed.

text/0152-telemetry-schemas.md
tigrannajaryan: Fixed.

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1158 @@
...
+
+If schema_url field is non-empty both in Resource* message and in the contained
+InstrumentationLibrary* message then the value in InstrumentationLibrary*
+message takes the precedence.
+
+## API and SDK Changes
+
+### Instrumentation Library Schema URL
+
+OpenTelemetry API need to be changed to allow getting a Tracer/Meter/LogEmitter
```
jkwatson, line 744:
```suggestion
The OpenTelemetry API needs to be changed to allow getting a Tracer/Meter/LogEmitter
```

tigrannajaryan: Fixed.

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1159 @@
...
+
+Telemetry Schemas are central to how we make this possible. Here is a summary of
+the proposal:
+
+- We introduce a file format for defining telemetry schemas.
+
+- Telemetry schemas are versioned. Over time the schema may evolve and telemetry
+  sources may emit data confirming to newer versions of the schema.
+
+- Telemetry schemas explicitly define transformations that are necessary to
```
jsuereth, line 105: +1

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1159 @@
...
+  convert telemetry data between different versions of the schema, provided that
+  such conversions are possible. When conversions are not possible it
+  constitutes a breaking change between versions.
+
+- Telemetry schemas are identified by Schema URLs.
+
+- Telemetry sources (e.g. instrumentation libraries) should include a schema URL
```
jsuereth, line 112: Not explicitly stated (AFAICT);  should the telemetry schema include the version number?

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1159 @@
...
+
+- Telemetry consumers should pay attention to the schema of the received
+  telemetry. If necessary, telemetry consumers may transform the telemetry data
+  from the received schema version to the target schema version as expected at
+  the point of use (e.g. a dashboard may define which schema version it
+  expects).
+
+- OpenTelemetry will publish a telemetry schema as part of the specification.
+  The schema will contain the list of transformations that semantic conventions
+  undergo. The schema will be available to be referred and downloaded at well
```
jsuereth, line 123: Should the "known URL" be the URL of the schema?

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1159 @@
...
+
+![Query Translate](img/0152-query-translate.png)
+
+### Collector-Assisted Schema Transformation
+
+Here is a somewhat different use case, where the backend is not aware of schemas
+and we rely on OpenTelemetry Collector to translate the telemetry to a schema
+that the backend expects to receive. The "Schema Translate Processor" is
+configured, the target schema_url is specified and all telemetry data that
+passes through the Collector is converted to that target schema:
```
jsuereth, line 198: Nit: should this only convert telemetry of a compatible schema (or previous schema version?)

I had thought "schema" would be namespaced/split apart in styles (e.g. HTTP), or is it meant to be ONE schema for ALL semantic conventions?

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1159 @@
...
+
+Schema URL is an identifier of a Schema. The URL specifies a location of a
+[Schema File](#schema-file) that can be retrieved (so it is a URL and not just a
+URI) using HTTP or HTTPS protocol.
+
+Fetching the specified URL may return an HTTP redirect status code. The fetcher
+MUST follow the HTTP standard and honour the redirect response and fetch the
+file from the redirected URL.
+
+The last part of the URL path is the version number of the schema.
```
jsuereth, line 212: Per my comment above, I think this should be in the list of requirements :)

jsuereth (commented):
i love this direction!  May take me a while to think through nuances of the schema-format and its evolution.   Specifically: I'm a bit concerned that you may be defining 'code' / 'behavior' in YAML around these transformations.

text/0152-telemetry-schemas.md
tigrannajaryan: Yes, it is defined later in the doc.

text/0152-telemetry-schemas.md
tigrannajaryan: Yes, the doc later explains what a Schema URL is in more details.

text/0152-telemetry-schemas.md
tigrannajaryan: There is one OpenTelemetry schema for all semantic conventions.

text/0152-telemetry-schemas.md
tigrannajaryan: I did not want to overload the Summary section with too many details. Let me see if I can add this.

tigrannajaryan (commented):
> Specifically: I'm a bit concerned that you may be defining 'code' / 'behavior' in YAML around these transformations.

That's exactly what it does. The transformation are high-level code. They are instructions that operate on telemetry data. The set of opcodes is defined by schema file specification. Can you please tell what your concerns are?

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1159 @@
...
+
+- Requires telemetry sources to be able to emit telemetry in multiple different
+  schema versions, which is a significant burden on the telemetry sources. This
+  is likely a dealbreaker.
+
+- Requires a communication channel from telemetry consumers to telemetry sources
+  which currently does not exist (all communication is currently
+  one-directional, from sources to consumers).
+
+## Prior Art
```
jmacd, line 1029: Fantastic. 😀

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1159 @@
...
+
+- How to make it easy for instrumentation libraries to refer to a particular
+  OpenTelemetry schema version and also make sure any semantic convention
+  helpers the library uses (e.g. constants that define the semantic conventions)
+  match exactly that same schema version? One possible solution is to introduce
+  helper packages per schema version that the libraries can use, e.g. constants
+  that define the semantic conventions and the corresponding schema version url.
+  This should be likely the topic for a follow-up OTEP.
+
+- Should we make it possible to include the entire Schema File in OTLP requests
```
jmacd, line 901: [Thinking about how JSON Schema encourages re-use with a `$ref` keyword], I could imagine a "Yes" answer here that does not require any more protocol change than what you have here.

We could specify to use a schema URL like `resource://some.identifier` and then use the `some.identifier` resource value as a schema--that might work except for resource value size restrictions.

In a logging system, (maybe, if it were up to me, ...), I might suggest the process log its schema early on in the process lifetime and then use a schema URL that refers backwards in the log.

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1159 @@
...
+   </td>
+  </tr>
+</table>
+
+## OTLP Changes
+
+To allow carrying the Schema URL in emitted telemetry it is necessary to add a
+schema_url field to OTLP messages.
+
+We add schema_url fields to the following messages:
```
jmacd, line 677: This is an interesting pattern, where we place a new optional field in one of two places, three times over.  Since it's likely to set a precedent, I have a couple of thought experiments: 

1. Instead of adding a new `schema_url` to the _ResourceSignal_ messages here, why not use a resource named `schema.url`? (JSON Schema uses `$schema`, I like that too). This would support schema URLS with no new fields.

2. Instead of adding a new `schema_url` to every of the _InstrumentationLibraryXXX_ messages, could we instead add another place for per-instrumentation-library resources to be expressed?  e.g., `opentelemetry.proto.resource.v1.Resource library_resource = 3;`. Would that be more flexible? I think it might be.

The reason for both of these questions is that I can imagine other kinds of property that are somehow special and if we add a new field for each one, we're in trouble. I'm thinking about information related to the metric collection interval and/or other kinds of processing that may have taken place as telemetry data is collected, information we may want to record either at the global level or at the per-library level.

In transportation systems, the term "Waybill" is used to document what is in transit in a multi-party exchange of goods. We have a similar situation here, where you want to express "what's inside" with a schema, and I strongly suspect there are other kinds of "what's inside" information ahead.

jmacd (approved):
Very enthusiastic, thanks @tigrannajaryan

jmacd:
I'm excited about the potential to use Telemetry Schemas inside logs data to as a way to enable Logs-to-Metrics, Logs-to-Span, and Logs-to-Resource transformations (:grin:).

text/0152-telemetry-schemas.md
austinlparker: I'd also offer a non-specific concern over increasing the size of outgoing messages from OpenTelemetry, especially from things like the browser runtime. To that point, I'd support the idea of making the schema a named resource vs. adding fields.

text/0152-telemetry-schemas.md
pyohannes: It would indeed be a burden on languages SDKS, however, I think when not providing this feature on SDK level we end up with inconsistent and unexpected user experiences.

E. g., the Jaeger and Zipkin exporters which every SDK ships by default wouldn't provide any schema mapping capabilities.

Apart from that I really like this proposal and the direction it's heading towards.

victlu:
How does this OTEP compare against more general Schema Evolution features available in protocols like Avro and/or Parquet? I assume we need something more purpose built for Telemetry community?

tigrannajaryan:
> How does this OTEP compare against more general Schema Evolution features available in protocols like Avro and/or Parquet? I assume we need something more purpose built for Telemetry community?

Avro and Protobuf (I am not familiar with Parquet) require full schema definition for data, while this proposal intentionally avoids that and deals with schema _changes_ only (for now). Other than that there are similarities such as being able to rename a field and retain compatibility (AFAIK, Avro via name aliases, Protobuf via field numbering).

This proposal is also extensible in the sense that arbitrary complex _changes_ to the schema may be introduced as supported transformation types in the future, which are more powerful than what Avro or Protobuf offer in terms of schema evolution.

So, where Avro or Protobuf may allow schema evolution more or less automagically, they are very limited in the sense of what is allowed to evolve, while this proposal is potentially more powerful at the cost of requiring explicit declaration of changes.

This proposal also has many specifics which make sense in the context of OpenTelemetry (i.e. schema version numbers, schema urls, downloadable schemas, etc), but are not necessary useful in a general purpose encoding/data transport such as Avro or Protobuf.

text/0152-telemetry-schemas.md
tigrannajaryan: I like this idea.

text/0152-telemetry-schemas.md
tigrannajaryan: @jmacd 

>Instead of adding a new schema_url to the ResourceSignal messages here, why not use a resource named schema.url? (JSON Schema uses $schema, I like that too). This would support schema URLS with no new fields.

Do you mean a resource attribute named "schema.url"? That would work. The only thing I am not quite sure about is whether it is fair to say that "the telemetry schema is an attribute of a Resource that emits the telemetry".

>Instead of adding a new schema_url to every of the InstrumentationLibraryXXX messages, could we instead add another place for per-instrumentation-library resources to be expressed? e.g., opentelemetry.proto.resource.v1.Resource library_resource = 3;. Would that be more flexible? I think it might be.

Alternatively, we could allow `InstrumentationLibrary` to have arbitrary attributes and record "schema.url" attribute there. I am not quite sure recording a Resource inside an InstrumentationLibraryXXX message reflects our understanding of what a Resource is.

I don't have a strong opinion, open for suggestions.

tigrannajaryan: > I'd also offer a non-specific concern over increasing the size of outgoing messages from OpenTelemetry, especially from things like the browser runtime. To that point, I'd support the idea of making the schema a named resource vs. adding fields.

@austinlparker size increase likely is negligible since the schema_url is recorded once per batch of telemetry emitted by instrumentation library.

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1159 @@
...
+More transformation types can be added in the future when there is a need to
+represent a particular change in the telemetry schema that is not possible with
+the current set of transformations. However, care should be taken to avoid
+introducing unnecessary or overly complicated transformation types, especially
+ones that cannot be applied on a local portion of telemetry data and requires a
+full state (e.g. transformations, such as aggregations, etc), since such
+transformations can place significant implementation burden on telemetry sources
+that wish to support the notion of telemetry schemas.
+
+## Alternates Considered
```
Oberon00, line 974: This is a rather complex and powerful approach. To justify it, I think there is one more alternative to consider (with corresponding subsection here): Have a strict policy to never rename an attribute once it becomes stable. While this may lead to sub-optimal names, it would be much simpler to implement and have not negative performance impact.

text/0152-telemetry-schemas.md
austinlparker: > > I'd also offer a non-specific concern over increasing the size of outgoing messages from OpenTelemetry, especially from things like the browser runtime. To that point, I'd support the idea of making the schema a named resource vs. adding fields.
> 
> @austinlparker size increase likely is negligible since the schema_url is recorded once per batch of telemetry emitted by instrumentation library.

Ah, ok, it read like you could apply multiple `schema.url` attributes. Looking at OTLP, it seems like adding this field to `InstrumentationLibrary` would make the most sense.

gramidt (approved):

text/0152-telemetry-schemas.md
tigrannajaryan: >Maybe we just handle them as attributes rather than as resources and build semantic conventions up for them.

Yes, this is a possible option.

text/0152-telemetry-schemas.md
tigrannajaryan: Reworded the corresponding item in Summary section.

text/0152-telemetry-schemas.md
tigrannajaryan: This is surely up for debate.
I think I have heard from enough sources that fossilizing semantic conventions and emitted telemetry schema is not an option. IMO, long term it is an impossible position. People make mistakes and want to be able to fix them. The eventual surface area of semantic convention and telemetry schemas as a whole is much bigger than the API.
Forcedly fixing the API is achievable (especially given that we have tools like compilers that help enforce it). Fixing the semantic convention I think is much more difficult (especially given that there is no fully machine-readable definition of Otel schema and there are no tools that can help enforce that schema). We could introduce such machine-readable schema and tools, but the effort for Otel will be bigger than what this proposal asks for and still won't be as useful for enabling the evolution of schemas.

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1160 @@
...
+independently over time, while continuously retaining the ability to correctly
+work together.
+
+Telemetry Schemas are central to how we make this possible. Here is a summary of
+the proposal:
+
+- We introduce a file format for defining telemetry schemas.
+
+- Telemetry schemas are versioned. Over time the schema may evolve and telemetry
+  sources may emit data confirming to newer versions of the schema.
```
pyohannes, line 103: Should Telemetry Schema versions correspond to versions of the OpenTelemetry spec? Or would those be two different (but linked) lines of versions?

text/0152-telemetry-schemas.md
iNikem: I would vote against re-implementing schema conversion in every language.

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1160 @@
...
+
+All OpenTelemetry instrumentation solutions will follow this schema.
+
+Changes to OpenTelemetry semantic conventions SHOULD be handled as described in
+the [Schema Version Number](#schema-version-number) section. For example,
+addition of a completely new optional attribute to the semantic conventions
+SHOULD NOT require a new telemetry schema version. Similarly, a removal of an
+optional attribute SHOULD NOT require a new schema version, removal or addition
+of a required attribute to an existing entity SHOULD require a new schema
+version. Generally, any changes to the semantic conventions schema that do not
```
iNikem, line 831: If we add a new required attribute to the semantic convention, how will a new schema version file look like? It still will be empty, right? Because there is no transformation that changes previous version into a new one?

text/0152-telemetry-schemas.md
weyert: That would require everyone to use the collector?

text/0152-telemetry-schemas.md
tigrannajaryan: There is no direct match between Otel spec version numbers and Otel schema version numbers. There is no need to have such coupling, the two should be able evolve independently. 
Yes, a particular Otel spec X version will naturally correspond a particular Otel schema version Y (since semantic conventions are part of the spec and schemas are tied to semantic conventions), however X may and likely will be different from Y.

text/0152-telemetry-schemas.md
tigrannajaryan: Correct. The schema "changes" will be empty. However, I believe it is still valuable to record it as a new schema version, simply to indicate that it is different. The telemetry consumers are free to use or ignore that fact.
In the future we may add ability to explicitly record in the "changes" section the fact that a required attribute was added.

text/0152-telemetry-schemas.md
tigrannajaryan: >That would require everyone to use the collector?

Either use a Collector or use a backend which understands schemas.

Alternatively, if there is no Collector and the backend does not understand schemas the system will operate just as it does today, without being aware of schemas. That is not a catastrophic situation, only things that rely on the schema may be broken, the rest should work. So the degraded behaviour in systems that do not understand schemas is no worse than we have today.

text/0152-telemetry-schemas.md
yurishkuro: This feels rather confusing to me. What's the argument against keeping schema and spec version number the same?

tigrannajaryan: If we want the version numbers the same then we have to align releases of the spec with changes in schema. I was thinking that keeping these decoupled gives us more freedom to evolve the 2 independently.

Not a strong opinion though. I think using the same version numbers would work too. It just means that in some cases we will release a new schema version when there are no changes to it (because it is time to release a new spec version but nothing is changed in semantic convention). That's fine, we just need to be careful not to forget to keep schema version number in sync with spec version number (an automated pre-release check would be preferable).
Likewise, sometimes there is probably nothing new in the spec other than changes to semantic conventions, so we will have to release a new spec version to get a new version number we can use for new schema.

If we feel keeping version numbers in sync is less confusing, then I don't mind it, that can work.

jkwatson: I think requiring a spec release for an updated schema would be a very good thing.

carlosalberto (approved):

text/0152-telemetry-schemas.md
pyohannes: I'd be in favor of sharing the same version numbers. I expect people writing instrumentation according to a certain version of the specification. It seems natural to me that this version is the same as the schema version.

text/0152-telemetry-schemas.md
Oberon00: That sounds reasonable, but I would add a subsection on this to the "Alternates Considered".

text/0152-telemetry-schemas.md
tigrannajaryan: The version number recommendations section I have in the doc will no longer work in that case. Let me think about this.

text/0152-telemetry-schemas.md
tigrannajaryan: Sounds good, I will add.

text/0152-telemetry-schemas.md
carlosalberto: Sharing them means we will have as many schema versions as API ones, which feels odd and I don't remember seeing a schema with that many versions.

text/0152-telemetry-schemas.md
tigrannajaryan: Added a subsection "Freeze Schema" in "Alternates Considered".

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1174 @@
...
+full state (e.g. transformations, such as aggregations, etc), since such
+transformations can place significant implementation burden on telemetry sources
+that wish to support the notion of telemetry schemas.
+
+## Alternates Considered
+
+### Freeze Schema
+
+Instead of introducing formal schemas, schema files and version we can require
+that the instrumentation once it is created never changes. Attributes are never
```
Oberon00, line 980: As long as the instrumentation conforms to the semantic conventions before and after, that should be fine, shouldn't it?

I think rather than freezing the schema completely, we only need to ban changes that would create an entry in the YAML file defined in this OTEP. So renaming is not allowed, adding new attributes is. Changing the meaning of an attribute is also not allowed, but I think this can also not automatically handled with the schema YAML, although the version specified there could then be used in the backend for version-specific handling.

tigrannajaryan: Yes, certain changes would be fine with the approach you propose. I still don't think this alternate approach is sufficient. I don't believe it solves the problems we want to solve. In my opinion being able to rename attributes or metrics is a basic requirement. Given how much time we spend arguing about attribute names here in Otel semantic conventions and how much uncertainty is there that the name is right even after the attribute is accepted I believe we will inevitably run into a situation when name changes are wanted.

I saw schema troubles in the past. For example with metrics emitted by Otel Collector, we renamed metrics and broke Collector dashboards in a proprietary monitoring tool. When schemas are not explicitly declared and are not a visible concept they still change and things break, just less explicitly visibly so.

Again, we can argue that we should be careful and come up with semantic conventions that we think are right and then never change them. I think it is an attempt to achieve perfection that is very difficult to do and at the same time increases the barrier to introducing semantic conventions because the cost of making a mistake is very high. I think we should lower that cost a bit. It should still be expensive to change semantic conventions (we don't want to encourage that too much), but it should not be impossible.

text/0152-telemetry-schemas.md
tigrannajaryan: I spent a bit time on this and I believe I agree that it is preferable to keep the schema version numbers in sync with the spec version numbers.

@carlosalberto I understand your concern. In some cases there will be no actual changes in the schema, but we will still bump the version number to keep it in sync with spec version number. It will mean adding one line to the schema YAML file. We can also automatically verify that this line is correctly added, so I think it is not a big problem and the trade-off is worth it to avoid confusion and the need to explain mismatching version numbers.
I will modify the proposal to reflect this.

tigrannajaryan: I changes sections "Schema Version Number" and "OpenTelemetry Schema", please have a look.

pyohannes: Thanks. I think it will make life easier for people writing instrumentation: they need to keep track of only two versions (OpenTelemetry spec and language specific API) instead of three (spec, language API, and telemetry schema).

pyohannes (approved):

text/0152-telemetry-schemas.md
tigrannajaryan: I have put some more thought into this. While I believe it is a possible option I think it a tiny bit less desirable than using a distinct protocol field in OTLP. If the schema url is recorded simply as an attribute we need to come up with a name for the that attribute, let's assume for example it is `schema.url`. But what if we want to change the name of this attribute? We can't, this will be the one attribute which is not described and is not handled by the schema. This becomes an exception. I do not quite like having exceptions like this.
I think having a distinct new field in OTLP that is part of the proto is warranted in this case since it is a meta information that describes the attributes, it is not the same kind of information as all other attributes of a resource are.
So, for this reason my slight preference is to keep my original proposal.
To be clear: this does not prevent us from introducing attributes for instrumentation libraries. But that's a separate topic and can be done in a separate OTEP unrelated to this one.

text/0152-telemetry-schemas.md
tigrannajaryan: @yurishkuro please have a look at the updated "API and SDK Changes" and see if you are OK with it.

text/0152-telemetry-schemas.md
tigrannajaryan: Taking a note for the future. Resolving for now, since it is not mandatory for this initial proposal.

text/0152-telemetry-schemas.md
tigrannajaryan: @yurishkuro @carlosalberto are you OK with the changes in "Schema Version Number" and "OpenTelemetry Schema" sections?

text/0152-telemetry-schemas.md
gramidt: I agree.

austinlparker: I concur with your logic (the "meta information that describes attributes" part). Perhaps you could add a note or errata that indicates why this information is special enough to warrant a new distinct field, in an attempt to keep this decision from establishing precedent? I think I'm more concerned about the potential for this decision to create a ripple effect, where we're introducing many versions of OTLP and causing frission for implementors and end-users alike.

tigrannajaryan: OK, Let me see if I can add something to the document.

To be clear: introducing `schema_url` as a new field in OTLP is a fully backwards compatible change. The default value for `schema_url` will be the empty string (as it normally works for new fields in Protobufs) which works great for our purposes. Empty `schema_url` indicates that the schema is unspecified (as you would expect from any pre-existing software that doesn't know about the concept of schema).
This is one of the fortunate cases when OTLP sender implementations and OTLP receiver implementations don't need to coordinate anything and can start supporting `schema_url` at their leisure. :-)

zenmoto (approved):
I really like this proposal. I like that it bakes in the ability to evolve both the schema contents and the schema capabilities as requirements for representing data evolve. I think that modeling the schema in an event-sourcing sort of way really lends to an extensibility and fluidity of the design that's missing in other schema systems I've used. My biggest misgiving about [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/README.md) is that they cannot be exhaustive but they are also difficult to extend without conflict. I think that the approach you take here neatly addresses this problem and adds much-needed flexibility to a critical portion of the design.

tigrannajaryan:
@tedsuo I think you were interested in this topic, please have a look.

jmacd:
I believe this PR has reached critical mass. Shall we merge this and continue with a new PR on the same document, a new OTEP, or prototypes?

tigrannajaryan:
My plan once this is merged is to start submitting PRs that add the appropriate bits to the specification repo. I already have a small [prototype implementation](https://github.com/tigrannajaryan/telemetry-schema) in Go, which I plan to contribute to our Go SDK and then the next step for me will be adding a schema processor to the Collector using this contributed prototype.

Let's keep this open for another 24 hours and we can merge. 

If anyone has objections to this OTEP please speak now. Once it is merged I will begin materializing it in the spec.

yurishkuro (approved):

text/0152-telemetry-schemas.md
carlosalberto: @tigrannajaryan I'm fine with it. We will have to validate it once this hits the Specification + implementation, so we can still revert if truly needed.

carlosalberto:
>  I already have a small prototype implementation in Go

Would be nice to have a prototype in another language (a dynamic one if possible), just to have further validation. We can discuss this after this PR is merged ;)

jsuereth (approved):
I still have reservations on our ability to define a cohesive migration language in YAML that doesn't eventually turn into gobbledy-gook (or something hard to maintain).

However, on the whole, I really like this proposal for stability and clear evolution.

tigrannajaryan:
> > I already have a small prototype implementation in Go
> 
> Would be nice to have a prototype in another language (a dynamic one if possible), just to have further validation. We can discuss this after this PR is merged ;)

Just to be clear.

Schema conversion is the process that can be CPU intensive. SDKs are not expected to implement schema conversion. This is a backend job or a Collector job. So, I think we don't need to see how slow is schema conversion in slow languages, people are just not supposed to do that in slow languages. My benchmark in Go is representative of what we will see in Collector. We may also want to benchmark in some other fast languages which people may use for backends (Java, etc), but I do not expect to see dramatically different results.

The work that SDKs are supposed to do, is to populate the schema_url field. This is a trivial amount of work and will most likely be below measurement errors in any benchmark that generate telemetry using our SDKs.

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1176 @@
...
+
+Note that if there is a schema url associated with instrumentation library it
+overrides the application-wide schema url as described [here](#otlp-changes).
+
+## OpenTelemetry Schema
+
+OpenTelemetry publishes it own schema at
+`https://opentelemetry.io/schemas/<version>`. The version number of the schema
+is the same as the specification version number which publishes the schema.
+Every time a new specification version is released a corresponding new schema
```
yurishkuro, line 794: It may be useful to require even stronger coupling. Since both schemas and semantic conventions are machine-readable, there could be a simple validation script that makes sure that the new version of the schema accurately reflects the changes to the spec, e.g. if an attribute was removed / renamed from the spec, something about that has to be present in the schema and we should block the release otherwise.

tigrannajaryan: Yes, validation is a good idea.

text/0152-telemetry-schemas.md
```diff
@@ -0,0 +1,1176 @@
...
+      # Definitions that apply to Metric data type.
+      changes:
+        - rename_metrics:
+            # map of key/values. The keys are the old metric name used
+            # in the previous version, the values are the new metric name
+            # starting from this version.
+            container.cpu.usage.total: cpu.usage.total
+            container.memory.usage.max: memory.usage.max
+
+        - rename_labels:
```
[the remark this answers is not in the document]
tigrannajaryan: Thanks. I'll wait until the spec is updated and will update here as well.

weyert:
> Schema conversion is the process that can be CPU intensive. SDKs are not expected to implement schema conversion. This is a backend job or a Collector job. So, I think we don't need to see how slow is schema conversion in slow languages, people are just not supposed to do that in slow languages. 

This makes it sound like Telemetry Schemas is a collector only thing. Does the specification typically discuss collector only features? This would exclude a whole group of people who don't or can use the collector. In that case would you expect vendors implement support for this?

I am worried that in the future there will be a dependency on schemas to do transformations while same SDKs or vendors aren't supporting it.

gramidt (approved):

tigrannajaryan:
> > Schema conversion is the process that can be CPU intensive. SDKs are not expected to implement schema conversion. This is a backend job or a Collector job. So, I think we don't need to see how slow is schema conversion in slow languages, people are just not supposed to do that in slow languages.
> 
> This makes it sound like Telemetry Schemas is a collector only thing. Does the specification typically discuss collector only features? This would exclude a whole group of people who don't or can use the collector. In that case would you expect vendors implement support for this?
> 
> I am worried that in the future there will be a dependency on schemas to do transformations while same SDKs or vendors aren't supporting it.

This was already briefly discussed [above](https://github.com/open-telemetry/oteps/pull/152#discussion_r612022387).

Schemas are not a Collector-only thing. As described in this OTEP, several parties (SDK, Collector, Backend) can participate in a particular way in a schema-aware system. See the diagrams and the explanation in the text: the Collector is an optional element that can help do schema conversions if the backend is not schema-aware. It is perfectly valid to have a system without the Collector and for the backend to handle the schema conversion.

Don't have a Collector and using a backend doesn't know how to handle schemas? That's the worst case and in that case you are exactly where you are today, no worse.

> I am worried that in the future there will be a dependency on schemas to do transformations while same SDKs or vendors aren't supporting it.

Schemas are an optional capability. If you have a backend which is aware of schemas you get nicer features. If the backend is not schema-aware you get what you have today, schemas can be fully ignored in that case. 

My view is that it is not SDK's job to do schema conversions. However, nothing prevents us from implementing such capabilities in SDKs in the future if we decide that it is necessary.

weyert:
Thank you for clarifying @tigrannajaryan. I really liking the idea.

bogdandrutu (approved):

bogdandrutu:
@weyert you can approve :) it will not be green but will still show as approve in the list

weyert (approved):

kumoroku (approved):

PR #153, opened 2021-04-23
New sample value to enable fetching from the Jaeger Collector

bbarin:


bbarin:
Hi there, I've already signed the CLA, however, this PR still fails due to missing signed CLA. Is this bounded to the email?

yurishkuro:
Curious to see if there's interest in this. Just fetching strategies is not enough, this needs support for all types of sampling configuration that Jaeger has, which is a non-trivial amount of work.

PR #156, opened 2021-05-15
 Counter, UpDownCounter, and Gauge instruments compared

jmacd:
This is written in support of https://github.com/open-telemetry/opentelemetry-specification/issues/1297.

### DIFF
--- a/text/metrics/0156-counter-gauge-model.md
+++ b/text/metrics/0156-counter-gauge-model.md
@@ -23,13 +23,13 @@ OpenTelemetry has a founding principal that the interface (API) should
 be decoupled from the implementation (SDK), thus the Metrics project
 set out to define the meaning of metrics API events.
 
-OpenTelemetry uses the term _temporality_ to describe how aggregations
-are accumulated across time, whether they are reset to zero with each
-interval (_delta_) or accumulated over a sequence of intervals
-(_cumulative_).  Both forms of temporality are considered important,
-as they offer a useful tradeoff between cost and reliability.  The
-data model specifies that a change of temporality does not change
-meaning.
+OpenTelemetry uses the term _temporality_ to describe how Sum
+aggregations are accumulated across time, whether they are reset to
+zero with each interval (_delta_) or accumulated over a sequence of
+intervals (_cumulative_).  Both forms of temporality are considered
+important, as they offer a useful tradeoff between cost and
+reliability.  The data model specifies that a change of temporality
+does not change meaning.
 
 OpenTelemetry recognizes both synchronous and asynchronous APIs are
 useful for reporting metrics, and each has unique advantages.  When
@@ -42,19 +42,32 @@ are the totals of a Sum aggregation (i.e., cumulatives).
 
 ## Glossary
 
-_Meaning_ what the events are saying
+_Meaning_: Metrics API events have a semantic definition that dictates
+the meaning of the event, in particular how to interpret the integer
+or floating point number value passed to the API.
 
-_Interpretation_ how we produce information from meaning
+_Interpretation_: How we extract information from metrics data using
+the semantics of the API and the semantics of the OTLP data points.
 
-_Metric instrument_ has a name
+_Metric instrument_ is a named instrument, belonging to an
+instrumentation library, declared with one of the OpenTelemtetry
+Metrics API instruments.  For the purpose of this text, it is a
+Counter, an UpDownCounter, or a Gauge.
 
-_Metric attribute_ is a qualifier, can have many, they are independent
+_Metric attributes_ can be applied to Metric API events, which allows
+interpreting the meaning of events using different subsets of
+attribute dimensions.
 
-_Metric data stream_ is a variable
+_Metric data stream_ is a collection of data points, written by a
+writer, having an identity that consists of the instrument's name, the
+instrumentation library, resource attributes, and metric attributes.
 
-_Metric data points_ are items in a stream
+_Metric data points_ are the items in a stream, each has a point
+kind. For the purpose of this text, the point kind is Sum or Gauge.
+Sum points have two options: Temporality and Monotonicity.
 
-_Metric timeseries_ are the outputs
+_Metric timeseries_ is the output of aggregating a stream of data
+points for a specific set of resource and attribute dimensions.
 
 ## Meaning and interpretation of Counter and UpDownCounter events
 
@@ -310,3 +323,34 @@ In some of these cases, it may be logical but practically impossible
 to use one or more Counter instruments in place of Gauges.  CPU
 utilization can be derived from a usage Counter.  Fan speed can be
 derived from a revolution Counter.
+
+## Summary
+
+The OpenTelemetry Metrics data model supports addition and removal of
+attributes in a way that preserves meaning.  This design gives
+developers the ability to introduce new attributes in a safe way.
+
+OpenTelemetry metrics developers are asked to consider whether they
+want an UpDownCounter or Gauge when making asynchronous measurements,
+and they should make this decision based on whether the default
+aggregation rule for UpDownCounter or Gauge preserves meaning.  This
+decision comes down to whether attributes are meant to subdivide a Sum
+point or qualify a Gauge point.
+
+The default aggregation rules for OpenTelemetry metrics data points
+ensure that meaning is preserved when removing attributes from a
+stream of metrics data.  The rules for reaggregation specify that
+attributes should be safely removed before aggregating with other
+metrics that are missing the same attributes, a process referred to as
+dimensional alignment.
+
+This design allows optional attributes to be included by the SDK in
+metric data when it is available, such as those extracted from
+TraceContext Baggage, in ways that consumers of the metrics data can
+interpret correctly.
+
+Having the ability to automatically remove attributes without changing
+the meaning of Counter, UpDownCounter, and Gauge metrics API events
+makes it possible for OpenTelemetry collectors to be configured with
+re-aggregation rules, which can be managed by users in order to limit
+collection costs.
--- a/text/metrics/0156-counter-gauge-model.md
+++ b/text/metrics/0156-counter-gauge-model.md
@@ -1,0 +1,301 @@
+# Counter, UpDownCounter, and Gauge instruments compared
+
+## Background
+
+The OpenTelemetry Metrics data model and API have introduced a new
+kind of data point, the non-monotonic Sum, and a corresponding
+instrument named UpDownCounter.
+
+OpenTelemetry has a founding principal that the interface (API) should
+be decoupled from the implementation (SDK), thus the Metrics project
+set out to given semantic meaning to the API events.  The events on
+Counter and Gauge instruments mean different things and submit to
+different kinds of interpretation.
+
+OpenTelemetry uses the term temporality to describe how aggregations
+are accumulated across time, whether they are reset to zero with each
+interval ("delta") or accumulated over a sequence of intervals
+("cumulative").  Both temporality forms are considered important, as
+they offer a useful tradeoff between cost and reliability.  The data
+model specifies that a change of temporality does not change meaning.
+
+OpenTelemetry recognizes both synchronous and asynchronous APIs are
+useful for reporting metrics, and each has unique advantages.  When
+used with Counter and UpDownCounter instruments, there is an assumed
+relationship between the aggregation temporality and the choice of
+synchronous or asynchronous API.  Inputs to synchronous
+(UpDown)Counter instruments are the changes of a Sum aggregation
+(i.e., deltas).  Inputs to asynchronous (UpDown)Counter instruments
+are the totals of a Sum aggregation (i.e., cumulatives).
+
+## Meaning vs. Interpretation
+
+The terms "Meaning" and "Interpretation" are used below to describe
+how semantics are conveyed from the producer of metric events,
+typically a developer working in the OpenTelemetry API, and the
+consumer of metric timeseries, typically the user of an observability
+product.
+
+- Meaning is the indicated significance of a metric event, _what_ information the producer is stating through the use of an OpenTelemetry API
+- Interpretation is a recipe for deriving information from a metric event, or _how_ the consumer can benefit from the stated information.
+
+## Meaning and interpretation of Sum points
+
+Sum points are taken to have meaning in a stream, independent of
+aggregation temporality, as follows:
+
+- Sum points are quantities that define a rate of change with respect to time
+- Rate-of-change over time may be used to derive a current total.
+
+The rate interpretation is preferred for monotonic Sum points, and the
+the current total interpretation is preferred for non-monotonic Sum
+points.  Both interpretations are meaningful and useful for both
+monotonic and non-monotonic Sum points.
+
+Sum points imply a linear scale of measurement.  A Sum value that is
+twice the amount of another actually means twice as much of something
+was counted.  Linear interpolation is considered to preserve meaning.
+
+## Meaning and interpretation of Gauge points
+
+Gauge points, which do not have temporality, are taken to have meaning
+in a stream as follows:
+
+- Gauge point values are individual measurements captured at an instant in time
+- Gauge points record the last known value in a series of individual measurements.
+
+Note that these two statements imply different interpretation for
+synchronous and asynchronous measurements.  When recording Gauge
+values through a synchronous API, the interpretation is "last known
+value", and when recording Gauge values through an asynchronous API
+the interpretation is "current value".
+
+The distinction between last known value (synchronous) and current
+value (asynchronous) is considered not significant in the data model.
+Contrasting with Sum points, less can be assumed about the
+measurements:
+
+- No implied linear scale of measurement, therefore
+- Rate-of-change may not be well defined
+- Ratios are not necessarily meaningful
+- Linear interpolation is not necessarily supported.
+
+## Comparing Gauge and non-monotonic cumulative Sum
+
+The non-monotonic Sum point, the one that naturally results from an
+asynchronous UpDownCounter instrument, appears similar in nature to a
+Gauge.  Both can be used to express the current value in a series.
+
+As stated, there is a difference between Sum and Gauge points in the
+data model that can be seen when adding and removing attributes from
+metric events.
+
+## Attributes are used for interpretation
+
+The meaning of an attribute is derived from the interpretation of the
+metric.  Attributes are used to logically restrict attention to a
+subset of metric events, therefore the use of attributes leads to
+additional interpretation.
+
+Because attributes function logically as event selectors, we are able
+to assert that the addition of an attribute does not by definition
+change the meaning of a metric event.  The addition of an attribute
+may create new timeseries, by producting a of greater number of
+distinct attribute sets, though the meaning is unchanged.
+
+For example, any metric event with no attributes:
+
+```
+gauge.Set(value)
+```
+
+can be extended by a new attribute, without changing its meaning or
+interpretation:
+
+```
+gauge.Set(value, { 'property': this.property })
+```
+
+## New measurements: Counter and UpDownCounter instruments
+
+Sum points have been defined to have linear scale of measurement,
+therefore Sum points can be subdivided.  A single Counter event can be
+logically replaced by multiple Counter events having an equal sum.
+This property allows the producer of metric events to introduce new
+measurements, while preserving existing interpretation.
+
+For example, it is reasonable to replace a single Counter event adding `x+y`:
+
+```
+counter.Add(x+y)
+```
+
+with separate counter events and one additional attribute:
+
+```
+counter.Add(x, { 'property': 'X' })
+counter.Add(y, { 'property': 'Y' })
+```
+
+This property for Sum points makes it possible to configure an
+instrumentation library with or without subdivided Sums, with no
+change in the meaning conveyed.
+
+## New measurements: Gauge instruments
+
+Gauge instruments, unlike Counter instruments, cannot be subdivided.
+Likewise, multiple Gauge measurements cannot be meaningfully combined
+using addition.  As meaning is derived from individual values, new
+measurements can be introduced using attributes to create distinct
+streams.
+
+By interpreting Gauge points as a distribution of current values, new
+measurements can be defined to be preserve meaning.
+
+For example, a Gauge for expressing a vehicle's speed relative to the
+ground can be expressed either as the speed of its midpoint or by an
+independent measurement of the speed of each wheel.
+
+```
+speedGauge.Set(vehicleSpeed)
+```
+
+This can be replaced by one Gauge per wheel, since wheel speed and
+vehicle speed each define vehicle speed relative to the ground:
+
+```
+for i := 0; i < 4; i++ {
+  speedGauge.Set(wheelSpeed[i], { 'wheel': i })
+}
+```
+
+This form of Gauge rewrite is generally useful when additional
+measurements offer more information and are drawn from the same
+distribution.
+
+## Meaning-preserving attribute erasure
+
+Several rules for rewriting metric events that preserve meaning have
+been shown above, focused on introducing new attributes and new
+measurements in ways do not change existing meaning and
+interpretation.
+
+Removing attributes from metric events does not, by definition, change
+their meaning, since attributes are interpreted as event selectors.
+Removing attributes from aggregated streams of OpenTelemetry Metrics
+data requires attention to the meaning being conveyed.
+
+Safe attribute erasure for OpenTelemetry Metrics streams is specified
+in a way that preserves meaning while removing only the forms of
+interpretation that made use of the erased attribute.
+
+Re-aggregation describes the process of combining OpenTelemetry
+metrics streams.  For re-aggregation to preserve meaning, Sum points
+must be combined by adding the inputs and Gauge points must be
+combined by forming a distribution of the inputs.
+
+Note that erasure of attributes is defined so that it reverses the
+effect of introducing new measurements, and meaning is preserved in
+both directions.  This explains the definition for default
+aggregations that should be applied when re-aggreation OpenTelemetry
+metrics streams.  Sum streams are re-aggregated to preserve the
+implied rate, while Gauge points are reggregated to preserve the
+implied distribution.
+
+## Conveying meaning to the user
+
+OpenTelemetry states a requirement separating the API from the
+implementation, and to do so we have defined the meaning of metrics
+API events.  To preserve meaning through stages of reaggregation, we
+have specified distinct default aggregation rules for Counter and
+Gauge streams.
+
+When attributes are used with Counter and Gauge instruments, every
+distinct combination of attribute values determines a separate
+OpenTelemetry metrics stream, and each stream conveys meaning
+independently.  Because meaning is independent from the attributes
+used, the user may wish to disregard some attributes when interpreting
+a stream of metrics, restricting their attention to a subset of
+attributes.
+
+In database systems, this process is refered to as a performing a
+"Group-By", where aggregation is used to combine streams within each
+distinct set of grouped attributes.  For the benefit of OpenTelemetry
+users, Metrics systems are encouraged to choose a a meaning-preserving
+aggregation when grouping metric streams to convey meaning to the
+user.
+
+When conveying meaning to the user by grouping and aggregating over a
+subset of attribute keys, the default aggregation selected should be
+one that preserves meaning.  For monotonic Counter instruments, this
+means conveying the combined rate of each group.  For UpDownCounter
+instruments, this means conveying the combined total of each group.
+For Gauge instruments, this means conveying the combined distribution
+of each group.
+
+## Choice of UpDownCounter or Gauge
+
+The OpenTelemetry UpDownCounter instrument resembles the Gauge
+instrument, but streams generated from these instruments apply
+different aggregation rules by default.  The choice of instrument
+should be made to ensure that the default aggregation rule preserves
+meaning, as that is the point of these definitions.
+
+Examining Gauge instruments in existing systems for anecdotal evidence
+suggests that a significant majority of Gauges should be written as
+UpDownCounters in OpenTelemetry.  Examples are given below.
+
+### UpDownCounter measurements
+
+UpDownCounter instruments are used for capturing quantities, where
+typical examples include:
+
+- Queue size
+- Memory size
+- Cache size
+- Active requests
+- Live object count
+
+To test that these quantities are suitable UpDownCounter measurements,
+verify that adding two inputs together logically produces another of
+the same type and scale of measurement.  A queue size plus a queue
+size yields a queue size, for example; add one count of live objects
+with another, and you have a count of live objects.  By choosing the
+UpDownCounter, developers ensure that the meaning conveyed is a sum,
+which ensures the correct rate interpretation.
+
+When interpreting total sums aggregated from UpDownCounter
+instruments, it is important to consider the set of contributing
+attributes, which determine the scale of measurement.  If one server
+outputs UpDownCounter data in two attribute dimensions while another
+uses three attribute diensions, the mean value is not a meaningful
+quantity.  The process of correcting mixed attribute dimensions for
+cumulative sums is referred to as _dimensional alignment_.
+
+### Gauge measurements
+
+Gauge instruments are used for capturing physical measurements,
+calculated ratios, and results of function evaluation.  For example:
+
+- CPU utilization
+- CPU temperature
+- Fan speed
+- Water pressure
+- Success/failure ratio
+
+To test that these are suitable Gauge measurements, verify that adding
+two inputs together does not logically produce a measurement of the
+same type.
+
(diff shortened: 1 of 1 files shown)


text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+
+The rate interpretation is preferred for monotonic Sum points, and the
+the current total interpretation is preferred for non-monotonic Sum
+points.  Both interpretations are meaningful and useful for both kinds
+of Sum point.
+
+Sum points imply a linear scale of measurement.  A Sum value that is
+twice the amount of another actually means twice as much of the
+variable was counted.  Linear interpolation is considered to preserve
+meaning.
```
ltyson, line 89: You might consider adding a concrete example of a real life metric to help the reader follow the precise but very technical description above (maybe even a diagram, if it lends itself to that). There are some examples I've seen in other places that you could probably just steal, e.g. https://lightstep.com/blog/opentelemetry-101-what-are-metrics/

I'm thinking that would help people like me who learn more readily through examples. Same below with Gauge events.

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+can be extended by a new attribute, without changing its meaning or
+altering any existing interpretation:
+
+```
+gauge.Set(value, { 'property': this.property })
+```
+
+## New measurements: Counter and UpDownCounter instruments
+
+Sum points have been defined to have linear scale of measurement,
```
ltyson, line 145: I had some trouble deciphering the phrase "linear scale of measurement". I understand the concept, after reading the rest of the paragraph, but I wonder if there's a different way to say this, maybe by adding an explanatory aside, something like this (although it probably doesn't use the right terminology)

"Sum points have been defined to have linear scale of measurement, meaning the same Sum point value could be obtained through many different combinations of metric event values. This property can also be applied in reverse, meaning that Sum points can be subdivided."

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+therefore Sum points can be subdivided.  A single Counter event can be
+logically replaced by multiple Counter events having an equal sum.
+This property allows the producer of metric events to introduce new
```
ltyson, line 148: maybe: "introduce new measurements" => "introduce new attributes that subdivide measurements"? (maybe that's the wrong terminology though)

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+```
+counter.Add(x, { 'property': 'X' })
+counter.Add(y, { 'property': 'Y' })
+```
+
+This property for Sum points makes it possible to configure an
+instrumentation library with or without subdivided Sums and to
+meaningfully aggregate data with a mixture of attributes.
+
+## New measurements: Gauge instruments
```
ltyson, line 168: Nice! This section is super clear and easy to understand.

nit: "New measurements" => "Adding new measurements" for clarity? Same below.

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+With Counter instruments, a new attribute can be introduced with
+additional measurements to subdivide a variable count.
+
+With Gauge instruments, a new attribute can be introduced with
+additional measurements to make multiple observations of a variable.
+
+The OpenTelemetry Metrics API introduces a new kind of instrument, the
+UpDownCounter, that behaves like a Counter, meaning that attributes
+subdivide the variable being counted, but their primary interpretation
+is like that of a Gauge.
```
ltyson, line 18: At this point I'm thinking, "ok, that's nice, but why should I care?" 

I would suggest starting this doc at the very top by briefly illustrating the problem that exists when you only have Counter and Gauge instruments, to help the reader see that things aren't fine the way they are, and motivate them to understand the problem and care about this solution.

Actually, you might just start with the use-case that you wrote up for me in our slack chat about this.

```
The user journey I'm after is this:
1. As a user I know nothing about metrics named random_oss_software.* but they were produced with OTel SDKs.
2. I click "Create a dashboard"
3. I enter the "random_oss.software.*" pattern
4. I get a good dashboard.
```

and then show at least one scenario (doesn't have to be exhaustive) where this is not possible to do without the UpDownCounter instrument.

Then you could move into the rest of the doc as written.

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+data.  Metric attributes add information without changing the value of
+a metric event.  Addition and removal of metric attributes can be
+accomplished safely by applying transformations that preserve meaning.
+
+Addition of attributes on a metric event can create new timeseries, by
+producting a of greater number of distinct attribute sets.  However,
+the meaning in the original events is preserved in the complete set of
+timeseries.
+
+Removing attributes from metric streams without changing meaning
```
ltyson, line 126: I'd suggest moving this down below the example, and adding a sentence afterward to draw attention to the problem, e.g. 

"...which means applying the natural aggregation function to merge metric streams. But how do we know which is the "natural" aggregation function? For Counters the answer is always SUM (for: reason), however for Gauges it might be SUM or MEAN depending on the semantics of the values represented by the particular metric. (for: reason). Therein lies the problem."

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+The distinction between last known value (synchronous) and current
+value (asynchronous) is considered not significant in the data model.
+Contrasting with Sum points, less can be assumed about the
+measurements.  No implied linear scale of measurement, therefore:
+
+- Rate-of-change may not be well defined
+- Ratios are not necessarily meaningful
+- Linear interpolation is not necessarily supported.
+
+## Attributes are used for interpretation
```
ltyson, line 114: This section header doesn't quite seem to flow naturally with the text, maybe call this section something 
like "Adding attributes to events for additional meaning" to help tie it to previous and subsequent sections.

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+instrumentation library, resource attributes, and metric attributes.
+
+_Metric data points_ are the items in a stream, each has a point
+kind. For the purpose of this text, the point kind is Sum or Gauge.
+Sum points have two options: Temporality and Monotonicity.
+
+_Metric timeseries_ is the output of aggregating a stream of data
+points for a specific set of resource and attribute dimensions.
+
+## Meaning and interpretation of Counter and UpDownCounter events
```
ltyson, line 72: I don't know if this is what you're going for, but if your main goal is to explain the need for UpDownCounter, it might make sense to omit it from this section and wait to present it later as a solution to the problem described in this doc. If that's the goal, it feels a bit premature to describe it here, because I as a reader don't yet know why it's necessary or useful and I get a bit lost.

ltyson (commented):
A few overall comments:
* the introduction feels too long and doesn't really make it clear what the problem is -- I think that's the weakest part right now
* I would suggest adding concrete examples of real-world metrics for each of the instrument types, to help example-based-learners (like me) to understand more easily

dashpole (approved):

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+Metrics API instruments.  For the purpose of this text, it is a
+Counter, an UpDownCounter, or a Gauge.
+
+_Metric attributes_ can be applied to Metric API events, which allows
+interpreting the meaning of events using different subsets of
+attribute dimensions.
+
+_Metric data stream_ is a collection of data points, written by a
+writer, having an identity that consists of the instrument's name, the
+instrumentation library, resource attributes, and metric attributes.
```
AjohnsonLS, line 63: Is this meant to help define an interface/standard or simply saying that a data stream encompasses all of the data and the source?

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+
+_Metric data points_ are the items in a stream, each has a point
+kind. For the purpose of this text, the point kind is Sum or Gauge.
+Sum points have two options: Temporality and Monotonicity.
```
AjohnsonLS, line 67: Temporality is previously defined, but not Monotonicity.

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+value", and when recording Gauge values through an asynchronous API
+the interpretation is "current value".
+
+The distinction between last known value (synchronous) and current
+value (asynchronous) is considered not significant in the data model.
+Contrasting with Sum points, less can be assumed about the
+measurements.  No implied linear scale of measurement, therefore:
+
+- Rate-of-change may not be well defined
+- Ratios are not necessarily meaningful
```
AjohnsonLS, line 111: These statements feel too strong to me. I can think of instances where gauges are very useful in context of rate, ratios, and trend modeling. However, I think it's important to say that Gauges are signed with their own time stamp. Valuable analysis likely needs some data alignment (aggregating by some method to regular time intervals) in order to make meaningful comparisons

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+## New measurements: Gauge instruments
+
+Gauge instruments, unlike Counter instruments, cannot be subdivided.
+Multiple Gauge measurements cannot be meaningfully combined using
+addition.  In the time dimension, Gauge instrument events are
+aggregated by taking the last value.
+
+The same aggregation can be applied when removing an attributes from
+metric streams forces reaggregation.  The most current value should be
+selected.  In case of identical timestamps, a random value should be
```
AjohnsonLS, line 177: The random value sample for time collisions is an important decision that should be clearly documented. There's a lot of behavior that surfaces when this happens that is difficult to understand. I do think it's a good solution, just needs to be clearly documented.

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+important, as they offer a useful tradeoff between cost and
+reliability.  The data model specifies that a change of temporality
+does not change meaning.
+
+OpenTelemetry recognizes both synchronous and asynchronous APIs are
+useful for reporting metrics, and each has unique advantages.  When
+used with Counter and UpDownCounter instruments, there is an assumed
+relationship between the aggregation temporality and the choice of
+synchronous or asynchronous API.  Inputs to synchronous
+(UpDown)Counter instruments are the changes of a Sum aggregation
```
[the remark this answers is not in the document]
reyang: +1.

I would suggest that we avoid "incremental" since it might suggest monotonicity (or we will have to put something like "incremental/decremental").
My suggestion would be "Inputs to synchronous (UpDown)Counter instruments are the deltas, while inputs to asynchronous instruments are always the absolute value (e.g. the total sum)."

text/metrics/0156-counter-gauge-model.md
ocelotl, line 18: I find this a bit hard to understand.

_The OpenTelemetry Metrics API introduces a new kind of instrument, the ￼UpDownCounter, that behaves like a Counter,_

Perfect, so far so good :+1: 

_meaning that attributes subdivide the variable being counted,_

It starts getting a bit confusing here. Who do these attributes belong to? I assume they belong to the UpDownCounter instrument. The _behaves like a Counter, meaning that attributes ￼subdivide the variable being counted_ tells me that the Counter attributes subdivide the variable being counted, and since UpDownCounter behaves like a Counter, then I understand that UpDownCounter has attributes that also subdivide the variable being counted.

_but their primary interpretation ￼is like that of a Gauge._

Ok, here it gets confusing. the usage of "but" here makes me understand that the difference between Counter and UpDownCounter is that the primary interpretation of the latter is like that of a Gauge. What does that mean? The difference between the words Counter and UpDownCounter is "UpDown" so, is the "UpDown" what a Gauge does? :shrug: I was expecting something like this instead _but the UpDownCounter is non monotonic_.

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+OpenTelemetry has a founding principal that the interface (API) should
+be decoupled from the implementation (SDK), thus the Metrics project
+set out to define the meaning of metrics API events.
+
+OpenTelemetry uses the term _temporality_ to describe how Sum
+aggregations are accumulated across time, whether they are reset to
+zero with each interval (_delta_) or accumulated over a sequence of
+intervals (_cumulative_).  Both forms of temporality are considered
+important, as they offer a useful tradeoff between cost and
+reliability.  The data model specifies that a change of temporality
```
ocelotl, line 31: The temporality of a sum aggregation can change? I mean, can it be delta now, cumulative later, then delta again?

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+does not change meaning.
+
+OpenTelemetry recognizes both synchronous and asynchronous APIs are
+useful for reporting metrics, and each has unique advantages.  When
+used with Counter and UpDownCounter instruments, there is an assumed
```
ocelotl, line 36: What is being used with counter and updowncounter instruments?

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+relationship between the aggregation temporality and the choice of
+synchronous or asynchronous API.  Inputs to synchronous
```
ocelotl, line 38: Is the API synchronous or asynchronous? Would it be more correct to say instruments are synchronous or asynchronous?

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+kind. For the purpose of this text, the point kind is Sum or Gauge.
+Sum points have two options: Temporality and Monotonicity.
+
+_Metric timeseries_ is the output of aggregating a stream of data
+points for a specific set of resource and attribute dimensions.
+
+## Meaning and interpretation of Counter and UpDownCounter events
+
+Counter and UpDownCounter instruments produce Sum metric data
+points that are taken to have meaning in a metric stream, independent
```
ocelotl, line 75: What does _that are taken to have meaning in a metric stream_ mean?

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+of the aggregation temporality, as follows:
+
+- Sum points are quantities that define a rate of change with respect to time
+- Rate-of-change over time combined with a reset time may be used to derive a current total.
+
+The rate interpretation is preferred for monotonic Sum points, and the
```
ocelotl, line 81: I think the 2 previous points are interpretations, right?

If that is the case, referring to one as "the rate interpretation" is a confusing because both points use the word rate. Better to number these interpretations and refer to them to the first or second one.

ocelotl, line 81:
```suggestion
The rate interpretation is preferred for monotonic Sum points, and
```

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+the current total interpretation is preferred for non-monotonic Sum
+points.  Both interpretations are meaningful and useful for both kinds
```
ocelotl, line 83: Bit confused here, first this says that one interpretation is preferred for a certain kind of sum point and the other one is preferred for the other kind of sum point. Then it says that both are "meaningful and useful" for both kinds of sum point. Then, why is one preferred over the other? :shrug:

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+of Sum point.
+
+Sum points imply a linear scale of measurement.  A Sum value that is
+twice the amount of another actually means twice as much of the
+variable was counted.  Linear interpolation is considered to preserve
```
ocelotl, line 88: Sorry, what do you mean with "linear interpolation is considered to preserve meaning"? Do you mean that it is possible to interpolate linearly between two sum points?

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+meaning.
+
+## Meaning and interpretation of Gauge events
+
+Gauge instruments produce Gauge metric data points are taken to
+have meaning in a metric stream as follows:
+
+- Gauge point values are individual measurements captured at an instant in time
```
ocelotl, line 96: Aren't sum points also individual measurements captured at an instant in time?

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+- Gauge points record the last known value in a series of individual measurements.
+
+Note that these two statements imply different interpretation for
+synchronous and asynchronous measurements.  When recording Gauge
+values through a synchronous API, the interpretation is "last known
+value", and when recording Gauge values through an asynchronous API
+the interpretation is "current value".
```
ocelotl, line 103: Hm, "last known value" and "current value" are very similar. It is hard to understand the difference between synchronous  gauge and asynchronous gauge using this concept.

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+
+The distinction between last known value (synchronous) and current
+value (asynchronous) is considered not significant in the data model.
```
ocelotl, line 106: Hm, but it is significant for the reader of this document to understand the difference between synchronous gauge and asynchronous gauge...

ocelotl (commented):
There are several things that are hard for me to understand. It is very likely that this is the case because of how little do I know of the concepts of metrics and the terminology used to describe them. Nevertheless, maybe this can be useful information to the author in case it is possible to rephrase some of this OTEP in order to make it a bit more clear to more inexperienced readers :slightly_smiling_face:

text/metrics/0156-counter-gauge-model.md
jsuereth: If you're talking version of  a service, possibly.   I could code service A to use DELTAs, then decide to switch to cumulatives in version 2.

Practically you shouldn't expect a service to be going back and forth between delta + cumulative points.

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+Gauge instruments produce Gauge metric data points are taken to
+have meaning in a metric stream as follows:
+
+- Gauge point values are individual measurements captured at an instant in time
+- Gauge points record the last known value in a series of individual measurements.
+
+Note that these two statements imply different interpretation for
+synchronous and asynchronous measurements.  When recording Gauge
+values through a synchronous API, the interpretation is "last known
+value", and when recording Gauge values through an asynchronous API
```
jsuereth, line 102: Not sure I buy this.  In a synchronous Gauge, recording a value is "Here's the current value".

The difference in Async vs. Sync is whether you can have the "current value" or "last known value" be sampled on-demand.

i.e. I think the terminology here is being phrased from an SDK/exporter perspective vs. the instrument's perspective (which I assume is closer to a user of metrics API)

text/metrics/0156-counter-gauge-model.md
jsuereth, line 106: I think you're implicitly saying that when you export metrics, you're setting the Timestamp for a synchronous Gauge to "export time", whereas the actual timestamp was when the synchronous instrument sent the last piece of data.

E.g. Reproting INterval
```
|       X       X           |
t0     t1      t2           t3
```

Here, the synchronous gauge is reporting its value at t1 and t2.  For the collection interval t0 -> t3, we report the value recorded at t2 *BUT* you're saying for DELTA + CUMULATIVE it would report its timestamp at t3, whereas an asynchronous instrument would just capture a value at t3.


I think this needs some kind of picture.

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+measurements in ways do not change existing meaning or alter existing
+interpretations.
+
+Removing attributes from metric events does not, by definition, change
+their meaning, since attributes are interpreted as event selectors.
+Removing attributes from aggregated streams of OpenTelemetry Metrics
+data requires attention to the meaning being conveyed.
+
+Safe attribute erasure for OpenTelemetry Metrics streams is specified
+in a way that preserves meaning while removing only the forms of
```
jsuereth, line 213: What if "removing the forms of interpretation that made use of the erased attribute" basically is destroying the practical value of the metric?


I think this is implying that removing an attribute from a metric is an ok thing to do (it' s not).  The meaning that IS preserved is relevant to reaggregation + processing, but could be disastrous to dashboards and users.

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+distinct combination of attribute values determines a separate
+OpenTelemetry metrics stream, and each stream conveys meaning
+independently.  Because meaning is independent from the attributes
+used, the user may wish to disregard some attributes when interpreting
+a stream of metrics, restricting their attention to a subset of
+attributes.
+
+In database systems, this process is refered to as a performing a
+"Group-By", where aggregation is used to combine streams within each
+distinct set of grouped attributes.  For the benefit of OpenTelemetry
```
jsuereth, line 247: This kind of falls over for me in practice.  Database Group-By is a full query language where NORMALLY you need to provide your own Aggregation semantic for every column of data *or* the query will fail.

What we're suggesting here is instead of:
```
SELECT AVG(count) FROM x GROUP BY label
```

The default aggregation function is *infferred* by otel based on the metric type.

Two things:

1. We shouldn't prevent users from using different aggregation functions in practice.  Indeed, once they make it to PromQL (or MQL for GCP), they'll be able to do whatever they want here and need to track their own meaning.
2.I assume a lot of the gymnastics we go through around default aggregation is to avoid exposing a query-like langauge for "rewrite rule" style collection behavior. 

My $.02 here is the focus on giving users a way to solve "rewrite rules" use cases is good.  Making it as easy as possible is good.  If we can't explain what *is and isn't safe* in very simple terms, we might be in trouble.   If the "meaning" we retain isn't the one users wanted to retain, then we're not really adding value.

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+users, Metrics systems are encouraged to choose a a meaning-preserving
+aggregation when grouping metric streams to convey meaning to the
+user.
+
+When conveying meaning to the user by grouping and aggregating over a
+subset of attribute keys, the default aggregation selected should be
+one that preserves meaning.  For monotonic Counter instruments, this
+means conveying the combined rate of each group.  For UpDownCounter
```
jsuereth, line 255: Why does monotonic counter instruments turn into a rate?  Did I miss an above description on how UpDown vs. Counter have different aggregation meaning?

jsuereth (commented):
I like a lot things in this document, but I think it suffers from a few things.


1. the bottom section outlining when to use Gauge vs. UpDownCounter is gold and should be teased apart or put first.

2. The notion of "safe operations" to perform regarding Attributes (like removal/addition) should be formalized as part of the DataModel.   I think you took a good first crack, but we may want to flesh out ALL of these things so that it's clear what operations are allowed in "telemetry schema" and which are considered "compatible"

3. The distinction of "last value" vs. "current value" is (currently) a bit confusingly worded.   It also makes me wonder if we should represent SyncGauge differently in OTLP regarding "start timestamp" and "current timestamp".   E.g. should we ALWAYS have sync gauge use the timestamp when value was recorded vs. when it is reported externally?

text/metrics/0156-counter-gauge-model.md
lakamsani: I read that line to mean that the terms synchronous and asynchronous when used with Counter and UpDownCounter. As in synchronous Counter vs asynchronous Counter etc. ([SumObserver](https://opentelemetry.lightstep.com/metrics/) as I read on Lightstep docs, also see the comparison table)

text/metrics/0156-counter-gauge-model.md
```diff
@@ -0,0 +1,356 @@
...
+
+_Interpretation_: How we extract information from metrics data using
+the semantics of the API and the semantics of the OTLP data points.
+
+_Metric instrument_ is a named instrument, belonging to an
+instrumentation library, declared with one of the OpenTelemtetry
+Metrics API instruments.  For the purpose of this text, it is a
+Counter, an UpDownCounter, or a Gauge.
+
+_Metric attributes_ can be applied to Metric API events, which allows
```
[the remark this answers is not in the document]
lakamsani: I would imagine attributes are specific to each metric so the spec wouldn't define what the attributes are. This is what is [referred to as labels](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L89) in the ASCII art in the spec but the agreed-upon name going forward is [attributes](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L312).

text/metrics/0156-counter-gauge-model.md
lakamsani: nit: `removing an attributes` => either "removing an attribute" or "removing attributes"

text/metrics/0156-counter-gauge-model.md
lakamsani: Maybe this [description](https://github.com/jmacd/oteps/blob/jmacd/countergauge/text/metrics/0156-counter-gauge-model.md#meaning-and-interpretation-of-counter-and-updowncounter-events)?

jmacd:
I will close this until such a point in time as someone convinces me it's needed.

ghannington-Rocket:
## Strong argument for exporting an Asynchronous UpDownCounter to OTLP as a non-monotonic sum rather than as a gauge?

In Stack Overflow, I recently asked the question
"[Strong argument for exporting an OpenTelemetry Asynchronous UpDownCounter to OTLP as a non-monotonic sum rather than as a gauge?](https://stackoverflow.com/q/79018263)".

A Stack Overflow user commented on my question with a link to this (closed) OTEP. (Thank you! 👍)

### In brief

While this OTEP is interesting, I didn't find an answer here to that question.

I invite answers to that question; either here, or (I guess, preferably) in Stack Overflow. (I'm interested in the answer, wherever it might be.)

### In detail

I actually have someone asking me now, "Why bother exporting values from Asynchronous UpDownCounter and Asynchronous Gauge instruments differently in OTLP? Why not just export them all in the OTLP as gauges?"

I don't have a good answer for them. As I wrote in that Stack Overflow question:

> "Because it generates clarity to the semantic of the metrics stream" \[a reason given in the OpenTelemetry docs topic Specs / OTel 1.37.0 / Metrics / [Supplementary Guidelines](https://opentelemetry.io/docs/specs/otel/metrics/supplementary-guidelines/)\] sounds like weak sauce to me, since, for example, the Collector transform processor seems not to care whether it's aggregrating gauges or sums.

No offense intended, but the following, from this OTEP:

> Metrics systems are encouraged to choose a a meaning-preserving aggregation when grouping metric streams to convey meaning to the user

sounds similarly weak to me ("are encouraged").

Especially, given this in the OTEP:

> Multiple Gauge measurements cannot be meaningfully combined using addition.

whereas, the Collector transform processor seems to have no problem applying a sum function to gauges; that is, combining multiple Gauge measurements using addition. Which, according to this OTEP, isn't a meaningful operation. (Please correct me if I'm wrong! I confess I haven't tried that in practice with gauges in the transform processor; for now, I'm just reading the [transform processor docs](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor).)

reyang:
> ## Strong argument for exporting an Asynchronous UpDownCounter to OTLP as a non-monotonic sum rather than as a gauge?

Non-monotonic sum has additional semantics that gauge doesn't provide. Keeping such semantic will help downstream consumers to pick a better experience if they want to do so.

> "Because it generates clarity to the semantic of the metrics stream" [a reason given in the OpenTelemetry docs topic Specs / OTel 1.37.0 / Metrics / [Supplementary Guidelines](https://opentelemetry.io/docs/specs/otel/metrics/supplementary-guidelines/)] sounds like weak sauce to me, since, for example, the Collector transform processor seems not to care whether it's aggregrating gauges or sums.

I want to understand more about this part "sounds like weak sauce". Would you share more about what makes you feel this is weak?

ghannington-Rocket:
Hi @reyang,

Thank you so much for responding to my comment, much appreciated.

You wrote:

> Non-monotonic sum has additional semantics that gauge doesn't provide.

Yes. I understand and acknowledge that. Specifically, in terms of the representation of that difference in OTLP (I doubt this is news to you, but just so you know that I know 🙂):
- A non-monotonic sum is represented in OTLP as a `sum` with `is_monotonic: false`
- A gauge is represented in OTLP as a `gauge`

But, what do you say to someone who shrugs at that "additional semantics" statement, and says, "So what? Show me an example where that makes a _practical difference_! Otherwise, I see no _practical reason_ to make that distinction in the OTLP; I'm going to continue exporting non-monotonic sums to OTLP as `gauge`." This is the situation I find myself in.

I went looking at OpenTelemetry Collector, expecting to find a clear example. Something that would stand out to me, that I could point to and say, "There! This is why the distinction is important in practice!".

But, so far, I've failed. In fact, what I've seen indicates that the OpenTelemetry transform processor agnostically gives users the freedom to pick a "bad" experience; to perform aggregrations that this OTEP characterizes as not meaningful; to add multiple gauge values. From the [OpenTelemetry Collector transform processor docs](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor):

> The following metric types can be aggregated:
>
> ...
> - gauge
>
> Supported aggregation functions are:
>
> ...
>
> - sum

You wrote:

> Would you share more about what makes you feel this is weak?

I can't see a strong *practical reason* to make this distinction in the OTLP. I want to, but I can't find one.

In particular:

- The OpenTelemetry Collector apparently allows you to add gauge values together, making a nonsense of the non-additive semantics of gauges.

- Telemetry backends collapse non-monotonic sums and gauges to the same single "gauge" type. (Corresponding to the [mapping to a timeseries data model](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#point-kinds) in the OpenTelemetry docs.)

Can you point me to actual examples of where these different representations in OTLP (`sum` with `is_monotonic: false` versus `gauge`) result in different behavior in a consumer of that OTLP, such as Collector or a telemetry backend? Where "different behavior" might, for example, include _restrictions_ on how the values are, or can be, aggregated?

reyang:
> Can you point me to actual examples of where these different representations in OTLP (`sum` with `is_monotonic: false` versus `gauge`) result in different behavior in a consumer of that OTLP, such as Collector or a telemetry backend? Where "different behavior" might, for example, include _restrictions_ on how the values are, or can be, aggregated?

I think telemetry vendors are in a better position to provide such examples (and this is where vendors actually compete).
I can share something that I'm trying to move to stable recently - cardinality limits https://github.com/open-telemetry/opentelemetry-specification/pull/4222. For non-monotonic sum, if the user run into overflow, https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#overflow-attribute, they will still have the accurate sum, that won't be available if gauge is used.

ghannington-Rocket:
Re: Strong argument for exporting an Asynchronous UpDownCounter to OTLP as a non-monotonic sum rather than as a gauge?

The strongest answer I can think of right now is based purely on semantics, rather than the behavior of any consumer.

By representing a numeric value in OTLP as a `gauge`, you are announcing to the "world" (consumers of the OTLP) that, to quote this OTEP:

> measurements cannot be meaningfully combined using addition

That is, the values are _non-additive_.

Whereas, non-monotonic cumulative sums *can* be meaningfully combined using addition. The values are _additive_.

Suppose an OTLP exporter faithfully exports non-additive values (from an Asynchronous Gauge instrument) to the big, wide world as `gauge` and additive values (from an Asynchronous UpDownCounter instrument) as `sum` with `is_monotonic: false` (and `aggregation_temporality: AGGREGATION_TEMPORALITY_CUMULATIVE`).

The first stop in the outside world might be OpenTelemetry Collector. Collector says, "You wanna aggregate those gauges by adding them together? No problem, buddy."

The last stop is a telemetry backend. All the backends I know say, "Sum, gauge, I don't care. It's all the same to me. You wanna add temperatures together? That's your business. Knock yourself out."

So much for semantics.

I've spent the last few years helping to forward asynchronous metrics from proprietary metrics systems (predating OpenTelemetry) to various backends, such as the Elastic Stack and Splunk.

In my experience (although, I acknowledge, my skills are no longer as current as they were in this area), those backends ingested numbers. Their data models didn't make any distinction between additive and non-additive values. It was up to you, the users, to understand the nature of your data.

In that sense, I like the idea of a data model that makes semantic distinctions such as "additive" and "non-additive".

I'm looking for compelling practical reasons to implement that semantic distinction. So far, I haven't found any.

PR #157, opened 2021-05-16
OTLP over message systems

alexvanboxel:


text/0000-otlp-over-messaging-systems.md
```diff
@@ -0,0 +1,110 @@
+# OTLP over messaging systems
```
arminru, line 1: Please make sure to rename your file to the PR # you've been assigned (0157).

text/0000-otlp-over-messaging-systems.md
```diff
@@ -0,0 +1,110 @@
...
+
+| value | trace | metric | log | description |
+|---|---|---|---|---|
+| `otlp_proto` (default) | X | X | X | protobuf serialized from the `Export(Trace/Metric/Log)ServiceRequest` |
+| `otlp_json` | X | X | X | proto3 json representation of a `Export(Trace/Metric/Log)ServiceRequest` |
+| `jaeger_proto` | X | - | - | the payload is deserialized to a single Jaeger proto `Span` |
+| `jaeger_json` | X | - | - | the payload is deserialized to a single Jaeger JSON Span using `jsonpb` |
+| `zipkin_proto` | X | - | - | the payload is deserialized into a list of Zipkin proto spans |
+| `zipkin_json` | X | - | - | the payload is deserialized into a list of Zipkin V2 JSON spans |
+| `zipkin_thrift` | X | - | - | the payload is deserialized into a list of Zipkin Thrift spans |
```
tigrannajaryan, line 70: These are unrelated to OTLP. I think the shouldn't be in this OTEP.

text/0000-otlp-over-messaging-systems.md
```diff
@@ -0,0 +1,110 @@
...
+Because the OTLP payload that is sent over messaging systems is
+well-defined, it’s easy to implement new systems consistently.
+An implementation should at least support `otlp_proto`, meaning that the
+payload is Protobuf serialized from `ExportTraceServiceRequest` for traces,
+`ExportMetricsServiceRequest` for metrics, or `ExportLogsServiceRequest` for
+logs.
+
+Optionally an implementation could support `otlp_json` or other alternative
+encodings like jaeger or zipkin payloads. Alternative encodings should be configured
+by an `encoding` field in the configuration file.
```
tigrannajaryan, line 42: What configuration file is this referring to? Does this assume that every implementation of OTLP over messaging system must have a configuration file? It seems like a stretch to me. Can it be a in-memory config option? Or an API parameter of a library that implements this OTEP?

text/0000-otlp-over-messaging-systems.md
```diff
@@ -0,0 +1,110 @@
...
+
+For the user it’s beneficial that both an exporter, and a receiver are
+implemented in the OpenTelemetry Collector, but SDK developers could
+also implement a language specific exporter.
+
+For log data, implementors of a receiver are encouraged to also support
+receiving raw data from a topic. This enables scenarios were
+non-OpenTelemetry log producers can produce a stream of logs, either
+structured or unstructured, on that topic and the receiver wraps the log
+data in a configurable resource.
```
tigrannajaryan, line 52: This seems unrelated to OTLP. Should it be in this OTEP?

text/0000-otlp-over-messaging-systems.md
```diff
@@ -0,0 +1,110 @@
...
+As the raw log data don’t have a resource attached to them, the receiver should add
+a generic resource and instrumentation library message around the raw messages. The
+instrumentation library should be set to the name of the receiver and the version to
+that of the collector.  Defining the exact resource should be done in the pipeline,
+using for example the `resourceprocessor`.
+
+## Prior art and alternatives
+
+The `kafkareceiver` and `kafkaexporter` already implement this OTLP as
+described in the OpenTelemetry Collector. The description in th OTEP
```
tigrannajaryan, line 98:
```suggestion
described in the OpenTelemetry Collector. The description in the OTEP
```

text/0000-otlp-over-messaging-systems.md
```diff
@@ -0,0 +1,110 @@
...
+makes both of them compliant without modification. Although it
```
tigrannajaryan, line 99: This is good. But let's also be aware this is not a must-have requirement. If we believe what Collector does is not the best we can do then we should be able to propose something different in this OTEP and fix the Collector (it is still in Beta and breaking changes are allowed).

text/0000-otlp-over-messaging-systems.md
```diff
@@ -0,0 +1,110 @@
...
+without conflicting with the implementation.
+
+## Open questions
+
+This proposal doesn't take advantage of attributes that some systems
+support. Should it? Would it be useful to look at the CloudEvents
+spec, to leverage the conventions for attributes?
+
+No guaranty of order could be given, because not all systems support order.
+Is that a problem?
```
tigrannajaryan, line 110: This should not be a problem since OTLP conceptually does not guarantee order of delivery in network protocol.

text/0000-otlp-over-messaging-systems.md
```diff
@@ -0,0 +1,110 @@
...
+receiving raw data from a topic. This enables scenarios were
+non-OpenTelemetry log producers can produce a stream of logs, either
+structured or unstructured, on that topic and the receiver wraps the log
+data in a configurable resource.
+
+## Internal details
+
+The default implementation must implement `otlp_proto`, meaning that the
+payload is Protobuf serialized from `ExportTraceServiceRequest` for traces,
+`ExportMetricsServiceRequest` for metrics, or `ExportLogsServiceRequest` for
```
tigrannajaryan, line 58: One thing that is not clear to me is if we want to include the signal type in the serialized payload so that he recipient can tell what signal it is. Since this OTEP does not define it it means the recipient that works with multiple signals needs another way to figure it out (often ends up using the channel name). I think this is an important capability that adds value and has no obvious downsides.

This is also an issue we want to solve for OTLP/File format: https://github.com/open-telemetry/opentelemetry-specification/issues/1443#issuecomment-780183624
It will be good to solve this here as well and do it in the same way we will do for OTLP/File so this likely requires to think through the OTLP/File design as well.

tigrannajaryan (changes requested):
I am temporarily blocking this since I have several questions that are not answered by the OTEP and I think they should be answered before we commit to making this part of the Otel spec.

text/0000-otlp-over-messaging-systems.md
aabmass: This doesn't help with the file format at all, but Kafka and Pubsub (maybe others) have key-value headers/attributes which are mentioned in "Open questions". The signal type and encoding (json or proto) could go in there

text/0000-otlp-over-messaging-systems.md
```diff
@@ -0,0 +1,110 @@
...
+Above you’ll find a non-exclusive list of possible encodings. Only `otlp_proto`
+is mandatory and the default.
+
+### Log specific collector receiver
+
+As the flow control is the hardest part of the implementations, adding a feature
+for reading raw log events avoids having an additional parallel
+implementation todo just that.
+
+The encoding field should be used to indicate that the data received is not the
```
aabmass, line 83: What is meant by "field" here?

text/0000-otlp-over-messaging-systems.md
```diff
@@ -0,0 +1,110 @@
...
+
+The `kafkareceiver` and `kafkaexporter` already implement this OTLP as
+described in the OpenTelemetry Collector. The description in th OTEP
+makes both of them compliant without modification. Although it
+doesn't implement the raw logging support but this could be added
+without conflicting with the implementation.
+
+## Open questions
+
+This proposal doesn't take advantage of attributes that some systems
```
aabmass, line 105: +1 I think it would be good to put encoding and signal type in there.

text/0000-otlp-over-messaging-systems.md
```diff
@@ -0,0 +1,110 @@
+# OTLP over messaging systems
+
+Best practices for transporting OTLP data over a message system.
```
aabmass, line 3: Do we have any performance best practices for message systems based on the existing Collector Kafka implementation? For example should you prefer to batch into many small messages or into fewer large messages?

text/0000-otlp-over-messaging-systems.md
```diff
@@ -0,0 +1,110 @@
...
+
+As the raw log data don’t have a resource attached to them, the receiver should add
+a generic resource and instrumentation library message around the raw messages. The
+instrumentation library should be set to the name of the receiver and the version to
+that of the collector.  Defining the exact resource should be done in the pipeline,
+using for example the `resourceprocessor`.
+
+## Prior art and alternatives
+
+The `kafkareceiver` and `kafkaexporter` already implement this OTLP as
```
jrcamp, line 97: Note that I'm proposing some changes to these:

https://github.com/open-telemetry/opentelemetry-collector/pull/3044
https://github.com/open-telemetry/opentelemetry-collector/pull/3200

Still  a WIP.

text/0000-otlp-over-messaging-systems.md
jrcamp, line 70: I'm proposing changes to these in PR linked below. The types would be something like `zipkinv1/thrift`, `zipkinv2/protobuf`. Aside from the formatting it treats zipkinv1 and zipkinv2 as two separate protocols instead of putting them both behind `zipkin`.

text/0000-otlp-over-messaging-systems.md
alexvanboxel: Maybe not, it would be nice though that implements do this. But maybe it shouldn't be in the spec.

text/0000-otlp-over-messaging-systems.md
alexvanboxel: I've added a comment in the OTLP/File discussion about CloudEvents. Certainly, if you want to allow multiplexing logs, traces, and metrics over a single topic CloudEvents is a good fit and it would align with this OTLP as well as the spec defines bindings for different messaging systems:

- [Kafka](https://github.com/cloudevents/spec/blob/v1.0.1/kafka-protocol-binding.md)
- [MQTT](https://github.com/cloudevents/spec/blob/v1.0.1/mqtt-protocol-binding.md)
- [NATS](https://github.com/cloudevents/spec/blob/v1.0.1/nats-protocol-binding.md)
- [Google Pubsub](https://github.com/google/knative-gcp/blob/main/docs/spec/pubsub-protocol-binding.md)

text/0000-otlp-over-messaging-systems.md
alexvanboxel: I would not make it part of the spec, just like it is not part of the OTLP/gRPC spec.

text/0000-otlp-over-messaging-systems.md
alexvanboxel: I think I'm going to remove the other encoding out of the spec as @tigrannajaryan  suggests

alexvanboxel: I will remove it, but I'll leave it here to give other people a chance to comment.

text/0000-otlp-over-messaging-systems.md
jsuereth: +1.   I think perhaps having a standard convention for knowing what signal is coming in

(either force separate channels for each signal, *OR* a standard header of some sort to disambiguate) would be good.  OTLP effectively has this in their GRPC definition via the separate services.

alexvanboxel: Well, gRPC can handle this over the same endpoint (a separate service is just a path in the HTTP/2 header).

I'm leaning toward the CloudEvents proposal, where you can have multiple (at least have the possibility to) types over one topic.

text/0000-otlp-over-messaging-systems.md
tigrannajaryan: @alexvanboxel I read the comment you left in the OTLP/File issue but I am a bit confused why it talks about the CloudEvents or Avro.

My comment here is about how an OTLP payload that is encoded as a Protobuf or JSON should be recorded in messaging systems such that the signal type information is included. I think we should keep it as simple as possible, new any additional concepts and technology introduced (CloudEvents or Avro) IMO is unnecessary complication.

If we want to record the type of the signal there are fairly ways to achieve it while staying with the existing set of technologies. For example we can add this message to OTLP Protobuf definitions:
```protobuf
// This may be also named simply Request or Payload or something like that
message ExportServiceRequest {
  oneof request {
    ExportTraceServiceRequest traces = 1;
    ExportMetricsServiceRequest metrics = 2;
    ExportLogsServiceRequest logs = 3;
  }
}
```
ExportServiceRequest can be then serialized and become the payload in the messaging system. Recipients can then deserialize and will be able to use the right signal in the "request" field.

alexvanboxel: @tigrannajaryan the consensus is that we want to signal the type of the message, so I'm already changing my application. Let me demonstrate what I mean by leveraging the CloudEvent spec (as it's a spec and not technology). In my Google Pubsub implementation, it describes what headers to use. So I could simple use:

```
func (receiver *pubsubReceiver) detectEncoding(attributes map[string]string) (Encoding, error) {
	ceType := attributes["ce-type"]
	ceContentType := attributes["ce-datacontenttype"]
	if strings.HasSuffix(ceContentType,"application/x-protobuf") {
		switch ceType {
		case "org.opentelemetry.otlp.traces.v1":
			return OtlpProtoTrace, nil
		case "org.opentelemetry.otlp.metrics.v1":
			return OtlpProtoMetric, nil
		case "org.opentelemetry.otlp.logs.v1":
			return OtlpProtoLog, nil
		}
	} else if strings.HasSuffix(ceContentType,"application/json") {
		switch ceType {
		case "org.opentelemetry.otlp.traces.v1":
			return OtlpJsonTrace, nil
		case "org.opentelemetry.otlp.metrics.v1":
			return OtlpJsonMetric, nil
		case "org.opentelemetry.otlp.logs.v1":
			return OtlpJsonLog, nil
		}
	}
	return Unknown, nil
}
```

Granted, it's heavier than a `oneOf`, but it leaves the flexibility of even signaling OTLP/Json encoding.

tedsuo:
@alexvanboxel  we are marking OTEPs which are no longer being updated as `stale`. We will close this PR in one week. If you wish to resurrect this PR, that is fine.

alexvanboxel:
I've neglected some of my OpenTelemetry work for the last few months. I will refocus some of my time back on open source. I will pick it back up, including the work in the collector.

carlosalberto:

PR #158, opened 2021-05-18
Configuration of Dynamic Span Creation vs Propagation

Xitric:
This OTEP proposes the ability to configure the level of detail of span creation from OpenTelemetry instrumentation, as an alternative to sampling for reducing the performance impact. It is a matter of balancing performance and expressiveness of traces.

The intention of this OTEP is mostly as a basis for further discussion around the subject, and the benefits or disadvantages of pursuing such functionality.


text/0158-dynamic-span-creation.md
```diff
@@ -0,0 +1,109 @@
...
+> During crises, operators may be willing to accept a decrease in performance in exchange for a larger wealth of information, but cannot act on this intent because of the static tracing instrumentation.
+
+## Explanation
+
+In order for operators to configure the expressiveness of tracing at runtime, they need a way to communicate with the instrumented applications.
+As inspired by Jaeger, this could be implemented by allowing for OpenTelemetry Collectors to inform applications about the desired level of detail, perhaps using polling. We believe that the level of detail (LoD) should be configurable at at least three levels:
+
+- `PROPAGATION_ONLY`: Only propoagate headers, and suppress the creation of all spans. This will work well with service meshes.
+- `EXTERNAL_SPANS`: Propagate headers, and generate additional client, server, consumer, and producer spans. This will enhance the traces generated by service meshes, but only when the performance impact can be justified.
+- `ALL_SPANS`: Propagate headers, and generate spans at any level of detail
```
yurishkuro, line 51: Specifying which spans to generate is one possible dimension of controlling trace verbosity. There are other dimensions, such as:
  * still capturing the spans but changing the level of details recorded about them, e.g. including / excluding numerous performance metrics (CPU cycles, memory allocs, etc.) which could be expensive to acquire
  * there are also opposite use cases that are very interested in perf metrics but not so much in the static span data such as span name and resource attributes, e.g. for evaluating resource utilization of a particular product feature
  * only capturing spans at certain places in the call graph. For example, some workflows might involve a lot of queries to storage systems (100:1 ratio of db spans to service spans is not unusual), and storage spans are both expensive to capture and not always needed. The verbosity controls can either suppress storage spans completely or replace them with client-side summaries.

text/0158-dynamic-span-creation.md
```diff
@@ -0,0 +1,109 @@
...
+| Envoy generates server span |                            |                              |
+| OTel span suppressed        | OTel generates server span | OTel generates server span   |
+| OTel span suppressed        | OTel span suppressed       | OTel generates internal span |
+| OTel span suppressed        | OTel generates client span | OTel generates client span   |
+| Envoy generates client span |                            |                              |
+
+## Internal details
+
+This proposal would entail adding a LoD configuration that can be passed to applications via the collector.
+During span creation, the instrumentation library would have to consider the current LoD, and decide whether to suppress the creation of spans, similarly to how it may decide whether a span should be sampled.
```
yurishkuro, line 66: I think in practice the desired verbosity level would need to be propagated in the trace context.

text/0158-dynamic-span-creation.md
```diff
@@ -0,0 +1,109 @@
...
+
+We propose an additional feature of OpenTelemetry, where instrumentation libraries can be configured, at runtime, to provide dynamic expressiveness of generated traces.
+The motivation is that during normal operation, the system generates spans at a coarse level of detail.
+In crisis situations, operators can configure the system to generate traces at an increased level of detail, to support their debugging efforts.
+This will ensure a minimal performance impact in day-to-day operations, while an increased performance impact can be accepted for short periods of time during crises.
+This is not unlike how logs are specified at various levels of detail, such as DEBUG, INFO, WARN, and ERROR.
+
+We base this motivation on a number of piror observations:
+
+- [In his Master's Thesis, Carosi](https://www.semanticscholar.org/paper/Protractor%3A-Leveraging-distributed-tracing-in-for-Carosi/708e776d9440abd56006a312168773fdc1ed9667) evaluates the performance impact on tracing applications with service meshes, and finds that the performance impact (in terms of response times) is negligible.
```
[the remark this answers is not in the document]
Xitric: I see why this was not clear enough. I have refactored the motivation, and hope that the new presentation is clearer.

text/0158-dynamic-span-creation.md
Xitric: I understand why it would be of interest to use the existing capability to transmit state in the trace context. Initially, the thought was to configure instrumentation verbosity on a service-by-service level, in which case I do not believe that the trace context is sufficient. However, we are also open to the idea of configuring trace verbosity per workflow, if that is more in line with the design of OpenTelemetry. In fact, workflow-based trace verbosity may even be the better option, and reduce the performance impact during debugging periods! We will consider this and update the proposal soon.

text/0158-dynamic-span-creation.md
Xitric: I agree, there are many dimensions to this. We have also experienced workflows with over 600 db spans. While I understand that this information may be of interest to some clients, our industry partners have a different use case. They mostly collect metrics via integrations with platforms such as AWS. For this reason, their use case for tracing is mostly for debugging and performance optimization by visualising and analyzing workflows. Common to both use cases is the need for a dynamic trace verbosity as opposed to a static degree of instrumentation.

I think we will need more feedback on use cases to properly design verbosity controls of general usefulness. Should we keep the discussion to this pull request, or open an issue somewhere?

tedsuo:
@Xitric  we are marking OTEPs which are no longer being updated as `stale`. We will close this PR in one week. If you wish to resurrect this PR, that is fine.

PR #159, opened 2021-05-20
Update 0146-metrics-prototype-scenarios.md

jkwatson:
Add a note about temperature recordings being potentially negative.

/cc @reyang

### DIFF
--- a/text/metrics/0146-metrics-prototype-scenarios.md
+++ b/text/metrics/0146-metrics-prototype-scenarios.md
@@ -181,6 +181,9 @@ speaking it is not considered as a "dimension" from the SDK perspective.
 | --------- | --------------- |
 | MachineA  | 65.3            |
 
+Note: Temperature may take on negative values. For this reason, you must choose the instrument carefully to ensure it is one that
+can accept negative recordings and that they can be aggregated appropriately.
+
 **Server humidity:**
 
 | Host Name | Humidity (%) |


arminru (approved):

text/metrics/0146-metrics-prototype-scenarios.md
```diff
@@ -181,6 +181,9 @@ speaking it is not considered as a "dimension" from the SDK perspective.
 | --------- | --------------- |
 | MachineA  | 65.3            |
 
+Note: Temperature may take on negative values. For this reason, you must choose the instrument carefully to ensure it is one that
```
jsuereth, line 184: Mandatory dumb comment:  "Why not use Kelvin?"

jsuereth (approved):

tigrannajaryan (approved):

text/metrics/0146-metrics-prototype-scenarios.md
arminru: I actually doubt that a server room would get into negative Fahrenheit (except if located in Arctic climate, during a power outage and if the device taking temperature samples using our OTel metrics SDK is battery-powered and still working) but it makes sense to also have a use case that covers negative values 🙂

arminru:
Turning the PR off and on again since the CI is stuck...

text/metrics/0146-metrics-prototype-scenarios.md
jkwatson: Your quantum computer better be keeping it's q-bits cold! Maybe the server room itself isn't below zero, but the "chip" temperature would definitely be wanting to be below zero C or F.

jmacd: > Mandatory dumb comment: "Why not use Kelvin?"

Real temperatures in Kelvin scale are around 300, where an exponential histogram has low precision. This is why the mention of a constant offset for histogram measurements came up yesterday. 

The reason this discussion started has to do with a requirement from OpenMetrics to know if the sum of events in a histogram is monotonic or not. We've resisted to this point making optional behaviors on API instruments, but it's a logical approach. Histogram instruments could have an option to accept negative values, it would be rarely used.
 
 If we go this way, the next logical question is: why not make UpDownCounter an option to Counter? I would support that, at this point, although it made me uncomfortable a couple of years ago.

jmacd (approved):

PR #161, opened 2021-06-01
Telemetry Schema and Resource

jsuereth:
Specify a mechanism for users of Resource to interact with Resource leveragingTelemetrySchema version for a stable API.

Goals of this OTEP:

- Exporters in the SDK have access to "stable" resource names.  I.e. I can code an exporter against version X of the telemetry schema
- Improve flexibiltiy in altering Resource semantic conventions through clearly defined (and implemented) conversion tooling.


text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,128 @@
+# Telemetry Schema and Resource
+
+Specify a mechnism for users of Resource to interact with Resource leveraging
+TelemetrySchema version for a stable API.
+
+## Motivation
```
punya, line 6: Could you add a couple of examples (real or made-up) illustrating the kinds of breaks that this OTEP would protect/mitigate?

jsuereth: done

Makefile
```diff
@@ -44,3 +44,7 @@ install-markdown-lint:
 markdown-lint:
 	markdownlint -c .markdownlint.yaml '**/*.md'
 
+
+.PHONY: check
+check: markdown-lint misspell markdown-link-check
+	echo "All checks complete"
```
tigrannajaryan, line 50: Nit: unrelated change.

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
...
+  next.version:
+    resources:
+      changes:
+        - rename_attributes:
+            attribute_map:
+              cloud.zone: cloud.availability_zone
+  previous.version:
+```
+
+If an exporter is defined expecting to find `cloud.zone` on `previous_version`,
```
tigrannajaryan, line 42: Do any exporters do this today? I wonder why the exporter needs to look for a specific attribute.

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
...
+    ... existing methods ...
+    /** The current schema version. */
+    public String getSchemaUrl();
+    /**
+     * Returns a new instance of Resource with labels compatible with the
+     * given schema url.
+     *
+     * <p> Returns an invalid Resource if conversion is not possible.
+     */
+    public Resource convertToSchema(String schemaUrl);
```
tigrannajaryan, line 80: Nice! This is an expected capability of schemas, so no surprises here.
I believe this should also be able to return an error since some conversions are not possible (e.g. between different schema families or if the versions are incompatible).

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
...
+}
+```
+
+This could be used in the `Resource.merge` method to determine if resources
+can be converted to a compatible schema.  Specfically resource can be merged
+IFF:
+
+- The schemas are compatible
+- Converting all `Resource` instances to the "newest" schema version returns
+  valid `Resource` objects.
```
tigrannajaryan, line 114: This is a very nice idea.

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
...
+The resulting resource will have the Schema URL calculated as follows:
+
+- If the old resource's Schema URL is empty then the resulting
+  resource's Schema URL will be set to the Schema URL of the updating resource
+- Else if the updating resource's Schema URL is empty then the resulting
+  resource's Schema URL will be set to the Schema URL of the old resource,
+- Else if the Schema URLs of the old and updating resources are the same then
+  that will be the Schema URL of the resulting resource,
+- **Else if the Schema urls of the old and updating resources are compatible**
+  **then the older resource will be converted to newer schema url and merged.**
```
tigrannajaryan, line 140: 👍

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
...
+  and updating resources are not empty and are different).
+
+### Resource conversion logic
+
+SDKs will be expected to provide a mechanism to apply transformations
+(both backwards and forwards) listed in `changes` for any compatible schemas.
+
+**Note: Until conversion logic is available in SDKs, any change to telemetry**
+**schema that relies on `changes` will be considered a breaking, i.e.**
+**they will lead to a major version bump.**
```
tigrannajaryan, line 151: "major version bump" of what?

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
...
+
+The major drawback to this proposal is the requirement of making all schema
+migration implementations available in all SDKs.  This is mitigated by the
+following:
+
+- Initially, we could prevent any `rename` alterations to schema for compatible
+  version numbers. `convertToSchema` would check for semver compatibility and
+  just return the original resource if compatible.
+- Alternatively, we can check the newer schema for any transformations in the
+  `all` or `resource` section before failing to convert.
```
tigrannajaryan, line 163: So this would be a "poor mans" conversion function, which just figures out that the conversion is noop and can just return the original and fail otherwise, right? I believe this should cover a very large number of real-world use cases of existing resource detectors.

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
...
+
+This should allow a slow-rollout of implementation by SDKs by increasing the
+number of successful `convertToSchema` operations they allow.
```
tigrannajaryan, line 166: 👍

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
...
+(both backwards and forwards) listed in `changes` for any compatible schemas.
+
+**Note: Until conversion logic is available in SDKs, any change to telemetry**
+**schema that relies on `changes` will be considered a breaking, i.e.**
+**they will lead to a major version bump.**
+
+## Trade-offs and mitigations
+
+The major drawback to this proposal is the requirement of making all schema
+migration implementations available in all SDKs.  This is mitigated by the
```
tigrannajaryan, line 156: Right. On the other hand, a major benefit of this is that we will end up with schema conversion implementations in many languages which will make it easier for vendors to adopt the concept of schemas on their backends. :-)

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Open questions
+
+- What kinds of operations on Resource should be considered compatible?
+- Is "downgrading" a schema version going to be "safe" for exporters?
+
+## Future possibilities
+
+The notion of telemetry-schema conversion could be expanded and adapted into
+its own set of functionality that SDKs or SDK-extensions (like resource
```
tigrannajaryan, line 195: I think we should make them independent part of SDKs, consumable separately if possible by code that does not care about the rest of Otel but needs to deal with schemas. It will be very nice though to make this official part of Otel libaries.

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
...
+same version of telemetry schema as the SDK and find a mechanism to warn/prevent
+older exporters from using a newer SDK.
+
+Another alternative is for SDKs to continue propagating resources as-is and
+have exporters issue errors when resource versions don't align with
+expectations.
+
+## Open questions
+
+- What kinds of operations on Resource should be considered compatible?
```
tigrannajaryan, line 189: This is defined by schema OTEP. Conversion between families is incompatible. Ambiguous "renames" are incompatible.

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
...
+- Is "downgrading" a schema version going to be "safe" for exporters?
```
tigrannajaryan, line 190: I think it is safe. Any reason to believe it may not?

tigrannajaryan (approved):
Very nice OTEP. Spot on with the spirit of schemas and nicely builds on top of it.

MadVikingGod:
Some general questions I have about this proposal:

tigrannajaryan:
>Do we expect this to be used by most users of the SDK? I can see the use in the collector, and backends, but other then resources what does an end user application or an instrumentation library get with this feature?

It will be used by the SDK itself to correctly merge the resources. It can be also used by an external code to do schema translation (e.g. the Collector or a backend).

>Could this be accomplished outside of the SDK, or maybe as a feature of the collector?
Schema conversion, once the data leaves the SDK, yes, can be done by the Collector (and it is a planned feature). However, the merging of the resources must happen in the SDK, there does not seem to be a way to avoid that.

>Does that mean that only URLs defined by the spec are the only valid ones, what if I want to make a convention for Digital Ocean that isn't part of current specs?

Yes, you can have your own schema that is not Otel schema. You will need to choose a schema family URL and publish the schema files for each version you release.

>What does newer mean in such a world? Is otel v1.4.0 newer than Oracle v1.1.2?

They are different schema families, not comparable. "Newer" only makes sense for schemas that belong to the same family.

>What does compatible mean in such a world? Is otel v1.4.0 compatible with Oracle v1.1.2?

No, not compatible, since they are different families.

>How does one convert from Otel v1.4.0 to Oracle v1.1.2?

You can't. 

Note: I planned a concept of [parent schemas](https://github.com/open-telemetry/oteps/blob/main/text/0152-telemetry-schemas.md#parent-schema) for the future. This would allow deriving Oracle schema from Otel schema and in that case possibly allow conversion. But it is not part of the current Schemas concept yet.

>How do we limit the amount of code we introduce by updating the conventions?

There is no code needed to be introduced when you update the conventions. The schema conversion code is capable of generically handling changes to convention. The rules defined in schema file drive this generic schema conversion code.

MadVikingGod:
> There is no code needed to be introduced when you update the conventions. The schema conversion code is capable of generically handling changes to convention. The rules defined in schema file drive this generic schema conversion code.

I'm trying to understand how an actual implementation would do this.  So does this mean that we are going to either embed the schema file or go fetch the schema file when needed?  Then the code that would be written for this would be an interpreter of the schema to dynamically convert one version to the next?

tigrannajaryan:
> > There is no code needed to be introduced when you update the conventions. The schema conversion code is capable of generically handling changes to convention. The rules defined in schema file drive this generic schema conversion code.
> 
> I'm trying to understand how an actual implementation would do this. So does this mean that we are going to either embed the schema file or go fetch the schema file when needed?

Both are valid approaches.

> Then the code that would be written for this would be an interpreter of the schema to dynamically convert one version to the next?

Yes, exactly.

If you haven't read the Schemas OTEP it may be useful to do it since it answers some of your questions: https://github.com/open-telemetry/oteps/blob/main/text/0152-telemetry-schemas.md

text/0161-telemetry-schema-and-resource.md
jsuereth: We do this for google cloud.  We're mapping from Otel Resource to cloud's "MonitoredResource" with a  well-defined schema.

jsuereth:
@MadVikingGod 
> Some general questions I have about this proposal:
> Does this need to be part of the SDK?

Yes, we need this for GCP exporters.

> Do we expect this to be used by most users of the SDK? I can see the use in the collector, and backends, but other then resources what does an end user application or an instrumentation library get with this feature?

SDK != End users.   Instrumentation library = API.    This needs to be in the SDK for exporters, and it's a code-only thing.  End users should be interacting MOSTLY via configuration or 'code configuration' on the sdk.

> Could this be accomplished outside of the SDK, or maybe as a feature of the collector?

We also need this in the collector.  This is an interface that should be provided to consumers of `Resource` in the SDK.

> Does that mean that only URLs defined by the spec are the only valid ones, what if I want to make a convention for Digital Ocean that isn't part of current specs?
> What does newer mean in such a world? Is otel v1.4.0 newer than Oracle v1.1.2?
> What does compatible mean in such a world? Is otel v1.4.0 compatible with Oracle v1.1.2?
> How does one convert from Otel v1.4.0 to Oracle v1.1.2?

I think tigran responded to most of these.

> Does this mean where ever this code lies (it's proposed for the SDK but see 1) we have to have some way to process these conversions? This has lots of implication like:
How does it do forward conversions for specs after the current version was released?
How do we limit the amount of code we introduce by updating the conventions?
How do we know what supported conversions are available for the current library?

Totally agree these are good questions.  I see this migration being a *temporary solution* to prevent ecosystem breakage as everyone adopts the latest telemetry schema.   Let's say a user is using the following components:

- OpenTelemtery API  1.1
- Instrumetantion for my hipster microservice framework, 1.2
- OpenTelemetry SDK   1.4
- Resource Detector for K8s   v1.1
- Google cloud exporter v1.0

If we take the approach we have today for Resources, it means I need to upgrade them ALL in lockstop for correct behavior.   To mitigate this, we've kept Resource detection components + GCP exporters in our own repositories and release them together, so it's easier to know if you have an issue.    

However, if folks use any other Resource Detector component besides the one we provide, even if it abides by semantic conventions, we can't tell if we're compatible.  Renaming attributes is *silent breakage* to some extent, and worse if the exporter is unable to declare which version of the schema it was using.

MadVikingGod:
So, let me start off I like the ideas presented here, but I think there are a few assumptions that have been made that aren't stated, and it's a bit light on how external conventions interact with the conventions discussed.

And as I'm thinking more on this I think we are solving two separate issues, How to merge Resource Conventions, and How to merge Telemetry Conventions.  They are both the same problem, but **where** we can solve them is different, as well as how we should solve them might be different. We might want to split the issue of resource migration and telemetry migration into different proposals.

The major concern right now with useability.  If I want to guarantee any single convention outside of the current semantic conventions, anything that uses that is unable to be mixed in with other data.  Meaning even if my convention is a superset of a semconv there is no way to upgrade other libraries to my convention. Nor can I downgrade to a different family of conventions.  

The other large concern is that this seems to be leading us to dynamic processing of arbitrary schemas.  While this can be made safe, it is a very challenging task to do so safely, and I would highly recommend against it. 


> SDK != End users. Instrumentation library = API.
Actually the End User is the only one who gets to choose the SDK and is the only entity guaranteed to have it at all.  As I've been asking about I think especially for the case of telemetry semconvs we don't need it in the SDK, and adding it for an exporter that may live in contrib seems like the wrong way to reason about this.  

For the resource portion, the user/library writers will know when they write the application what conventions will be used.  I think we shouldn't have any complicated logic around transforming those but push for a way for an application author to know:
1. At compile time that different conventions that may be used aren't stepping on each other.
2. At upgrade time that taking a newer version doesn't mask something they have done.

For the telemetry portion, I think a stand-alone library is much more appropriate. SDKs don't provide any ingest logic and this is an ingest problem.

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
...
+## Internal details
+
+There are three major changes to SDKs for this proposal.
+
+### Understanding schema compatibility
+
+Here we propose simply providing an implementation of
+[SemVer](https://semver.org/) for isCompatible, and using the
+the [ordering rules](https://semver.org/#spec-item-11) for isNewer.  This would
+pull the version number out of the schema url.
```
MadVikingGod, line 125: As proposed only schemas from the same Family can be compatible with each other.  Could a field in the schema like `parent` be used to enable cross Family compatibility?

tigrannajaryan: Yes, within the same tree of families. Parent schemas are discussed here as a future possibility: https://github.com/open-telemetry/oteps/blob/main/text/0152-telemetry-schemas.md#parent-schema
I don't think we want it now though. It is too early, the concept of schemas is very new and I think we need to wait for some real world feedback before we evolve it in this direction.

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
...
+
+Specify a mechnism for users of Resource to interact with Resource leveraging
+TelemetrySchema version for a stable API.
+
+## Motivation
+
+In the [current definition](https://github.com/open-telemetry/opentelemetry-specification/pull/1692)
+of Resource + Telemetry schema interaction, the specification focuses mostly on
+how to provide a consistent "storage" of Resource within the SDK. However,
+from a usage standpoint, how do consumers gain stability across changes to
```
iNikem, line 11: Who is a "consumer" here? End-user or backend receiving telemetry or SDK component, such as exporter?

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
...
+```java
+public abstract class Resource {
+    ... existing methods ...
+    /** The current schema version. */
+    public String getSchemaUrl();
+    /**
+     * Returns a new instance of Resource with labels compatible with the
+     * given schema url.
+     *
+     * <p> Returns an invalid Resource if conversion is not possible.
```
iNikem, line 78: Do we have a definition of "invalid Resource"?

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
...
+- If the old resource's Schema URL is empty then the resulting
+  resource's Schema URL will be set to the Schema URL of the updating resource
+- Else if the updating resource's Schema URL is empty then the resulting
+  resource's Schema URL will be set to the Schema URL of the old resource,
+- Else if the Schema URLs of the old and updating resources are the same then
+  that will be the Schema URL of the resulting resource,
+- **Else if the Schema urls of the old and updating resources are compatible**
+  **then the older resource will be converted to newer schema url and merged.**
+- Else this is a merging error (this is the case when the Schema URL of the old
+  and updating resources are not empty and are different).
```
iNikem, line 142:
```suggestion
  and updating resources are not empty and are not compatible).
```

text/0161-telemetry-schema-and-resource.md
iNikem: Not sure that it will be easy for backend vendors to reuse parts of the SDK

text/0161-telemetry-schema-and-resource.md
jkwatson: Since Resource resolution happens at application startup, I'm nervous about having SDKs have to go and fetch multiple schemas from the public internet as a part of the startup process, in order to perform the conversion. There are failure modes here, and general startup-time issues to be considered (especially, for example, in mobile or FAAS implementations).

tigrannajaryan: > Since Resource resolution happens at application startup, I'm nervous about having SDKs have to go and fetch multiple schemas from the public internet as a part of the startup process, in order to perform the conversion.

That's a good point. The SDKs can have the schema file that they correspond to embedded with the source code so that there is no need to fetch anything for detectors that are part of the SDK. If each SDK includes the latest Otel schema file then any Otel schema conversions can happen without network access.
Custom, non-Otel schema conversions will still require a network fetching. Perhaps we also provide an alternate `convertToSchema` function which accepts a schema file instead of schema URL?

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
...
+  valid `Resource` objects.
+
+## Internal details
+
+There are three major changes to SDKs for this proposal.
+
+### Understanding schema compatibility
+
+Here we propose simply providing an implementation of
+[SemVer](https://semver.org/) for isCompatible, and using the
```
aabmass, line 123: I'm a little confused on this. If you had a schema rule like the one above:

```yaml
  next.version:
    resources:
      changes:
        - rename_attributes:
            attribute_map:
              cloud.zone: cloud.availability_zone
  previous.version:
```

since there is a transform between versions, would the two schemas be compatible and have the same major version?

text/0161-telemetry-schema-and-resource.md
jsuereth: SDK component (e.g. exporter)

text/0161-telemetry-schema-and-resource.md
jsuereth: I don't think we do. That's how things are specified today though (with no definition of what invalid means)

text/0161-telemetry-schema-and-resource.md
jsuereth: It's not clear to me what the answer to this is, which is why I opened: https://github.com/open-telemetry/oteps/pull/162

IF the answer to the quesiton is "yes they're compatible", then I think this OTEP resolves *some* major pain that kind of change would cause.

jsuereth:
@MadVikingGod You state:
> For the telemetry portion, I think a stand-alone library is much more appropriate. SDKs don't provide any ingest logic and this is an ingest problem.

I'd disagree with the later part.  `Exporter`s are ingestion adapters and such this is their problem and they live as SDK components.  I'd be ok if this were a (stable) library an exporter can rely on, but it needs to be baked in.

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,196 @@
+# Telemetry Schema and Resource
+
+Specify a mechnism for users of Resource to interact with Resource leveraging
```
reyang, line 3:
```suggestion
Specify a mechanism for users of Resource to interact with Resource leveraging
```

tigrannajaryan:
@jsuereth do you plan to continue working on this OTEP?

jsuereth:
@tigrannajaryan Yes, tabled it temporarily to work more directly with metrics.  I still think this is an unresolved issue w/ semantic conventions + telemetry.

tigrannajaryan:
> @tigrannajaryan Yes, tabled it temporarily to work more directly with metrics. I still think this is an unresolved issue w/ semantic conventions + telemetry.

@jsuereth I believe we need this change. It evolves the schemas in the right direction.

text/0161-telemetry-schema-and-resource.md
jsuereth: Exactly.

text/0161-telemetry-schema-and-resource.md
Aneurysm9: I don't think that an alternate conversion method should be required to work from a file, but an alternate URL->bytes function could be provided.  I'd imagine possibly a three tier lookup: 1) embedded in the application by the SDK, 2) on disk in a well-known or specified location 3) over the network through traditional URL resolution.

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,203 @@
...
+The resulting resource will have the Schema URL calculated as follows:
+
+- If the old resource's Schema URL is empty then the resulting
+  resource's Schema URL will be set to the Schema URL of the updating resource
+- Else if the updating resource's Schema URL is empty then the resulting
+  resource's Schema URL will be set to the Schema URL of the old resource,
+- Else if the Schema URLs of the old and updating resources are the same then
+  that will be the Schema URL of the resulting resource,
+- **Else if the Schema urls of the old and updating resources are compatible**
+  **then the older resource will be converted to newer schema url and merged.**
```
Aneurysm9, line 147: It seems that "older resource" may be used in two senses here.  One being the target of a merge operation and the other being the resource with the lesser schema version.  Utilizing a different term for one or both of those senses may help avoid confusion.

text/0161-telemetry-schema-and-resource.md
Aneurysm9: Would this check be limited to attributes actually present on the resources being evaluated, or would the presence of any transformation, even for non-present attributes, prevent merging/migration?

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,203 @@
...
+| Resource Detector | <-> | SDK | <-> | Exporter |
++-------------------+     +-----+     +----------+
+```
+
+## Explanation
+
+We propose a new set of SDK features:
+
+- A mechanism to "migrate" a resource up/down compatible schema changes.
+- A mechanism to determine if two schema URLs allow compatible changes.
```
bogdandrutu, line 60: Is this a long term issue or just a short term need, I had the impression that we should always do backwards compatible changes. So we can always go up, sometimes we may not be able to go down.

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,203 @@
...
++-------------------+     +-----+     +----------+
+| Resource Detector | <-> | SDK | <-> | Exporter |
++-------------------+     +-----+     +----------+
+```
+
+## Explanation
+
+We propose a new set of SDK features:
+
+- A mechanism to "migrate" a resource up/down compatible schema changes.
```
bogdandrutu, line 59: I think we should provide only the "go up" version, and enforce everyone to use the most recent (latest) schema. The exporter in your case should be able to adopt (maybe) the rules to work on the Resource schema (if newer) vs downgrading the Resource.

text/0161-telemetry-schema-and-resource.md
bogdandrutu: Per this thread conversation, it may be better to start with a solution that "file"/"bytes" for the schema are provided, and delegate to the user of the API to ensure that is available somewhere to grab. Some users may embed the files in their application, some may embed the schema bytes, etc.

@jkwatson had a very good point, and probably we should postpone the API that fetches from the internet.

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,203 @@
...
+     */
+    public Resource convertToSchema(File schemaFile);
+}
+```
+
+This would allow a consumer of resources to ensure a stable view of resource
+labels, e.g.
+
+```java
+public static String MY_SCHEMA_VERSION="https://opentelemetry.io/schemas/1.1.0";
```
bogdandrutu, line 95: Not necessary related question: Is this the expected value for the proto fields or just the string number "1.1.0"?

/cc @tigrannajaryan

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,203 @@
...
+    public static boolean isCompatible(String oldUrl, String newUrl);
+    public static boolean isNewer(String testUrl, String baseUrl);
+}
+```
+
+This could be used in the `Resource.merge` method to determine if resources
+can be converted to a compatible schema.  Specfically resource can be merged
+IFF:
+
+- The schemas are compatible
```
bogdandrutu, line 119: Same question, is this something that we need to worry? Should we assume that converting to the "newest" is always possible?

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,203 @@
...
+- Converting all `Resource` instances to the "newest" schema version returns
+  valid `Resource` objects.
+
+## Internal details
+
+There are three major changes to SDKs for this proposal.
+
+### Understanding schema compatibility
```
bogdandrutu, line 127: Is this section needed because currently the schema url does not respect semver for the version number? I though that was implicit since the schema url version number corresponds to the specs version which follows semver, probably my assumption is wrong.

text/0161-telemetry-schema-and-resource.md
bogdandrutu: Discussed this in the specs SIG, probably very limited code can be shared since the SDKs don't work on the protos but on internal data compatible with the protos.

text/0161-telemetry-schema-and-resource.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Consuming resources at specific schema version
+
+The first is done by providing a new method against `Resource` in the SDK that
+will transform the attributes as necessary to match a given schema version.
+
+e.g. In Java you might have the following for resource:
+
+```java
+public abstract class Resource {
```
bogdandrutu, line 70: Should we say that similar changes will be applied to other places where schemaurl is used, like spans/metrics/logs?

text/0161-telemetry-schema-and-resource.md
bogdandrutu: You need this because you want to be able to also convert Spans/Metrics/Logs that also have schemaURL.

bogdandrutu (commented):
Overall LGTM, I think definitely a prototype is needed :)

My major concern is about providing the downgrading functionality since right now we require changes to be backwards compatible (semver does the same) but not forward compatible.

Please consider to "resolve" all the comments if the discussion got to the resolution point.

text/0149-exponential-histogram.md
```diff
@@ -0,0 +1,162 @@
+# Add exponential bucketing to histogram protobuf
```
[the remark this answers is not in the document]
jsuereth: merge conflict.

text/0161-telemetry-schema-and-resource.md
jsuereth: This is an issue of how users consume exporters + resource detecters in the SDK.  Both are pluggable components and we can expect resource detection to advance ahead of exporters, so the forward-compatible behavior is needed in that kind of distribution environment.

text/0161-telemetry-schema-and-resource.md
jsuereth: The Schema URL could be for a different specification entirely.   This OTEP proposed a level of compatibility we could not assume before.

This is needed so we CAN assume semver for the default otel schema.

text/0161-telemetry-schema-and-resource.md
jsuereth: Eventually, that would be the goal.

text/0161-telemetry-schema-and-resource.md
tigrannajaryan: We should be comfortable with both up and down directions. The schema files unambiguously defines whether it is safe to go in a particular direction.

text/0161-telemetry-schema-and-resource.md
tigrannajaryan: I agree with the notion of providing the local schema instead of fetching from the URL. We may go a step further and define `Schema` as an opaque object that is loadable from file or from a URL. This opens up the possibility of performance improvements (parse and compile the schema once such that the subsequent conversion calls are fast). See e.g. [prototype here](https://github.com/tigrannajaryan/telemetry-schema/tree/main/schema) that has "parse/compile" and "convert" as separate operations.

jmacd (approved):
Thanks, this is well thought out.

pyohannes (approved):

tigrannajaryan:
@jsuereth do you plan to continue working on this OTEP?

tedsuo:
@jsuereth I assume we still need this OTEP?

jsuereth:
@tedsuo I think we still need it but I've put effort on hold to address higher-priority items.  I think we still have time to revisit this post some semconv stability, and this needs two more approvers.

carlosalberto:

PR #163, opened 2021-06-12
AndreaMaria_webvitals/open_telemetry/oteps

Andipandi111:


tigrannajaryan:
@open-telemetry/specs-approvers anybody knows why this PR was allowed to run the status checks? I assume the protection that prevents first time contributors from running workflows should have prevented it? Did somebody approve the workflow or we need to enable some project setting for the protection to be enabled?

PR #167, opened 2021-07-05
Add remote sampler spec

pavolloffay:
Signed-off-by: Pavol Loffay <p.loffay@gmail.com>

### DIFF
--- a/text/experimental/remote-sampling.md
+++ b/text/experimental/remote-sampling.md
@@ -1,0 +1,146 @@
+# Remote Sampling for SDK
+
+This is a proposal to add remotely configurable sampling to OpenTelemetry SDK.
+
+## Motivation
+
+Remote sampling configuration is a useful feature that allows users configuring sampling in a central place (e.g. collector) for
+fleet of running OpenTelemetry SDKs. Apart from a central configuration place the remote sampling offers fine-grained
+per-operation sampling configuration.
+
+Remote sampling is already widely used and adopted in other open-source tracing systems such as [Jaeger](https://www.jaegertracing.io/docs/1.23/sampling/).
+
+## Explanation
+
+This draft is related to [0121-config-service](https://github.com/open-telemetry/oteps/blob/main/text/experimental/0121-config-service.md)
+
+Goals:
+
+* supply configuration from a remote source e.g. OpenTelemetry collector to SDK
+* update configuration in SDK at runtime
+* define remote sampler required configuration
+
+## Internal details
+
+```protobuf
+// Sampling configuration for a single service.
+message ServiceSamplingResponse {
+  // Default sampler when no operation_sampling is matched.
+  Sampler default_sampler = 1;
+  // A list of sampling configurations for a specific operations.
+  // First matching configuration is used.
+  repeated OperationSampling operation_sampling = 2;
+}
+
+// Sampling configuration for an operation.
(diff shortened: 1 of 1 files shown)


text/experimental/remote-sampling.md
```diff
@@ -0,0 +1,146 @@
...
+        "param": 0.0
+      }
+    ]
+  }
+}
+```
+
+## Open questions
+
+* Should we support regex match on the span name or attributes? It can be useful for matching `http.url` attribute.
```
anuraaga, line 142: Yes :) Also for matching against Baggage.

pavolloffay: I would be more careful with baggage. The sampler API does not explicitly take it https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span-creation. However I know it can be extracted from the context.

I would prefer to add regex support in the second iteration.

text/experimental/remote-sampling.md
```diff
@@ -0,0 +1,146 @@
...
+Goals:
+
+* supply configuration from a remote source e.g. OpenTelemetry collector to SDK
+* update configuration in SDK at runtime
+* define remote sampler required configuration
+
+## Internal details
+
+```protobuf
+// Sampling configuration for a single service.
```
joe-elliott, line 26: is there an expectation of protobuf? this is not a high throughput endpoint and json would be easier for human readability.

text/experimental/remote-sampling.md
```diff
@@ -0,0 +1,146 @@
...
+  // see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration
+  string name = 1; // TODO this can be enum (always_on, always_off, parentbased_traceidratio...)
+  // Sampler argument, equivalent to OTEL_TRACES_SAMPLER_ARG. Example 0.25 for traceidratio sampler.
+  string arg = 2;
+}
+
+// Request to get sampling configuration.
+message RemoteSamplingRequest {
+  string service_name = 1;
+  // TODO shall we add version and/or environment?
```
joe-elliott, line 62: instead of specific things like "version" or "environment" i think arbitrary key/value matching makes sense.

i think this feature would be quite interesting, but unsure how widely used it would be.

pavolloffay:
Related request for regex-based matching for Jaeger remote sampler https://github.com/jaegertracing/jaeger/issues/3133

pavolloffay:
I am closing this and I will open a PR to document Jaeger remote sampler in the spec.

The work for remote sampler in the OTES will be divided into a control plane and the sampler implementation.

willarmiros:
@pavolloffay was the PR for Jaeger remote sampler this one? https://github.com/open-telemetry/opentelemetry-specification/pull/1791

Do you still have an intention of defining a more generic remote sampling specification, or is that on hold for now because the jaeger remote sampler is included in the spec?

pavolloffay:
There are plans for a native OTEL remote sampler. The sampling WG is going to focus on it in the future.

willarmiros:
Sounds good, I will drop in on a sampling SIG meeting in the future. FWIW, here is the design doc for the X-Ray remote sampler in otel: https://docs.google.com/document/d/11V1CDr6eLoMq_3cK1bUm0GzYEyhEiI_83QEJMUFfu9g/edit?usp=sharing

PR #168, opened 2021-07-23
Specify how to propagate head sampling probability

jmacd:
Recently a piece of text was removed from https://github.com/open-telemetry/oteps/pull/148, which contained more than one proposal. That OTEP now specifies only how to record Sampler information in the span. This OTEP specifies how to propagate head trace sampling probability, so that child spans can be counted on the fly. This specification would allow `ParentBased` samplers to output sampling probabilities in a way that supports span-to-metrics pipelines, for example.

### DIFF
--- a/text/trace/0000-sampling-propagation.md
+++ b/text/trace/0000-sampling-propagation.md
@@ -1,0 +1,126 @@
+# Propagate head trace sampling probability
+
+Propose extending the W3C trace context `traceparent` to convey head trace sampling probability.
+
+## Motivation
+
+The head trace probability is useful in child contexts to be able to
+record the effective sampling probability in child spans.  This is
+documented in [OTEP 148](TODO: after merging), which establishes
+semantic conventions for conveying the adjusted count of a span via
+attributes recorded with the span.  When a sampling decision is based
+on the parent's context, the effective sampling probability, which
+determines the child's adjusted count, cannot be recorded without
+propagating it through the context.
+
+We propose to propagate the trace sampling probability that is in
+effect whenever the [W3C
+sampled](https://www.w3.org/TR/trace-context/#sampled-flag) flag is
+set by extending the `traceparent`.
+
+## Explanation
+
+To limit the cost of this extension, to ensure that it is widely
+supported, and for statistical reasons documented below, we propose to
+limit head tracing probability to powers of two.  This limits the
+available head sampling probabilities to 1/2, 1/4, 1/8, and so on, and
+we can compactly encode these probabilities as small integers using
+the negative base-2 logarithm of the effective probability.
+
+For example, the value 2 corresponds with 1-in-4 sampling, the value
+10 corresponds with 1-in-1024 sampling.
+
+Wheres the [version-0 W3C trace context `traceparent`
+header](https://www.w3.org/TR/trace-context/#examples-of-http-traceparent-headers)
+is a concatenation of four fields, this proposal would upgrade
+`traceparent` to version 1:
+
+```
+traceparent: (version)-(trace_id)-(span_id)-(flags)
+```
+
+The version 1 `traceparent` header will use a new field named `log-count`, i.e.,:
+
+```
+traceparent: (version)-(trace_id)-(span_id)-(flags)-(log-count)
+```
+
+where `log-count` is the encoded negative base-2 logarithm of
+sampling probability, which is the base-2 logarithm of the adjusted
+count of a child span created in this context (i.e., the logarithm of
+the effective count, thus "log-count").  To compute the adjusted count
+of a child span created in this context, use `2^log-count`.  A
+log-count of `0` corresponds with `(2^0)=1`, thus 0 conveys a context
+with probability 1.
+
+The sampling probability of a context is independent from whether it
+is sampled.  We consider it [useful to convey sampling probability
+even when unsampled]() as it can be used to estimate the potential
+overhead of starting new sampled traces.
+
+### Examples
+
+These are extended [from the W3C
+examples](https://www.w3.org/TR/trace-context/#examples-of-http-traceparent-headers):
+
+```
+Value = 00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01-05
+base16(version) = 00
+base16(trace-id) = 4bf92f3577b34da6a3ce929d0e0e4736
+base16(parent-id) = 00f067aa0ba902b7
+base16(trace-flags) = 01  // sampled
+base16(log-count) = 05  // head probability is 2^-5.
+```
+
+```
+Value = 00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-00-00
+base16(version) = 00
+base16(trace-id) = 4bf92f3577b34da6a3ce929d0e0e4736
+base16(parent-id) = 00f067aa0ba902b7
+base16(trace-flags) = 00  // not sampled
+base16(log-count) = 10 // head probability is 2^-16
+```
+
+We are able to express sampling probabilities as small as 2^-255 using
+just 3 bytes per `traceparent`.
+
+## Internal details
+
+A use known as "inflationary sampling" from Google's Dapper system is
+documented in [OTEP 148](TODO: inflationary sampling section).  This
+is is used to justify propagating the head sampling probability even
+when unsampled.
+
+[An algorithm for making statistical inferance from partially-sampled
+traces has been published](https://arxiv.org/pdf/2107.07703.pdf) that
+explains how to work with power-of-2 sampling rates.  The reasoning
+behind restricting the set of sampling rates is:
+
+- Lowers the cost of propagating head sampling probability
+- Makes math involving partial traces tractable
+
+## Trade-offs and mitigations
+
+Restricting head sampling rates to powers of two does not limit tail
+Samplers from using arbitrary probabilities.
+
+Restricting head sampling rates to powers of two does not limit
+Samplers from using arbitrary effective probabilities over a period of
+time.  For example, choosing 1/2 sampling half of the time and 1/4
+sampling half of the time leads to an effective sampling rate of 3/8.
+
+## Prior art and alternatives
+
+Google's Dapper system propagated a field in its trace context called
+"inverse_probability", which is equivalent to adjusted count.  This
+proposal uses the base-2 logarithm of adjusted count to save space
+
+## Open questions
+
+This OTEP suggests how to modify the W3C trace context to accomodate
+sampling in OpenTelemetry.  [OTEP 148](TODO) suggests semantic
+conventions for encoding adjusted count in a Span, but neither text
+specifies how to modify the built-in Samplers to produce the proposed
+new `traceparent` field so that the `ParentBased` Sampler can
+correctly set the proposed `sampler.adjusted_count` attribute.  This
+will be future work.
--- a/text/trace/0000-sampling-propagation.md
+++ b/text/trace/0000-sampling-propagation.md
@@ -14,9 +14,9 @@ determines the child's adjusted count, cannot be recorded without
 propagating it through the context.
 
 We propose to propagate the trace sampling probability that is in
-effect whenever the [W3C
-sampled](https://www.w3.org/TR/trace-context/#sampled-flag) flag is
-set by extending the `traceparent`.
+effect alongside the [W3C
+sampled](https://www.w3.org/TR/trace-context/#sampled-flag) flag by
+extending the `traceparent`.
 
 ## Explanation
 
@@ -32,14 +32,14 @@ For example, the value 2 corresponds with 1-in-4 sampling, the value
 
 Wheres the [version-0 W3C trace context `traceparent`
 header](https://www.w3.org/TR/trace-context/#examples-of-http-traceparent-headers)
-is a concatenation of four fields, this proposal would upgrade
-`traceparent` to version 1:
+is a concatenation of four fields,
 
 ```
 traceparent: (version)-(trace_id)-(span_id)-(flags)
 ```
 
-The version 1 `traceparent` header will use a new field named `log-count`, i.e.,:
+This proposal would upgrade `traceparent` to version 1 with a new
+field named `log-count`,
 
 ```
 traceparent: (version)-(trace_id)-(span_id)-(flags)-(log-count)
--- a/text/trace/0000-sampling-propagation.md
+++ b/text/trace/0000-sampling-propagation.md
@@ -6,12 +6,12 @@ Propose extending the W3C trace context `traceparent` to convey head trace sampl
 
 The head trace probability is useful in child contexts to be able to
 record the effective sampling probability in child spans.  This is
-documented in [OTEP 148](TODO: after merging), which establishes
-semantic conventions for conveying the adjusted count of a span via
-attributes recorded with the span.  When a sampling decision is based
-on the parent's context, the effective sampling probability, which
-determines the child's adjusted count, cannot be recorded without
-propagating it through the context.
+documented in [OTEP 148](TODO), which establishes semantic conventions
+for conveying the adjusted count of a span via attributes recorded
+with the span.  When a sampling decision is based on the parent's
+context, the effective sampling probability, which determines the
+child's adjusted count, cannot be recorded without propagating it
+through the context.
 
 We propose to propagate the trace sampling probability that is in
 effect alongside the [W3C
@@ -86,18 +86,18 @@ just 3 bytes per `traceparent`.
 
 ## Internal details
 
+The reasoning behind restricting the set of sampling rates is that it:
+
+- Lowers the cost of propagating head sampling probability
+- Makes math involving partial traces tractable.
+
 A use known as "inflationary sampling" from Google's Dapper system is
-documented in [OTEP 148](TODO: inflationary sampling section).  This
-is is used to justify propagating the head sampling probability even
-when unsampled.
+documented in [OTEP 148](TODO).  This is is used to justify
+propagating the head sampling probability even when unsampled.
 
 [An algorithm for making statistical inferance from partially-sampled
 traces has been published](https://arxiv.org/pdf/2107.07703.pdf) that
-explains how to work with power-of-2 sampling rates.  The reasoning
-behind restricting the set of sampling rates is:
-
-- Lowers the cost of propagating head sampling probability
-- Makes math involving partial traces tractable
+explains how to work with power-of-2 sampling rates.
 
 ## Trade-offs and mitigations
 
--- a/text/trace/0000-sampling-propagation.md
+++ b/text/trace/0000-sampling-propagation.md
@@ -64,8 +64,8 @@ These are extended [from the W3C
 examples](https://www.w3.org/TR/trace-context/#examples-of-http-traceparent-headers):
 
 ```
-Value = 00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01-05
-base16(version) = 00
+Value = 01-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01-05
+base16(version) = 01
 base16(trace-id) = 4bf92f3577b34da6a3ce929d0e0e4736
 base16(parent-id) = 00f067aa0ba902b7
 base16(trace-flags) = 01  // sampled
@@ -73,12 +73,12 @@ base16(log-count) = 05  // head probability is 2^-5.
 ```
 
 ```
-Value = 00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-00-00
-base16(version) = 00
+Value = 01-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-00-00
+base16(version) = 01
 base16(trace-id) = 4bf92f3577b34da6a3ce929d0e0e4736
 base16(parent-id) = 00f067aa0ba902b7
 base16(trace-flags) = 00  // not sampled
-base16(log-count) = 10 // head probability is 2^-16
+base16(log-count) = 11 // head probability is 2^-17
 ```
 
 We are able to express sampling probabilities as small as 2^-255 using


text/trace/0000-sampling-propagation.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Open questions
+
+This OTEP suggests how to modify the W3C trace context to accomodate
+sampling in OpenTelemetry.  [OTEP 148](TODO) suggests semantic
+conventions for encoding adjusted count in a Span, but neither text
+specifies how to modify the built-in Samplers to produce the proposed
+new `traceparent` field so that the `ParentBased` Sampler can
+correctly set the proposed `sampler.adjusted_count` attribute.  This
+will be future work.
```
oertl, line 126: I support this proposal. A related question to this is whether we also want to extend `traceparent` by a geometrically distributed random value with a success probability of 1/2 as described in https://arxiv.org/pdf/2107.07703.pdf (section 2.8) and as needed for trace ID ratio based sampling (if we want to avoid hashing the trace ID for every span). This random value could also be represented by a single byte and would have the following meaning:

* 0 would indicate that only spans with sampling rate equal to 1 (with `log-count` = 0) are recorded. 
* 1 would indicate that only spans with a sampling rate greater than or equal to 1/2 (with `log-count` <= 1) are recorded. 
* 2 would indicate that only spans with a sampling rate greater than or equal to 1/4 (with `log-count` <= 2) are recorded. 
* 3 would indicate that only spans with a sampling rate greater than or equal to 1/8 (with `log-count` <= 3) are recorded. 
* ...

As a consequence, the sampled flag would become obsolete as it can be directly derived from a comparison of `log-count` with the random value. Since the geometrically distributed random value would require adding another field and hence also a further upgrade of the `traceparent` version, it should be discussed, if we would like to achieve both at once.

Btw, the proposed geometrically distributed random value is a discrete version of the `sampling score` as proposed in https://github.com/open-telemetry/oteps/pull/135.

text/trace/0000-sampling-propagation.md
```diff
@@ -0,0 +1,126 @@
...
+sampling probability, which is the base-2 logarithm of the adjusted
+count of a child span created in this context (i.e., the logarithm of
+the effective count, thus "log-count").  To compute the adjusted count
+of a child span created in this context, use `2^log-count`.  A
+log-count of `0` corresponds with `(2^0)=1`, thus 0 conveys a context
+with probability 1.
+
+The sampling probability of a context is independent from whether it
+is sampled.  We consider it [useful to convey sampling probability
+even when unsampled]() as it can be used to estimate the potential
```
carlosalberto, line 58: Link?

text/trace/0000-sampling-propagation.md
```diff
@@ -0,0 +1,126 @@
+# Propagate head trace sampling probability
+
+Propose extending the W3C trace context `traceparent` to convey head trace sampling probability.
+
+## Motivation
+
+The head trace probability is useful in child contexts to be able to
+record the effective sampling probability in child spans.  This is
+documented in [OTEP 148](TODO), which establishes semantic conventions
```
carlosalberto, line 9: Missing link ;)

jmacd: I do not like to place links to active PRs, so I would wait for OTEP 148 to merge before finishing these. The link is to Dapper's inflationary sampler which requires knowing the probability of negative decisions in order to properly inflate the probability at a new root.

text/trace/0000-sampling-propagation.md
jmacd: Dappers inflationary sampler is described in OTEP 148 https://github.com/open-telemetry/oteps/blob/88d38b5e82e54317c692a62e730e7aef030dfac2/text/trace/0148-sampling-probability.md#dappers-inflationary-sampler

text/trace/0000-sampling-propagation.md
jmacd: @oertl Thank you. 

> the sampled flag would become obsolete as it can be directly derived from a comparison of `log-count` with the random value

Your proposal seems to require that `TraceIDRatio` be the only probability sampler, which leads me to wonder why we would ever propagate the `is-sampled` bit. I think we have to keep a way for the parent to signal to the child that a span should be recorded without giving a probability, although these spans are difficult to count.

Your proposal suggests to me the use of a new flag to signal probabilistic sampling as opposed to non-probabilistic sampling. If we accept your proposal and keep the is-sampled bit for non-probability sampling, do we still need to propagate the head probability?

I will have to think about this a bit more. 

> discrete version of the `sampling score` as proposed in #135

This makes sense. It makes me wonder, though, if we have a way to generate the correct randomness, what's really keeping us from using random bits in TraceID? If we have random bits in the TraceID, the number-of-leading-zeros test can be applied directly, right?

jmacd: @oertl You proposed something related

> whether we also want to extend traceparent by a geometrically distributed random value

and I am trying to determine whether to include more text in this OTEP. It's tempting to call these independent solutions to independent problems, but maybe that is not the case.

One way of framing the problem is this. We want to ensure that the `sampler.adjusted_count` attribute proposed in #170 (previously #148) is always correct regardless of sampler configuration, as long as each participant acts in good faith. In this framing, the two problems are connected because we have to anticipate a system that uses a mixture of `AlwaysOn`, `AlwaysOff`, `Parent`, and `TraceIDRatio` samplers. 

At a minimum we need two new specifications involving W3C trace context and/or how it is used in OTel:

1. We have to propagate the head sampling probability when `Parent` is used
2. We have to specify a how to obtain randomness when `TraceIDRatio` is used

We also need to show that there is a reasonable expectation that `sampler.adjusted_count` is always set correctly if Samplers are updated according to this specification, no matter which one of them is used. I will work on a revision to this OTEP to clarify this intended outcome.

jmacd: @oertl I've been considering my own question above, 

> if we have a way to generate the correct randomness, what's really keeping us from using random bits in TraceID?

and I gather that the answer is -- it requires less randomness with your proposal. In the version where we add a single byte to convey randomness, the expected number of random bits required is 2 in this case (for the reader, I believe we're talking about the mean of a https://en.wikipedia.org/wiki/Geometric_distribution with p=0.5). Can you confirm?

What you propose, in practice, might take a buffer of random bits and for each new trace context root created it would set a new `random` field in the W3C trace context, as in:

```
traceparent: (version)-(trace_id)-(span_id)-(flags)-(log_count)(random)
```

with a new flag `randomized`, value 0x4 to support a new field.

```
// Logic for TraceID root sampler
 r := 0
 for {
     if nextRandomBit() {
       break // The expected value of r is 2
     }
    r++
}
// Set random to r for child contexts that will use their own TraceID probability
// Set is-sampled based on r and the root's configured TraceID probability
// Set log_count to the negative base-2 logarithm of the root's configured TraceID probability
```

Please confirm my understanding, thanks!

oertl:
@jmacd 

> This makes sense. It makes me wonder, though, if we have a way to generate the correct randomness, what's really keeping us from using random bits in TraceID? If we have random bits in the TraceID, the number-of-leading-zeros test can be applied directly, right?

Yes, if the trace ID is specified to be truly random, it would be possible to use it without hashing.

> ...and I gather that the answer is -- it requires less randomness with your proposal. In the version where we add a single byte to convey randomness, the expected number of random bits required is 2 in this case (for the reader, I believe we're talking about the mean of a https://en.wikipedia.org/wiki/Geometric_distribution with p=0.5). Can you confirm?

Yes, the given pseudocode generates a geometrically distributed random number with p=0.5 and takes just 2 random bits on average. 

What I was trying to say is that all samplers (`AlwaysOn`, `AlwaysOff`, `Parent`, and `TraceIDRatio`) can be implemented using the same logic as the `TraceIDRatio` sampler, provided that the sampling rate of the parent is propagated. They only differ how the sampling rate is chosen for each span. For `AlwaysOn` it is 1, for `AlwaysOff` it is 0, for `Parent` it is the parent sampling rate, and for `TraceIDRatio` it is the fixed sampling rate. One could therefore say that a sampler should only be responsible for the selection of the sampling rate. The sampling decision is then always made according to the trace ID ratio logic.

To better reflect that in the interface, I would even redefine the sampler interface which is currently a function mapping to a boolean indicating the sampling decision. If it is defined as a function that maps each individual span to the corresponding sampling rate, the sampler implementations would be very simple. For example, `TraceIDRatio` would just return its fixed sampling rate and `Parent` would return the sampling rate of its parent. The final sampling decision would be made outside of the sampler following a specified and fixed logic based on the shared random number (or trace ID) and the sampling rate only.

The proposed redefined sampler interface would have following benefits:
* Any sampling decision will be consistent, because the same logic is used to get a sampling decision based on the sampling rate and the shared random number (or trace ID).
* The sampling rate and therefore the adjusted count is known for all spans as the interface enforces returning a sampling rate for every span.
* For estimation and further processing it is sufficient to know the sampling rate. It is not necessary to know which sampler was used to determine the sampling rate. What matters is just the sampling rate. This makes recording the sampler name (which would introduce additional overhead and which was mentioned here https://github.com/open-telemetry/oteps/pull/148#issuecomment-885323810) obsolete.
* Furthermore, new samplers which, for example, choose sampling rates more dynamically than `TraceIDRatio` based on span attributes, are easy to implement and would not break estimation algorithms.

text/trace/0000-sampling-propagation.md
jmacd: (confirmed below: https://github.com/open-telemetry/oteps/pull/168#issuecomment-888369265)

jmacd:
@oertl Thank you. I definitely understand your proposal. You've described a definite simplification in the Sampler interface, which is an SDK specification topic that I was hoping to avoid. I agree with your summary, the overall Sampler outcome would be easier to reason about with the change you suggest, but I am not sure that legacy uses of the API will agree. The amount of energy it will make this happen may exceed the benefit produced.

I am trying to keep these OTEPs separate: 

#170 is about the data model for Span (formerly known as #148)
#168 is about uses the W3C trace context (this PR)

Changing the Sampler API would be another issue entirely.

I will revise this PR based on the discussion in https://github.com/open-telemetry/oteps/pull/168#discussion_r677859626, which means combining the two pieces of additional information that go into the `traceparent` to both include the head probability and the randomly distributed value, as they are both apparently needed to complete a functional span-to-metrics pipeline.

yurishkuro:
Personally I would prefer to separate the propagation proposal from the W3C format discussion. Whatever happens with this OTEP is not going to have any (administrative) impact on changing the W3C spec, the discussion would need to happen there. On the other hand, the actual functionality is not dependent on changing the W3C format because it can be achieved by using `tracestate`, which would be completely in the hands of OTEL & if adopted & successful would serve as a strong motivation for upgrading the `traceparent` spec.

jmacd:
@oertl and @yurishkuro 

> This makes recording the sampler name (which would introduce additional overhead and which was mentioned here -- #148) obsolete.

In my proposal, in the text of of OTEP 170 (now), the `sampler.name` attribute is only a MUST when the adjusted count is unknown, which would only result from legacy trace contexts (if we succeed). I see no reason to record the `sampler.name` otherwise, especially if as you suggest the `TraceIDRatio` decision is used consistently.

jmacd:
@yurishkuro Thank you for your guidance. I have revised this proposal only to use `tracestate`.

@oertl I have revised this to include your proposal with a syntax like:

```
tracestate: otelprob=PPRR
```

where PP is the base16 probability value and RR is the base16 randomness value. Please take another look.

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,181 @@
...
+information we can convey a consistent sampling decision for sampling
+rates as small as 2^-255.
+
+### Proposed `tracestate` syntax
+
+The consistent sampling decision and head trace sampling probability
+will be propagated using four bytes of base16 content, as follows:
+
+```
+tracestate: otelprob=PPRR
```
yurishkuro, line 87: This does not match w3c spec (cf. https://www.w3.org/TR/trace-context/#combined-header-value), the expectation is that a vendor uses only a single key

> Example: `vendorname1=opaqueValue1,vendorname2=opaqueValue2`

We need to pick a vendor key for all of OTEL (`otel` or even just `ot`), and then decide on the encoding format for parts/attributes (which w3c does not prescribe)), for example:

```
tracestate: ot=p:PP;r:RR
tracestate: otel=pr:PPRR
```
Colons and semi-colons are allowed in values, but we need to consider if some future parts of state may need to contain (semi)colons.

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,181 @@
...
+For example, the probability value 2 corresponds with 1-in-4 sampling,
+the probability value 10 corresponds with 1-in-1024 sampling.  Using
+one byte of information we can convey sampling rates as small as 2^-255.
+
+### Random value
+
+With head trace sampling probabilities limited to powers of two, the
+amount of randomness needed per trace context is limited.  A
+consistent sampling decision is accomplished by propagating a
+geometrically distributed random variable with shape parameter `1/2`,
```
yurishkuro, line 52: > geometrically distributed random variable

define it?

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,181 @@
...
+requiring only two bits of randomness on average per trace.  See
```
yurishkuro, line 53: not following this

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,181 @@
...
+r := 0
+for {
+  if nextRandomBit() {
+    break // The expected value of r is 2
+  }
+  r++
+}
+```
+
+This can be computed from a stream of random bits as the number of
```
yurishkuro, line 71: if the intent is to use traceIdRatio sampler, this section is completely ignoring the randomness already present in the trace id

oertl (approved):

text/trace/0168-sampling-propagation.md
jmacd: Yes. This has been discussed in past meetings of the Sampling SIG, which led to this issue: https://github.com/open-telemetry/opentelemetry-specification/issues/1826

The OTel specification says very little about how random a TraceID should be. @carlosalberto wrote:

> One detail to be aware of here is that some vendors may be providing their existing ids, so we have to make sure this works in all scenarios.

if we are not able to dictate the degree of actual randomness, we may not be able to use the TraceID for effective sampling by itself. The other options that have been discussed are this option (in this PR, to attach a little additional randomness) and to specify a standard hashing function to use. Of those options, this option seems easiest and most reliable. 

Just saying "16 random bytes" doesn't say enough. If we say that every bit of the 128-bit TraceID is independently and identically random with p=0.5, then the random property of that ID we are trying to compute is the number of leading zeros. If we can assert that every bit is perfectly random, then we do not need this additional random variable to be propagated.

carlosalberto (approved):

text/trace/0168-sampling-propagation.md
carlosalberto: Sounds like a plan. Shall we define _first_ in the Specification the way to populate `tracestate` ourselves @yurishkuro? Happy to do that, but I'd like if that's what you foresee.

yurishkuro: @carlosalberto there are two independent things - how the tracestate format should look, and how SDK should interact with it. The latter is more complicated and I would say requires a prototype first, because there may be multiple touch points in the SDK that would need to mutate tracestate.

jmacd:
I have prototyped changes described in this OTEP [here](https://github.com/open-telemetry/opentelemetry-go/compare/main...jmacd:jmacd/propagate?expand=1). I was able to make this change without modifying the OTel-Go `Sampler` API.

The prototype raises several issues around specifying how to handle errors, and I think we can specify the behavior  of errors in such a way that consumers get an unambiguous signal about sampling.

The greatest question the prototype raises for me is how to handle zero probability. In the prototype, I've placed a limit on the range of the geometric random variable with (as in @oertl's paper) the value 0 representing a span that meets every sampling threshold, 1 representing a span that is sampled 50% of the time, and so on.  I believe we should limit this value to a reasonable number--I used 64 as the maximum value of this number and I think it best if we let this maximum value represent zero probability. This makes 63 the greatest value, which corresponds with an adjusted count of 2^63. The question this raises for me is how to handle the boundary condition--in practice it probably does not matter, but how should a Sampler behave if it sees a string of 64 leading zeros? (Should it reject the trial and start again? I think this is correct.) 

To summarize: random value 64 represents a probability of zero, not of 2^-64. @oertl what do you think?

oertl:
@jmacd I would limit the smallest sampling with 1/2^62 as 2^62 is the largest power of two that can be represented with a signed 64-bit integer (important to represent the adjusted count in languages like Java which only have signed data types). As a consequence, the exponent of the sampling rate X could be encoded with just 6 bits. The value X=63 is still free and can be used to represent a sampling rate of zero.

The number of leading zeros of a 64bit random number gives values from 0,1,2,...,64. I would trim to values 0,1,2,...,62 such that 6 bits are also sufficient to encode the propagated random number R. Here are the possible values with corresponding probabilities

- R=0 occurs with 1/2^1 probability and all spans are sampled with sampling rate >= 1/2^0 
- R=1 occurs with 1/2^2 probability and all spans are sampled with sampling rate >= 1/2^1
- ...
- R=61 occurs with 1/2^62 probability and all spans are sampled with sampling rate >= 1/2^61
- R=62 occurs with 1/2^62 probability and all spans are sampled with sampling rate >= 1/2^62

Hence, a span is sampled if X <= R. If X=63 sampling will never occur, because the maximum value of R is 62.

text/trace/0168-sampling-propagation.md
dyladan: I agree with yuri that `tracestate: otelprob=PPRR` is not the intention of the w3c spec and that we should have a single key and define our own structure within the value. `ot` or `otel` both seem fine to me. Yuri's proposed semicolon-delimited value format `key1:value1;key2:value2` seems reasonable enough to me. I would probably prefer `pr:PPRR` to `p:PP;r:RR` because you need both keys in order for the value to be useful and the first option is shorter and makes it more clear the values are tightly coupled.

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,181 @@
...
+- Makes math involving partial traces tractable.
+
+[An algorithm for making statistical inference from partially-sampled
+traces has been published](https://arxiv.org/pdf/2107.07703.pdf) that
+explains how to work with a limited number of power-of-2 sampling rates.
+
+### Behavior of the `TraceIDRatioBased` Sampler
+
+The Sampler must be configured with a power-of-two probability
+`P=2^-S`.
```
dyladan, line 130: Defined limit of `S<=62` per @oertl comment?

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,181 @@
...
+limit valid head sampling probabilities to `P=2^-254` or greater to
+address this ambiguity.
+
+If the context is a new root, the initial `tracestate` must be created
+using geometrically-distributed random value `R` (as described above,
+with maximum value 254) and the initial head probability value `S`.
+
+If the context is not a new root, output a new `tracestate` with the
+same `R` value as the parent context, using the Sampler's own value of
+`S` for the head probability.
```
dyladan, line 145: Just to be clear, I am propagating tracestate with my own S value, not the value I received from the parent?

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,181 @@
...
+
+The consistent sampling decision and head trace sampling probability
+will be propagated using four bytes of base16 content, as follows:
+
+```
+tracestate: otelprob=PPRR
+```
+
+where `PP` are two bytes of base16 probability value and `RR` are two
+bytes of base16 random value.
```
dyladan, line 91: If we limit to `2^-62` per @oertl suggestion, does it make sense still to use base 16? Base 10 is more human friendly and they both will take 2 characters to represent the max value so there is no space savings to use base 16.

text/trace/0168-sampling-propagation.md
jmacd: @carlosalberto has a draft on this: https://github.com/open-telemetry/opentelemetry-specification/pull/1852

text/trace/0168-sampling-propagation.md
jmacd: I've rewritten this passage, hoping to clarify.
```
A
consistent sampling decision is accomplished by propagating a specific
random variable.  The random variable is a described by a discrete
geometric distribution having shape parameter `1/2`, listed below:

| Value | Probability |
| ----- | ----------- |
| 0 | 1/2 |
| 1 | 1/4 |
| 2 | 1/8 |
| 3 | 1/16 |
| 4 | 1/32 |
| ... | ... |
| N | 1/(2^(N+1)) |

See [Estimation from Partially Sampled Distributed
Traces](https://arxiv.org/pdf/2107.07703.pdf) section 2.8 for a
detailed explanation.

Such a random variable `r` can be generated using the following
pseudocode:

```
r := 0
for {
  if nextRandomBit() {
    break // The expected value of r is 2
  }
  r++
}
```
```

text/trace/0168-sampling-propagation.md
jmacd: I've written this OTEP under the assumption that we can assume nothing definite about TraceID randomness at the individual bit level. I believe it would take a W3C revision to do better than this. Although I am willing to continue updating this document with more possibilities if necessary, I think that's out of scope here.

text/trace/0168-sampling-propagation.md
jmacd: I've changed the text to refer to "6 bits" of randomness, with 63 reserved to mean zero probability.

text/trace/0168-sampling-propagation.md
jmacd: Yes, this is what I mean by "the Sampler's own value of `S`". I'll try to improve this text.

text/trace/0168-sampling-propagation.md
jmacd: Yes, this is a good point. When @carlosalberto and I discussed this very point, we also considered that if these values were moved into the `traceparent` header, where they could really take just a byte or two, it would be more consistent to use base16 like the rest of the traceparent.

jmacd:
See https://github.com/w3c/trace-context/issues/463

text/trace/0168-sampling-propagation.md
jmacd: I've improved this text.

text/trace/0168-sampling-propagation.md
jmacd: I've added to the text that this proposal is about how we propagate, not about tracestate syntax. Whatever we decide for tracestate will dictate what we do here.

text/trace/0168-sampling-propagation.md
jmacd: I've added to the caveats section this note about W3C and traceID randomness.

jmacd:
I've updated this OTEP after all the feedback, mentioning the prototype and the discussion in this week's W3C meeting.

jmacd:
While reviewing the `Span` protocol for [OTEP 170](https://github.com/open-telemetry/oteps/pull/170) I realized that `Span` encodes its `tracestate`

For the proposed `otel` vendor key within `tracestate` here, I would specify to DROP that entry (or at least the `r:` and `p:` sub-fields) before serializing a `Span` as those values are not directly useful without knowing which Sampler was used.

jmacd:
I have updated this PR to support a minimum sampling probability of 2^-61. This allows the Span field described in #170 to use six bits of information, not 7, to encode the additional Unknown probability value.

jmacd:
Discussed in the Sampling SIG:
- Do we want to make propagation of r and p values always on? (This costs ~30 bytes per context, in this proposal)
- Can we make this opt-in, or opt-out?
- Can we look forward to a future where `traceparent` version 1 supports this with ~3 bytes per context.

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,317 @@
...
+| 5         | 16             |
+| 6         | 32             |
+| 7         | 0              |
+
+Note there are only 6 non-zero, non-unknown values for the adjusted
+count. Thus there are six defined values of `R` and `S`.  The
+following table shows `R` and the corresponding selection probability,
+along with the calculated adjusted count for each `S`:
+
+| `R` value | `R` selection probability | `S=0` | `S=1` | `S=2` | `S=4` | `S=5` | `S=6` |
```
oertl, line 249: column `S=3` is missing

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,317 @@
...
+the W3C `sampled` flag and copies the incoming `tracestate` keys to
+the child context.
+
+### Behavior of the `AlwaysOn` Sampler
+
+The `AlwaysOn` Sampler behaves the same as `TraceIDRatioBased` with `P=1` (i.e., `S=0`)
+
+### Behavior of the `AlwaysOff` Sampler
+
+The `AlwaysOff` Sampler behaves the same as `TraceIDRatioBased` with `P=0` (i.e., `S=62`).
```
jsuereth, line 221: IIUC - This means selection probability is MAX VALUE vs. 0 (63). Is that a typo?

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,317 @@
...
+
+Restricting head sampling rates to powers of two does not limit
+Samplers from using arbitrary effective probabilities over a period of
+time.  For example, choosing 1/2 sampling half of the time and 1/4
+sampling half of the time leads to an effective sampling rate of 3/8.
+
+## Prior art and alternatives
+
+Google's Dapper system propagated a field in its trace context called
+"inverse_probability", which is equivalent to adjusted count.  This
```
jsuereth, line 315: This may have been true at one point, but is not now.

jsuereth (approved):
My comments are non-blocking on the general direction of this OTEP.

The open question in the comments (can this be opt-in, how many bytes, w3c context) can be resolved when hitting the specifcation.

text/trace/0168-sampling-propagation.md
jmacd: There were a number of off-by-one errors stemming from changes following https://github.com/open-telemetry/opentelemetry-specification/pull/1899. Fixed.

text/trace/0168-sampling-propagation.md
jmacd: 🔪

jmacd:
I revised this OTEP with more content and more examples, following discussion that took place in https://github.com/open-telemetry/opentelemetry-specification/pull/1899.

No significant changes have been made, only correcting off-by-one errors now that the `p` value of 0 indicates Unknown probability.

I added specification for several corner cases, suggesting that `p` can be omitted when not sampled and that `p` should not be respected when `r` is not set, for example.

The final version of this text uses:

`s` where sampling probability is `2**s`
`p` where 0 is unknown, 63 is zero, and 1-62 are known non-zeros
`r` in the range [0, 61]

and the sampled flag equivalent to `p-1 < r+1 && p < 63`. I think this definition makes the formulas most readable, since `s` has a natural form, `p` is the same as for the Span's `log_head_adjusted_count` (from OTEP 170), and `r` equals the number of leading zeros in a random bit string.

Please take a careful look @oertl, thanks!

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,368 @@
...
+
+## Summary
+
+The following table summarizes how the three Sampler cases behave with
+respect to the incoming and outgoing values for `p`, `r`, and
+`sampled`:
+
+| Sampler                | Incoming `r` | Incoming `p` | Incoming `sampled` | Outgoing `r`    | Outgoing `p`   | Outgoing `sampled` |
+| --                     | --           | --           | --                 | --              | --             | --                 |
+| Parent                 | unused       | expected     | respected          | passed through  | passed through | passed through     |
```
oertl, line 277: Parent-based sampler should not pass through inconsistent values of `r`, `p`, and `sampled` (see my comment below).

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,368 @@
...
+`tracestate` with an `otel` vendor tag, which will require
+(separately) [specifying how the OpenTelemetry project uses
+`tracestate` itself](https://github.com/open-telemetry/opentelemetry-specification/pull/1852).
+
+## Explanation
+
+Two pieces of information are needed to convey consistent head trace
+sampling probability:
+
+1. The head trace sampling probability
```
reyang, line 35: nit: add a `.` (or remove the `.` from the 2nd item)

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,368 @@
...
+
+For example, the value 3 means there were three leading zeros and
+corresponds with being sampled at probabilities 1-in-1 through 1-in-8
+but not at probabilities 1-in-16 and smaller.
+
+### Proposed `tracestate` syntax
+
+The consistent sampling randomness valuw (`r`) and and head sampling
+probability value (`p`) will be propagated using two bytes of base16 content
+for each of the two fields, as follows,
```
reyang, line 128:
```suggestion
for each of the two fields, as follows:
```

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,368 @@
...
+The following `tracestate` value:
+
+```
+tracestate: otel=r:0a;p:03
+```
+
+translates to
+
+```
+base16(probability) = 03 // 1-in-8 head probability
```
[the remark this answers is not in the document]
reyang: and the 2nd line is 1-in-512

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,368 @@
...
+but not at probabilities 1-in-16 and smaller.
+
+### Proposed `tracestate` syntax
+
+The consistent sampling randomness valuw (`r`) and and head sampling
+probability value (`p`) will be propagated using two bytes of base16 content
+for each of the two fields, as follows,
+
+```
+tracestate: otel=p:PP;r:RR
```
reyang, line 131: tracestate key value pairs might get [removed based on FIFO](https://www.w3.org/TR/trace-context/#tracestate-limits), do we have recommendation for OpenTelemetry SDKs such as "the OpenTelemetry SDKs are encouraged to preserve otel tracestate as much as possible, for example, always pop it to 1st key-value pair of tracestate" (which might be a bit off from the W3C TraceContext perspective).

reyang (approved):
LGTM assuming the off-by-one issues mentioned by @oertl will be fixed.

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,368 @@
...
+that cannot be generated by the built-in samplers.  The case where
+sampled is true with `p=63` indicating 0% probability may be used when
+recording spans that were selected by a different sampler while a
+probability sampler is also in use.  These cases are known as "zero
+adjusted count" contexts which are sampled with 0% probability.
+
+The case where sampled is false with `p=1` indicating 100% probability
+is an illogical condition.  See [Propagating `p` when
+unsampled](#propagating-p-when-unsampled) below.
+
```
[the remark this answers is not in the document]
jmacd: I prefer the first of your remedies, but want to refine it to accommodate the zero probability (`p=63`) as being allowed as a special case to allow sampling with zero adjusted count. Other than this special case, the outgoing `p` should be dropped, which is effectively 0 (unknown), when the invariant is broken. I've rewritten this section, please check. Thanks for pointing out how to simplify this!

text/trace/0168-sampling-propagation.md
jmacd: I rewrote this as "checked and passed through", then added a table of range checks that should be applied, then documented the invariant as you stated it. I want there to be a special case for `sampled` is true and `p=63` to allow intentional sampling with zero adjusted count. I would be glad to prototype a composite sampler that applies either the probability sampler's decision (resulting in p in [1..62]) or a non-probability sampler's decision (effectively `p=63`, saying "collect this for some reason but don't count it") to show this in action.

jmacd:
@open-telemetry/specs-trace-approvers please take a look. The feedback has been addressed and the corner cases are explicitly spelled out. Thank you @oertl and @reyang.

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,391 @@
...
+(separately) [specifying how the OpenTelemetry project uses
+`tracestate` itself](https://github.com/open-telemetry/opentelemetry-specification/pull/1852).
+
+## Explanation
+
+Two pieces of information are needed to convey consistent head trace
+sampling probability:
+
+1. The head trace sampling probability.
+2. Source of consistent sampling decisions.
```
yurishkuro, line 36: styling suggestion:
```suggestion
1. p-value representing the head trace sampling probability.
2. r-value representing the "randomness" as the source of consistent sampling decisions.
```
and refer to these as p-value and r-value instead of full names, because full names are misleading (p=2 is not a "probability value" by itself, but a specially encoded representation of such)

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,391 @@
...
+### Behavior of the `AlwaysOff` Sampler
+
+The `AlwaysOff` Sampler behaves the same as `TraceIDRatioBased` with
+zero probability (i.e., `p=63`).
+
+## Worked 3-bit example
+
+The behavior of these tables can be verified by hand using a smaller
+example.  The following table shows how these equations work where
+`r`, `p`, and `s` are limited to 3 bits.
```
yurishkuro, line 232:
```suggestion
`r`, `p`, and `s` are limited to 3 bits instead of 6 bits.
```

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,391 @@
...
+
+### Randomness value
+
+With head trace sampling probabilities limited to powers of two, the
+amount of randomness needed per trace context is limited.  A
+consistent sampling decision is accomplished by propagating a specific
+random variable denoted `r`.  The random variable is described by a
+(truncated) geometric distribution having shape parameter `1/2`, listed below:
+
+| `r` Value        | Selection Probability | Sampling probability   |
```
yurishkuro, line 91: I don't know how many people will be able to follow this. How about starting with a simpler explanation? Like generating a uniformly distributed 61-bit(ish) random value R such that a classic probabilistic sampler with probability P would sample if R <= P. That's the baseline that probably everyone will understand. Then we take into account that if P is discrete as 2**-s, then R can be similarly discreet, and its encoding can be reduced to counting the number of leading 0 bits (a couple concrete examples would probably help a lot).

The fact that r is a geometric distribution - is that relevant? Sure, we can mention it, after the more common sense explanation.

Finally, why do we need to list THAT algorithm (nextRandomness)? That's not how it will be implemented in practice in the SDKs. Why not show the more realistic algorithm that generates an n-bit random number and does bit-shifting to produce r-value?

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,391 @@
...
+- Lowers the cost of propagating head sampling probability
+- Limits the number of random bits required
+- Avoids floating-point to integer rounding errors
+- Makes math involving partial traces tractable.
+
+[An algorithm for making statistical inference from partially-sampled
+traces has been published](https://arxiv.org/pdf/2107.07703.pdf) that
+explains how to work with a limited number of power-of-2 sampling rates.
+
+### Behavior of the `TraceIDRatioBased` Sampler
```
yurishkuro, line 179: I would recommend introducing a new sampler with a different name and not touching the TraceIDRatioBased at all, because the actual mechanism has nothing to do with "trace-id ratio" anymore.

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,391 @@
...
+| 2         | 2              |
+| 3         | 4              |
+| 4         | 8              |
+| 5         | 16             |
+| 6         | 32             |
+| 7         | 0              |
+
+Note there are only 6 non-zero, non-unknown values for the adjusted
+count. Thus there are six defined values of `r` and `s`.  The
+following table shows `r` and the corresponding selection probability,
```
yurishkuro, line 251: Why do we care about "`r` selection probability"? What is the objective - to generate r-value or to sample the trace?

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,391 @@
...
+
+## Trade-offs and mitigations
+
+### Not using TraceID randomness
+
+It would be possible, if TraceID were specified to have at least 61
+uniform random bits, to compute the randomness value described above
+as the number of leading zeros among those 61 random bits.
+
+This proposal requires modifying the W3C traceparent specification,
```
yurishkuro, line 339:
```suggestion
However, this would require modifying the W3C traceparent specification,
```

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,391 @@
...
+therefore we do not propose to use bits of the TraceID.
+
+[This issue has been filed with the W3C trace context group.](https://github.com/w3c/trace-context/issues/463)
```
yurishkuro, line 342: the linked issue is not about guaranteeing randomness of TraceID.

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,391 @@
...
+  r := 0
+  for r < 61 && nextRandomBit() == false {
+    r++
+  }
+  return R
+}
+```
+
+This can be computed from a stream of random bits as the number of
+leading zeros using efficient instructions on modern computer
```
yurishkuro, line 119: > leading zeros

or trailing zeros

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,391 @@
...
+
+If `sampled` is false and the invariant is bilated, drop `p` from the
+outgoing context to convey unknown head probability.  Set
+`log_head_adjusted_count` to 0.
+
+The case where `sampled` is true with `p=63` indicating 0% probability
+may by regarded as a special case to allow zero adjusted count
+sampling, which permits non-probabilistic sampling to take place in
+the presence of probability sampling.  Set `log_head_adjusted_count`
+to 63.
```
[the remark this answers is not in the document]
jmacd: In words, I think we should support a sampling policy like "Sample at 3.125% AND at least 1 per hour". A probability sampler will compute the 3.125% sample (i.e., 2^-5 or `p=6`) and a non-probability sampler will compute the one per hour sample with `p=0`. We can spell out rules for composing these, and this is why I have referred to the important special case of zero adjusted count. The probability sampler outputs an adjusted count of zero for any span it does not select. If another sampler decides to sample, we have the important special case of a sampled trace with zero adjusted count -- i.e., a non-probabilistic sample.

Stepping back, there is a higher-level invariant guiding this discussion, one stating that the sum of adjusted counts equals the true number of spans in the population. The `sampled <=> p-1<=r` invariant only needs to be applied when `p<63` because when `p=63` we have by definition broken the higher-level invariant: any time you intentionally choose zero-probability sampling you break the higher-level invariant, so we may as well exclude the lower-level invariant at that point. Thus, the special case of `sampled=true` and `p=63`. We would still want to have `r` defined in that case, because a child span somewhere in the trace could change the decision and therefore begin using a non-zero adjusted count for a sub-trace.

text/trace/0168-sampling-propagation.md
jmacd: What would you propose that we do with `TraceIDRatioBased`, then? It has TODOs in the specification and the implementations we have rely on unspecified use of TraceID bits. If the best we can do is not touch the `TraceIDRatioBased` logic, then I propose we name it to `IncosistentProbabilityBased` and leave it as deprecated with the existing caveat that it should only be used at the root.

From the perspective of this OTEP, traces rooted in `InconsistentProbabilityBased` samplers would not be able to convey head probability.

From the perspective of OTEP 170, all spans in such a trace would have to record unknown adjusted count. We could _in the future_ resume discussions about a `sampler.adjusted_count` attribute to allow arbitrary sampling probability at the root span while propagating unknown head probability. Consumers could then use `InconsistentProbabilityBased` samplers w/ arbitrary probability at the root of their trace, perform trace assembly, and then perform span-to-metrics over sampled traces.

As for naming the proposed new Sampler, @oertl how do you feel about `ConsistentProbabilityBased`? 

(Actually, do we need all these "Based" strings anyway, could we just talk about "Parent", "TraceIDRatio", "ConsisterntProbability", ...?)

yurishkuro: > What would you propose that we do with TraceIDRatioBased, then

Deprecated it? Since the spec never said how it needs to be implemented, do existing SDKs even have this sampler today? And if they do, how can it be consistent across langs?

text/trace/0168-sampling-propagation.md
jmacd: Thank you. I've updated this text with a more gentle introduction and removed some of the mathy words. I've changed the example code to use `rand.Int63()` and `bits.LeadingZeros64()` from the Golang standard library, also posted some code to compute it several other ways and compare the costs:

https://gist.github.com/jmacd/79c38c1056035c52f6fff7b7fc071274

Assuming an SDK uses at least 61 bits of *independent* randomness in the TraceID, it should be easy to compute in any case. The reason for this exposition is that not all TraceIDs will have independent randomness, and in that case all they need to compute an r-value, on average, are two random bits.

text/trace/0168-sampling-propagation.md
jmacd: Included in the gist.

text/trace/0168-sampling-propagation.md
jmacd: OK. Yes the SDKs have implemented this Sampler because among the builtin samplers, it's the only one useful as a root other than always-on and always-off. It's just not specified in a consistent way. That's why I proposed it _could_ be kept as a sampler for non-power-of-two-probabilities, which the legacy TraceIDRatioBased sampler supports, but I think it would be better to replace it and round the input sampling probabilities to powers-of-two. It's fine to choose a new name, my goal is that we end up with just one behavior.

Could we just call this single new sampler the ProbabilitySampler? Could we merge this OTEP and debate names and whether to deprecate the legacy behavior in another PR?

text/trace/0168-sampling-propagation.md
jmacd: I think it's appropriate in an OTEP to explain technical details, they don't need to go into the specification. The reason we want to see the probability of each r-value is that it helps convince ourselves that the math works out. The r-value selection probabilities are being multiplied by adjusted counts in the example that is being worked out, that is why they are here.

Above, I've explained that an SDK is not required to use independent random bits in its TraceID. r-values can be calculated in any way that generates this distribution.

jmacd:
@open-telemetry/specs-trace-approvers 

This OTEP has received a lot of review and we are down to a number of questions that I feel should be resolved as we move this work into the specification--this PR and #170 could merge. The remaining questions:

1. Should the Sampler described in OTEPs #168 and #170 replace the currently underspecified TraceIDRatioBased sampler, or
2. Should the old TraceIDRatioBased sampler be deprecated or possibly kept under a new name
3. Should the propagation implied by OTEP #168 be opt-out or opt-in.  If this is an option, what is the option named? 

If you agree that these are the only remaining questions, please review and approve both OTEPs and we can move on to debating these topics.

text/trace/0168-sampling-propagation.md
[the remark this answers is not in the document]
jmacd: @yurishkuro and @pavolloffay please confirm that a policy like "Sample at 3.125% AND at least 1 per hour"  is a realistic sampler configuration in today's Jaeger sampler? This is a composite sampler made up of a probability sampler and a rate-limited sampler, for example.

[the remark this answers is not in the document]
yurishkuro: > please confirm that a policy like "Sample at 3.125% AND at least 1 per hour" is a realistic sampler configuration in today's Jaeger sampler

Confirmed. The "at least 1 per hour" part is useful when using adaptive sampling backend which works best when it's aware of all endpoints in the architecture, including those with low QPS that may otherwise never be sampled by the probability sampler (because there can be several orders of magnitude difference in QPS so the conservative default sampling probability has to be very low).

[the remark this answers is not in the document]
jmacd: The rule for composing more than one `log_head_adjusted_count` value will say that it is OK to combine any number of Unknown p-values with a single known p-value and to `OR` their `sampled` flags together.

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,419 @@
...
+
+When propagated, the "p-value" as it is known will be interpreted as
+shown in the folowing table.  The p-value for known sampling
+probabilities is the negative base-2 logarithm of the probability,
+offset by +1 to so that the 0 p-value can be treated as unknown (for
+backwards compatibility):
+
+| p-value | Head Probability | Note                                                 |
+| -----   | -----------      | ----                                                 |
+| 0       | Unknown          | Do not propagate `p=0`, instead omit from tracestate |
```
jmacd, line 65: Non-blocking comment for @jsuereth: Note there is a question about field presence here. If OTel adopts the use of proto3 field presence then we can distinguish between 0 and "not present", that would change (and simplify) the backwards compatibility story. Then p-values would equal the effective `s` value and the invariant would be written `sampled <=> p <= r`.

text/trace/0168-sampling-propagation.md
yurishkuro: I would prefer more special name like "ConsistentSampler" over "ProbabilitySampler", as the letter is pretty common in pre-OTEL SDKs and means "flip a p-weighted coin"

yurishkuro (approved):
lgtm. Still a bit concerned by the complexity of the explanation, would be great to have a few SDK maintainers to review this for understandability.

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,419 @@
...
+import (
+	"math/rand"
+	"math/bits"
+)
+
+func nextRValueLeading() int {
+	x := uint64(rand.Int63()) // 63 least-significant bits are random
+	y := x << 1 | 0x7         // 61 most-significant bits are random
+	return bits.LeadingZeros64(y)
+}
```
yurishkuro, line 146: alternative (may be off-by-one), which is language and std-lib neutral:
```go
const topbit = uint64(1) << 61
x := uint64(rand.Int63())
for n := 0; n < 61; n++ {
    if x & topbit == topbit {
        return n
    }
    x = x << 1
}
return 61
```

yurishkuro, line 146: alternative which is language & std-lib neutral (may be off by one):
```go
x := uint64(rand.Int63())
for r := 0; r < 61; r++ {
    if x & 0x1 == 0x1 {
        return r
    }
}
return 61
```

jmacd: I added this as a second example, computing the number of trailing zeros. I added it to the gist and verified that it computes the same result as the version using `bits.TrailingZeros64()`.

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,439 @@
...
+```
+
+Or we may compute the number of trialing zeros, for example:
```
yurishkuro, line 150:
```suggestion
Or we may compute the number of trailing zeros instead, for example:
```

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,439 @@
...
+
+```golang
+import (
+	"math/rand"
+)
+
+func nextRValueLeading() int {
```
yurishkuro, line 157:
```suggestion
func nextRValueTrailing() int {
```

yurishkuro (approved):

text/trace/0168-sampling-propagation.md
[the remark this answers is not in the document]
jmacd: We can recommend the use of adaptive probability sampling and we've shown a couple of examples of this in action, though we know that it's impossible to implement a hard rate limit without introducing bias.

My proposal allows easily combining non-probabilistic samplers (which all produce unknowns) with a single probabilistic sampler so that adjusted counts are always known, with no modification to the other samplers. It may be better to use an adaptive strategy, but it's a very different approach.

> switches once in an hour to a sampling rate of 100%.

This is true, but it doesn't give me a good feeling. Faced with non-probabilistic Samplers, we can either force them to become probabilistic or accommodate known zeros. I think that both are viable, but it is more difficult to reason about the data produced by an adaptive sampler, and if we're not careful there may be temporal or systematic bias produced. It's out-of-scope go into a detailed analysis of this here, but merely the fact that sophisticated analysis is required suggests that some users will better off with a simple approach.

> kind of unnatural anyway

This may just be me, but it feels natural to use zero for this application. We can add any number of zeros to a sum without changing the expected value.

I agree that there is something unnatural about sampling at 0%, because it is not possible to preserve an expected value of the sum when doing so. It is biased by definition. But for the application at hand, we are sampling with a non-zero probability and the use of zero meaningfully conveys that decision in a way that allows non-probabilistic samplers to exist. I'm not against adaptive sampling, it's just more complex and I'd like to offer both options.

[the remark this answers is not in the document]
oertl: @jmacd

> combining non-probabilistic samplers (which all produce unknowns)

I am still trying to understand that. Does combining multiple samplers (one probabilistic and multiple non-probabilistic samplers) mean applying them all to the same span? If yes, is it possible that multiple samplers decide to sample the same span at the same time? If so, which value is propagated for `p`? The propabilistic sampler would set `p<63` while the non-probabilistic sampler would set `p=63`.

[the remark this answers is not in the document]
jmacd: > while the non-probabilistic sampler would set p=63.

**The non-probabilistic sampler would set `p=0` for unknown.**

The composite sampler will see a number of unknowns `p=0` and a single known `p>0`. The resulting `p` is the known value (63 or less than 63, just not 0), while `sampled` may be true or false.

In other words, as long as we have one probability sampler we always know the adjusted count. We can combine the one probability sampler with any number of non-probability samplers.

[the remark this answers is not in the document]
oertl: I am not talking about the adjusted count, but what is propagated in the `tracestate`. Non-probabilistic samplers are required to set `sampled = true` and `p = 63` (sampling rate = 0) in order to ensure that all subsequent parent-based samplers also report zero adjusted counts while probabilistic samplers are required to set `sampled = true` and `1 <= p < 63` (sampling rate > 0). As there is only one `tracestate` for the same span, there is a conflict. It must be somehow ensured that non-probabilistic samplers do never overwrite the `tracestate` of the probabilistic samplers if they are applied to the same span.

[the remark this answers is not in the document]
oertl: The only reason for zero adjusted counts that I can see is the parallel use of multiple samplers applied to the same spans. In this case it must be ensured that only one of the samplers delivers non-zero adjusted count values in order to get unbiased results. 

@jmacd
> I think that both are viable, but it is more difficult to reason about the data produced by an adaptive sampler, and if we're not careful there may be temporal or systematic bias produced. It's out-of-scope go into a detailed analysis of this here, but merely the fact that sophisticated analysis is required suggests that some users will better off with a simple approach.

The simultaneous application of multiple samplers to the same span makes things much more complicated (see also my previous  comment) than a single sampler that chooses the sampling rate dynamically. Any sampler implementation that picks some nonzero sampling rate (possbily different for every single span and dependent on some internal logic) and makes the sampling decision according to `sampled <=> p-1 <= r` will be unbiased by definition. Actually, this allows a lot of freedom how to sample. (See my prototype implementations. Any implementation of [AbstractConsistentSampler](https://github.com/dynatrace-research/opentelemetry-sampling-poc/blob/main/poc/src/main/java/com/dynatrace/research/otelsampling/sampling/AbstractConsistentSampler.java) will be unbiased. Furthermore, different samplers can be easily composed as demonstrated by the [ComposedSampler](https://github.com/dynatrace-research/opentelemetry-sampling-poc/blob/main/poc/src/main/java/com/dynatrace/research/otelsampling/sampling/ComposedSampler.java)). Reasoning about the data is easier. There is the advantage that adjusted counts will be always greater than or equal  to 1 and that there will never be any duplicates. In the presence of multiple samplers, the same span might be sampled more than once, once with a non-zero adjusted count and the other times with a zero adjusted count.

I think using multiple samplers in parallel to sample the same span should be generally avoided.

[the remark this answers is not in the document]
jmacd: > The only reason for zero adjusted counts that I can see is the parallel use of multiple samplers applied to the same spans. In this case it must be ensured that only one of the samplers delivers non-zero adjusted count values in order to get unbiased results.

You wrote "In this case it must be ensured that only one of the samplers delivers **non-zero** adjusted count values in order to get unbiased results." (emphasis mine) but this is NOT what I'm describing. I am saying "it must be ensured that only one of the samplers is a probability sampler."

I am using the expression "non-probability sampler" to describe a decision that is not based in probability and therefore **MUST have Unknown ​probability**.  The only composition I have described is a single probability sampler **having a single Known probability** combined with any number of non-probability samplers. The combination allows for any number of Unknowns and a single Known. 

We are not specifying how to multiple probability samplers in parallel.  The `log_head_adjusted_count` field specified in OTEP 170 is specified as a sample of all spans. As it covers all spans in "the" population, there could not be a need for more than one of them.

If a user did want to use multiple probability samplers in parallel, then I would presume: (1) they are independent, (2) they cover sub-populations of spans. In an early draft of OTEP 170 I spelled out how this might look, e.g., with span attributes like `sampler.name.adjusted_count` for Sampler named `name`. This could be specified later, and when that time comes it will become natural to set `log_head_adjusted_count=63` because there are multiple parallel probability samplers.

For now, the specified reason to use `log_head_adjusted_count=63` is that you have an always-off sampler.

I will update this PR with more information on composing Known and Unknown probability values, since there is an immediate use for combining probability and non-probability samplers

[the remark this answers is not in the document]
oertl: @jmacd I finally figured out a use case which justifies to break the invariant `sampled <=> p-1 <= r`. Maybe this is exactly what you meant. Basically, if `sampled`, `p`, and `r` are all known, the invariant `sampled <=> p-1 <= r` should be never violated when propagating them to child spans.

However, a valid use case is, for example, to sample up to 100 spans every minute. This can be achieved with a reservoir sampling buffer of size 100 that is reset every minute. It is possible to do reservoir sampling in a consistent way (see [ReservoirSampler.java](https://github.com/dynatrace-research/opentelemetry-sampling-poc/blob/main/poc/src/main/java/com/dynatrace/research/otelsampling/sampling/ReservoirSampler.java)). The problem is that the sampling decision is not known at propagation time. Only at the end of every minute, it is known which spans survived in the reservoir and got finally sampled. This means that neither `sampled` nor `p` can be correctly set and propagated. Therefore, reservoir sampling is incompatible with parent-based sampling which requires `sampled` and `p` of the parent to be known. (Reservoir sampling is compatible with all other consistent built-in samplers that do not depend on the parent sampling decision.) Since `sampled` is not an optional field, `sampled=unknown` cannot be directly expressed. A workaround would be to set  `sampled=true` together with `p=63` (sampling probability = 0%) which is a contradiction and violates the invariant, but can be used to express the special situation where both `sampled` and `p` are not known at propagation time.

I have the feeling that we are mixing up `p` and `log_head_adjusted_count` in our discussion. As far as I understand, `p` is the value that is propagated to child spans, while `log_head_adjusted_count` describes the adjusted count of sampled spans. The parent-based sampler uses the incoming `p` to set `log_head_adjusted_count=p`, if the span gets sampled. I am still not aware of any use case for reporting spans with `log_head_adjusted_count=63` (which corresponds to adjusted count = 0). I can imagine cases for `log_head_adjusted_count=0` (which corresponds to adjusted count = unknown). For example, when a parent-based sampler did not receive a valid `p` value while `sampled=true`.

(The only situation I can imagine where adjusted count = 0 could be useful is when multiple samplers are simultaneously applied to the same span. In this case only one primary sampler is allowed to report nonzero adjusted counts while all others have to set adjusted count = 0 for their sampled spans to avoid bias. However, I think you agreed that using multiple samplers in parallel does not make much sense.)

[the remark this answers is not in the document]
jmacd: This thread is too long for anyone to read. :-)

I think we're actually agreeing about things and have become befuddled with our own terminology, so I've updated the text of recently-merged #170 with detail to help clarify this topic. The new PR #175 contains a table that specifies how to compose samplers.

I realized I never responded to some of the `AbstractConsistentSampler` and `ComposedSampler` above. In the new table in #175 I've written that to combine two consistent samplers, take the `min()` p-value (as you stated), but it also covers the case of combining known and unknown values as I've stressed it should.

jmacd:
@open-telemetry/specs-trace-approvers as announced in the Monday maintainer's meeting, this will proceed next week unless objections are raised here. There was one lengthy review comment thread (https://github.com/open-telemetry/oteps/pull/168#discussion_r704955698) that I resolved by opening a PR to revise OTEP 170 with more detail on how to compose Samplers, please see #175.

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,454 @@
...
+value of `s+1` in the range [1, 62].  If the sampling probability is
+zero (the special case where `s` is undefined), use `p=63` the
+specified value for zero probability.
+
+In both cases, set the sampled bit if the outgoing `p` minus one is
+less than or equal to the outgoing `r` (i.e., `p-1 <= r`).
+
+If the context is not a new root and the incoming context's r-value
+is not set, the implementation SHOULD notify the user of an error
+condition and follow the incoming context's `sampled` flag.
```
bogdandrutu, line 248: This statement makes this feature "required" for all deployments, which I think it should not be.

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,454 @@
...
+### Not using TraceID hashing
+
+It would be possible to make a consistent sampling decision by hashing
+the TraceID, but we feel such an approach is not sufficient for making
+unbiased sampling decisions.  It is seen as a relatively difficult
+task to define and specify a good enough hashing function, much less
+to have it implemented in multiple languages.
+
+Hashing is also computationally expensive. This proposal uses extra
+data to avoid the computational cost of hashing TraceIDs.
```
bogdandrutu, line 416: I would disagree with this argument, it is more expensive to parse the strings from the tracestate, communicate them over network, etc. than a hash of a 16-byte array (at least this is my intuition).

bogdandrutu (changes requested):
One big question that I don't get the answer from this OTEP is if this is a required feature. I feel that you need this to be required to work (especially when the trace travers multiple customers), but I also feel this is an extra overhead and complicated logic for cases when the customer does not care about this.

bogdandrutu:
@jmacd please fix the check-errors.

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,454 @@
...
+| N       | 2**(-N+1)        | 1 in 2**(N-1)                                        |
+| ...     | ...              |                                                      |
+| 61      | 2**-60           |                                                      |
+| 62      | 2**-61           |                                                      |
+| 63      | 0                | Maximum encoded value                                |
+
+[Described in OTEP
+170](https://github.com/open-telemetry/oteps/pull/170), the
+`ParentBased` sampler will use the incoming context's p-value as
+specified here to set the span's `log_head_adjusted_count` field.
```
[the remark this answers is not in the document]
jmacd: As you note, if the `tracestate` contains this information (and more) then why would we store it in a field? My hope is that this use of `tracestate` demonstrates value such that W3C would accept a version-1 `traceparent` that includes the same information for 1-3 extra bytes, depending on the specifics. If W3C would also specify that 62 specific bits of the TraceID are required to be uniform random, that alleviates the need for an `r` variable, which means only requiring the `p` variable _which is only required when sampled_ so the total cost w/ a version-1 `traceparent` is very low--that's why `log_head_adjusted_count` is a dedicated field.

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,454 @@
...
+but not at probabilities 1-in-16 and smaller.
+
+### Proposed `tracestate` syntax
+
+The consistent sampling r-value (`r`) and and head sampling
+probability p-value (`p`) will be propagated using two bytes of base16
+content for each of the two fields, as follows:
+
+```
+tracestate: otel=p:PP;r:RR
```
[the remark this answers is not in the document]
jmacd: Stated below "proposal should be taken as a recommendation". After https://github.com/open-telemetry/opentelemetry-specification/pull/1852 makes progress we can update our OTEPs if that will help.

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,454 @@
...
+```
+
+where `PP` are two bytes of base16 p-value and `RR` are two bytes of
+base16 r-value.  These values are omitted when they are unknown.
```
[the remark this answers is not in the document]
jmacd: Again, this is just speculative recommendation. Let's put this question in https://github.com/open-telemetry/opentelemetry-specification/pull/1852. 

The W3C uses two bytes for "flags" despite having only 1 bit in use, so I imagined following their precedent and using two base16 bytes. Ideally, the version-1 W3C `traceparent` TraceID has 62 well-specified uniform random bits and then the additional `p` value can be had for as little as 1 byte in most cases (i.e., unknown probability and 100% sampling through 2**-14 sampling would require only one byte).

text/trace/0168-sampling-propagation.md
jmacd: This level of detail about specific error handling, I think, can be debated in https://github.com/open-telemetry/opentelemetry-specification/pull/1899.

text/trace/0168-sampling-propagation.md
jmacd: @oertl has defended this point in the past with a link to https://github.com/rurban/smhasher, which shows how expensive and failure-prone a number of hashing functions are. My position is that the math of computing adjusted counts from inverse probabilities is likely to have artifacts as a result of low-quality hashing. The better long-term solution is a W3C `traceparent` with 62 specified uniform random bits.

jmacd:
> One big question that I don't get the answer from this OTEP is if this is a required feature.

Discussed in "Default behavior". I'd prefer this were on-by-default, but with the present `tracestate` solution it's expensive. Maybe after there's a W3C traceparent, where this costs 1-3 bytes per context, this can be on by default. I'll accept off-by-default for now-- Lightstep's OTel launchers would turn this on by default. I imagine "yet another" environment variable to say whether the `TraceIDRatioBased` sampler generates `r` or not.

text/trace/0168-sampling-propagation.md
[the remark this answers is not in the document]
bogdandrutu: Please make sure the comment is capture there.

text/trace/0168-sampling-propagation.md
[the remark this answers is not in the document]
bogdandrutu: We already have tracestate so I see no point in adding the new field, that is what I am trying to say. Unless you say that the tracestate for this entry should not be sent then I see no point in having the new field

text/trace/0168-sampling-propagation.md
bogdandrutu: Please make sure the comment is capture there.

bogdandrutu:
@jmacd where should we discuss/agree if this is or not the default behavior?

text/trace/0168-sampling-propagation.md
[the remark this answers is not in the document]
bogdandrutu: https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L89

carlosalberto:
@bogdandrutu Just FYI there's a SIG call on 8 minutes, although don't know who package are you today ;)

text/trace/0168-sampling-propagation.md
```diff
@@ -0,0 +1,454 @@
...
+170](https://github.com/open-telemetry/oteps/blob/main/text/trace/0170-sampling-probability.md) has discussed
+the use of a `sampler.adjusted_count` attribute that would not be
+limited to power-of-two values.  Discussion about how to represent the
+effective adjusted count for tail-sampled Spans belongs in [OTEP
+170](https://github.com/open-telemetry/oteps/blob/main/text/trace/0170-sampling-probability.md), not this OTEP.
+
+Restricting head sampling rates to powers of two does not limit
+Samplers from using arbitrary effective probabilities over a period of
+time.  For example, choosing 1/2 sampling half of the time and 1/4
+sampling half of the time leads to an effective sampling rate of 3/8.
```
yurishkuro, line 431: It may be better to pick a different example. p=3/8 looks neat, but what if I wanted p=7/16? I would need to pick p=1/2 vs. p=1/4 in proportion 3:1
```
1/2 * 0.75 + 1/4 * 0.25 = 7/16
```

text/trace/0168-sampling-propagation.md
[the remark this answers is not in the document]
jmacd: Fixed.

text/trace/0168-sampling-propagation.md
PeterF778: Perhaps an even better example would be to show that a more traditional 5% sampling rate can be obtained by sampling with 1/16 probability 60% of the time and with 1/32 probability 40% of the time.

jmacd: Applied.

jmacd: Done!

bogdandrutu (approved):
Unblocking, since @jmacd promised that the default behaviors will be discussed in specs.

PR #170, opened 2021-07-27
Probability sampling: Encode Span's head-adjusted count

jmacd:
This text is reproduced from #148, a review that became too long after several revisions to the document.

This proposes specification text for a `sampler.name` and `sampler.adjusted_count` attribute, used to convey sampling probability in recorded `SpanData` messages.

A companion OTEP #168 discusses how to propagate sampling probability when using the `Parent` sampler.


jmacd:
@oertl @yurishkuro @paulosman this document is unchanged from the most recent discussion in #148. I thought that opening a new PR would improve our chances of getting this specification merged.

This specification includes a lot of background to establish that there are many ways to sample and count spans, but all of them involve recording an adjusted count.

text/trace/0170-sampling-probability.md
```diff
@@ -0,0 +1,677 @@
...
+| AlwaysOff    | No    | Not applicable | Spans are not recorded              |
+| ParentBased  | Maybe | `Parent`       | Adjusted count requires propagation |
+| TraceIDRatio | Yes   | `TraceIDRatio` |                                     |
+```
+
+Note that the `AlwaysOn` and `AlwaysOff` Samplers do not need to
+record their names, since they are indistinguishable from not having
+a Sampler configured.  When there is no `sampler.name` attribute
+present and a Span is recorded, it should be counted as one span
+(i.e., count == adjusted_count).
```
yurishkuro, line 652: I find this unnecessarily complicated / confusing. Too many exceptions. I think a simpler approach is
  1. to have AlwaysOn sampler record count of 1, so that spans can be processed normally, without special-casing it,
  2. the absence of the adjusted_count attr to mean the weight is unknown, special processing is needed
  3. I would prefer `sampler.name=Parent` to be implied if not specified, since it's the most common behavior (but I realize it in itself is an exception)

yurishkuro (commented):
I am in favor of merging the narrative, but I do not full agree with the proposed spec changes.

text/trace/0170-sampling-probability.md
jmacd: Here is the general rule I am aiming for:

**The absence of `sampler.adjusted_count` implies either adjusted count is 1 or is unknown, depending on whether `sampler.name` is defined.**

If `sampler.name` is defined and `sampler.adjusted_count` is not, you must have an external way to find the adjusted count (if you care--this will become legacy behavior). OTel defines only one such external mechanism (i.e., how to get the adjusted count when `sampler.name=Parent`).

The rationale behind omitting "adjusted count" when it equals 1 is that it is an adjustment to the natural count of one individual Span. There is no adjustment when the count is 1, so no reason to record an adjusted count. This implies that we can omit the `sampler.name` attribute when the count is known to equal 1 because it is equivalent to no sampling. 

> since it's the most common behavior

The default Sampler is `ParentBased(root=AlwaysOn)`. Given the specification I am proposing, the default SpanData will not include `sampler.name` or `sampler.adjusted_count` attributes, because all spans have adjusted count of 1 in a `ParentBased(root=AlwaysOn)` configuration.

text/trace/0170-sampling-probability.md
```diff
@@ -0,0 +1,677 @@
...
+| Attribute | Type | Description | Examples | Required |
+|---------- | ---- | ----------- | -------- | -------- |
+| `sampler.adjusted_count` | number | Effective count of the span. | 10       | No  |
+| `sampler.name`           | string | The name of the Sampler.     | `Parent` | Yes |
+
+For the built-in samplers, the following names are specified:
+
+| Built-in Sampler | Sets `sampler.adjusted_count`? | `sampler.name` | Notes  |
+| ---------------- | ------------------------------ | -------------- | ------ |
+| AlwaysOn     | No    | Not applicable | Sampling attributes are not used    |
```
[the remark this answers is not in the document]
jmacd: I think this is answered in https://github.com/open-telemetry/oteps/pull/170#discussion_r678474697. There is no need to record an adjustment when the count is 1.

text/trace/0170-sampling-probability.md
```diff
@@ -0,0 +1,677 @@
...
+# Semantic conventions for Sampled spans
+
+This document defines how to describe an what sampling was performed
+when recording a span that has had sampling logic applied.
+
+Span sampling attributes support computing metrics about spans that
+are part of a sampled trace from knowing their sampling inclusion
+probability.
+
+The _adjusted count_ of a span is defined as follows:
```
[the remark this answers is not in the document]
jmacd: I think this is addressed in https://github.com/open-telemetry/oteps/pull/170#discussion_r678474697. The natural count of the Span is 1, which should be used when there is no adjustment.

text/trace/0170-sampling-probability.md
```diff
@@ -0,0 +1,677 @@
...
+
+- Adjusted count equals zero when inclusion probability equals zero
```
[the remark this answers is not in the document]
jmacd: The 0 value for adjusted count is meaningful and allows recording non-probabilistically selected Spans while performing probability sampling. In an earlier draft of #148 I had proposed a convention for supporting multiple independent probability samples, in which case -- when more than one Sampler may cause a Span to be recorded -- a zero value for adjusted count should be used to signify a value that is recorded but not selected in a particular sample.

I removed the detail about supporting multiple samples because OTel doesn't need to specify how to do more than one mechanism. The mechanism I believe OTel should require is the system-wide sample, which is the sample of all traces you get by default when turning on an OTel SDK. Thus, I am proposing a single set of `sampler.name` and `sampler.adjusted_count` attributes. Vendors and systems that want to use more than one sampler may end up recording spans with `sampler.adjusted_count=0` if they are selected by alternate samplers but not included in the system-wide sample.

text/trace/0170-sampling-probability.md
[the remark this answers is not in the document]
yurishkuro: I think "natural count" is somewhat subjective. For sites that use traditional probabilistic head-based sampling with propagation of sampling decision, the "natural count" for spans would be 1/p. These sites would also be more sensitive (due to cost) to including additional span attributes than the sites that can "afford" 100% sampling (e.g. Jaeger SDKs only record sampling params on the root span). I am weary of making a choice in the spec that would penalize one group at the expense of the other. We may be forced to make such a choice (unless we provide for alternative means of capturing the info), in which case it would make sense to poll many users who would be affected by it.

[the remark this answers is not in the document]
jmacd: > I think "natural count" is somewhat subjective.

You are calling an adjusted count natural. I feel that is incorrect. There is only one event corresponding with each Span, thus the natural event count is 1.

Your position, if I understand it, is that the cost of reporting adjusted count when it is not 1, which supports span-to-metrics pipelines, is not worth the expense of having always on. I am open to other's input on this.

I would like to question the argument that a system that adopts probability sampling because 100% of spans is too expensive will object to the incremental cost of reporting adjusted count.  A slight reduction of the sampling rate will compensate for the additional cost of supporting span-to-metrics. I am interested in knowing more about how existing systems compute metrics about spans if they are forced to assemble traces before generating those metrics. Do any such span-to-metrics pipelines exist for Jaeger systems?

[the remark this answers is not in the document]
yurishkuro: > I am interested in knowing more about how existing systems compute metrics about spans if they are forced to assemble traces before generating those metrics.

What's the problem with assembling traces first? It introduces an ingestion delay, but then the actual insights one gets from these "metrics" are coupled with availability of traces anyway.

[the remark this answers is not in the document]
jmacd: This is about more than ingestion delay. We have to add the root span's duration to the ingestion delay before metrics can be derived for the entire trace. If the root span never completes or is dropped, an entire trace worth of spans can't be counted.

You could also ask: why not just use metrics in the first place? Answer 1: Spans contain logs, we get information about the population of log events from a these span metrics. Answer 2: This data stream supports high-cardinality metric pipelines.

This was covered in the narrative above, especially lines 436-440 and 463-469. "Spans are countable" is a requirement I am standing behind, because real-time span-to-metrics requires it.

text/trace/0170-sampling-probability.md
jmacd: I have revised the text in order to present the rationale in a better way.

text/trace/0170-sampling-probability.md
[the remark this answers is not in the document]
jmacd: I have added a sentence to explain why zero adjusted count is useful and a sentence to explain the connection between expected-value and probability.

paulosman:
> @oertl @yurishkuro @paulosman this document is unchanged from the most recent discussion in #148. I thought that opening a new PR would improve our chances of getting this specification merged.
> 
> This specification includes a lot of background to establish that there are many ways to sample and count spans, but all of them involve recording an adjusted count.

Thank you. I can at least provide a vendor perspective, as a consumer of sampled telemetry events. 

Adding an adjusted count solves a real problem for us: At the moment we depend on the return value from [ShouldSample](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#shouldsample) including attributes that are added to the span. We have custom plugins that add a sample rate to this list of attributes, which is then read at ingest and used to make estimates about the population. If no adjusted count is included, we assume a value of 1. 

This works but requires customers to use custom sampling plugins that provide no real value above the ability to communicate the adjusted count (the plugins use a deterministic algorithm that is similar to the `TraceIdRatioBased` sampler). Having the adjusted count be part of the spec would allow us to get rid of the need for these plugins.

For tail sampling, we accept OTLP data to our sampling proxy and then convert the data to our HTTP API which includes an adjusted count. We'd prefer to be able to do end-to-end OTLP, which the spec change described in this OTEP would allow us to do.

So :+1: from me. Thanks again for writing this up.

paulosman (approved):

text/trace/0170-sampling-probability.md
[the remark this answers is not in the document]
jmacd: I'm resolving this comment because I believe I have address the question. 

> insights one gets from these "metrics" are coupled with availability of traces anyway.

This could be a source of confusion. The goal when performing span-to-metrics is to learn about the population of spans in their own right. At Lightstep, when we see an uninteresting span we might just count it an move on. 

I've completed a prototype of this code and OTEP #168 by now, and will update #168 with more detail about propagation. @yurishkuro please let me know if there is something more I can explain or address about the span data model conventions in this PR, and I will update #168 tomorrow. Thanks.

jmacd:
@open-telemetry/specs-trace-approvers Please take a look. 

I've completed a prototype of this proposal in https://github.com/open-telemetry/opentelemetry-go/compare/main...jmacd:jmacd/propagate?expand=1.  The text in this PR matches the prototype.

text/trace/0170-sampling-probability.md
```diff
@@ -0,0 +1,717 @@
...
+#### Probability Sampling Attributes (Proposed text)
+
+The `sampler.adjusted_count` attribute, when set, MUST equal an
+unbiased estimate of the number of representative spans in the
+population of spans in the system.
+
+The _exported count_ associated with a span is defined as either 1 or
+0, depending on whether the span is exported and thus counted.  The
+exported count is 1 if the span is exported (because it will be
+counted) and 0 if the span is not exported.
```
yurishkuro, line 644: I suggest removing this notion of "exported count", as it is confusing and not particularly useful - it's only used in the table where depending on the context it's always 0 or 1.

jmacd: Done. Thanks.

jmacd:
There is an error condition implied by #168 that would occur when the random variable `R` is not received in the trace state, [here](https://github.com/open-telemetry/opentelemetry-go/pull/2177/files#diff-3f1cead504691608d45bc4e9a09c8485f2b8da14e6b619391b42c737b52a8d75R557).

This illustrates the benefits of deriving `R` from the bits of the TraceID. If we depend on `tracestate` to make a consistent decision and do not receive that field, the `TraceIDRatio` Sampler will not know how to make its sampling decision. We could fall back to whatever behavior was being implemented by each SDK before this specification, which might be labeled `ProbabilitySampler`. This would lead to a situation where non-root spans are sampled with known adjusted count, but without having been selected using a consistent decision. 

As specified, @oertl, the `sampler.adjusted_count` attribute can be correctly unbiased without requiring the sampling decision to be made in a consistent way. If an arbitrary Sampler records adjusted counts in this way, how much do you care to know that another sampler was used? None of the built-in samplers would do this, except for a hypothetical TraceIDRatio Sampler when the random variable `R` is not received in the trace state. We can avoid any built-in Samplers doing this by adding requirements to the TraceID, probably via W3C `traceparent`.

carlosalberto (approved):

yurishkuro:
Suggestion: we may want to consider adding top-level Span fields for count and sampler name, given how important these concepts are and the space savings we'd achieve by recording values only.

jmacd:
@yurishkuro thanks for the suggestion -- it raises a few questions. The value represented by a `sampler.adjusted_count` attribute could take either integer or floating-point values and the zero value is recognizable as distinct from an unset value. One way to preserve all of these features in the OTLP Span message is to represent the field as a `opentelemetry.proto.common.v1.KeyValue`. The protocol field could have the following text, for example.

```
  // Sampler adjustment is recorded by the Sampler that was in effect to produce this
  // span.  The following values are recognized:
  //
  //   If `sampler_adjustment.key` is non-empty, it is the sampler Name.  If it is
  //   one of the OpenTelemetry built-in Samplers, the following string is recognized:
  //   - "parent": the ParentBased sampler was used.  If this is set and value is missing,
  //     it means the adjusted counted is unknown.
  //   The `sampler_adjustment.key` value MAY be omitted when the adjusted count is known.
  //
  //   If `sampler_adjustment.value` is set to a non-negative numeric value, it is
  //   taken as the adjusted count of the Span being recorded.  In this case, the Span
  //   SHOULD have been selected using an unbiased technique, in which case the adjusted
  //   count conveys the effective representative count of the Span after sampling.
  //   - (int64) the Span has an integer adjusted count
  //   - (double) the Span has an floating-point adjusted count.
  //   - other Value types are unrecognized.
  //   Negative values are considered invalid.
  opentelemetry.proto.common.v1.KeyValue sampler_adjustment = 17;
```

How would you feel about this? 

@oertl has made the suggestion that _if we restrict adjusted counts to powers-of-two_ then we could transmit just a small integer to convey sampling. I worry this limits our ability to tail sample on the collection path in arbitrary ways, but this option could be combined with another alternative, such as to respect `sampler.adjusted_count` when present in order to support floating-point values. That would be a field something like:

```
  // Log adjusted count is the logarithm of adjusted count for this
  // span, with the following recognized values.
  //
  // 0: The zero value represents an adjusted of 1.  The zero value is
  //    equivalent to an unset field and SHOULD be omitted.
  //
  // 1..62: Values 1 through 62 represent an adjusted count of 2^(value)
  //    are equivalent to setting `sampler.adjusted_count`.
  //
  // 63: Value 63 (0x3f) represents an adjusted count of zero, equivalent
  //    to setting `sampler.adjusted_count=0`.
  //
  // 64: Value 64 (0x100) represents an unknown adjusted count, equivalent
  //    to setting any non-empty `sampler.name` to indicate an unknown sampler 
  //    (e.g., ParentBased).
  uint32 log_adjusted_count = 17;
```

I don't know of a tail-sampling algorithm that can be rate-limited and respect power-of-two sampling rates, although @oertl hinted at one in Thursday's Sampling SIG meeting.

yurishkuro:
> opentelemetry.proto.common.v1.KeyValue.

I would probably just define a new type with proper field names instead of overloading/reusing KeyValue. And since this is for out of band data collection, I am not sure power-of-2 values provide enough value.

oertl:
@jmacd, @yurishkuro I think tail-based sampling should be considered completely independent of (consistent) span-level sampling. I propose to keep the adjusted counts due to span-level sampling and due to tail-based sampling completely apart. The span sampling rate could be limited to powers of two, while tail-based sampling could use any sampling rate. I guess it is much easier for vendors to control tail-based sampling anyway as this happens at the collector compared to span-level sampling??? If this assumption is true, I think it makes sense to allow vendors to choose any tail-based sampling rate they want.

jmacd:
@yurishkuro @oertl Taking both of these pieces of feedback together yields something like:

```
  message SamplerDetail {
    // Log head adjusted count is the logarithm of adjusted count for this
    // span as calculated at the head, with the following recognized values.
    //
    // 0: The zero value represents an adjusted of 1.  The zero value is
    //    equivalent to an unset field and SHOULD be omitted.
    //
    // 1..62: Values 1 through 62 represent an adjusted count of 2^(value)
    //
    // 63: Value 63 (0x3f) represents an adjusted count of zero.
    //
    // 64: Value 64 (0x40) represents an unknown adjusted count.
    //
    // Values greater than 64 or less than zero are unrecognized.
    uint32 log_head_adjusted_count = 1;

    // Final adjusted count is the effective adjusted count of this
    // Span after tail sampling.  When this field is set to a non-zero
    // value, it SHOULD be taken as an accurate means of counting the
    // span as opposed to the inferring its count from the
    // log_head_adjusted_count field.
    float final_adjusted_count = 2;

    // Final sampler name optionally describes the most recent Sampler
    // that was used in selecting a Span.  This field SHOULD be set
    // when the adjusted count implied by log_head_adjusted_count will
    // result in an accurate count of spans.  When the final sampler
    // name is set, it means the log_head_adjusted_count field SHOULD
    // NOT be considered accurate for counting the span population
    // because further sampling has been applied.
    string final_sampler_name = 3;
  }

  SamplerDetail sampler_detail;
```

The only capability discussed above that is lost in this encoding is the ability for a tail sampler to express a zero adjusted count from a tail sampler as distinct from an unknown adjusted count. Since we're talking about custom logic in a tail sampler anyway, I'm comfortable not specifying any difference between unknown and zero.

The way this is written, the built-in Samplers will all result in Spans with `log_head_adjusted_count`. Tail samplers can be written to output arbitrary or unknown adjusted counts according to the specification already drafted here, meaning they should set their name and not their adjusted count of they are not probabilistic in nature.

jmacd:
I will update this PR based on the discussion in Thursday's Sampling SIG meeting, where we concluded that Tail sampling is interfering with our objectives and can be dropped at this time. Any kind of tail sampling that wants to be involved in span-to-metrics will require further specification, probably can be based off the proposals that were developed in this OTEP.

jmacd:
I revised this OTEP based on the Thursday Sampling SIG discussion, in which it was proposed that we introduce just a single integer field to encode head sampling probability, ignoring (for the time being) the topic of tail sampling.  The proposed field would take 65 possible values, as follows:

| Value | Head Adjusted Count |
| ----- | ---------------- |
| 0 | _Unknown_ |
| 1 | 1 |
| 2 | 2 |
| 3 | 4 |
| 4 | 8 |
| 5 | 16 |
| 6 | 32 |
| ... | ... |
| X | 2^(X-1) |
| ... | ... |
| 63 | 2^62 |
| 64 | 0 |

and the `SpanData` would be updated with a new field in `v1/trace.proto`.

```
  // Log-head-adjusted count is the logarithm of adjusted count for
  // this span as calculated at the head, offset by +1, with the
  // following recognized values.
  //
  // 0: The zero value represents an UNKNOWN adjusted count.
  //    Consumers of these Spans cannot cannot compute span metrics.
  //
  // 1: An adjusted count of 1.
  // 
  // 2-63: Values 2 through 63 represent an adjusted count of 2^(Value-1)
  //
  // 64: Value 64 represents an adjusted count of zero.
  //
  // Values greater than 64 are unrecognized.
  uint32 log_head_adjusted_count = <next_tag>;
```

oertl:
@jmacd, maybe it is better to sacrify the sampling rate 1/2^62 (which will never be used in practice I guess) in order to fit the state into 6 bits?

jmacd:
> @jmacd, maybe it is better to sacrifice the sampling rate 1/2^62 (which will never be used in practice I guess) in order to fit the state into 6 bits?

I agree, and have applied this change in both OTEPs.

tedsuo (approved):
This OTEP looks GTM, any small changes could be worked out during the spec process.

yurishkuro (approved):

text/trace/0170-sampling-probability.md
```diff
@@ -0,0 +1,705 @@
...
+unset.  Thus, the 0 value shall mean unknown adjusted count.
+
+The OTEP 168 proposal for _propagating_ head sampling probability uses
+6 bits of information, with 62 ordinary values, one zero value, and a
+single unused value.
+
+Here, we propose a biased encoding for head sampling probability equal
+to 1 plus the `P` value as proposed in OTEP 168.  The proposed span
+field, a biased base-2 logarithm of the adjusted count, is named
+simply `log_head_adjusted_count` and still requires 6 bits of
```
oertl, line 610: I am wondering, if `log_head_adjusted_count` should be rather named `log_parent_adjusted_count`. Consider, a trace A -> B -> C where A and B are consistently sampled using different sampling rates and C uses a parent sampler that samples the span if the parent B is sampled. I guess we want to know the sampling rate of B for span C. Therefore, the parent sampling rate needs to be propagated and not that of the head.

yurishkuro: I think the intention is to describe *current* span and abstract away the origin of adjusted count calculation.

jmacd: We discussed this and agreed to take this question to spec PR #1899. This OTEP can merge either way.
Terms that we consider alternatives to "head":
"parent" (as in this comment)
"initial" (another option)

text/trace/0170-sampling-probability.md
```diff
@@ -0,0 +1,705 @@
...
+with specific interpretation for conveying head sampling probability.
+
+### Span sampling
+
+Example use-cases for probability sampling of spans generally involve
+generating metrics from spans.
+
+#### Sample spans to Counter Metric
+
+For every complete span it receives, the example processor will synthesize
```
jsuereth, line 90: I think there was a logical leap here for me.  Are we discussing a `SpanProcessor`, or just a conceptual processor of spans that generates metric data?

text/trace/0170-sampling-probability.md
```diff
@@ -0,0 +1,705 @@
...
+adjusted count of 10.  Assuming the sample was selected using an
+unbiased algorithm, we can interpret this event as having an expected
+count of `100/0.1 = 1000`.
+
+#### Metric cardinality limiter
+
+A metrics processor can be configured to limit cardinality for a
+single metric name, allowing no more than K distinct label sets per
+export interval.  The export interval is fixed to a short interval so
+that a complete set of distinct labels can be stored temporarily.
```
jsuereth, line 164: I think this one may need a full example (or visual).

jsuereth (approved):

reyang (approved):

text/trace/0170-sampling-probability.md
jmacd: How about simply `log_adjusted_count`?

jmacd:
The only remaining question here, possibly, is whether the Span field name should be `log_head_adjusted_count` or something similar. I propose that we merge this OTEP. 😀

PR #172, opened 2021-08-26
Instrumentation layers and suppressing duplicates

lmolkova:
Proposal to create clarity on instrumentation layers interaction and how to suppress duplicate layers (e.g. multiple instrumented  HTTP clients layers).

cc @trask

### DIFF
--- a/text/trace/0189-instrumentation-layers-suppression.md
+++ b/text/trace/0189-instrumentation-layers-suppression.md
@@ -18,7 +18,33 @@ This document describes approach for instrumentation layers, suppressing duplica
 - Semantic Conventions: Client libraries instrumentation MUST make context current to enable correlation with underlying layers of instrumentation
 - OTel SDK SHOULD allow suppression strategy configuration:
   - suppress nested by kind (e.g. only one CLIENT allowed)
-  - suppress nested by kind + type (only one HTTP CLIENT allowed, but outer DB -> nested HTTP is ok)
+  - suppress nested by kind + convention it follows (only one HTTP CLIENT allowed, but outer DB -> nested HTTP is ok)
+
+#### SpanKey API
+
+SpanKey allows to read/write span to context.
+Defines known SpanKey, shared between instrumentations (static singletons):  HTTP, RPC, DB, Messaging.
+
+#### Example
+
+- HTTP Client 1:
+  - Gets HTTP CLIENT span from context: `SpanKey.HTTP_CLIENT.fromContextOrNull(ctx)`
+  - No HTTP client span on the context:
+    - start span
+    - store span in context: `SpanKey.HTTP_CLIENT.storeInContext(ctx, span)`
+    - Make `ctx` current
+- Http Client 2:
+  - Gets HTTP CLIENT span from context: `SpanKey.HTTP_CLIENT.fromContextOrNull(Context.current())`
+  - HTTP client span is already there: do not instrument
+
+Suppression logic is configurable and encapsulated in `SpanKey` - instrumentation should not depend on configuration, e.g.:
+
+- suppressing by kind only - context key does not distinguish conventions within kind
+  - `SpanKey.HTTP_CLIENT.fromContextOrNull` returns CLIENT span
+  - `SpanKey.HTTP_CLIENT.storeInContext` stores span in CLIENT span context key
+- suppressing by kind + convention - context key is per convention and kind
+  - `SpanKey.HTTP_CLIENT.fromContextOrNull` returns HTTP CLIENT span
+  - `SpanKey.HTTP_CLIENT.storeInContext` stores span in CLIENT + convention span context key
 
 ## Internal details
 
--- a/text/trace/0189-instrumentation-layers-suppression.md
+++ b/text/trace/0189-instrumentation-layers-suppression.md
@@ -1,0 +1,127 @@
+# Instrumentation Layers and Suppression
+
+This document describes approach for instrumentation layers, suppressing duplicate layers and unambiguously enriching spans.
+
+## Motivation
+
+- Provide clarity for instrumentation layers: e.g. DB calls on top of REST API
+- Give mechanism to suppress instrumentation layers for the same type: e.g. multiple instrumented HTTP clients using each other.
+- Give mechanism to enrich specific spans unambiguously: e.g. HTTP server span with routing information
+
+## Explanation
+
+### Spec changes proposal
+
+- Semantic Conventions: Each span MUST follow one convention, specific to the call it describes
+- Trace API: Add `SpanKey` API that gets span following specific convention from the context (e.g. `SpanKey.HTTP_CLIENT.fromContextOrNull(context)`).
+- Semantic Conventions: instrumentation MUST back off if span of same kind and following same contention is already in the context by using `ContextKey` API.
+- Semantic Conventions: Client libraries instrumentation MUST make context current to enable correlation with underlying layers of instrumentation
+- OTel SDK SHOULD allow suppression strategy configuration:
+  - suppress nested by kind (e.g. only one CLIENT allowed)
+  - suppress nested by kind + type (only one HTTP CLIENT allowed, but outer DB -> nested HTTP is ok)
+
+## Internal details
+
+Client libraries frequently use common protocols (HTTP, gRPC, DB drivers) to perform RPC calls, which are usually instrumented by OpenTelemetry.
+At the same time, client library is rarely a thin client and may need its own instrumentation to
+
+- connect traces to application code
+- provide extra context:
+  - duration of composite operations
+  - overall result of all operation
+  - any extra library-specific information not available on transport call span
+
+Both, client library 'logical' and transport 'physical' spans are useful. They also rarely can be combined together because they have 1:many relationship.
+
+So instrumentations form *layers*, where each layer follows specific convention (or describes certain library).
+
+*Example*:
+
+- HTTP SERVER span
+  - DB CLIENT call - 1
+    - HTTP CLIENT call - 1
+      - DNS CLIENT
+      - TLS CLIENT
+    - HTTP CLIENT call - 2
+
+There are two HTTP client spans under DB call, they are children of DB client spans.  DB spans follow DB semantics only, HTTP spans similarly only follow HTTP semantics. If there are other layers of instrumentation (TLS) - it happens under HTTP client spans.
+
+### Duplication problem
+
+Duplication is a common issue in auto-instrumentation:
+
+- e.g. HTTP clients frequently are built on top of other HTTP clients, making multiple layers of HTTP spans
+- Libraries may decide to add native instrumentation for common protocols like HTTP or gRPC:
+  - to support legacy correlation protocols
+  - to make better decisions failures (e.g. 404, 409)
+  - give better library-specific context
+  - support users that can't or don't want to use auto-instrumentation
+
+So what happens in reality without attempts to suppress duplicates:
+
+- HTTP SERVER span (middleware)
+  - HTTP SERVER span (servlet)
+    - Controller INTERNAL span
+      - HTTP CLIENT call - 1 (Google HTTP client)
+        - HTTP CLIENT call - 1 (Apache HTTP client)
+
+#### Proposed solution
+
+Disallow multiple layers of the same instrumentation, i.e. above picture translates into:
+
+- HTTP SERVER span (middleware)
+  - Controller INTERNAL span
+    - HTTP CLIENT call - 1 (Google HTTP client)
+
+To do so, instrumentation:
+
+- checks if span with same kind + type is registered on context already
+  - yes: backs off, never starting a span
+  - no: starts a span and registers it on the context
+
+Registration is done by writing a span on the context under the key. For this to work between different instrumentations (native and auto), the API to access spans must be in Trace API.
+
+Same mechanism can be used by users/instrumentations to enrich spans, e.g. add route info to HTTP server span (current span is ambiguous)
+
+### Configuration
+
+Suppression strategy should be configurable:
+
+- backends don't always support nested CLIENT spans (extra hints needed for Application Map to show outbound connection)
+- users may prefer to reduce verbosity and costs by suppressing spans of same kind
+
+So two strategies should be supported:
+
+- suppress all nested of same kind
+- suppress all nested of same kind + type (default?)
+
+### Implementation
+
+https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanKey.java
+
+## Trade-offs and mitigations
+
+Trace API change is needed to support native library instrumentations - taking dependency on unstable experimental instrumentation API (or common contrib code) is not a good option. Instrumentation API is a good temporary place until we can put it in Trace API, native instrumentation can use reflection to access `SpanKey` in instrumentation API.
+
+## Prior art and alternatives
+
+- Terminal context: suppressing anything below
+- Exposing spans stack and allowing to walk it accessing span properties
+- Suppress all nested spans of same kind
+- Make logical calls INTERNAL
+
+Discussions:
+
+- https://github.com/open-telemetry/opentelemetry-specification/issues/1767
+- https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/903
+- https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/465
+- https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1822
+- https://github.com/open-telemetry/opentelemetry-specification/issues/526
+- https://github.com/open-telemetry/opentelemetry-python-contrib/issues/369
+- https://github.com/open-telemetry/opentelemetry-python-contrib/issues/445
+- https://github.com/open-telemetry/opentelemetry-python-contrib/issues/456
+
+## Open questions
+
+- Backends need hint to separate logical CLIENT spans from physical ones
+- Good default (suppress by kind or kind + type)
--- a/text/trace/0189-instrumentation-layers-suppression.md
+++ b/text/trace/0189-instrumentation-layers-suppression.md
@@ -172,7 +172,7 @@ So following strategies should be supported:
 
 #### Trace API
 
-Proof of concept in Java: https://github.com/lmolkova/opentelemetry-java/pull/1
+[Proof of concept in Java](https://github.com/lmolkova/opentelemetry-java/pull/1)
 
 - introduces a new `SuppressedSpan` implementation. It's almost the same as `PropagatingSpan` (i.e. sampled-out), with the difference that `makeCurrent` is noop
   - `PropagatingSpan` can't be reused since sampling out usually assumes to sample out this and all following downstream spans. Sampled out span becomes current and causes side-effects (span.current().setAttribute()) not compatible with suppression.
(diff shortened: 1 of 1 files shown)


text/trace/0189-instrumentation-layers-suppression.md
```diff
@@ -0,0 +1,127 @@
...
+- Give mechanism to suppress instrumentation layers for the same type: e.g. multiple instrumented HTTP clients using each other.
+- Give mechanism to enrich specific spans unambiguously: e.g. HTTP server span with routing information
+
+## Explanation
+
+### Spec changes proposal
+
+- Semantic Conventions: Each span MUST follow one convention, specific to the call it describes
+- Trace API: Add `SpanKey` API that gets span following specific convention from the context (e.g. `SpanKey.HTTP_CLIENT.fromContextOrNull(context)`).
+- Semantic Conventions: instrumentation MUST back off if span of same kind and following same contention is already in the context by using `ContextKey` API.
```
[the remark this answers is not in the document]
iNikem: I honestly don't know. We do have clients who want to have as much information as possible.

Do you think that having this "no-suppression" option will complicate the design considerably? If not, I think we should not take the option away from the clients. We cannot predict the future :)

If this is actually complicated, then maybe having lower non-http layer is actually enough...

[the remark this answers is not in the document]
lmolkova: Yes, there's no additional complexity with adding no-suppression option. I think it can be useful for debugging instrumentation issues. I'll add it to the doc later this week and hope we'll get a chance to discuss it on one of the instrumentation sigs

text/trace/0189-instrumentation-layers-suppression.md
```diff
@@ -0,0 +1,154 @@
...
+
+- Provide clarity for instrumentation layers: e.g. DB calls on top of REST API
+- Give mechanism to suppress instrumentation layers for the same convention: e.g. multiple instrumented HTTP clients using each other.
+- Give mechanism to enrich specific spans unambiguously: e.g. HTTP server span with routing information
+
+## Explanation
+
+### Spec changes proposal
+
+- Semantic Conventions: Each span MUST follow at most one convention, specific to the call it describes.
```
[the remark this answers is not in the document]
Oberon00: Some semantic conventions explicitly request others: E.g. FaaS with HTTP trigger requests HTTP. General semantic conventions (e.g. `net.*`, `code.*`) expect to be used with others.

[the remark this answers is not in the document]
pyohannes: I think one has to be a bit more discerning here.

Firstly, one needs to distinguish between semantic conventions for resources and semantic conventions for tracing. As far as I understand, in this context we're only talking about semantic conventions for tracing.

Regarding semantic conventions for tracing, I think some are composable, whereas others are not. I get the impression that most of the [general attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md) are meant to be used in a composable way. However, I also think you're right that it makes things hard for backends and visualization tools if there are spans with both HTTP and DB attributes.

That said, I think currently semantic conventions are much too murky to ask for exclusively one convention per span. Probably the level of composability needs to be defined for each area of semantic conventions, and cannot be formulated generally.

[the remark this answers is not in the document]
lmolkova: > Some semantic conventions explicitly request others: E.g. FaaS with HTTP trigger requests HTTP. General semantic conventions (e.g. net.*, code.*) expect to be used with others.

Agreed. Assuming convention explicitly includes another one - this is just fine.

>  in this context we're only talking about semantic conventions for tracing.
yes

> currently semantic conventions are much too murky to ask for exclusively one convention per span.
I agree that currently, we don't have enough clarity. I suggest we start to improve that and if we agree this is the right direction, start aligning conventions with it.

text/trace/0189-instrumentation-layers-suppression.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Explanation
+
+### Spec changes proposal
+
+- Tracing Semantic Conventions: Each span MUST follow a single (besides general) convention, specific to the call it describes.
+- Trace API: Add `SpanKey` API that
+  - checks if similar span already exists on the context (e.g. `SpanKey.HTTP_CLIENT.exists(context)`)
+  - gets span following specific convention from the context (e.g. `SpanKey.HTTP_CLIENT.fromContextOrNull(context)`).
+- Tracing Semantic Conventions: instrumentation MUST back off if span of same kind and following same contention is already exists on the context by using `SpanKey` API.
```
pyohannes, line 19:
```suggestion
- Tracing Semantic Conventions: instrumentation MUST back off if span of same kind and following same convention is already exists on the context by using `SpanKey` API.

```

text/trace/0189-instrumentation-layers-suppression.md
```diff
@@ -0,0 +1,180 @@
+# Instrumentation Layers and Suppression
+
+This document describes approach for tracing instrumentation layers, suppressing duplicate layers and unambiguously enriching spans.
```
pyohannes, line 3: I think it's worth mentioning that the same problem will exists for metrics, even more aggravated (just thinking about a metric "Request Duration" in nested HTTP client spans).

The solution proposed in this document could also be used to suppress those "duplicate" metrics.

text/trace/0189-instrumentation-layers-suppression.md
```diff
@@ -0,0 +1,180 @@
...
+- Provide clarity for instrumentation layers: e.g. DB calls on top of REST API
+- Give mechanism to suppress instrumentation layers for the same convention: e.g. multiple instrumented HTTP clients using each other.
+- Give mechanism to enrich specific spans unambiguously: e.g. HTTP server span with routing information
+
+## Explanation
+
+### Spec changes proposal
+
+- Tracing Semantic Conventions: Each span MUST follow a single (besides general) convention, specific to the call it describes.
+- Trace API: Add `SpanKey` API that
```
pyohannes, line 16: As discussed today: alternatively consider moving this logic into the SDK. This would involve fewer API changes and probably result in a more robust solution (otherwise we'd need to rely on every library instrumentor properly calling `SpanKey.*.exists`).

lmolkova:
Based on the discussions on the instrumentation SIG meeting, I've changed this OTEP to minimize tracing API changes. With the current approach, instrumentation specifies `type` at span creation time and SDK returns a *suppressed* span that's very similar to sampled-out, except it does never become current.

Example: https://github.com/lmolkova/opentelemetry-java/pull/1/files#diff-48014f0437bde243933a6c8902a1f386fab6c78d6777019f24db1951e0070fc4

Open questions:
- messaging spans don't work great with it and I suggest working on it in messaging spec (there are more reasons than clear layering and suppression)
- adding optional optimization API to let instrumentation check first if span should be started
- defaults
- naming improvements are always welcome

text/trace/0189-instrumentation-layers-suppression.md
lmolkova: I suggest keeping it separate for now  - we need to have some clarity around metrics instrumentation - if it's the responsibility of every library or a central component before we can tackle metrics suppression.

pyohannes: Agreed. I just think we shouldn't have two different suppression mechanisms for metrics and traces, if possible.

text/trace/0189-instrumentation-layers-suppression.md
```diff
@@ -0,0 +1,222 @@
+# Instrumentation Layers and Suppression
+
+This document describes approach for tracing instrumentation layers and suppressing duplicate layers.
+
+## Motivation
+
+- Provide clarity for instrumentation layers: e.g. DB calls on top of REST API
+- Give mechanism to suppress instrumentation layers for the same convention: e.g. multiple instrumented HTTP clients using each other.
```
jmacd, line 8: Why are there cases of multiple instrumentation HTTP clients using each other? That sounds like an odd configuration of clients. I wonder if when combining these clients, one can disable instrumentation of the other?

The DB call on top fo a REST call -- you propose to suppress the REST instrumentation, I take it? (Could the DB be configured with an un-instrumented REST client, instead?) 

I wonder if it would be possible to replace `InstrumentationType` with `InstrumentationLibrary.Name`. Could you accomplish the same functional behavior, if for all spans with a non-remote parent you knew the InstrumentationLibrary of the nearest not-suppressed ancestor? It might make the proposal both simpler and more general: InstrumentationLibrary is already defined; it means you don't need an expandable enum.

lmolkova: > Why are there cases of multiple instrumentation HTTP clients using each other? 

1. One case is manual/native + auto: https://github.com/open-telemetry/opentelemetry-specification/issues/1767
2. Also, HTTP clients test to use each other (python, js, java all have one or another way to suppress HTTP instrumentation)
- in some cases, they are pluggable, i.e. higher level don't really know if the lower level is instrumented - this is more popular in SERVER spans
- in some cases,  I guess higher-level clients instrumentation is not necessary

> The DB call on top fo a REST call -- you propose to suppress the REST instrumentation, I take it? (Could the DB be configured with an un-instrumented REST client, instead?)

I suggest that 
1. DB and rest are two layers, i.e. they are both spans  (and not DB attributes stamped on HTTP span)
2. Depending on the configuration we can either
    - suppress nested HTTP (to reduce noise if users/backends don't want it)
    - not suppress HTTP and have full picture


> InstrumentationType with InstrumentationLibrary.Name

I'd love that, however `InstrumentationLibrary.Name` could be `GoogleHttpClient` or `ApacheHttpClient` while they both implement HTTP semantic conventions. The alternative could be to set `InstrumentationType` on the tracer resource in addition to `InstrumentationLibrary.Name` and version. I'd also put the semantic convention version it follows there.

lmolkova: > InstrumentationType with InstrumentationLibrary.Name

Experimented with this idea: added another 'type' parameter to library instrumentation info when tracer is being built:

- it makes tracers type-aware, i.e. tracers should not emit spans of different types. Which makes sense, but a new thing. 
-  it's even more subtle than the span creation option.  App and lib devs would have a hard time paying attention and will end up with multi-type tracers (i.e. no suppression).

I updated the PR https://github.com/lmolkova/opentelemetry-java/pull/1.

Will bring it up at the Tue Instrumentation SIG meeting.

I also entertained the idea of having a single tracer per type and returning noop tracers if one with the same type is registered - but it depends on the tracer creation time and is not stable.

text/trace/0189-instrumentation-layers-suppression.md
```diff
@@ -0,0 +1,222 @@
...
+- introduces a new `SuppressedSpan` implementation. It's almost the same as `PropagatingSpan` (i.e. sampled-out), with the difference that `makeCurrent` is noop
+  - `PropagatingSpan` can't be reused since sampling out usually assumes to sample out this and all following downstream spans. Sampled out span becomes current and causes side-effects (span.current().setAttribute()) not compatible with suppression.
+- adds extendable `InstrumentationType` and `SpanBuilder.setType(InstrumentationType type)`
+- adds optional optimization `Tracer.shouldStartSpan(name, kind, type)`
+
+#### Instrumentation API
+
+[Instrumentation API in Java implementation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanKey.java) with suppression by type.
+
+## Trade-offs and mitigation
```
MovieStoreGuy, line 186: I would like to more included in this section, I find it hard to suggest this to internal users without understanding the implications to the data.

MovieStoreGuy (commented):
I have left a few questions in sections I'd like some more clarification.

My main points come down to, what data would a service potential discard? Can I still get enough information to investigate issues within my service? What is potential reduction in outgoing are we expecting to see?

text/trace/0189-instrumentation-layers-suppression.md
```diff
@@ -0,0 +1,222 @@
...
+### Spec changes proposal
+
+- Tracing Semantic Conventions: Each span MUST follow a single (besides general and potentially composite), convention, specific to the call it describes. It MUST specify it when creating a span.
+- Tracing Semantic Conventions: Client libraries instrumentation MUST make context current to enable correlation with underlying layers of instrumentation
+- Trace API: Add `InstrumentationType` expandable enum with predefined values for known trace conventions (HTTP, RPC, DB, Messaging (see open questions)). *Type* is just a convention name.
+- Trace SDK: Add span creation option to set `InstrumentationType`
+  - During span creation, checks if span should be suppressed (if there is another one for kind + type on the parent `Context`) and if there is, returns a *suppressed* span, which is
+    - non-recording
+    - propagating (carries parent context)
+    - does not become current (i.e. `makeCurrent` call with it is no-op)
```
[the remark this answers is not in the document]
lmolkova: Suppressed spans are never reported and are basically duplicates. If we'd make them current, following may happen:
- HTTP tracing middleware creates *Span-1*
- HTTP servlet instrumentation creates *Span-2* - suppressed
- users do `Span.current().setAttribute` - they expect it to be on *Span1* and not *Span2* (which they never even know was suppressed).

text/trace/0189-instrumentation-layers-suppression.md
```diff
@@ -0,0 +1,222 @@
...
+  - HTTP client continues instrumentation with suppressed span
+    - adds attributes: noop
+    - injects context: duplicate work, since parent context is already injected and suppressed span carries the same context
+    - makes suppressed span current: current is not modified, scope is noop
+- Instrumentation - TLS (imaginary, in scope of span2)
+  - works as if it's in scope of span1
+  - current is span1
+  - if explicit parent was used (span2), it carries span1 context
+
+Suppression logic is configurable and encapsulated in span builder SDK implementation - specific library instrumentation should not depend on configuration, e.g.:
```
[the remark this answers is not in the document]
lmolkova: libraries never think about the configuration. I believe suppression by kind + type is a good default since it shows all layers suppressing duplicate instrumentations only. 
I don't have too strong an opinion - vendors/users should be able to easily change it.

text/trace/0189-instrumentation-layers-suppression.md
```diff
@@ -0,0 +1,222 @@
...
+      - HTTP CLIENT call - 1 (Google HTTP client)
+        - HTTP CLIENT call - 1 (Apache HTTP client)
+
+#### Proposed solution
+
+Suppress inner layers of the same instrumentation, i.e. above picture translates into:
+
+- HTTP SERVER span (middleware)
+  - Controller INTERNAL span
+    - HTTP CLIENT call - 1 (Google HTTP client)
```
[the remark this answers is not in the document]
lmolkova: We assume that two different HTTP clients send almost identical spans (if they follow OTel HTTP spec). The duration might be slightly different, but essentially they report the same thing. So we'll keep the outer span and fully disregard the inner span.

text/trace/0189-instrumentation-layers-suppression.md
```diff
@@ -0,0 +1,222 @@
...
+### Configuration
+
+Suppression strategy should be configurable on SDK side - instrumentation does not need to know about it.
+
+Configuration is needed since:
+
+- backends don't always support nested CLIENT spans (extra hints needed for Application Map to show outbound connection)
+- users may prefer to reduce verbosity and costs by suppressing spans of same kind
+
+So following strategies should be supported:
```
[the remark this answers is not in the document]
lmolkova: Yes, e.g. you have DB running over HTTP.

- CLIENT-only suppression will cause all inner CLIENT spans to be suppressed. Our conventions make each CLIENT span have enough information to identify which endpoint it calls into. DB spans have endpoint info. None of the HTTP details would be recorded (no status codes, tries, etc), only the overall result will be set on DB span. You'd choose this configuration if you only care about the logical results and don't care much about transport or if you're saving costs at all means. If you do HTTP requests without outer DB client call, they will be recorded as usual.
- CLIENT + type suppression would keep DB and nested HTTP spans. If will only suppress duplicate HTTP client spans if multiple instrumentations work together. You'd choose this configuration if you want more details.

I understand your concern about suppression in general, what I hear from users some of them may want to sacrifice a bit of verbosity to save money. And kind + type suppression only suppresses duplicate data for those who can afford it.

text/trace/0189-instrumentation-layers-suppression.md
lmolkova: what do you want to see in this section? How to choose a suppression strategy setting for a service assuming it will make it to the spec or what are the trade-offs of introducing these API changes?

text/trace/0189-instrumentation-layers-suppression.md
```diff
@@ -0,0 +1,222 @@
...
+## Motivation
+
+- Provide clarity for instrumentation layers: e.g. DB calls on top of REST API
+- Give mechanism to suppress instrumentation layers for the same convention: e.g. multiple instrumented HTTP clients using each other.
+
+## Explanation
+
+### Spec changes proposal
+
+- Tracing Semantic Conventions: Each span MUST follow a single (besides general and potentially composite), convention, specific to the call it describes. It MUST specify it when creating a span.
```
pyohannes, line 14: As part of this, we'd also need to define which semantic conventions are composable and which are not.

text/trace/0189-instrumentation-layers-suppression.md
```diff
@@ -0,0 +1,222 @@
+# Instrumentation Layers and Suppression
+
+This document describes approach for tracing instrumentation layers and suppressing duplicate layers.
+
+## Motivation
+
+- Provide clarity for instrumentation layers: e.g. DB calls on top of REST API
```
pyohannes, line 7: Related to the questions about a definition of layers: there are discussing about actually modelling HTTP request with multiple spans (for e. g. retries or forwards). Should we aim to cover those cases with a suppression proposal?

text/trace/0189-instrumentation-layers-suppression.md
lmolkova: Some semantic conventions define it (e.g. [Couch DB](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md) wants to have HTTP attributes on DB spans). Unless it's already explicitly defined, we say conventions are not composable. 
It looks like we should allow multiple types on the API surface if we want to allow that.

text/trace/0189-instrumentation-layers-suppression.md
lmolkova: nope, for now, I assume we have one layer of HTTP spans (if we'll have more for logical + physical, I'd treat them as two different layers) - they can't carry the same information when both are enabled.

HTTP instrumentation would need configuration to emit 1) logical or 2) physical or 3) both to avoid redundancy and doubling of costs. I don't see a way to achieve it with suppression

text/trace/0189-instrumentation-layers-suppression.md
MovieStoreGuy: Wooh, sorry. I forgot I had made a comment here.

Something that tells me the risk / potential loss of data that I am expecting.

tedsuo:
@trask @lmolkova is this something the HTTP semconv working group would like to pick up again?

trask:
> @trask @lmolkova is this something the HTTP semconv working group would like to pick up again?

the scope of the current incarnation of the HTTP semconv working group is "just" stability. we discussed this and believe it can be addressed post-stability.

PR #173, opened 2021-08-31
Proposed scenarios and roadmap for messaging semantic conventions for tracing

pyohannes:
[Semantic conventions for messaging systems for tracing](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md) are available, but are in an experimental state. A [workgroup focusing on messaging semantic conventions](https://github.com/open-telemetry/community/pull/819) will work on bringing the existing semantic conventions for messaging to a stable state. The workgroup meets on **Thursdays at 8AM PST**.

This documents proposes a scope for an initial stable version of messaging semantic conventions, as well as a roadmap.  It should serve as a starting point for initial discussions in the workgroup and, once agreed on, define the further agenda of the workgroup.


Oberon00:
Consider moving this PR out of the draft state ("Ready for Review"). If it's in draft state, that signals that you don't need feedback on it yet except from people you explicitly requested it from. Maybe that is the case here since you asked for feedback in the Messaging Conventions WG, but if you also want feedback from outside the WG, I'd mark it as "ready for review". Unless there are some changes you plan to do before.

text/trace/0157-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,184 @@
...
+           +--------------+       |
+                  ^               v
+                  .            Process
+                  .               |
+                  .               v
+                  . . . . . .  Settle
+```
+
+1. The producer creates a message.
+2. The producer publishes the message to an intermediary.
```
Oberon00, line 88: What is not present in this OTEP but defined in the current semantic conventions is the concept of the "(logical) message destination" of which the current conventions know the two kinds "queue" (1 to 1) and "topic" (1 to many).
Or maybe the concept of "intermediary" is supposed to capture not only the concept of a "broker" (aka "physical message destination" in the current conventions) but also the logical message destination?
Would it make sense to split the Intermediary up?

Quoting the current conventions:

> A message is sent by a message *producer* to:
>
> * Physically: some message *broker* (which can be e.g., a single server, or a cluster, or a local process reached via IPC). The broker handles the actual delivery, re-delivery, persistence, etc. In some messaging systems the broker may be identical or co-located with (some) message consumers.
With Apache Kafka, the physical broker a message is written to depends on the number of partitions, and which broker is the *leader* of the partition the record is written to.
> * Logically: some particular message *destination*.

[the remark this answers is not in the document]
pyohannes: With the terminology (described in the section above) we tried to keep as close as possible to [CloudEvents](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#intermediary), which is another CNCF project. I think that's the closest we currently get to a vendor-independent industry standard for describing events (message in the wider sense of this document).

CloudEvents defines and uses the term "intermediary". It also uses "publish" over "send".

I will add a remark to the "Terminology" section that the terms are derived from the CloudEvents spec. Just to make it clear, and to give people an option to disagree.

pyohannes: > What is not present in this OTEP but defined in the current semantic conventions is the concept of the "(logical) message destination" of which the current conventions know the two kinds "queue" (1 to 1) and "topic" (1 to many).

I wonder how much this overlaps with the distinct scenarios of individual settlement and checkpoint settlement described below. Let's discuss this in the next meeting.

However,  I think whether one publishes to a queue or to a topic doesn't alter the model of the 5 stages. This can rather be captured in different scenarios that are based on this model.

text/trace/0157-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,184 @@
...
+Publish -> | INTERMEDIARY | -> Receive
+           +--------------+       |
+                  ^               v
+                  .            Process
+                  .               |
+                  .               v
+                  . . . . . .  Settle
+```
+
+1. The producer creates a message.
```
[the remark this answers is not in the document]
pyohannes: > The Create step is left out in current conventions. Since no networking is involved in creation of the message (only in publishing), I did not consider it of interest for messaging semantic convention.

This OTEP is not about proposing an alternative to the existing conventions, but just to clearly establish a shared conceptual model and a set of scenarios we want to cover. We might still decide to model the "Create" and "Publish" steps by a single span, or to not have a span for the "Create" stage at all.

I think it makes sense to have the "Create" and "Publish" stages separate in a conceptual model. 

> From the WG meeting it sounded like Create is actually supposed to mean "logical publish operation" (potentially with retries) while Publish is the "physical publish operation".

That's my understanding, seems we have to make this clearer in the document. For each message, "Create" can only happen once, whereas "Publish" can happen multiple times (due to retries). "Create" (as well as "Process") are agnostic of the underlying protocol (AMQP, MQTT, HTTP), whereas all the other stages are not.

> That is an interesting topic but should maybe be left out initially, as this is already being discussed both generally (#1306) and for HTTP (#1747)

100 % agreed, whatever we come up with must be consistent with HTTP and other similar areas.

text/trace/0157-messaging-semantic-conventions.md
Oberon00: I agree that it makes no fundamental difference with what steps are involved, although there are probably many differences if you look a bit closer.
I was less concerned about the steps and more about the entities involved. E.g. if there is a problem with message delivery, it might be because some configured max queue (logical destination) length is exceeded. Or there might be a problem with the "physical" server you send the message to (network overloaded, etc). I think the model should capture not only the  steps performed, but also the important entities that are involved in each step.
That said, I think it _could_ make sense to focus on the steps first, because these involve span relationships / context propagation / ... which are probably more difficult to get right than the entitities which will probably come down to some span attributes here and there.

pyohannes:
> Consider moving this PR out of the draft state ("Ready for Review").

Done. I think the first workgroup meeting showed enough alignment on the document so we can have a wider discussion.

Examples are still missing and will be added once we discussed the scenarios.

text/trace/0157-messaging-semantic-conventions.md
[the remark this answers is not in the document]
blumamir: > Since no networking is involved in creation of the message (only in publishing), I did not consider it of interest for messaging semantic convention

I want to also point that sometimes publishing is performed in batches. there is a single publish operation (span) that can contain multiple messages. For example, SQS supports this [send-message-batch operation](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/sqs/send-message-batch.html)

By excluding a dedicated span for "per message" creation, the following issues might come up:

## `Message Attributes`
Some of the collected attributes are relevant only for the context of a message. for example:
- message id
- conversation id
- instrumentations that allow extended capabilities of capturing payloads and metadata

Attaching these attributes to the publish batch operation will require to set the values as arrays, or exclude them altogether. This behaviour can be problematic at least in these cases:
- confusing to end users and backend processing pipelines, which should be prepared for attributes arriving as either single value (single message publish), or array value (batch publish), and correlate the array indices from different attributes to get a complete set for a single message. If displayed in a raw format (for example, in jaeger), it can be visually hard to observe these long attribute arrays and exract insights from them.
- not symmetric with the "consuming" part which does collect a "per message" span. Although it is possible to handle this asymmetry in code, it does produce a more complex and hard to understand flows.

## `Processing Links`
At least for current conventions, the process span has a span link to the publisher. If the publisher is for batch of messages, then multiple processors for different messages will link to the same span, which again can cause very complex and non-trivial processing of traces.

## `Parietal Failure`
When doing batch publish, the status of the response can be either failure for the entire operation (for example, bad credentials, or network issue), or we can get "success" for some messages and "failure" for others. For example, in SQS, if you do a batch publish of 10 messages, and one message is above the payload limit, the response will contain `Successful` array with 9 message ids, and `Failed` array with 1 message id.
If we don't have the create "per-message" reference, it becomes less straight forward to record this information into publish single-span attributes. I would expect the semantic conventions to support the correct representation of this usecase which I believe best describe when you can reference each message with its own span.

These are real-world problems that I encountered while implementing and processing messaging systems traces.

text/trace/0157-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,184 @@
...
+2. The producer publishes the message to an intermediary.
+3. The consumer receives the message from an intermediary.
+4. The consumer processes the message.
+5. The consumer settles the message by notifying the intermediary that the
+   message was processed successfully.
```
blumamir, line 92: The settlement phase is optional.
For example, in rabbitMQ, consumer client can be configured with the `noAck` flag which means the settlement is automatically applied by the broker:

noAck (boolean): if true, the broker won’t expect an acknowledgement of messages delivered to this consumer; i.e., it will dequeue messages as soon as they’ve been sent down the wire. Defaults to false (i.e., you will be expected to acknowledge messages).

Pub-Sub systems (for example: mqtt with QOS 0/1 and amazon SNS) also do not, or can be configured to not apply any application-level settlement as far as I know. once the message is received by the consumer, it is considered settled.

text/trace/0157-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,184 @@
...
+| [Stability guarantees](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#not-defined-semantic-conventions-stability) for semantic conventions are approved and merged. This is not strictly related to semantic conventions for messaging but is a prerequisite for stabilizing any semantic conventions. | 09/30/2021 |
+| An OTEP proposing a set of attributes and conventions covering the scenarios in this document is approved and merged. | 10/29/2021 |
+| Proposed specification changes are verified by prototypes for the scenarios and examples below. | 11/15/2021 |
+| The [specification for messaging semantic conventions for tracing](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md) are updated according to the OTEP mentioned above and are declared [stable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#stable). | 11/30/2021 |
+
+## Terminology
+
+### Message
+
+A "message" is a transport wrapper for the transfer of information between two
```
kenfinnigan, line 36: Not sure "transport" is appropriate as it's entirely possible for messages to be transferred between parties in process, which doesn't involve any transport.

For instance, Quarkus has the concept of "in memory" message channels that are equivalent to Kafka and AMQP but are entirely internal to an application

text/trace/0157-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,184 @@
...
+or more parties. The information is a combination of data and metadata.
```
kenfinnigan, line 37: Can we say it's between two parties? It's possible for messages to be sent to a destination and never read again. Or read at some far off point in the future when there was no concept of what party would receive the message

kenfinnigan, line 37: Should we be utilizing the messaging semantics of "envelope" instead of "wrapper" and "payload" instead of "data"?

text/trace/0157-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,184 @@
...
+In the strict sense, a _message_ is an item of data that is sent to a specific
+destination, whereas an _event_ is a signal emitted by a component upon
+reaching a given state. This document is agnostic of those differences and uses
+the term "message" in a wider sense to cover both concepts.
+
+### Producer
+
+The "producer" is a specific instance, process or device that creates and
+publishes a message. "Publishing" is the process of sending a message or batch
+to the intermediary or consumer.
```
kenfinnigan, line 51: This should probably reference "Publishing" of the message to a destination, where the destination could either be a consumer or an intermediary.

text/trace/0157-messaging-semantic-conventions.md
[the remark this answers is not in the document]
pyohannes: Thanks @blumamir, I agree. I put this on the agenda for tomorrow's meeting, so we can close on this.

text/trace/0157-messaging-semantic-conventions.md
kenfinnigan: On this topic, Quarkus has the ability to acknowledge a message before it's actually processed. Not sure how we can represent it in the flow above, but we need to think about these alternative flows

text/trace/0157-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+#### Examples
+
+1. RabbitMQ: TODO
+
+### Checkpoint-based settlement
+
+Checkpoint-based settlement systems are likely to consume multiple messages at
+once. Messages are created, published (possibly) in batches with other
```
kenfinnigan, line 125: I'm not sure about this sentence.

With Kafka, my understanding is publishing would usually not be batched and if there is a manual batching being done I'm not sure it can be done with unrelated messages. I think it would only be possible to batch messages that were being sent to the same logical destination, which would make the messages of the same type but different payloads

text/trace/0157-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,184 @@
...
+unrelated messages, and delivered in (reshuffled) batches.
+
+Messages are processed as a stream and settled to specific checkpoints. A
+checkpoint points to a position of the stream up to which messages were
+processed and settled. Messages cannot be settled individually, instead, the
+checkpoint needs to be forwarded.
```
kenfinnigan, line 131: I don't think that's strictly true.

Though impacting performance, it is possible with Kafka to settle each message when the processing is complete

text/trace/0157-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,184 @@
...
+Solving this problem requires a solution for sampling based on span links,
+which is not in scope for this OTEP.
+
+### Instrumenting intermediaries
+
+Instrumenting intermediaries can be valuable for debugging configuration or
+performance issues, or for detecting specific intermediary failures.
+
+Stable semantic conventions for instrumenting intermediaries can be provided at
+a future point in time, but are not in scope for this OTEP.
```
kenfinnigan, line 172: Not sure if it's worth mentioning here, but I know there are efforts underway for Kafka to provide its own metrics collection process which might lead to tracing as well in the future

text/trace/0157-messaging-semantic-conventions.md
pyohannes: > That said, I think it could make sense to focus on the steps first, because these involve span relationships / context propagation / ... which are probably more difficult to get right than the entitities which will probably come down to some span attributes here and there.

@Oberon00 When you talk about entities, do you have something in mind that goes beyond the conceptual producer/intermediary/consumer entities? I'm wondering whether we can define entities more fine-grained in such a high-model, as more fine-grained distinctions might be specific to different messaging systems.

text/trace/0157-messaging-semantic-conventions.md
pyohannes: "Individual settlement" means settling a message independent of the position of the message in the queue/stream.

In a Kafka message stream, you cannot settle message B before message A, if message B is at a position later in the stream (past A).

I'll try to make that more clear in the document.

text/trace/0157-messaging-semantic-conventions.md
pyohannes: I tried to avoid the concept of a "destination". Events differ from messages in the sense that events aren't sent to a specific destination, but are just emitted, agnostic of who receives or acts upon the event.

Despite the misleading title, these semantic conventions aim to cover both messaging and eventing system (I tried to capture that in the definition of "message" above).

text/trace/0157-messaging-semantic-conventions.md
pyohannes: Thanks. I added those possible variations (settling before processing, and settling bundled with receiving).

text/trace/0157-messaging-semantic-conventions.md
pyohannes: In my understanding, the messaging semantic conventions aren't applicable to "in memory" message channels. I added this to the list of "Open questions", and I will also put it on the agenda for the next meeting.

text/trace/0157-messaging-semantic-conventions.md
pyohannes: I removed the "between to or more parties".

text/trace/0157-messaging-semantic-conventions.md
pyohannes: You're right, it's misleading. I removed the reference to publishing in batches.

text/trace/0157-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,210 @@
...
+                  . . . . . .  Settle
+```
+
+1. The producer creates a message.
+2. The producer publishes the message to an intermediary.
+3. The consumer receives the message from an intermediary.
+4. The consumer processes the message.
+5. The consumer settles the message by notifying the intermediary that the
+   message was processed. In some cases, the message is settled before it is
+   processed, or it is settled automatically when it is received.
```
blumamir, line 99: The settlement is not necessarily related to any receive.
Some messaging scenarios involve "broadcasting" the message, so whoever is listening receives the message upon transmission, but it is possible that no consumer is registered and the message is simply discarded / settled without any consumer receiving it.

blumamir: for example: redis pub/sub, socket.io rooms, mqtt with QOS 0/1

blumamir: or alternativly, message can be transmitted to consumer, and settled without any confirmation that it was indeed received by the consumer (mqtt QOS=0).
These types of broadcast systems simply iterate on the current active subscribers, try to send the message to each of them and settle it without any feedback from the subscribers.

text/trace/0157-messaging-semantic-conventions.md
kenfinnigan: Saw the definition of "message", but I interpreted that to refer to different people interchangeably using "message" and "event" to mean the same thing.

Is the intention for these conventions to be used for event streaming and event-driven architectures? If so, it might be worth explicitly calling that out

text/trace/0157-messaging-semantic-conventions.md
kenfinnigan: Thanks

text/trace/0157-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,210 @@
...
+
+   Each of the producers produces a continuous stream of messages.
+
+## Open questions
+
+The following areas are considered out-of-scope of a first stable release of
+semantic conventions for messaging. While not being explicitly considered for
+a first stable release, it is important to ensure that this first stable
+release can serve as a solid foundation for further improvements in these areas.
+
```
lmolkova, line 169: Can we briefly talk about spec boundaries? E.g. if the application produces/consumes/routes batches of 100x messages, tracing them might not be a great option because of the performance, storage needs of expressing 100x of links, and gantt-charts not being really useful.

text/trace/0157-messaging-semantic-conventions.md
pyohannes: I added this to the workgroup meeting notes for further discussion.

text/trace/0157-messaging-semantic-conventions.md
pyohannes: We definitely need to talk about that, but I'd prefer to do so once we agreed on a set of scenarios. Reasoning about boundaries already requires us to consider and evaluate possible solutions.

text/trace/0157-messaging-semantic-conventions.md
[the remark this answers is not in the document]
kenfinnigan: After thinking on this, and some discussions in today's WG meeting, I would agree it makes sense to separate the Create and Publish stages.

Create is a distinguishable step as there is the possibility to batch many Create steps (and thus messages) into a single Publish step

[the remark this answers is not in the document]
lmolkova: Here's what we discussed today at the meeting:

Separation is needed for:
- batch send (we need unique context per message to trace them individually)
- creating messages, buffering, and publishing in the background as an optimization
- retries on publish (we should also agree on how they are represented)

Should we merge create/publish when there is one message?
- retries?
- one scenario (separate create/publish) is easier to implement and build UX for  

We'll need to define how this separate Create stage looks like (is it a span? if it's not, how we can generate and record a unique context without span) if we agree on the direction.

@Oberon00 does it make sense for you? Do you want to discuss it in more detail?

text/trace/0157-messaging-semantic-conventions.md
kenfinnigan: After discussion in today's WG meeting, I was asked to provide some more details on "in memory" to indicate how they align with messaging with a broker.

This behavior is based on my experience with Quarkus, so it may differ slightly from other frameworks.

An "in memory" channel has a unique name, within a single application, to define a source or destination of a message. With Quarkus, the only difference between an "in memory" channel and an external channel is the necessary configuration for the external broker, the code to consume/process/publish a message is identical.

With respect to the existing semantic conventions for messaging, there are attributes that would be equally applicable to "in memory" channels as much as ones interacting with an intermediary, such as `messaging.destination`. However, there are attributes that don't currently fit, or which might require an additional value, such as `messaging.destination`, which could possibly add `internal`.

If it's necessary for "in memory" channels to be out of scope for an initial release due to time constraints, that's ok provided it's taken into account by not including changes that would make it unworkable in the future. But I do think "in memory" should be in scope for these conventions as it can be just as important to observe the internal workings of a reactive application utilizing streams and event processing as a synchronous application with method calls.

I would equate "in memory" channel spans as the reactive equivalent to spans of synchronous method calls between services.

pyohannes: > If it's necessary for "in memory" channels to be out of scope for an initial release due to time constraints, that's ok provided it's taken into account by not including changes that would make it unworkable in the future.

That sounds good to me. Do you think we could add Quarkus as specific product, alongside Kafka, RabbitMQ and others? I think it will be easy to support certain in-memory queues that behave like messaging systems, but I'm a bit hesitant to aim for general support for in-memory queues.

text/trace/0157-messaging-semantic-conventions.md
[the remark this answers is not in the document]
pyohannes: Closing this conversation, because I think there's agreement that it's helpful to have the create/publish stages separate on a conceptual level.

We will have further discussions (probably based on another OTEP) about whether and how we want to reflect that on span creation.

text/trace/0157-messaging-semantic-conventions.md
pyohannes: Yes, I think we want to cover both event streaming and event-driven architectures. I'll put it on the agenda for the next meeting and then add clarifications according to the discussion.

text/trace/0157-messaging-semantic-conventions.md
kenfinnigan: Not sure if Quarkus really fits as a product from that perspective, maybe "Reactive Streams"? Might need some thought

pyohannes: "In-memory" message channels are added to the Open Questions section, which we will go through in the workgroup once we're through the scenarios.

As long as Quarkus is covered by the scenarios, we should be able to support it with this semantic conventions. Probably we'd need to work on the wording: "reactive streams" sounds like a better category than "in-memory message channels". The latter, in my understanding, also incorporates channels as a programming language construct (like in Go), we don't want to apply the conventions to those.

text/trace/0157-messaging-semantic-conventions.md
pyohannes: I'm open to those changes, if they provide more clarity. I'd like to be consistent with the CloudEvents terminology (which I tried to base this terminology on). I will put this on the agenda as a minor-priority item, and I will leave this conversation open until we get to discuss it in the workgroup.

text/trace/0157-messaging-semantic-conventions.md
pyohannes: A link to the proposal: https://cwiki.apache.org/confluence/display/KAFKA/KIP-714%3A+Client+metrics+and+observability

text/trace/0157-messaging-semantic-conventions.md
pyohannes: According to the discussion in the workgroup,  I added a note that the settlement stage can be optional in exceptional cases.

text/trace/0157-messaging-semantic-conventions.md
pyohannes: @kenfinnigan I tried to capture the results of our discussion in the document, adding this to the in-memory queues open question:
> While the semantic conventions might be used to instrument in-memory queues or channels which mimic the architecture of inter-application messaging systems, generally supporting in-memory queues or channels is considered out of scope.

text/trace/0157-messaging-semantic-conventions.md
pyohannes: I explicitly added the terms "event streaming" and "event-driven architectures" to the motivation section.

text/trace/0157-messaging-semantic-conventions.md
pyohannes: I added this to the open question about sampling, making clear that link-based sampling is out of scope, but that questions about the number of links in a single trace need to be discussed and considered by the conventions:
> However, having a too high number of span links in a single trace or having too many traces linked together can make the visualization and analysis of traces inefficient. This problem is not related to sampling and needs to be addressed by the semantic conventions.

text/trace/0157-messaging-semantic-conventions.md
pyohannes: I applied the suggested changes.

blumamir (approved):

text/trace/0157-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,238 @@
...
+### Metrics
+
+Messaging semantic conventions for tracing and for metrics overlap and should
+be as consistent as possible. However, semantic conventions for metrics will be
+handled separately and are not in scope for this OTEP.
+
+### In-memory queues or channels
+
+Messaging semantic conventions are not meant for instrumenting in-memory queues
```
kenfinnigan, line 225: @pyohannes I don't really agree with the statement "are not meant for instrumenting in-memory".

I'd prefer we said it can be used for in-memory, but explicitly catering to in-memory is out of scope for the otep.

The way it's worded at present gives the impression that it would be wrong to use them, which I don't think is the case

pyohannes: @kenfinnigan I re-worded this section, trying to emphasize that everything falling in one of the scenarios laid out in the previous section should be supported (no matter whether the intermediary is a separate process or not).

kenfinnigan: @pyohannes I'm good with it mentioning alignment with the scenarios, but saying in-memory channel is explicitly excluded from the conventions is a problem in my view.

With Quarkus, our use of an in-memory channel very much aligns with the scenarios outlined. The only difference is the intermediary is framework code and not a broker, but explicitly excluding in-memory channel indicates that what Quarkus does shouldn't follow the conventions.

I'd like the conventions to be applicable to the Quarkus use case as I would prefer not to have different conventions for reactive messaging in Quarkus between situations where AMQP or Kafka is the intermediary, vs the intermediary being an in-memory channel.

Maybe the wording needs to explicitly allow a reactive stream consisting of publishers and subscribers as being acceptable? That's the Quarkus use case for in-memory. Does the wording help at all?

pyohannes: I removed the example mentioning the in-memory queue. I'm not concerned about the conventions not being applicable to Quarkus. I definitely see it more in the camp of full-fledged messaging systems than as a pattern of low-level asynchronous message passing. And I think the scenarios definitely apply to it.

text/trace/0157-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+
+1. The producer creates a message.
+2. The producer publishes the message to an intermediary.
+3. The consumer receives the message from an intermediary.
+4. The consumer processes the message.
+5. The consumer settles the message by notifying the intermediary that the
+   message was processed. In exceptional cases (fire-and-forget), the
+   settlement stage does not exist.
+
+The messaging semantic conventions need to define how to model those stages in
```
anuraaga, line 111: Eventually it will be nice to further augment this with real world flows perhaps through cloud services. From what I understand, messaging can behave vastly different for scenarios like S3 - SQS - Lambda vs Kinesis vs DynamoDB streams, most importantly whether batching happens or not. It'll be good to knock out as many of these real world flows as possible and detail out user expectations for each.

pyohannes: > It'll be good to knock out as many of these real world flows as possible and detail out user expectations for each.

Agreed. We expect quite some prototyping and verification work once we start to work through the conventions. Any help in doing this for SQS and Kinesis would be very welcome ...

joaopgrassi (approved):
I'm far from a messaging specialist, but from the meetings I joined and after reading all the discussions here, I think this is looking good! Thanks for all the effort in bringing this forward! Giving my "approval"

lmolkova (approved):

text/trace/0173-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+event data in common formats to provide interoperability across services,
+platforms and systems.
+
+### Message
+
+A "message" is a transport envelope for the transfer of information. The
+information is a combination of a payload and metadata.  Messages may be
+transferred directly between parties or via one or more intermediaries.  A
+message may carry annotations that are directed at intermediaries on the
+message path.  Messages are uniquely identifiable.
```
arminru, line 56: Does this mean each message has a unique ID or is this meant to be more conceptual (i.e., that each message is an individual with individual properties).

text/trace/0173-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+publishes a message. "Publishing" is the process of sending a message or batch
+to the intermediary or consumer.
+
+### Consumer
+
+A "consumer" receives the message and acts upon it. It uses the context and
+data to execute some logic, which might lead to the occurrence of new events.
+
+The consumer receives, processes, and settles a message. "Receiving" is the
+process of obtaining a message from the intermediary, "processing" is the
```
arminru, line 75:
```suggestion
process of obtaining a message from the intermediary or producer, "processing" is the
```
Since direct communication is possible as well as per the definition in the Message definition above.

text/trace/0173-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+process of acting on the information a message contains, "settling" is the
+process of notifying an intermediary that a message was processed successfully.
```
arminru, line 77:
```suggestion
process of notifying an intermediary or producer that a message was processed successfully.
```

text/trace/0173-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+CloudEvents is hosted by the CNCF and provides a specification for describing
+event data in common formats to provide interoperability across services,
+platforms and systems.
+
+### Message
+
+A "message" is a transport envelope for the transfer of information. The
+information is a combination of a payload and metadata.  Messages may be
+transferred directly between parties or via one or more intermediaries.  A
+message may carry annotations that are directed at intermediaries on the
```
arminru, line 55: Can the metadata be directed at consumers as well or only intermediaries?

text/trace/0173-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+                  . . . . . .  Settle
+```
+
+1. The producer creates a message.
+2. The producer publishes the message to an intermediary.
+3. The consumer receives the message from an intermediary.
+4. The consumer processes the message.
+5. The consumer settles the message by notifying the intermediary that the
+   message was processed. In exceptional cases (fire-and-forget), the
+   settlement stage does not exist.
```
arminru, line 109: Is this really that exceptional or is both equally valid?

text/trace/0173-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+                  .               |
+                  .               v
+                  . . . . . .  Settle
+```
+
+1. The producer creates a message.
+2. The producer publishes the message to an intermediary.
+3. The consumer receives the message from an intermediary.
+4. The consumer processes the message.
+5. The consumer settles the message by notifying the intermediary that the
```
arminru, line 107: Similar to the above - is the intermediary always present or optional (i.e., in direct communication  between the two peers)?

text/trace/0173-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+   message was processed. In exceptional cases (fire-and-forget), the
+   settlement stage does not exist.
+
+The messaging semantic conventions need to define how to model those stages in
+traces, how to propagate context, and how to enrich traces with attributes.
+Failures and retries need to be handled in all stages that interface with the
+intermediary (publish, receive and settle) and will be covered by general
+instrumentation guidance.
```
arminru, line 115: General as in for all kinds of purposes or in general _for messaging_ only?

text/trace/0173-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+
+Transport batching can be treated as a special case: messages can be
+transported together as an optimization, but are produced and consumed
+individually.
+
+```
++---------+ +---------+ +---------+ +---------+ +---------+ +---------+
+|Message A| |Message B| |Message C| |Message D| |Message E| |Message F|
++---------+ +---------+ +---------+ +---------+ +---------+ +---------+
+ Settled                 Settled                             Settled
```
arminru, line 136: This figure is not clear to me. Is there an x-axis that's supposed to present time or does this just show a mapping of `message`->`settled: yes/no` and could just as well be a table?

text/trace/0173-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+
+   * 1 producer, 1 queue, 2 consumers
+   * 1 producer, fanout exchange to 2 queues, 2 consumers
+   * 2 producers, fanout exchange to 2 queues, 2 consumers
+
+   Each of the producers continuously produces messages.
+
+### Checkpoint-based settlement
+
+Messages are processed as a stream and settled to specific checkpoints. A
```
arminru, line 152: `to` or `at`? (I'm not a native speaker 🙂)

text/trace/0173-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+without also settling message A).
+
+```
+                               Checkpoint
+                                   | 
+                                   v                                
++---------+ +---------+ +---------+ +---------+ +---------+ +---------+
+|Message A| |Message B| |Message C| |Message D| |Message E| |Message F|
++---------+ +---------+ +---------+ +---------+ +---------+ +---------+
+                     <---  Settled
```
arminru, line 171: Looking at this figure now the one above is more clear to me, maybe you could give the former some description.

text/trace/0173-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+```
+
+#### Examples
+
+1. The following configurations should be instrumented and tested for Kafka or
+   a similar messaging system:
+
+   * 1 producer, 2 consumers in the same consumer group
+   * 1 producer, 2 consumers in different consumer groups
+   * 2 producers, 2 consumers in the same consumer group
```
arminru, line 181: "Consumer groups" are not defined in this document yet. Should they be?

text/trace/0173-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+inefficient. This problem is not related to sampling and needs to be addressed
+by the semantic conventions.
+
+### Instrumenting intermediaries
+
+Instrumenting intermediaries can be valuable for debugging configuration or
+performance issues, or for detecting specific intermediary failures.
+
+Stable semantic conventions for instrumenting intermediaries can be provided at
+a future point in time, but are not in scope for this OTEP.
```
arminru, line 215: Might be worth mentioning that you still aim to provide instrumentations that work well without the need to have intermediaries instrumented in first place.

text/trace/0173-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+
+### Metrics
+
+Messaging semantic conventions for tracing and for metrics overlap and should
+be as consistent as possible. However, semantic conventions for metrics will be
+handled separately and are not in scope for this OTEP.
+
+### Asynchronous messaging passing in the wider sense
+
+Asynchronous messaging passing in the wider sense is a communication method
```
arminru, line 225:
```suggestion
### Asynchronous message passing in the wider sense

Asynchronous message passing in the wider sense is a communication method
```

text/trace/0173-messaging-semantic-conventions.md
pyohannes: More the former ("each message has a unique ID"). I'm not quite sure what the latter ("each message is an individual with individual properties") would mean, sounds to me like "each message is unique", which doesn't make much sense.

The idea behind this is, that when one looks at messages in different places (producers, consumers, observability backends), one can tell whether the messages are the same or not.

It is put in a more opaque way, as "a unique ID" suggests a single attribute, whereas a message can also be uniquely identifiable by a combination of attributes.

text/trace/0173-messaging-semantic-conventions.md
pyohannes: Yes, metadata can be directed both at intermediaries and consumers. In some cases, consumers look at the metadata before deciding whether to deserialize a message.

I tried to make that clear. I also removed the use of the term "annotation", using metadata consistently.

text/trace/0173-messaging-semantic-conventions.md
pyohannes: I added a textual description.

text/trace/0173-messaging-semantic-conventions.md
pyohannes: This was discussed in the workgroup. We decided that the messaging system semantic conventions require at least one intermediary, and I updated the definition of "message" accordingly:

> Messages are transferred via one or more intermediaries.

Other message passing scenarios should be covered by different conventions (refer to the out-of-scope section "Asynchronous message passing in the wider sense").

text/trace/0173-messaging-semantic-conventions.md
pyohannes: We assume it's always present.

text/trace/0173-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,255 @@
...
+Messaging semantic conventions for tracing and for metrics overlap and should
+be as consistent as possible. However, semantic conventions for metrics will be
+handled separately and are not in scope for this OTEP.
+
+### Asynchronous message passing in the wider sense
+
+Asynchronous message passing in the wider sense is a communication method
+wherein the system puts a message in a queue or channel and does not require an
+immediate response to continue processing. This can range from utilizing a
+simple queue implementation to a full-fledged messaging system.
```
arminru, line 241: Please add some examples of what kind of systems/scenarios are meant here so it's clearer which particular frameworks are left out of scope for the messaging conventions. The description of https://github.com/open-telemetry/opentelemetry-specification/pull/1582 might be a good reference there.

pyohannes: I added some examples, and a reference to the PR.

trask (approved):

text/trace/0173-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,263 @@
...
+available sampling capabilities of OpenTelemetry, it is not possible to ensure
+that a set of linked traces is sampled. As a result, it is unlikely to sample a
+set of traces that covers the complete path a message takes.
+
+Solving this problem requires a solution for sampling based on span links,
+which is not in scope for this OTEP.
+
+However, having a too high number of span links in a single trace or having too
+many traces linked together can make the visualization and analysis of traces
+inefficient. This problem is not related to sampling and needs to be addressed
```
tigrannajaryan, line 217: It is not clear how this problem can be addressed by semantic conventions. Does this intend to say that minimizing the number of the links is a goal when designing semantic conventions?

text/trace/0173-messaging-semantic-conventions.md
```diff
@@ -0,0 +1,263 @@
...
+6. The [specification for messaging semantic conventions for tracing](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md)
+   are updated according to the OTEP mentioned above and are declared
+   [stable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#stable).
+
+The steps in the roadmap don't necessarily need to happen in the given order,
+some steps can be worked on in parallel.
+
+## Terminology
+
+To leverage existing standards, the terminology used in this document is based
```
tigrannajaryan, line 47: I was confused when I read this the first time. I had the impression we will use CloudEvents specification as a way to model messaging events. I understand now that that is not the case, we are merely borrowing the terminology from CloudEvents (which is good). Perhaps delete the "To leverage existing standards" part (that's what confused me, I thought we are leveraging much more than just the terminology). The rest of the sentence is sufficient.

PR #174, opened 2021-09-10
Scope and roadmap for bringing the existing HTTP semantic conventions for tracing to an initial stable state

denisivan0v:
[Semantic conventions for tracing for HTTP](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md) are available, but are in an experimental state.

A [workgroup focusing on HTTP conventions](https://github.com/open-telemetry/community/issues/826) is requested and will work on bringing the existing semantic conventions for HTTP to a stable state once established.

This documents proposes a scope for an initial stable version of HTTP semantic conventions, as well as a roadmap. It should serve as a starting point for initial discussions in the workgroup and, once agreed on, define the further agenda of the workgroup.

### DIFF
--- a/text/trace/0174-http-semantic-conventions.md
+++ b/text/trace/0174-http-semantic-conventions.md
@@ -50,7 +50,7 @@ Scenarios and open questions mentioned below must be addressed via separate PRs.
 
 ### Error status defaults
 
-4xx responses are no longer create error status codes n case of
+4xx responses are no longer create error status codes in case of
 `SpanKind.SERVER`. It seems reasonable to define the same/similar behavior
 for `SpanKind.CLIENT`.
 
--- a/text/trace/0174-http-semantic-conventions.md
+++ b/text/trace/0174-http-semantic-conventions.md
@@ -48,6 +48,12 @@ some steps can be worked on in parallel.
 
 Scenarios and open questions mentioned below must be addressed via separate PRs.
 
+### Error status defaults
+
+4xx responses are no longer create error status codes n case of
+`SpanKind.SERVER`. It seems reasonable to define the same/similar behavior
+for `SpanKind.CLIENT`.
+
 ### Required attribute sets
 
 > At least one of the following sets of attributes is required:
@@ -79,16 +85,18 @@ too long. E.g., was 500/timeout from the final destination or a proxy?
 ### Security concerns
 
 Some attributes can contain potentially sensitive information. Most likely, by
-default web frameworks/http clients should not expose that.
+default web frameworks/http clients should not expose that. For v1.0 these
+attributes can be explicitly called out.
 
 For example, `http.target` has a query string that may contain credentials.
 
 ## Scope for vNext: scenarios and open questions
 
-### Error status
+### Error status configuration
 
-Per current spec 4xx must result in span with Error status. In many cases
-404/409 error criteria depends on the app though.
+In many cases 4xx error criteria depends on the app (e.g., for 404/409). As an
+end user, I might want to have an ability to override existing defaults and
+define what HTTP status codes count as errors.
 
 ### Optional attributes
 
--- a/text/trace/0174-http-semantic-conventions.md
+++ b/text/trace/0174-http-semantic-conventions.md
@@ -31,6 +31,8 @@ and thus to ship and use stable instrumentation.
 
 ## Scope: scenarios and open questions
 
+> NOTE. The scope defined here is subject for discussions and can be adjusted.
+
 Scenarios and open questions mentioned below must be addressed via separate PRs.
 
 ### Error status
@@ -69,17 +71,17 @@ required and available pre-sampling should be provided)
 * To make it efficient for noop case, need a hint for instrumentation
 (e.g., `GlobalOTel.isEnabled()`) that SDK is present and configured before
 creating pre-sampling attributes.
- 
+
 ### Context propagation needs explanation
 * Reusing instances of client HTTP requests between tries (it’s likely, so clean
-up context before making a call).
+  up context before making a call).
   
 ### WebSockets/Long-polling and streaming
 Anything we can do better here? In many cases connection has app-lifetime,
 messages are independent - can we explain to users how to do manual tracing
 for individual messages? Do span events per message make sense at all?
 Need some real-life/expertize here.
- 
+
 ### Request/Response body (technically out-of-scope, but we should have an idea how to let users do it)
 There is a lot of user feedback that they want it, but
 
--- a/text/trace/0174-http-semantic-conventions.md
+++ b/text/trace/0174-http-semantic-conventions.md
@@ -24,13 +24,22 @@ experimental semantic conventions for HTTP, but does not propose solutions.
 
 ## Roadmap for v1.0
 
-| Description | Done By     |
-|-------------|-------------|
-| This OTEP, consisting of scenarios/open questions and a proposed roadmap, is approved and merged. | 11/30/2021 |
-| [Stability guarantees](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#not-defined-semantic-conventions-stability) for semantic conventions are approved and merged. This is not strictly related to semantic conventions for HTTP but is a prerequisite for stabilizing any semantic conventions. | 11/30/2021 |
-| Separate PRs covering the scenarios and open questions in this document are approved and merged. | 01/31/2022 |
-| Proposed specification changes are verified by prototypes for the scenarios and examples below. | 02/28/2022 |
-| The [specification for HTTP semantic conventions for tracing](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md) is fully updated according to this OTEP and declared [stable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#stable). | 03/31/2022 |
+1. This OTEP, consisting of scenarios/open questions and a proposed roadmap, is
+   approved and merged.
+2. [Stability guarantees](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#not-defined-semantic-conventions-stability)
+   for semantic conventions are approved and merged. This is not strictly related
+   to semantic conventions for HTTP but is a prerequisite for stabilizing any
+   semantic conventions.
+3. Separate PRs addressing the scenarios and open questions listed in this
+   document are approved and merged.
+4. Proposed specification changes are verified by prototypes for the scenarios
+   and examples below.
+5. The [specification for HTTP semantic conventions for tracing](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md)
+   are updated according to this OTEP and are declared
+   [stable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#stable).
+
+The steps in the roadmap don't necessarily need to happen in the given order,
+some steps can be worked on in parallel.
 
 ## Scope for v1.0: scenarios and open questions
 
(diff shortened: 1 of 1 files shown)


text/trace/0174-http-semantic-conventions.md
```diff
@@ -0,0 +1,127 @@
...
+stable state is a crucial step for users and instrumentation authors, as it
+allows them to rely on [stability guarantees](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#not-defined-semantic-conventions-stability),
+and thus to ship and use stable instrumentation.
+
+## Roadmap
+
+| Description | Done By     |
+|-------------|-------------|
+| This OTEP, consisting of scenarios/open questions and a proposed roadmap, is approved and merged. | 09/30/2021 |
+| [Stability guarantees](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#not-defined-semantic-conventions-stability) for semantic conventions are approved and merged. This is not strictly related to semantic conventions for HTTP but is a prerequisite for stabilizing any semantic conventions. | 09/30/2021 |
```
jsuereth, line 27: Will this include JUST Resource + Trace semantic conventions or also metrics?

text/trace/0174-http-semantic-conventions.md
```diff
@@ -0,0 +1,127 @@
...
+Scenarios and open questions mentioned below must be addressed via separate PRs.
+
+### Error status
+
+Per current spec 4xx must result in span with Error status. In many cases
+404/409 error criteria depends on the app though.
+
+### Required attribute sets
+
+> At least one of the following sets of attributes is required:
```
jsuereth, line 45: Does this apply to metrics or just traces?

SHOULD we aim for unity between the two in attribute sets, mark this as explicitly out of scope?

text/trace/0174-http-semantic-conventions.md
denisivan0v: The goal of the OTEP is to bring HTTP semantic conventions for tracing to a stable state. Will update the PR title.

text/trace/0174-http-semantic-conventions.md
denisivan0v: Just traces. Metrics are out of scope of this OTEP.

lmolkova:
Adding security concerns to the scope: `http.target` has a query string that may have credentials, by default web frameworks/http clients should not expose potentially sensitive information.

Oberon00:
@lmolkova 
> http.target has a query string that may have credentials

I have never heard of anybody including credentials in a query string. HTTP basic auth (user:password@hostname) is already addressed in the spec.

On the other hand, Java frameworks have been using the otherwise very obscure "matrix param" format with semicolons to include a session ID: https://stackoverflow.com/a/23600711/2128694 (URLs like `http://www.example.com/context;JSESSIONID=12345?query_param1=ABC` may look familiar).

In any case, http.target is security-sensitive even if you remove the query string. Nowadays, things that would previously have been put into query strings are often included in the path through HTTP route templates like `/search/<term>` instead of `/search?q=<term>?. User IDs, etc. are also often included.

Hostnames (subdomains) may also be user or tenant-specific. IPs are PII. Probably there is more.

lmolkova:
> I have never heard of anybody including credentials in a query string. HTTP basic auth (user:password@hostname) is already addressed in the spec.

https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview#sas-token
https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html#RESTAuthenticationQueryStringAuth


Should we make the target a required attribute if it's sensitive? Perhaps we should define the default level of sanitization and how it can be customzed?

I don't want to go into the technical discussion here. This otep currently summarizes gaps/concerns on current spec on the way to stability, not proposes solutions.

tedsuo:
just FYI, changing 4xx responses to no longer create error status codes on the server has already happened. I'd suggest putting "Error status default" as in-scope, and leave "configuration" as vNext.

text/trace/0174-http-semantic-conventions.md
```diff
@@ -0,0 +1,166 @@
...
+
+Redirects: users may need observability into what server hop had an error/took
+too long. E.g., was 500/timeout from the final destination or a proxy?
+
+### Context propagation needs explanation
+
+* Reusing instances of client HTTP requests between tries (it’s likely, so clean
+  up context before making a call).
+
+### Security concerns
```
jamesmoessis, line 79: Discussed in the instrumentation SIG: To what extent do we address this for v1?

MovieStoreGuy, line 79: We should explicitly call out attributes here that have potential to contain sensitive information (PII, UGC, heaven forbid auth values).

In the past, we've had issues we services:
- Leaking PIU as part of the URI (either as part of the path or the query string)
- Exposing the IP address of the client
- Including usernames as part of attributes 

With attributes that have potential to expose sensitive information, we can address in future conventions with a broader scope on how it should be handled.

denisivan0v: For v1 we can at least identify which attributes can potentially contain sensitive information and mark them accordingly. Added this to the OTEP.

denisivan0v: Yes, I agree that as a first step we can extend attribute [data model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/trace/http.yaml) to add `potentially_sensitive=true/false` property (or similar). Going forward, we might want to come up with an approach on how to allow end users to sanitize/remove that data. Added this to the OTEP. @cobly please feel free to link this OTEP from the issue/PR addressing this for v1.

denisivan0v:
>  I'd suggest putting "Error status default" as in-scope, and leave "configuration" as vNext.

@tedsuo I've updated the OTEP adding the following in-scope for v1:
```
4xx responses are no longer create error status codes in case of `SpanKind.SERVER`. 
It seems reasonable to define the same/similar behavior for `SpanKind.CLIENT`.
```

An the following for the vNext:
```
In many cases 4xx error criteria depends on the app (e.g., for 404/409). As an
end user, I might want to have an ability to override existing defaults and
define what HTTP status codes count as errors.
```

text/trace/0174-http-semantic-conventions.md
MovieStoreGuy: We are currently going through an internally draft at the moment to bounce ideas of each other, however, I'll raise the OTEP by the end of next week at the latest.

MovieStoreGuy (approved):

MovieStoreGuy:
So far this addition makes sense as an initial steps towards a first stable release with vNext incrementally adding more improvements.

tedsuo (approved):
Thanks for organizing this!

anuraaga (approved):

trask (approved):

yurishkuro (commented):
I think this PR mixes the problem of defining roadmap items (eg "we need to decide what/if attrs are required") with the specific solutions ("these are required"). Many of the problems have their own tickets with ongoing discussions, so here we could just say "we need to resolve issue #xxxx".

denisivan0v:
Thanks @yurishkuro. Within this PR/OTEP I'm aiming to outline the scope for v1.0 (and for vNext), so it's mostly about "we need to decide...". And specific solutions are proposed and being discussed as separated issued/PRs - I've added links to those to this OTEP.

yurishkuro:
> Within this PR/OTEP I'm aiming to outline the scope for v1.0 (and for vNext), so it's mostly about "we need to decide..."

This is not how the PR is written though. You have specific section that contain very concrete (and not yet agreed to) prescriptions like "these are required attributes" or "this is how retries must be handled". They do not at all read like "these are things we need to decide in other tickets". To me it's a blocker for approving this OTEP.

denisivan0v:
@yurishkuro Fair enough. I've re-phrased sentences to avoid providing concrete prescriptions in this OTEP. Do you now see any section that might be improved further?

text/trace/0174-http-semantic-conventions.md
jamesmoessis: Potentially this should be out of scope for v1, on 2nd thoughts. Addressing security should be thought through very carefully, so maybe it should be addressed separately.

jamesmoessis: ^ https://github.com/open-telemetry/oteps/pull/187

cobly: 100%. v1 doesn't need to be feature rich, just useable and reliable. We can fast follow with incremental additions.

denisivan0v:
@lmolkova Should the span suppression topic be added in scope for v1.0? /cc @tedsuo

text/trace/0174-http-semantic-conventions.md
denisivan0v: Discussed this at Instrumentation SIG meeting on Nov 30 (https://docs.google.com/document/d/1dWHhyXnfVife-cQ2DW5-d5Ldp1Lq8Rre2UsHpyo8cEE/edit#):
image

It seems important for v1 to at least identify attributes that can potentially contain sensitive information, so customers can setup in-proc or out-of-proc processing of such attributes based on their knowledge of the domain.

Going forward, we can incrementally add more details on how exactly that data could be processed/sanitized.

jamesmoessis (approved):
My prior concerns have been resolved. After SIG discussion, I am happy with what this PR proposes as an initial stable state.

lmolkova:
> @lmolkova Should the span suppression topic be added in scope for v1.0

I don't think suppression is a blocker for HTTP semantic conventions.

lmolkova (approved):

text/trace/0174-http-semantic-conventions.md
```diff
@@ -0,0 +1,186 @@
...
+some steps can be worked on in parallel.
+
+## Scope for v1.0: scenarios and open questions
+
+> NOTE. The scope defined here is subject for discussions and can be adjusted
+  until this OTEP is merged.
+
+Scenarios and open questions mentioned below must be addressed via separate PRs.
+
+### Error status defaults
```
jamesmoessis, line 51: This part seems kind of prescriptive too. Is there somewhere that this is documented as the accepted decision? I think you could make a case for client 4xx spans being UNSET or ERROR. I agree that UNSET is probably better due to the edge cases. Either way, we need to get consensus in the community, has there been any prior discussion on this topic?

jmacd (approved):
Looks ready to merge.

tigrannajaryan (approved):
This looks good to me as a roadmap. I assume the actual semantic conventions will be fleshed out in the specification repo.

text/trace/0174-http-semantic-conventions.md
carlosalberto: Ping @jamesmoessis @denisivan0v

jamesmoessis: I don't think this conversation should block merging, so if it's ready go ahead and merge. I will raise PR in the spec regarding error statuses and we can discuss there.

denisivan0v: @jamesmoessis that's correct, we need to get consensus on this in the community - that's the reason we have this item in scope. Currently, the spec states:

> For HTTP status codes in the 4xx range span status MUST be left unset in case of SpanKind.SERVER and MUST be set to Error in case of SpanKind.CLIENT.

However as you pointed out, there are some edge cases when it would be better to keep status as UNSET (e.g. 409 CONFLICT just indicates that data was modified by other client - which is not a client error; the same could be applicable to 429 or even 404 in case a resource not yet created).

I'd suggest to keep this item and discuss this topic on the next Instrumentation SIG meeting.

Does this help? Please let me know if it works for you and this discussion can be resolved.

denisivan0v: @jamesmoessis you added your comment while I was typing mine :) Looks, we are on the same page here, resolving the discussion.

denisivan0v:
@tigrannajaryan that's right, items in scope are going to be addressed via separate PRs in the specification repo.

@tedsuo can you please help to merge this PR?

PR #178, opened 2021-09-15
Define Mapping of Arbitrary Data to OTLP AnyValue

tigrannajaryan:
The mapping is necessary to correctly implement Logging Library SDKs such that
values mapped are unambiguous, consistent across languages and implementations.


tigrannajaryan:
@jkwatson @zenmoto PTAL.

text/logs/0178-mapping-to-otlp-anyvalue.md
```diff
@@ -0,0 +1,167 @@
...
+#### Associative Arrays
+
+Values that represent an associative arrays with unique keys (also often known
+as maps, dictionaries or key-value stores) SHOULD be converted to AnyValue's
+[kvlist_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L36)
+field.
+
+If the keys of the source array are not strings they MUST be converted to
+strings by any means available, often via a toString() or stringify functions
+available in programming languages, ensuring that the uniqueness of the key is
```
yurishkuro, line 113: "Ensuring uniqueness" doesn't specify the method of doing so. What if instead we simply refer to the following text on handling non-unique keys?

text/logs/0178-mapping-to-otlp-anyvalue.md
```diff
@@ -0,0 +1,167 @@
...
+preserved.
+
+The elements of the source array SHOULD be converted to AnyValue recursively.
+
+If the associative array may contain more than one value for a given key (i.e.
```
yurishkuro, line 118: Add header "duplicate keys"

yurishkuro (approved):

text/logs/0178-mapping-to-otlp-anyvalue.md
```diff
@@ -0,0 +1,167 @@
+# Mapping Arbitrary Data to OTLP AnyValue
+
+This document defines how to map (convert) arbitrary data (e.g. in-memory
+objects) to OTLP's AnyValue.
```
jkwatson, line 4: Do we want to call out that this is specifically for the log datamodel's "body"?

text/logs/0178-mapping-to-otlp-anyvalue.md
```diff
@@ -0,0 +1,167 @@
...
+field using decimal representation.
+
+#### Enumerations
+
+Integer values which belong to a limited enumerated set (e.g. to a Java
+[enum](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html)) SHOULD be
+converted to AnyValue's
+[int_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L33)
+field with the value representing enum's associated value when such associated
+value is naturally obtainable.
```
jkwatson, line 42: Hmm. the specific example of `enum` in Java is problematic, as we have been implored repeatedly not to assume that the ordinal values will be stable or useful for serialization purposes. I think Java's enums should be mapped to their name, perhaps prefixed with the actual enum Type in some way.

text/logs/0178-mapping-to-otlp-anyvalue.md
```diff
@@ -0,0 +1,167 @@
...
+
+Values that represent ordered sequences of other values (such as
+[arrays](https://docs.oracle.com/javase/specs/jls/se7/html/jls-10.html),
+[vectors](https://en.cppreference.com/w/cpp/container/vector), ordered
+[lists](https://docs.python.org/3/tutorial/datastructures.html#more-on-lists),
+[slices](https://golang.org/ref/spec#Slice_types)) SHOULD be converted to
+AnyValue's
+[array_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L35)
+field. String Values and Byte Sequences are an exception from this rule (see
+above).
```
jkwatson, line 102: If the array is heterogenous (like `Object[]` in Java), is this still true? Are we ok with the contents of the AnyValue array type being hetergenous?

text/logs/0178-mapping-to-otlp-anyvalue.md
tigrannajaryan: Changed to prefer symbolic name over ordinal number.

text/logs/0178-mapping-to-otlp-anyvalue.md
jsuereth: +1

text/logs/0178-mapping-to-otlp-anyvalue.md
tigrannajaryan: Reworded and added examples.

text/logs/0178-mapping-to-otlp-anyvalue.md
tigrannajaryan: Made "Associative Arrays With Non-Unique Keys" a separate section.

tigrannajaryan:
@open-telemetry/specs-approvers Please review. This is needed by Logging SDK for now but there is no reason this cannot be extended to Span, Resource or SpanEvent attributes in the future if we decide so. It is best that we make it future-proof and ensure it is reasonable for all signals (even if we don't need it now).

tigrannajaryan:
@open-telemetry/specs-logs-approvers please review.

text/logs/0178-mapping-to-otlp-anyvalue.md
```diff
@@ -0,0 +1,217 @@
...
+[string_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L31)
+field.
+
+String values which not valid Unicode sequences SHOULD be converted to
+AnyValue's
+[bytes_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L37)
+with the bytes representing the string in the original order and format of the
+source string.
+
+#### Bytes Sequences
```
pmm-sumo, line 81: Should we provide some guidelines on how to interpret the bytes sequence? Maybe we could add a semantic convention for an attribute that could store information about the data format (e.g. following [media type](https://en.wikipedia.org/wiki/Media_type) or using [Body Meta-Data](https://docs.google.com/document/d/1ZExye1lW43owwaxcbjOvl0P2qER-UaYd_MyxItc2h0k/edit#heading=h.lsix8inep3nb))

We could introduce a standard attribute name for body (e.g. `body.media.type`) and a convention for attributes (e.g. for attribute `foo`, bytes could be stored into `foo.bytes` while type into `foo.media.type`). That might add some complexity but also provide the much needed information on how the data should be interpreted

kumoroku: unsurprisingly, I like the idea of having a way to annotate what is supposed to be found in a byte array (or even, depending on implementation in a string) :)

I think that we could turn this into another work item as suitable (another OTEP?), independently from approving this document.

kumoroku (approved):

text/logs/0178-mapping-to-otlp-anyvalue.md
tigrannajaryan: I agree. I think we can have a semantic convention for recording the Body's media type.

pmm-sumo: Should we make it part of this oteps, a separate one or just add to semantic conventions eventually?

tigrannajaryan: I think we can simply submit a separate semantic convention PR against spec repo. Not sure if an OTEP is necessary, but if we think the description of the semantic convention is going to be non-trivial maybe do an OTEP.

pmm-sumo (approved):

text/logs/0178-mapping-to-otlp-anyvalue.md
```diff
@@ -0,0 +1,217 @@
...
+                ]}}
+            },
+        ]
+    }
+}
+```
+
+### Other Values
+
+Any other values not listed above SHOULD be converted to AnyValue's
```
djaglowski, line 181: Should we define an expected mapping for a set? 

[Many languages](https://en.wikipedia.org/wiki/Set_(abstract_data_type)#Language_support) have a representation, so it seems we can either establish an imperfect but consistent mapping, or expect to have a variety of inconsistent mappings.

Perhaps the best representation would be as a `kvlist_value`, where each `key` represents an item in the set and each `value` is `empty`.

pmm-sumo: > Should we define an expected mapping for a set?
> 
> [Many languages](https://en.wikipedia.org/wiki/Set_(abstract_data_type)#Language_support) have a representation, so it seems we can either establish an imperfect but consistent mapping, or expect to have a variety of inconsistent mappings.

That's a good point

> Perhaps the best representation would be as a `kvlist_value`, where each `key` represents an item in the set and each `value` is `empty`.

This would limit the `key` type to `string` though, wouldn't it? Perhaps we could simply recommend using Arrays for sets?

djaglowski: > This would limit the type though as key must be string

Yeah, that's true. 

> recommend using Arrays for sets

I agree with this.

text/logs/0178-mapping-to-otlp-anyvalue.md
```diff
@@ -0,0 +1,217 @@
...
+
+Floating point values which are outside the range or precision of IEEE 754
+64-bit floating point numbers (e.g. IEEE 128-bit floating bit values) SHOULD be
+converted to AnyValue's
+[string_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L31)
+field using decimal floating point representation.
+
+#### String Values
+
+String values which are valid Unicode sequences SHOULD be converted to
```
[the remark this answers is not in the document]
tigrannajaryan: I am unable to find which version of Unicode Protobuf adheres to, so this probably needs to remain unspecified for now.

text/logs/0178-mapping-to-otlp-anyvalue.md
```diff
@@ -0,0 +1,217 @@
...
+If the symbolic name of the enumeration is not possible to obtain the
+implementation SHOULD map enumeration's value to AnyValue's
+[int_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L33)
+field set equal to enum's ordinal number when such ordinal number is naturally
+obtainable.
+
+If the ordinal value is also not possible to obtain the enumeration SHOULD be
+converted to AnyValue's
+[bytes_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L37)
+field in any manner that the implementation deems reasonable.
```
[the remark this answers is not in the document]
tigrannajaryan: >should have a unique representation within the resulting set of bytes_value's.

If it is not possible to obtain the symbolic name or the ordinal value of the enum then we are probably in serious trouble. In such circumstances it is not clear if it is possible at all to construct a unique representation. I am not sure requiring it is going to be helpful. Worst case we should be ready to see something like `unknown` in OTLP, regardless of the input enum's value.

text/logs/0178-mapping-to-otlp-anyvalue.md
tigrannajaryan: Added a section for Sets.

tigrannajaryan:
Lint failures will be fixed by https://github.com/open-telemetry/oteps/pull/196 which needs to be merged before this PR.

tigrannajaryan:

djaglowski (approved):

tigrannajaryan:
@open-telemetry/specs-trace-approvers please review.

jsuereth (approved):

tigrannajaryan:
@open-telemetry/specs-approvers last chance to review/comment/object before I merge this and start moving to the spec. This is going to apply to all signals.

dashpole (approved):

text/0178-mapping-to-otlp-anyvalue.md
```diff
@@ -0,0 +1,227 @@
...
+field using decimal floating point representation.
+
+#### String Values
+
+String values which are valid Unicode sequences SHOULD be converted to
+AnyValue's
+[string_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L31)
+field.
+
+String values which not valid Unicode sequences SHOULD be converted to
```
MrAlias, line 75:
```suggestion
String values which are not valid Unicode sequences SHOULD be converted to
```

text/0178-mapping-to-otlp-anyvalue.md
```diff
@@ -0,0 +1,227 @@
...
+
+Floating point values which are outside the range or precision of IEEE 754
+64-bit floating point numbers (e.g. IEEE 128-bit floating bit values) SHOULD be
+converted to AnyValue's
+[string_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L31)
+field using decimal floating point representation.
+
+#### String Values
+
+String values which are valid Unicode sequences SHOULD be converted to
```
MrAlias, line 70: Does an encoding of Unicode need to be specified?

text/0178-mapping-to-otlp-anyvalue.md
```diff
@@ -0,0 +1,227 @@
...
+the converted values are unambiguous and consistent across languages and
+implementations.
+
+## Explanation
+
+[AnyValue](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L27)
+is capable of representing primitive and structured data of certain types.
+
+Implementations that have a source data in any form, such as in-memory objects
+or data coming from other formats that needs to be converted to AnyValue SHOULD
```
MrAlias, line 18: With the goal being to have consistency across implementation do we need to explain when it is valid to not follow this recommendation? Or, should this be made into a requirement?

text/0178-mapping-to-otlp-anyvalue.md
```diff
@@ -0,0 +1,227 @@
...
+#### Integer Values
+
+Integer values which are within the range of 64 bit signed numbers
+[-2^63..2^63-1] SHOULD be converted to AnyValue's
+[int_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L33)
+field.
+
+Integer values which are outside the range of 64 bit signed numbers SHOULD be
+converted to AnyValue's
+[string_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L31)
```
MrAlias, line 32: This mapping of many types into a single "string_value" type means this transformation is not going to be reversible.

Based on the PR description, when we say we want an unambiguous transformation we are only talking about one direction, right?

text/0178-mapping-to-otlp-anyvalue.md
tigrannajaryan: I don't think so. OTLP Protobuf requires UTF8, but the source from which we map can be any other Unicode encoding and can be converted to UTF8. So I think saying "valid Unicode" is a sufficient restruction.

text/0178-mapping-to-otlp-anyvalue.md
tigrannajaryan: Hard to tell what such explanation can look like. I used the "SHOULD" qualifier because I can't prove these recommendations are applicable to every possible use case.

text/0178-mapping-to-otlp-anyvalue.md
tigrannajaryan: Correct. Not reversible. Only "unambiguous" in the sense that it attempts to clearly say what to do. Maybe it is not the right word to use.

text/0178-mapping-to-otlp-anyvalue.md
tigrannajaryan: Fixed.

text/0178-mapping-to-otlp-anyvalue.md
MrAlias: It seems like there is a compatibility issue with implementations not conforming to this specification, different implementations for the same operations would produce different telemetry. This seems to warrant requiring these statements instead of recommending them.

This isn't a blocker. I just want to make sure we understand the implications. It is also important to keep in mind that softening a requirement to a recommendation is backwards compatible. The other way is not.

MrAlias (approved):
:rocket:

text/0178-mapping-to-otlp-anyvalue.md
```diff
@@ -0,0 +1,227 @@
...
+source string.
+
+#### Bytes Sequences
+
+Byte sequences (e.g. Go's `[]byte` slice or raw byte content of a file) SHOULD
+be converted to AnyValue's
+[bytes_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L37)
+field.
+
+### Composite Values
```
zenmoto, line 88: It might be prudent to explicitly call out at this level of detail that composite values can contain values of differing types, though it also does seem clear from the type definitions.

PR #180, opened 2021-10-05
Fix missing apply_to_spans in OTEP 0152

tigrannajaryan:
The OTEP text correctly says:
>In addition it is also possible to optionally specify spans and events that the transformation should apply to.

However, the block after that text is missing the apply_to_spans
that the text alludes to. I wrote the text correctly but made a
mistake when copy/pasting the yaml from my experimental codebase.

I consider this to be a non-breaking change. It is a bug fix since
the tag in yaml is supposed to be there according to the text.

Also clarified that the condition is an AND.


tigrannajaryan:
Alternatively, if we do consider it a breaking change and not a bug fix, I will change the text to match the yaml. (But I think bug fix is the right approach in this case).

tigrannajaryan:
As discussed in spec SIG meeting, this is not a breaking change it is not yet part of the spec, so we are allowed to fix the OTEP before merging it with the spec.

carlosalberto (approved):

tigrannajaryan:
https://github.com/open-telemetry/opentelemetry-go/pull/2267 is blocked on this.

jmacd (approved):

PR #181, opened 2021-10-05
Rename "labels" to "attributes" OTEP 0152

tigrannajaryan:
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/1989

OTEP 0152 predates renaming of labels to attributes in the spec. Unfortunately
the schema file format still uses the term "label" in the metrics section this way:
```yaml
    metrics:
      changes:
        - rename_labels:
            label_map:
              # map of key/values. The keys are the old label name used
              # in the previous version, the values are the new label name
              # starting from this version.

            apply_to_metrics:
              # Optional. If it is missing the transformation is applied
              # to all metrics. If it is present the transformation is applied
              # only to the metrics with the name that is found in the sequence
              # specified below.
```

We don't want the word `label` in the schema files, instead we want the word
`attribute` to be used instead.

Discussed in spec meeting, decided:
- OTEPs which are not part of the spec repo are not considered to be final,
  or be an actual part of stable spec (OTEPs are not labeled "stable" in any way).
- Modify the OTEP, rename `label` to `attribute` (this commit).
- Merge the OTEP with the spec.

### DIFF
--- a/text/0152-telemetry-schemas.md
+++ b/text/0152-telemetry-schemas.md
@@ -18,7 +18,7 @@
     * [rename_attributes Transformation](#rename_attributes-transformation-1)
   * [metrics Section](#metrics-section)
     * [rename_metrics Transformation](#rename_metrics-transformation)
-    * [rename_labels Transformation](#rename_labels-transformation)
+    * [rename_attributes Transformation](#rename_attributes-transformation)
   * [logs Section](#logs-section)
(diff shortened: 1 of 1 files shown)


carlosalberto (approved):

iNikem (approved):

tigrannajaryan:
@open-telemetry/specs-metrics-approvers @open-telemetry/specs-approvers please review.

PR #182, opened 2021-10-25
Export SpanContext.IsRemote in OTLP

axw:
Proposal to update OTLP to indicate whether a span's parent is remote.


axw:
@open-telemetry/specs-approvers can I please get some reviews on this?

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,73 @@
...
+## Motivation
+
+It is sometimes useful to post-process or visualise only entry-point spans: spans which either have no parent (trace roots), or which have a remote parent.
+For example, the Elastic APM solution highlights entry-point spans (Elastic APM refers to these as "transactions") and surfaces these as top-level operations
+in its user interface.
+
+Currently, the only way entry-point spans can be identified is using (lack of) parent ID, and span kind. Relying on span kind can lead to invalid assumptions,
+particularly with relation to `CONSUMER` messaging spans. Using the [batch receiving example](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#batch-receiving)
+in the messaging semantic conventions, `Span Recv1` should be the only entry point into `Process C`. If we assume `CONSUMER` spans are always entry-point spans,
+then this leads to `Span Proc1` and `Span Proc2` being incorrectly classified as entry-point spans. For messaging spans we might also take into account the
```
iNikem, line 14: Why incorrectly? `Span Proc1` and `Span Proc2` essentially participate in distributed operation (in this case via links, not parent but anyway). If span context is propagated with consumed messages and you want to visualise that trace, then these spans will be effectively "local roots" or "entry points".

axw: > (in this case via links, not parent but anyway)

Just to be clear, I'm talking about the "Batch receiving" example and not "Batch processing". In the former example, the "process" spans are children of the "receive" parent, which is part of the same process.

I meant incorrect in that `Span Proc1` and `Span Proc2` would both considered "local roots", but also have process-local parent -- these are in conflict.

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,73 @@
...
+// If the parent span remoteness property is known then the "ok" result will be true,
+// and false otherwise.
+func (ms Span) ParentSpanIsRemote() (remote bool, ok bool)
+```
+
+## Trade-offs and mitigations
+
+None identified.
+
+## Prior art and alternatives
```
tigrannajaryan, line 48: Another alternate is to add a semantic convention to record this flag. Assuming this is a minority of spans it may be more efficient (instead of increasing in-memory size for every span).

axw: Good point. Thanks, I've added this to the proposal.

It really depends on the ratio, which will vary by application. It would be less efficient for a higher ratio of entry-point spans because of the greater overhead of key-value attributes -- both in their in-memory and wire representations.

Is this a significant concern? At least in theory, the proposed flag could be represented with a single byte (0=unset, 1=true, 2=false).

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,80 @@
...
+Currently, the only way entry-point spans can be identified is using (lack of) parent ID, and span kind. Relying on span kind can lead to invalid assumptions,
+particularly with relation to `CONSUMER` messaging spans. Using the [batch receiving example](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#batch-receiving)
+in the messaging semantic conventions, `Span Recv1` should be the only entry point into `Process C`. If we assume `CONSUMER` spans are always entry-point spans,
+then this leads to `Span Proc1` and `Span Proc2` being incorrectly classified as entry-point spans. For messaging spans we might also take into account the
+`messaging.operation` attribute to tell these apart, however `messaging.operation` is not required; and this would not satisfy other scenarios such as actively
+polling a message queue, which would result in a `CONSUMER` span which has a non-remote parent span.
+
+## Explanation
+
+The OTLP encoding for spans has a boolean `parent_span_is_remote` field for identifying whether a span's parent is remote or not.
```
anuraaga, line 20: I have seen the local root terminology used a lot in tracing and thing `is_local_root` could be a simpler term, though I appreciate the explicitness of `parent_span_is_remote` so it's ok too.

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,80 @@
...
+All OpenTelemetry SDKs populate this field, and backends may use it to identify a span as being an entry-point span.
+A span can be considered an entry-point span if it has no parent (`parent_span_id` is empty), or if `parent_span_is_remote` is true.
+
+## Internal details
+
+The first part would be to update the trace protobuf, adding a `boolean parent_span_is_remote` field to the
+[`Span` message](https://github.com/open-telemetry/opentelemetry-proto/blob/b43e9b18b76abf3ee040164b55b9c355217151f3/opentelemetry/proto/trace/v1/trace.proto#L84).
+
+OpenTelemetry SDKs already track whether a span's parent is remote in [`SpanContext`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#isremote).
```
anuraaga, line 29:
```suggestion
OpenTelemetry SDKs already track whether a span has been propagated from a remote in [`SpanContext`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#isremote).
```

The SDKs track whether a "span is remote", not specifically the span's parent. When creating a new span, then the tracer can check whether the parent span is remote and fill in the proposed field. The spirit of this line, that it is just a one line change in SDKs to fill the data, is correct but want to make sure the semantics of `is_remote` aren't confused

anuraaga (approved):
This is an important piece of information and aside from some small comments this OTEP looks like a good idea to me.

I have a small preference for using a field instead of attribute since we currently don't have any data modeling attributes I think, they are all informational. Memory analysis below but tl;dr, even if there was a bigger impact it seems too small to change this above fact about our data modeling.

Note that there likely won't be any significant memory impact on SDKs since we already have `SpanContext.isRemote` - in Java, we just modify here

https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/traces/SpanMarshaler.java#L43

to store `parentContext` instead of `parentId` and serialize based on it

https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/traces/SpanMarshaler.java#L117

so there is zero impact on non-local-root spans. Where SDKs use generated protobuf objects, the relatively short-lived proto Span object will have one extra field only, but no long term storage should be needed.

For backends that receive the data, it is likely that there will be a byte or four (depending on alignment) of extra long term memory usage. Anyways it doesn't seem enough to factor into decision making.

text/0182-otlp-remote-parent.md
axw: I like `is_local_root` too, and it aligns with how we would use it. I chose `parent_span_is_remote` to align with `SpanContext.isRemote` in the API. I'm fine with either.

text/0182-otlp-remote-parent.md
axw: Maybe it's just my pedantic brain but I find that a bit confusing: spans aren't propagated, span context is. Let me see if we can come to a compromise...

The [docs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#isremote) say:

> An API called IsRemote, that returns a boolean value, which is true if the SpanContext was propagated from a remote parent

How about I just repeat that here?

```suggestion
[`SpanContext.IsRemote`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#isremote) identifies whether span context has been propagated from a remote parent.
```

anuraaga: OTel went with the approach of `NonRecordingSpan` to hold a propagated span context

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#wrapping-a-spancontext-in-a-span

It's true that only a context is propagated but the SDKs happen to treat things a little differently

anuraaga: That being said your correction seems fine too

text/0182-otlp-remote-parent.md
axw: @iNikem did that clear things up? Any concerns still, or can this be resolved?

anuraaga:
@open-telemetry/specs-approvers Can you please review this otep?

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,80 @@
...
+## Motivation
+
+It is sometimes useful to post-process or visualise only entry-point spans: spans which either have no parent (trace roots), or which have a remote parent.
+For example, the Elastic APM solution highlights entry-point spans (Elastic APM refers to these as "transactions") and surfaces these as top-level operations
+in its user interface.
+
+Currently, the only way entry-point spans can be identified is using (lack of) parent ID, and span kind. Relying on span kind can lead to invalid assumptions,
+particularly with relation to `CONSUMER` messaging spans. Using the [batch receiving example](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#batch-receiving)
+in the messaging semantic conventions, `Span Recv1` should be the only entry point into `Process C`. If we assume `CONSUMER` spans are always entry-point spans,
+then this leads to `Span Proc1` and `Span Proc2` being incorrectly classified as entry-point spans. For messaging spans we might also take into account the
```
tigrannajaryan, line 14: Why are `Span Proc1` and `Span Proc2` classified as entry-point spans? They have a parent (Span Recv1), so why are they classified as entry-point spans at all?

text/0182-otlp-remote-parent.md
tigrannajaryan: It is not clear to me.

What is the desirable outcome? Which of all these spans you want to be classified as "entry-point"?

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,80 @@
...
+
+Update OTLP to indicate whether a span's parent is remote.
+
+## Motivation
+
+It is sometimes useful to post-process or visualise only entry-point spans: spans which either have no parent (trace roots), or which have a remote parent.
+For example, the Elastic APM solution highlights entry-point spans (Elastic APM refers to these as "transactions") and surfaces these as top-level operations
+in its user interface.
+
+Currently, the only way entry-point spans can be identified is using (lack of) parent ID, and span kind. Relying on span kind can lead to invalid assumptions,
```
pyohannes, line 11: If you're looking at a complete trace, you could identify "transactions" by grouping spans according to the [`service.name`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service) attribute (which is required). If a parent has a different `service.name`, it is very likely a remote parent.

I guess you want to identify spans with remote parents just by looking at isolated spans? If so, it would be good to add this as an explicit requirement (and maybe add the `service.name` based solution as alternative).

text/0182-otlp-remote-parent.md
discostu105: This approach can only classify "entry spans" according to one parent definition. If there are multiple (regular parent, links) then the approach struggles. It would have to be clarified that this "remote parent" information only covers regular parent information.

text/0182-otlp-remote-parent.md
discostu105: This would happen under the assumption that your algorithm simply classifies all `CONSUMER` spans as entry-spans.

text/0182-otlp-remote-parent.md
discostu105, line 11:
```suggestion
In a system where only one span is processed at the time, without building up a DAG of spans to identify the entry-span, the only way entry-point spans can be identified is using (lack of) parent ID, and span kind. Relying on span kind can lead to invalid assumptions,
```

text/0182-otlp-remote-parent.md
tigrannajaryan: I still don't understand. Can you please list exactly which spans from the referenced "Batch receiving" example are desirable to be classified as "entry-point"? I don't understand what the goal is here.

text/0182-otlp-remote-parent.md
axw: Thank you!

axw: Thanks, good call. That assumes that there won't be two services with the same name, but that's probably a fair assumption. I believe this is addressed by this suggestion (please let me know otherwise): https://github.com/open-telemetry/oteps/pull/182/files/7a14ec0e23ea13d90cc8f19d14e8f8dd04a68d82#r812340831

text/0182-otlp-remote-parent.md
axw: @tigrannajaryan sorry, I'll try to restate in more detail here. Once we've come to a shared understanding I'll update the text.

> Can you please list exactly which spans from the referenced "Batch receiving" example are desirable to be classified as "entry-point"?

`Span Prod1`, `Span Prod2`, and `Span Recv1`

Typically I would expect to see producer spans as children of other span (like a background task, or an HTTP request), in which case `Span Prod1` and `Span Prod2` would no longer be classified as entry-point. They're "entry-point" by definition in the OTEP text because they're trace roots.

More generally: the goal is to identify the spans which represent a request that is entering a service, or originating within a service.

Examples of entry point spans:
 - an incoming HTTP request (entering a service)
 - a message is received from a queue (entering a service)
 - a background task (originating within a service -- trace root)

Examples of non-entry point spans:
 - an outgoing HTTP request
 - a database query
 - message processing (comes _after_ receiving; the receiver span is the entry-point to the service)

Put another way: a trace is a DAG, using only span ID / parent ID (i.e. ignoring links); each service forms a subgraph, and each subgraph has its own [local root](https://github.com/open-telemetry/oteps/pull/182/files#r785611800), or entry point.

The "Batch receiving" example has three traces: Span_Prod1, Span_Prod2, and Span_Recv1+Span_Proc1+Span_Proc3. I've grouped the spans by service/subgraph, and added a border to the local root/entry-point spans:

```mermaid
graph TD
    subgraph producer1
    Span_Prod1
    end

    subgraph producer2
    Span_Prod2
    end

    subgraph batch_receiver
    Span_Recv1 --> Span_Proc1
    Span_Recv1 --> Span_Proc2
    Span_Proc1
    Span_Proc2
    end

    style Span_Prod1 stroke-width:4
    style Span_Prod2 stroke-width:4
    style Span_Recv1 stroke-width:4
```

This is probably not the most illustrative example, and I may have made things confusing by focusing on an edge case. Here's an alternative, but still contrived, example with non-batch receiving messaging spans:

```mermaid
graph TD
    subgraph comments_service
    POST_comments(POST /comment)
    POST_comments --> comments_send(comments send)
    end

    subgraph auth_service
    POST_comments --> POST_auth(POST /auth)
    POST_auth --> LDAP
    end

    subgraph user_details_service
    POST_comments --> GET_user_details(GET /user_details)
    GET_user_details --> SELECT_users(SELECT FROM users)
    end
    
    subgraph comments_inserter
    comments_send --> comments_receive(comments receive)
    comments_receive --> comments_process(comments process)
    comments_process --> INSERT_comments(INSERT INTO comments)
    end

    style POST_comments stroke-width:4
    style POST_auth stroke-width:4
    style GET_user_details stroke-width:4
    style comments_receive stroke-width:4
```

The reason why I focused on the "bulk receiving" example is because we can use other properties to identify entry-points for the more common traces:
 - trace roots are of course trivial to identify
 - server-side request spans can be identified by SERVER span kind
 - by assuming that a CONSUMER span is always an entry point, we can classify the "comments receive" span as an entry point

That assumption is the bit that we want to get rid of, because it's obviously wrong in some cases. Relying on span kind also feels a little brittle, as instrumentation libraries may make a mistake and mess things up.

tigrannajaryan: >Span Prod1, Span Prod2, and Span Recv1

All 3 of these spans are parentless. Unless I am missing something the fact that they are parentless is sufficient to classify them as "entry-point". Why is there a need for additional `IsRemote` flag? I think this example hurts more than it helps. It does not show why `IsRemote` flag is justified.

>Here's an alternative, but still contrived, example with non-batch receiving messaging spans

This is indeed what I was missing. It would be great to have this example in the OTEP instead of the Batch Receiving example which confuses me.

>the goal is to identify the spans which represent a request that is entering a service, or originating within a service.

Can we have this in the OTEP text? This seems like **the** goal yet not mentioned so clearly in the OTEP.

--
 
Question 1: do we have the means for instrumentation to easily identify that the span that they are creating has a remote parent? Is it easy to populate the `IsRemote` flag? The OTEP says how the flag will be used, how it will be recorded in the protocol, but it does not say how the instrumentations will be able to correctly set the value of this flag.

Question 2: why can't the backend identify entry points without the `IsRemote` flag? In the example shown the span can be classified as having a remote parent for example if its parent's `service.name` is different. Can a heuristic like this be sufficient to achieve the goal without adding a new flag?

axw: Apologies for the delay in responding. I have updated the text and replaced the example.

> Question 1: do we have the means for instrumentation to easily identify that the span that they are creating has a remote parent? Is it easy to populate the IsRemote flag? The OTEP says how the flag will be used, how it will be recorded in the protocol, but it does not say how the instrumentations will be able to correctly set the value of this flag.

Instrumentation does not need to do anything, it's already part of the API and SDKs. From the "Internal details" section:

[`SpanContext.IsRemote`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#isremote) identifies whether span context has been propagated from a remote parent.
The OTLP exporter in each SDK would need to be updated to record this in the new `parent_span_is_remote` field.

> Question 2: why can't the backend identify entry points without the IsRemote flag? In the example shown the span can be classified as having a remote parent for example if its parent's service.name is different. Can a heuristic like this be sufficient to achieve the goal without adding a new flag?

That would work, but assumes the backend can see all spans at the time it needs to make the determination whether a span is an entry-point or not; ours cannot. This was also addressed in https://github.com/open-telemetry/oteps/pull/182#discussion_r812340831, please let me know if I should elaborate further.

tigrannajaryan:
@axw sorry for late reply, this fall off my radar (and thanks for reminding). The proposal looks reasonable to me.

If we end up adding `flags` field to Span in OTLP as [proposed here](https://github.com/open-telemetry/opentelemetry-proto/pull/384/files) I think we can record the `IsRemote` as a bit in the `flags` (which makes it zero cost in memory and very small cost on the wire).

Oberon00:
> If we end up adding flags field to Span in OTLP as [proposed here](https://github.com/open-telemetry/opentelemetry-proto/pull/384/files) I think we can record the IsRemote as a bit in the flags

This is a very nice idea that may lead to much trouble and head-scratching if whatever bit we choose gets a meaning assigned in W3C. Unless we make our "flags" larger than the W3C trace flags, but then it's not zero-cost.

tigrannajaryan:
> > If we end up adding flags field to Span in OTLP as [proposed here](https://github.com/open-telemetry/opentelemetry-proto/pull/384/files) I think we can record the IsRemote as a bit in the flags
> 
> This is a very nice idea that may lead to much trouble and head-scratching if whatever bit we choose gets a meaning assigned in W3C. Unless we make our "flags" larger than the W3C trace flags, but then it's not zero-cost (and in theory, the W3C trace flags field might have more bits an a future traceparent header version).

Our flags _are_ already larger than W3C trace flags. We [reserve 8 bits for W3C trace flags](https://github.com/open-telemetry/opentelemetry-proto/blob/c31b2d8fd2e84684fa09de862e0fd7c3aa95f7ed/opentelemetry/proto/logs/v1/logs.proto#L209) (of which only 1 bit is defined by W3C currently) and the remaining 24 bits in our flags are available for anything else we want to use it for. If W3C in the future decides to come up with more than 8 bits of flags they have to do it in a very explicit way, in which case we can also follow what they do.

axw:
@bogdandrutu ICYMI Tigran's comment above (https://github.com/open-telemetry/oteps/pull/182#issuecomment-1132111698):

> I think you discussed this in OpenCensus too. PTAL.

Could you please take a look?

jack-berg (approved):

alanwest (approved):

text/0182-otlp-remote-parent.md
axw: Marking this resolved, please let me know if you still have any doubts about this.

willarmiros (approved):
This proposal would be very useful to us at AWS X-Ray and more generally as stated in the OTEP. To give the concrete example with our backend, X-Ray distinguishes spans as either [segments or subsegments](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html). Segments represent a microservice, which is semantically equivalent to a span whose parent is remote. Without this attribute exposed on the span, we have to do some [ugly guesswork](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/dd2141b5f6c498b7d817013886a22c3c1c4084bf/exporter/awsxrayexporter/internal/translator/segment.go#L87-L93) to determine if a span should be a segment or subsegment, leading to inaccuracies.

axw:
Thanks for chiming in @willarmiros, and to everyone who has reviewed so far.

I'd also love to get this merged. IIANM, there's still 3 more reviews required. I suppose the grey check approvals are not code owners, and don't count towards the required "4 approving reviews".

@open-telemetry/specs-approvers please take a look. If more clarification is needed, please let me know. I'd be happy to have a meeting if it helps, bearing in mind I'm UTC+8.

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Explanation
+
+The OTLP encoding for spans has a boolean `parent_span_is_remote` field for identifying whether a span's parent is remote or not.
+All OpenTelemetry SDKs populate this field, and backends may use it to identify a span as being an entry-point span.
+A span can be considered an entry-point span if it has no parent (`parent_span_id` is empty), or if `parent_span_is_remote` is true.
+
+## Internal details
+
+The first part would be to update the trace protobuf, adding a `boolean parent_span_is_remote` field to the
```
yurishkuro, line 54: L48 says "The OTLP encoding for spans has a boolean `parent_span_is_remote` field", what does it refer to if not protobuf?

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,108 @@
...
+[`Span` message](https://github.com/open-telemetry/opentelemetry-proto/blob/b43e9b18b76abf3ee040164b55b9c355217151f3/opentelemetry/proto/trace/v1/trace.proto#L84).
+
+[`SpanContext.IsRemote`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#isremote) identifies whether span context has been propagated from a remote parent.
+The OTLP exporter in each SDK would need to be updated to record this in the new `parent_span_is_remote` field.
```
yurishkuro, line 58: I don't think this would be sufficient. `SpanContext.IsRemote` can only be true for the _parent span context_, passed to StartSpan. Once the new span is started, the parent span context may be lost (reduced to `parent_span_id` field), so the Exporter won't have access to it.

text/0182-otlp-remote-parent.md
axw: Not sure I understand your question, let's see:

This is the implementation detail of L48. In case it's not clear: L48 is written "as though it was already implemented and you were explaining it to a user", per the template: https://github.com/open-telemetry/oteps/blob/main/0000-template.md#explanation

text/0182-otlp-remote-parent.md
axw: > Once the new span is started, the parent span context may be lost (reduced to parent_span_id field), so the Exporter won't have access to it.

Are you saying that `SpanContext.IsRemote` can be wrong? Do you have an example in mind where only the parent span ID is retained in the parent span context? The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#isremote) doesn't suggest anything like that.

If that's the case, then I suppose implementers would have to record the parent span context's `SpanContext.IsRemote` property on the child span at creation time.

yurishkuro: Parent span context is usually used to create the current span's context, by copying flags and trace id, and then to populate the `parent_span_id` field (and NOT copying IsRemote flag). After that the parent context is not needed for the current span, so by the time the span is finished and sent to exporters the parent's IsRemote indicator is lost unless it is captured explicitly in the span. In either case, `span.context.IsRemote` is always going to be false, by definition.

Oberon00: > After that the parent context is not needed for the current span, so by the time the span is finished and sent to exporters the parent's IsRemote indicator is lost unless it is captured explicitly in the span.

Depending on how you read the spec for ReadableSpan, the full parent SpanContext must be captured, and most (all?) SDKs do it, or capture at least a `has_remote_parent` boolean (not is_remote but has_remote_parent!). E.g. see https://github.com/dotnet/runtime/issues/65660.

<https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#additional-span-interfaces>:

> Readable span: A function receiving this as argument MUST be able to access all information that was added to the span, as listed [in the API spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span).

and at the linked API spec:

> Spans encapsulate:
> ...
> * A parent span in the form of a [Span](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span), [SpanContext](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spancontext), or null

yurishkuro: So it may be implied, but the spec does not explicitly require an API like hasRemoteParent, are you suggesting to add it?

Oberon00: Yes, I think that would be a good idea.

axw: > After that the parent context is not needed for the current span, so by the time the span is finished and sent to exporters the parent's IsRemote indicator is lost unless it is captured explicitly in the span.

Doesn't that conflict with the wording that @Oberon00 referenced above? i.e. `MUST be able to access all information that was added to the span` - which to me means that the parent SpanContext (as originally specified, fully intact) must be available on a span when it reaches the exporter.

> In either case, span.context.IsRemote is always going to be false, by definition.

Sorry, I think I was unclear my description. What I meant was that the span's *parent* `SpanContext.IsRemote()` would be used by the exporter. It the parent SpanContext is required to be accessible in a ReadableSpan, then we can use its IsRemote property.

Oberon00: If you interpret the spec strictly (which you should normally do), the full Context should be available which includes the whole parent Span (not readable). While that would totally work, I don't think any SDK has implemented it that way (.NET has the full parent ReadWriteSpan if there was one provided though, but .NET is divergent in other ways). So I think it would make sense to clarify this point in a non-breaking way, so that most current SDK implementation can be considered conformant without requiring changes. Unless someone has a strong use case for wanting the full parent Context available.

blumamir (approved):
I support adding this and will for sure use it. 

Currently, we (aspecto) rely on `service.name`+`service.instance.id` tuple to distinguish between different services. If we see `client`/`producer` followed by `server`/`consumer` on the same name + instance id tuple, we assume the "incoming" span is a transaction root. With this new flag, we could simplify our logic and be more robust to the edge cases and the variety out there. 

One side note related to this topic - I think the `isRemote` flag is also missing from the [links in protobuf](https://github.com/open-telemetry/opentelemetry-proto/blob/2bdeb79c7646d205c3e05322789afe30cb9ebaa8/opentelemetry/proto/trace/v1/trace.proto#L220) as well

```proto
  message Link {
    bytes trace_id = 1;
    bytes span_id = 2;
    string trace_state = 3;
    repeated opentelemetry.proto.common.v1.KeyValue attributes = 4;
    uint32 dropped_attributes_count = 5;
  }
```

The link [records a `SpanContext`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#specifying-links) so when users create a link they can specify the `isRemote` property and distinguish between remote links and links to another span in the same process. This attribute is described in the spec, and implemented in SDKs (at least in JS) but is not encoded in otlp and I think it should.

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Motivation
+
+It is sometimes useful to post-process or visualise only entry-point spans: spans which either have no parent (trace roots), or which have a remote parent.
+For example, the Elastic APM solution highlights entry-point spans (Elastic APM refers to these as "transactions") and surfaces these as top-level operations
+in its user interface.
+
+The goal is to identify the spans which represent a request that is entering a service, or originating within a service, without having to first assemble the
+complete distributed trace as a DAG (Directed Acyclic Graph). It is trivially possible to identify trace roots, but it is not possible to identify spans with
+remote parents.
```
pyohannes, line 13: I think in the context of this OTEP, it's worth clarifying why the existing `SpanKind` cannot be use to achieve the desired outcome.

This is text from the [`SpanKind` specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spankind):
> The first property described by `SpanKind` reflects whether the Span is a "logical" remote child or parent. By "logical", we mean that the span is logically a remote child or parent, from the point of view of the library that is being instrumented. Spans with a remote parent are interesting because they are sources of external load. Spans with a remote child are interesting because they reflect a non-local system dependency.

Given this and the example below, couldn't one treat every `SERVER` and `CONSUMER` span as spans with a remote parent? If not, then I think this OTEP should also propose an adapted specification of  `SpanKind`. Otherwise the specification would suggest that a span has both `SpanKind` which reflects whether the span is a "logical" remote child or parent, and `ParentSpanIsRemote`, which also reflects whether the span is a remote child.

AlexanderWert: Unfortunately, the specification on `SpanKind` remains unclear for the `CONSUMER` SpanKind, as you can see [in the summary table](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spankind):

> image

`Remote Incoming` is a `maybe` for a `CONSUMER` span. So, to solve that ambiguity I see the following alternatives:

- Introduce a new attribute `parent_span_is_remote` and API `SpanContext.HasRemoteParent` as proposed in this OTEP. (And I agree that would potentially require to adapt / refine the `SpanKind` specification as well to clarify the relation between `SpanKind` and `parent_span_is_remote`.)
- Revise the values for `SpanKind` (especially for `CONSUMER`) to make it unambiguous wrt. `Remote Incoming`. However, this would be a breaking change for spans that are reported as of now as `CONSUMER` spans.

pyohannes: Would you be fine adding the second alternative as an alternative solution to the OTEP? And also adding the need to clarify the relationship between `SpanKind` and `parent_span_is_remote` as an open question?

AlexanderWert: @pyohannes Thanks, added as requested.

pyohannes (approved):

text/0182-otlp-remote-parent.md
Oberon00: I think "remote incoming" is always "maybe" for Server and Consumer. E.g. you can have nested server spans in the same process for different framework levels

(Aside: Sync/Async also cannot be determined by the parent, it depends on what the client/producer does)

AlexanderWert: @Oberon00 Are you saying that `SpanKind` is not suitable to encode the `hasRemoteParent` information? And that we should rather make it explicit with a new `hasRemoteParent` in the API?

Oberon00: > Are you saying that SpanKind is not suitable to encode the hasRemoteParent information?

Yes, SpanKing is definitely not suitable for this (it could be used as input for a heuristic at best). That is is also what the original comment in this thread is saying

AlexanderWert:
@open-telemetry/specs-approvers Is anything missing in this proposal to get approvals / reviews?

There seems to be quite some interest for this and also https://github.com/open-telemetry/opentelemetry-proto/pull/384 seems to be waiting on this.

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,124 @@
...
+The OTLP encoding for spans has a boolean `parent_span_is_remote` field for identifying whether a span's parent is remote or not.
+All OpenTelemetry SDKs populate this field, and backends may use it to identify a span as being an entry-point span.
+A span can be considered an entry-point span if it has no parent (`parent_span_id` is empty), or if `parent_span_is_remote` is true.
+
+## Internal details
+
+The first part would be to update the trace protobuf, adding a `boolean parent_span_is_remote` field to the
+[`Span` message](https://github.com/open-telemetry/opentelemetry-proto/blob/b43e9b18b76abf3ee040164b55b9c355217151f3/opentelemetry/proto/trace/v1/trace.proto#L84).
+
+[`SpanContext.IsRemote`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#isremote) identifies whether span context has been propagated from a remote parent.
```
rapphil, line 57: Are we implicitly assuming that every time we read context from context propagators, the `SpanContext`should have `IsRemote=true`?

If yes, Do we need to call this out in this proposal?

Oberon00: This is not "implicitly assumed", it is already part of the spec. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#isremote

Note that this OTEP does not propose any new concepts, it simply makes an existing concept exported via OTLP.

AlexanderWert: One thing that came out from the discussions above is that in addition will need to add a new `hasRemoteParent` API (https://github.com/open-telemetry/oteps/pull/182#discussion_r970918867), so will update this OTEP to make it clear

Oberon00: Thanks for referencing that, apparently I had promised but forgotten to submit a suggestion (it will be only a small clarification though)

text/0182-otlp-remote-parent.md
Oberon00: I had forgotten about this. Now created https://github.com/open-telemetry/opentelemetry-specification/pull/3257

axw:
This was closed in error -- https://github.com/open-telemetry/opentelemetry-specification/pull/3257 is a prerequisite clarification, it doesn't resolve this PR overall.

Can someone with permissions please reopen? (I can't.)

Oberon00:
It seems that the syntax https://github.com/open-telemetry/oteps/pull/182#discussion_r971857784 resolves the whole PR. Sorry for that.

AlexanderWert:
- With https://github.com/open-telemetry/opentelemetry-specification/pull/3257 being merged, the parent SpanContext (including the parent's `IsRemote` information) is specified as something that must be exposed on the `ReadableSpan`. So, [this comment](https://github.com/open-telemetry/oteps/pull/182#discussion_r969725628) should be resolved.
- [Here](https://github.com/open-telemetry/oteps/pull/182#discussion_r1108718501), @Oberon00 explained very well, why SpanKind (with its _**"logical"**_ remote parent-child relationship ) is not suitable for carrying information about the actual `isRemoteParent` information

@open-telemetry/specs-approvers as the main discussion points have been addressed, can we proceed with this proposal to expose the `parent->isRemote` information (as a protobuf field) on a serialized span and merge this proposal?

trask (approved):

bogdandrutu (approved):

tsloughter:
Carrying on from the Zoom chat today in the Spec SIG meeting:

I found that my question was covered here https://github.com/open-telemetry/oteps/pull/182/files#r969043021 and specifically resolved by the patch to the spec here https://github.com/open-telemetry/opentelemetry-specification/pull/3257

> MUST expose at least the full parent [SpanContext](api.md#spancontext).

So in the case of Erlang at least we simply need to update to comply with the latest spec and include a copy of the parent SpanContext in the Span and not just the SpanId.

The other discussion was around whether `IsRemote` really means "remote" since it could be propagated from within the same process -- example being propagating a context from C++ called through the JNI or an Erlang NIF.

But what "is remote" I don't think is seen by anyone as anything to block this OTEP, rather a discussion to be had separately that was prompted by this OTEP.

Oberon00:
IsRemote today means IsExtracted, or "span ID not guaranteed to be generated by this TracerProvider instance"

carlosalberto:
I suggest we have the discussion regarding the naming ("is remote or not") as part of making this into the Specification (where we will have to update the proto files with the proper documentation docs anyway).

carlosalberto:
@open-telemetry/specs-approvers We have enough approvals so I will merge in the next couple of days. Please raise your voice if you want to do a review before that.

carlosalberto:
@axw We are ready to merge, but we need to get the lint check to pass. Can you check what's happening? https://app.circleci.com/pipelines/github/open-telemetry/oteps/1593/workflows/eadc0058-1700-47a1-b96a-60abff61c475/jobs/2588

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,124 @@
...
+The main benefit of this approach is that it additionally enables backends to group spans by their process subgraph.
+
+### Alternative 2: introduce a semantic convention attribute to identify entry-point spans
+
+As an alternative to adding a new field to spans, a new semantic convention attribute could be added to only entry-point spans.
+
+This approach would avoid increasing the memory footprint of all spans, but would have a greater memory footprint for entry-point spans.
+The benefit of this approach would therefore depend on the ratio of entry-point to internal spans, and may even be more expensive.
+
+### Alternative 3: extend SpanKind values
```
graphaelli, line 102:
```suggestion
### Alternative 3: extend SpanKind values
```

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,124 @@
...
+As an alternative to adding a new field to spans, a new semantic convention attribute could be added to only entry-point spans.
+
+This approach would avoid increasing the memory footprint of all spans, but would have a greater memory footprint for entry-point spans.
+The benefit of this approach would therefore depend on the ratio of entry-point to internal spans, and may even be more expensive.
+
+### Alternative 3: extend SpanKind values 
+
+Another alternative is to extend the SpanKind values to unambiguously define when a CONSUMER span has a remote parent or a local parent (e.g. with the message polling use case). 
+
+For example, introducing a new SpanKind (e.g. `AMBIENT_CONSUMER`) that would have a clear `no` on the `Remote-Incoming` property of the SpanKind, and `REMOTE_CONSUMER` would have a clear `yes` on the `Remote-Incoming` property of the SpanKind. The downside of this approach is that it is a breaking on the semantics of `CONSUMER` spans.
```
graphaelli, line 106:
```suggestion
For example, introducing a new SpanKind (e.g. `AMBIENT_CONSUMER`) that would have a clear `no` on the `Remote-Incoming` property of the SpanKind, and `REMOTE_CONSUMER` would have a clear `yes` on the `Remote-Incoming` property of the SpanKind. The downside of this approach is that it is a breaking on the semantics of `CONSUMER` spans.
```

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,124 @@
...
+## Open questions
+
+### Relation between `parent_span_is_remote` and `SpanKind`
+The specification for `SpanKind` describes the following:
+
+```
+The first property described by SpanKind reflects whether the Span is a "logical" remote child or parent ...
+```
+
+However, the specification stay ambiguous for the `CONSUMER` span kind with respect to the property of the "logical" remote parent.
```
graphaelli, line 118:
```suggestion
However, the specification stay ambiguous for the `CONSUMER` span kind with respect to the property of the "logical" remote parent.
```

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Relation between `parent_span_is_remote` and `SpanKind`
+The specification for `SpanKind` describes the following:
+
+```
+The first property described by SpanKind reflects whether the Span is a "logical" remote child or parent ...
+```
+
+However, the specification stay ambiguous for the `CONSUMER` span kind with respect to the property of the "logical" remote parent. 
+Nevertheless, the proposed field `parent_span_is_remote` has some overlap with that `SpanKind` property.
```
graphaelli, line 119:
```suggestion
Nevertheless, the proposed field `parent_span_is_remote` has some overlap with that `SpanKind` property.
```

graphaelli (changes requested):
satisfy markdown-lint

graphaelli:
@carlosalberto would you accept the suggestions to address the markdown linting issues?

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,124 @@
...
+### Alternative 3: extend SpanKind values
+
+Another alternative is to extend the SpanKind values to unambiguously define when a CONSUMER span has a remote parent or a local parent (e.g. with the message polling use case). 
+
+For example, introducing a new SpanKind (e.g. `AMBIENT_CONSUMER`) that would have a clear `no` on the `Remote-Incoming` property of the SpanKind, and `REMOTE_CONSUMER` would have a clear `yes` on the `Remote-Incoming` property of the SpanKind. The downside of this approach is that it is a breaking on the semantics of `CONSUMER` spans.
+
+
+## Open questions
+
+### Relation between `parent_span_is_remote` and `SpanKind`
```
graphaelli, line 111:
```suggestion
### Relation between `parent_span_is_remote` and `SpanKind`

```

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,124 @@
...
+This approach would avoid increasing the memory footprint of all spans, but would have a greater memory footprint for entry-point spans.
+The benefit of this approach would therefore depend on the ratio of entry-point to internal spans, and may even be more expensive.
+
+### Alternative 3: extend SpanKind values
+
+Another alternative is to extend the SpanKind values to unambiguously define when a CONSUMER span has a remote parent or a local parent (e.g. with the message polling use case). 
+
+For example, introducing a new SpanKind (e.g. `AMBIENT_CONSUMER`) that would have a clear `no` on the `Remote-Incoming` property of the SpanKind, and `REMOTE_CONSUMER` would have a clear `yes` on the `Remote-Incoming` property of the SpanKind. The downside of this approach is that it is a breaking on the semantics of `CONSUMER` spans.
+
+
```
graphaelli, line 108:
```suggestion
```

text/0182-otlp-remote-parent.md
```diff
@@ -0,0 +1,124 @@
...
+### Alternative 2: introduce a semantic convention attribute to identify entry-point spans
+
+As an alternative to adding a new field to spans, a new semantic convention attribute could be added to only entry-point spans.
+
+This approach would avoid increasing the memory footprint of all spans, but would have a greater memory footprint for entry-point spans.
+The benefit of this approach would therefore depend on the ratio of entry-point to internal spans, and may even be more expensive.
+
+### Alternative 3: extend SpanKind values
+
+Another alternative is to extend the SpanKind values to unambiguously define when a CONSUMER span has a remote parent or a local parent (e.g. with the message polling use case).
```
graphaelli, line 104:
```suggestion
Another alternative is to extend the SpanKind values to unambiguously define when a CONSUMER span has a remote parent or a local parent (e.g. with the message polling use case).
```

graphaelli (changes requested):
another round of markdown-lint-fix

carlosalberto:
Merged. Please start preparing the PR(s) to include this change in the Spec + proto repositories.

PR #185, opened 2021-11-10
Semantic conventions for Uptime Monitoring

jsuereth:
A proposal and guide.


text/metrics/0185-uptime-monitoring-semantic-conventions.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Restart Count
+
+restart_count is a monotonic sum of the number of times that a process has restarted. This metric should be generated from an external observer of the system.  The start timestamp of this metric is the start time of whatever process is observing restarts.
+A process *may* report its own restarts, but likely this would need to be done via a DELTA sum which is aggregated by some external observer.
+
+## Trade-offs and mitigations
+
+The biggest tradeoff here is defining `uptime` metrics as non-montonic sums vs. either pure gauge or non-montonic sums. The fundmental question here is whether default sum-based aggregation is meaningful for this metric, in addition to the default-query-capabilities of common backends for cumulative sums. The proposal trades-off allowing an external observer to monitor uptime (with resets) in addition to common assumptions on querying rates for cumulative sums.
```
dustinlessard-wf, line 80: nit, small typo on _non-montonic_

text/metrics/0185-uptime-monitoring-semantic-conventions.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Internal details
+
+We propose the following metrics be used to track uptime within OpenTelemetry:
+
+| Name                   | Description                  | Units | Instrument Type              |
+| ---------------------- | ---------------------------- | ----- | -----------------------------|
+| *.uptime               | Seconds since last restart   | s     | Asynchronous Gauge           |
+| *.health               | Availability flag.           | 1     | Asynchronous Gauge           |
+| *.restart_count        | Number of restarts.          | 1     | Asynchronous Counter         |
```
tigrannajaryan, line 60: There is already a `restart_count` _attribute_ defined for resources: https://github.com/open-telemetry/opentelemetry-specification/blob/a25d5f03ab58ecf88c09f635df97d2328b5ba237/specification/resource/semantic_conventions/k8s.md#container
It would be useful to tell how these two are related (if they are).

text/metrics/0185-uptime-monitoring-semantic-conventions.md
```diff
@@ -0,0 +1,104 @@
...
+Since supervisor processes like this do not often exist, other ways to partially achieve this are to configure an alert on the average uptime over a long time period (e.g. "average uptime over 60m window < 15m"), and/or alert when a metric is absent.
+
+## Internal details
+
+We propose the following metrics be used to track uptime within OpenTelemetry:
+
+| Name                   | Description                  | Units | Instrument Type              |
+| ---------------------- | ---------------------------- | ----- | -----------------------------|
+| *.uptime               | Seconds since last restart   | s     | Asynchronous Gauge           |
+| *.health               | Availability flag.           | 1     | Asynchronous Gauge           |
```
tigrannajaryan, line 59: The paragraph above nicely tells the drawbacks of "health" as a metric. As oposed to that I like kubernetes's approach. The "liveliness" and "readyness" are easier to define precisely: "liveliness" means "I am alive, let me run" (and the opposite of that means "I am in trouble, need help, restart me, do something"), and "readyness" means "I can now accept a workload".
I have a hard time assigning a similarly precise meaning to the "Healthy" metric. What does "availability" mean?
Given this, should we perhaps avoid "heath" as a metric altogether and perhaps instead use "ready" and "lively" (or whatever the names) metrics?

text/metrics/0185-uptime-monitoring-semantic-conventions.md
```diff
@@ -0,0 +1,104 @@
...
+
+Since supervisor processes like this do not often exist, other ways to partially achieve this are to configure an alert on the average uptime over a long time period (e.g. "average uptime over 60m window < 15m"), and/or alert when a metric is absent.
+
+## Internal details
+
+We propose the following metrics be used to track uptime within OpenTelemetry:
+
+| Name                   | Description                  | Units | Instrument Type              |
+| ---------------------- | ---------------------------- | ----- | -----------------------------|
+| *.uptime               | Seconds since last restart   | s     | Asynchronous Gauge           |
```
tigrannajaryan, line 58: The [general recommendations](https://github.com/open-telemetry/opentelemetry-specification/tree/a25d5f03ab58ecf88c09f635df97d2328b5ba237/specification/metrics/semantic_conventions#instrument-naming) appear to say that `.time` suffix needs a dot. Should this be `*.up.time` if we follow the recommendation?

text/metrics/0185-uptime-monitoring-semantic-conventions.md
jsuereth: From a metric standpoint, having restart_count as a resource attribute is really really bad.  I'm suprised no one commented on that, but I think it should be dropped, as it violates resource identity and causes high cardinality.


The TL;DR: though is that restart_count is something you'd want to alert on, and for metric-systems that means it needs to be a metric.  They'd likely be the *same* value, but one is a resource attribute the other is a metric data point.

text/metrics/0185-uptime-monitoring-semantic-conventions.md
jsuereth: This is meant to be the `readyness` flag.  We could split this into two with readyness being recommended, and `liveliness` being optional.

text/metrics/0185-uptime-monitoring-semantic-conventions.md
tigrannajaryan: > From a metric standpoint, having restart_count as a resource attribute is really really bad. I'm suprised no one commented on that, but I think it should be dropped, as it violates resource identity and causes high cardinality.

It was discussed, see https://github.com/open-telemetry/opentelemetry-specification/pull/1945#pullrequestreview-759394466

The conclusion was that for a k8s container in a pod it is an identifying attribute, not a metric.

text/metrics/0185-uptime-monitoring-semantic-conventions.md
reyang: It seems `*.time` is normally used for things that have additive property.

jmacd:
I'd like to consider an alternative not mentioned in this document, and I'm not sure where to propose it.

Instead of two metrics, "health" and "uptime", I propose a single non-monotonic Sum named "alive". This data type requires that a start time be included with the measurement, unlike Gauge. The difference between the start time and the measurement time is the process uptime. 

I have made this proposal already in connection with https://github.com/open-telemetry/opentelemetry-specification/issues/1078, where I pointed out that we can implement service discovery in a push-based metrics system by joining this "alive" metric with information retrieved by service discovery.

text/metrics/0185-uptime-monitoring-semantic-conventions.md
```diff
@@ -0,0 +1,105 @@
+# Guide to Uptime Monitoring for Metric Semantic convetions
+
+A guide to best practices and use cases around uptime monitoring using metrics.
+
+Original authors:  Keith Jordy (@kjordy) and Quentin Smith (@quentinmit), adapted for OpenTelemetry by @jsuereth.
+
+## Motivation
+
+Why should we make this change? What new value would it bring? What use cases does it enable?
```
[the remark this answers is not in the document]
tigrannajaryan: I think this just the copy of the template: https://github.com/open-telemetry/oteps/blob/main/0000-template.md
@jsuereth probably forgot to update this section :-)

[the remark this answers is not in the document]
jsuereth: I did, will update later in the week (a bit overloaded today)

jsuereth:
@jmacd Commented offline, but recording here for posterity.

> Instead of two metrics, "health" and "uptime", I propose a single non-monotonic Sum named "alive" with value 1. This data type requires that a start time be included with the measurement, unlike Gauge. The difference between the start time and the measurement time is the process uptime.

From a pure collection standpoint, I like a lot of what this brings, however I think we need to take an end-to-end focus.  Specficially: "Can I write a query / dashboard / alert to solve the stated use cases?"

AFAICT, with known backends/query languages (Prometheus, Graphite, etc.) it's hard to pull the data back out, specifically the "Seconds since start" value in PromQL.  We should make sure we have an answer to that.

text/metrics/0185-uptime-monitoring-semantic-conventions.md
quentinmit: The Kubernetes restart count is not actually part of a unique-across-time identifier:

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#containerstatus-v1-core

> The number of times the container has been restarted, currently based on the number of dead containers that have not yet been removed. Note that this is calculated from dead containers. But those containers are subject to garbage collection. This value will get capped at 5 by GC.

Arguably that means it's not even useful as a metric, but it's certainly not a unique identifier.

text/metrics/0185-uptime-monitoring-semantic-conventions.md
quentinmit: I like the naming of `readyness` (`readiness`?) and `liveliness`, but if we're going down the road of multiple types, there actually could be arbitrary numbers of health status. For example, a backend could serve internal and external clients, and be unready for external clients but ready for internal clients. Can we just call it `readyness` and say "add an attribute if there are multiple distinct ready states"?

text/metrics/0185-uptime-monitoring-semantic-conventions.md
evantorrie: also `fundmental` => `fundamental`

tedsuo:
@jsuereth how important/relevant is this OTEP? Please assign an appropriate priority, or close if it's old and we no longer need it.

tomasmota:
What is the state of this? It is still not clear to me how to implement this in otel. I suppose uptime is ok, but the health metric as 1|0 makes it not so useful. Should I then just do uptime for both, and only update health if the checks succeed?

Is it not a common use-case that most services would need this in some way? Or are people just relying directly on kubernetes checks instead? I understand that metric such as ops/sec. are much better, but not all services are doing stuff all the time, so this is much needed for those.

I had made an issue on this but closed it expecting this might progress. https://github.com/open-telemetry/opentelemetry-specification/issues/2923

tedsuo:
@jsuereth is this stale, or is semconv currently working on this?

Manuelraa:
I would also be interested in this. A generic `up` metric for creation of generic uptime alerting would be awesome.
Especially having it in the standard itself and e.g. integrated to OpenTelemetry Collector.

carlosalberto:

PR #186, opened 2021-11-10
Support per-Tracer configuration through Tracer Provider

tsloughter:
Support setting configuration (Sampler, IdGenerator, Span Limits, etc) on individual Tracers that overrides the shared configuration provided by the Tracer Provider.

### DIFF
--- a/text/0186-per-tracer-configuration.md
+++ b/text/0186-per-tracer-configuration.md
@@ -1,0 +1,124 @@
+# Per Tracer Configuration
+
+Support setting configuration (Sampler, IdGenerator, Span Limits, etc) on
+individual Tracers that overrides the shared configuration provided by the
+Tracer Provider.
+
+## Motivation
+
+- Allows Sampler decisions based on Instrumentation Library without adding it as
+  an explicit argument to `Sample` calls. This resolves [Issue
+  1588](https://github.com/open-telemetry/opentelemetry-specification/issues/1588).
+  
+- Provides a way to disable Tracing in a library or for any Named Tracer.
+
+This is already mentioned in the specification:
+
+> If configuration must be stored per-tracer (such as disabling a certain
+> tracer), the tracer could, for example, do a look-up with its
+> name+version+schema_url in a map in the TracerProvider, or the TracerProvider
+> could maintain a registry of all returned Tracers and actively update their
+> configuration if it changes.
+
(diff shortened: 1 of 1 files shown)


text/0186-per-tracer-configuration.md
```diff
@@ -0,0 +1,124 @@
+# Per Tracer Configuration
+
+Support setting configuration (Sampler, IdGenerator, Span Limits, etc) on
```
pyohannes, line 3: Could you give an exhaustive list of configuration to be shared? Or is it any kind of configuration, including processors exporters, and resources?

tsloughter: It is any configuration. But I don't believe that includes the resource? The resource is part of the Provider but it isn't a configuration value passed by the user when building a Provider, right?

I used `etc` because it should cover any future additional configuration added to the Tracer Provider.

I left out explicitly including Span Processors because it is a bit of a pain point I cover later in the OTEP, but I could add it back to the list just to be complete.

And the Exporter is configured as part of the Processor so it is based on if those are configurable or not.

text/0186-per-tracer-configuration.md
```diff
@@ -0,0 +1,124 @@
...
+
+> some applications may want to or have to use multiple TracerProvider
+> instances, e.g. to have different configuration (like SpanProcessors) for each
+> (and consequently for the Tracers obtained from them)
+
+While this proposal does not require removing this capability it does replace
+the necessity of using multiple Tracer Providers in order to have different
+configurations. Most existing implementations do not offer the
+capability of creating multiple Tracer Providers and may essentially prohibit
+it. Allowing configuration to be done through the Tracer Provider based on the
```
Aneurysm9, line 76: Is this true?  I don't recall anything in the spec that would lead to implementations that are incapable of supporting multiple tracer providers.  While multiple providers do present the issue noted below WRT sharing exporters, I think that any problem caused by sharing exporters can be solved without new specification (i.e., treated as implementation details).

tsloughter: There isn't anything in the spec that would necessarily lead to it. Some, like Java, support multiple providers only through having multiple SDKs.

I don't think leaving how exporters are shared up to each implementation would be a nice user experience, and definitely not lead to consistent support.

tedsuo:
Closed at @tsloughter's request :)

PR #187, opened 2021-11-11
Data Classifications for resources and attributes

MovieStoreGuy:
The idea behind this that not all telemetry data is the same with regards to importance, how it is processed, and what is supported by the down stream vendors.

This approach would allow for efficient checking of data to then apply various configuration that is required by your organisation.

Some sections are raw and left as dot points pending discussion to help expand upon them.


anuraaga:
This seems related to this spec issue https://github.com/open-telemetry/opentelemetry-specification/issues/2114

At a glance, classifications like sensitive, high-cardinality do seem more useful when creating attribute views than just optional and required.

/cc @lmolkova

yurishkuro (commented):
A few high level thoughts:
* the word "resource" has specific meaning in OTEL, you're overloading it by saying "resource classification". You are classifying attributes, not resources.
* I think the classifications can be part of the schema files that we already support, I don't see why std attributes need to be classified in user code.
* I'm fairly certain there will be big pushback on making this an SDK feature, why not do it in the collector where it can be implemented once?
* before an OTEP like this is accepted there needs to be a working prototype with benchmarks

MovieStoreGuy:
> the word "resource" has specific meaning in OTEL, you're overloading it by saying "resource classification". You are classifying attributes, not resources.

Both are being classified here, attributes have their own classification along with their the resource sending it. The reason being is that if you can not allow resources with PD or UGC to be exported to a vendor you can drop the resource or remove the attributes that contain PD or UGC. 

Classifying the resource is important because it does not require searching / validating attributes or other fields of the resource. 

> I think the classifications can be part of the schema files that we already support, I don't see why std attributes need to be classified in user code.

I don't see an easy pathway to success by doing this, and I maintaining it within the schema (semantic convention) alone would create a lot of friction trying to release changes in it. 

> I'm fairly certain there will be big pushback on making this an SDK feature, why not do it in the collector where it can be implemented once?

Some organisations can not allow for PD, UGC to leave the application so only implementing in the collector already restricts those users from adopting. 

> before an OTEP like this is accepted there needs to be a working prototype with benchmarks
I currently have a rather stripped down version that I had used to test my assumptions, the scenarios that I had replicated where:
- Regex filter (removing attributes that matched a pattern)
- Lookup based filter (have a list of known strings to exactly match)
- Classification Filtering (searching for attributes that are either PD or UGC)
- Resource Filtering (removing resources if they can PD or UGC)

Using Golang's benchmark, I had got the following results:
```
BenchmarkBaselineFilter/Collection-Size-100-12       	                                   40220	                 28404 ns/op
BenchmarkLookupFilter/Collection-Size-100-12         	                                   65653	                  18251 ns/op
BenchmarkRegxFilter/Collection-Size-100-12           	                                        835	             1408659 ns/op
BenchmarkClassificationAttributeFilter/Collection-Size-100-12      	           50774	                 23447 ns/op
BenchmarkClasssificationResourceFilter/Collection-Size-100-12     	1000000000	         0.0000052 ns/op
```

Simply performing a range over an attribute collection of size 100 was `28404 ns/op`, the time being a lookup filter containing exact string to match at `18251 ns/op` with classification attribute filtering taking `23447 ns/op` (approx in the middle of the two benchmarks).

However, using Resource classification filtering, it can filter out resource values sub nanosecond times.

MovieStoreGuy:
By the end of the day, I'll publish my mini prototype so others can play with themselves

tsloughter:
> I'm fairly certain there will be big pushback on making this an SDK feature, why not do it in the collector where it can be implemented once?

Are you sure? Not only does not everyone use the collector but that would still mean sending the PII data to the collector.

tigrannajaryan:
> * I think the classifications can be part of the schema files that we already support, I don't see why std attributes need to be classified in user code.

+1 to this. I think it is worth entertaining this alternate to see what the tradeoffs are. The SDKs already contain pre-built dictionaries of standard attributes (in Go SDK it is in the `semconv` package). We can include the classification in these dictionaries and avoid the need to:

- Call Classify() every time the attribute is added to a span/log/metrc.
- Send classification information over the network.

For non-standard attributes it may still be necessary to do what this OTEP suggests. I don't know how important it is though, perhaps standard attributes cover the majority of cases where classification matters (maybe not, I may be wrong).

>I don't see an easy pathway to success by doing this, and I maintaining it within the schema (semantic convention) alone would create a lot of friction trying to release changes in it.

On the other hand by NOT making it part of the semantic conventions there is a big risk of having inconsistent behaviour in different SDKs, where each instrumentation author makes their own classification decisions.

jamesmoessis:
> I think the classifications can be part of the schema files that we already support, I don't see why std attributes need to be classified in user code. 

Maybe there could be sensible defaults which can be overridden. The semantic convention, or the instrumentation author can't know which attributes will be sensitive because it changes between applications. For example, `net.peer.ip` may or may not be PII depending on whether the peer is a user's device or another server. Only the application owner can know that information.

MovieStoreGuy:
As I mentioned earlier, here is the prototype I had mentioned https://github.com/MovieStoreGuy/data-classifier, it is a rather stripped down only preserving attributes and resource (without any of the values or nested fields). 

> +1 to this. I think it is worth entertaining this alternate to see what the tradeoffs are. The SDKs already contain pre-built dictionaries of standard attributes (in Go SDK it is in the semconv package). We can include the classification in these dictionaries and avoid the need to:

Right, I think you're referring to the Classification Values here? 
I am happy for them to be defined as part of the semantic convention since the values themselves should be rather static over the longevity of the project. (High Cardinality, PII, or UGC data will not stop being a problem).
 
To your point of:
> Send classification information over the network.

In the scenario where you have mixed mode deployments of running compute nods, lambdas and users running experiments / tests locally. Not sending the classification would mean that the internal gateway / proxy collector that is forwarding the data can not filter known classifications that is not supported by the system being exported to. Assuming that application owners ignored classification filtering in their exporter, it would mean that resources or attributes that are can not be classified without impacting processing performance. 

Reactive classification is extremely computational expensive with a lot of operational overhead, proactive classifications reduces the need to categorise the data in order to filter it. 

Moreover, I mention within the OTEP that half the space requested (64bits) can be reserved for Vendors to enable advanced features like smaller retention windows, improve resource catalogs when filtering for telemetry (SLO resources would be more important that high cardinality resources for example). The vendor can extend their exporter to do additional actions when these classifications are set.

Classifications can be limited to the application context and no included as part of the wire datagram, however, I see a lot value including it in order to help to build central means of ensuring resource data isn't sent to the wrong place.

>  I don't know how important it is though, perhaps standard attributes cover the majority of cases where classification matters (maybe not, I may be wrong).

Our current adoption of Open Telemetry in Atlassian requires heavy stream process in order to filter out the standard attributes that are high cardinality, and we also wanting to extend the semantic convention to  support the idea of real user monitoring (RUM) which is currently not defined in the semantic convention so being able to add classification to those attributes and resources mean we can correctly experiment with the idea without having to add additional processing requirements into a our internal telemetry streaming service. 

> On the other hand by NOT making it part of the semantic conventions there is a big risk of having inconsistent behaviour in different SDKs, where each instrumentation author makes their own classification decisions.

I agree, the values of the classifications has to be consistent across language and the semantic convention is the best place to define it.

MovieStoreGuy:
To @jamesmoessis point, classifications are coupled to the application deployment.

Meaning that application sitting on the network edge would define `net.peer.ip` as PII however, an internal service behind the firewall would not consider it PII but both may consider it high cardinality.

yurishkuro:
> Some organisations can not allow for PD, UGC to leave the application so only implementing in the collector already restricts those users from adopting.

Are those organizations prepared to commit significant engineering capacity to developing and maintaining this non-trivial capability in every language supported by OTel? On the other hand, a deployment of the main app with collector as a sidecar communicating over local network (i.e. in practice over local unix socket) - why would that be considered "leaving the application"?

MovieStoreGuy:
Hey @yurishkuro ,

> Are those organizations prepared to commit significant engineering capacity to developing and maintaining this non-trivial capability in every language supported by OTel?

I am not sure your intent behind this comment and I want to understand more :) 
I am fully understand that a lot of the language projects are struggling to reach feature complete and adding this would add more to those backlogs however the intent of this OTEP is to discuss its merits.

I wanted to understand what part of this OTEP is non-trivial? 
As I understand the problem, adding classification support into Otel is at most adding less than 10 required methods into the SDK which their implementation is small. 
Adding the classification values to the semantic convention which is mostly auto generated stubs per language.

I can see managing the classifications within the semantic convention and how the convention would adopt it to be non trivial and these can be expanded upon further here :) 

Per your commitment comment, if the change is as simple as I think then it shouldn't require a significant timeline. I am happy to take on making a collector extension to support this idea as that has been my main project that I have contributed to.

> On the other hand, a deployment of the main app with collector as a sidecar communicating over local network (i.e. in practice over local unix socket) 

I am not certain that communicating via unix socket is supported by the collector? Let me check that :) 

> Why would that be considered "leaving the application"?

Once it has been sent via the application, it is no longer the application that has control of the data and that is where the concern is.

text/0187-data-classification.md
```diff
@@ -0,0 +1,323 @@
...
+
+A staring list can look like the following :
+
+| Classification                     | Purpose                                                                                               | BitShift  | Hint Value (base 16)  | bit mask             |
+|----------------------------------- |------------------------------------------------------------------------------------------------------ |---------- |---------------------- |--------------------- |
+| No Value                           | Is the default value when un set                                                                      | 0 << 0    | 0x0000                | 0000 0000 0000 0000  |
+| Ephemeral                          | The attributes that are short lived and have high potential to change over extended periods of time.  | 1 << 0    | 0x0001                | 0000 0000 0000 0001  |
+| High Cardinality                   | The value is an unbounded set                                                                         | 1 << 1    | 0x0002                | 0000 0000 0000 0001  |
```
arminru, line 184:
```suggestion
| High Cardinality                   | The value is an unbounded set                                                                         | 1 << 1    | 0x0002                | 0000 0000 0000 0010  |
```

pirgeo:
I like the idea. I am, however, worried a little bit about metric aggregations, since attributes are part of the metric identity. Dropping them can require re-aggregation of data (e.g. https://github.com/open-telemetry/opentelemetry-specification/issues/1297). I am not sure how easy or hard that is to add to existing metrics SDKs or if that would require a lot of additional processing on the collector side (@jmacd, @reyang, @jsuereth, do you have thoughts on this?)

MovieStoreGuy:
@pirgeo , I think you can get away with this without the need to re-aggregate.

You can follow similar principle of MapReduce to get around it, meaning:
- Filter the data (Resource or Attributes)
- Update any fields or resources (in the event of rewrite attributes or resources)
- Aggregate and combine classification

In short, I think this could simply be a decorator on the exporter or considered an optional extension.

MovieStoreGuy:
It looks like a simplified version of this has been started here https://github.com/open-telemetry/oteps/pull/100 where this proposal incapsulates all data classifications. 

Perhaps that we should join these efforts?

text/0187-data-classification.md
```diff
@@ -0,0 +1,323 @@
...
+
+There was a conversation that was related here on attributes that should be subject to data regulations Proposal: Add Sensitive Data LabelsOPEN , however, this approach would superseded what is mentioned there and be applied more broadly to the project.
+
+The alternatives that were considered are:
+
+- The Semantic Convention defines classification for known attributes
+  - this was no considered heavily since it misses USERS (developers) whom are using the SDK natively and then would need to manage their own Semantic Convention that would also need to be understood by the down stream processors and vendors
+    - Does not account for version drift with a breaking change involved
+    - Does a processor need to know all versions?
+    - Does that mean values can not be changed in the semantic convention?
```
kenfinnigan, line 289: Another downside of doing it in semantic conventions goes to your point above that different organizations will have potentially different classifications. Reaching a common agreement on classification for the semantic conventions would require lengthy discussion, with a good chance of many attributes not reaching a consensus decision on a classification

text/0187-data-classification.md
```diff
@@ -0,0 +1,323 @@
...
+
+Updating the SDK to include the convention of data classification would mean that users can extend their current telemetry via an opt in process to enable enhanced controls over the telemetry data without impacting the developer experience. Instrumentation authors are not limited by what attributes they can use when developing their telemetry and not a hindrance for existing standards within Open Telemetry. Organisations can define what resource classifications are allowed to be processed by an observability suite, what is subject to additional processing, and understand the types of data being sent. For example, all high cardinality data can be sent to on prem short term storage and then forwarded to a vendor with high cardinality attributes removed so that it can be viewed over a longer time period to show trends.
+
+Having the ability to set data classification on resources means that:
+
+Known high cardinality attributes can be filtered out to reduce the cost of Observability suites.
+
+Attributes containing User Generated Content, Personal Data or sensitive data can be rerouted into a slower processing queue to redact, anonymise, or drop resource values.
```
jpkrohling, line 15:
```suggestion
Attributes containing User Generated Content (UGC), Personal Data (PD) or sensitive data can be rerouted into a slower processing queue to redact, anonymise, or drop resource values.
```

text/0187-data-classification.md
```diff
@@ -0,0 +1,323 @@
...
+
+Associate resource telemetry to Service Level Objective / Service Level Indicator (SLO/SLI) so that observability systems can automatically identify data that should be alerted on.  
+
+Ensure that sparse data is not sampled out by mistake
+
+This approach can also be extended by vendors to enable more advanced features in their tool such as high resolution, short retention, data residency, or an endless list of features that can be adopted.
+
+## Explanation
+
+A service that adopts using resource classifications allows the open telemetry exporters to route data, remove problematic attributes, or filter values being sent to downstream systems. Using classifications simplify the existing processes since it does not use a lookup table that needs to be maintained, regular expressions needing to be maintained, required user intervention of what is acceptable or not, or needing to validate the entire resource object thus meaning no performance impact.
```
jpkrohling, line 25:
```suggestion
A service that adopts using resource classifications allows the OpenTelemetry exporters to route data, remove problematic attributes, or filter values being sent to downstream systems. Using classifications simplify the existing processes since it does not use a lookup table that needs to be maintained, regular expressions needing to be maintained, required user intervention of what is acceptable or not, or needing to validate the entire resource object thus meaning no performance impact.  
```

text/0187-data-classification.md
```diff
@@ -0,0 +1,323 @@
...
+
+Updating the SDK to include the convention of data classification would mean that users can extend their current telemetry via an opt in process to enable enhanced controls over the telemetry data without impacting the developer experience. Instrumentation authors are not limited by what attributes they can use when developing their telemetry and not a hindrance for existing standards within Open Telemetry. Organisations can define what resource classifications are allowed to be processed by an observability suite, what is subject to additional processing, and understand the types of data being sent. For example, all high cardinality data can be sent to on prem short term storage and then forwarded to a vendor with high cardinality attributes removed so that it can be viewed over a longer time period to show trends.
```
jpkrohling, line 9:
```suggestion
Updating the SDK to include the convention of data classification would mean that users can extend their current telemetry via an opt in process to enable enhanced controls over the telemetry data without impacting the developer experience. Instrumentation authors are not limited by what attributes they can use when developing their telemetry and not a hindrance for existing standards within OpenTelemetry. Organisations can define what resource classifications are allowed to be processed by an observability suite, what is subject to additional processing, and understand the types of data being sent. For example, all high cardinality data can be sent to on prem short term storage and then forwarded to a vendor with high cardinality attributes removed so that it can be viewed over a longer time period to show trends.
```

text/0187-data-classification.md
```diff
@@ -0,0 +1,323 @@
...
+
+Some ideas of future possibilities have been eluded to throughout the proposal, the potential changes that would be of great impact are:
+
+- Vendor based classification controls
+- Vendor neutral data protection mechanisms
+- Enforcing Data Regulations on telemetry
+- Classification based routing
+  - An exporter can only with certain set of classifications
+
+Thinking more broadly, it would mean that Open Telemetry could easily extend to support analytical based telemetry on user interactions while adhering to data regulation or organisational policies.
```
jpkrohling, line 323:
```suggestion
Thinking more broadly, it would mean that OpenTelemetry could easily extend to support analytical based telemetry on user interactions while adhering to data regulation or organisational policies.
```

text/0187-data-classification.md
```diff
@@ -0,0 +1,323 @@
...
+  }))
+}
+```
+
+Once classification have been added to the attribute keys, a tracing sampler can
+
+```go
+func (ts TailBasedSampler) Sample(spans trace.Spans) (export trace.Spans) {
+  for _, s := range spans {
+    if classification.Match(s.Classification(), classification.Ephermal) {
```
jpkrohling, line 73:
```suggestion
    if classification.Match(s.Classification(), classification.Ephemeral) {
```

text/0187-data-classification.md
```diff
@@ -0,0 +1,323 @@
...
+      export.Append(s)
+      // Always retain spans that contain ephermal attributes
```
jpkrohling, line 75:
```suggestion
      // Always retain spans that contain ephemeral attributes
```

text/0187-data-classification.md
```diff
@@ -0,0 +1,323 @@
...
+| High Cardinality                   | The value is an unbounded set                                                                         | 1 << 1    | 0x0002                | 0000 0000 0000 0010  |
+| Sensitive Value                    | The value MAY contain information that requires sanitisation                                          | 1 << 2    | 0x0004                | 0000 0000 0000 0100  |
+| Personal Identifiable Information  | The value DOES contain Personal Identifiable Information                                              | 1 << 3    | 0x0008                | 0000 0000 0000 1000  |
+| User Generated Content             | The value DOES contain User Generated Content                                                         | 1 << 4    | 0x0010                | 0000 0000 0001 0000  |
+| Service Level Objective            | The value is used to track a service level objective                                                  | 1 << 5    | 0x0020                | 0000 0000 0010 0000  |
```
jpkrohling, line 188: I'm not sure I would include this one, at least not for the first iteration. My impression is that this wouldn't have the same audience as the other values.

jpkrohling (commented):
Despite this being one more thing people will end up learning in order to do instrumentation, I like this proposal, especially as it's optional for current and future users.

text/0187-data-classification.md
MovieStoreGuy: For Service Level Object? Or classifications themselves?

jpkrohling: I mean only the SLO classification.

tedsuo:
@MovieStoreGuy we are marking OTEPs which are no longer being updated as `stale`. We will close this PR in one week. If you wish to resurrect this PR, that is fine.

MovieStoreGuy:
No worries, I don't have much bandwidth to work on this now but I would love to pick this up at some point.

PR #195, opened 2022-01-27
Environment Variable Resource Detection in Kubernetes

dashpole:
This OTEP proposes using a consistent set of environment variables to detect resource attributes available through the Kubernetes downward API.  The Kubernetes downward API is designed to serve this function and is the most consistent way to detect Kubernetes resource attributes.  This will improve user experience, and enable re-use of the same Kubernetes configuration across languages and when using the collector.

### DIFF
--- a/text/0195-kubernetes-resource-detection.md
+++ b/text/0195-kubernetes-resource-detection.md
@@ -16,10 +16,10 @@ Method 3 is already implemented today as the `k8sattributesprocessor`, but requi
 
 Not all attributes in the [k8s semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/resource/k8s.yaml) are supported by the Kubernetes downward API.  The OpenTelemetry attributes that could be supported ([api reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#envvarsource-v1-core)) are:
 
-* `kubernetes.pod.name`
-* `kubernetes.pod.uid`
-* `kubernetes.namespace.name`
-* `kubernetes.node.name`
+* `k8s.pod.name`
+* `k8s.pod.uid`
+* `k8s.namespace.name`
+* `k8s.node.name`
 
 The current state of detecting downward-api-based resources differs, depending on where detection is being done:
 
@@ -82,10 +82,10 @@ SDK's should add a Kubernetes environment variable-based resource detector.  The
 
 | Environment Variable | Semantic Convention |
 | ---- | ---- |
-| KUBERNETES_POD_NAME | kubernetes.pod.name |
-| KUBERNETES_POD_UID | kubernetes.pod.uid |
-| KUBERNETES_NAMESPACE_NAME | kubernetes.namespace.name |
-| KUBERNETES_NODE_NAME | kubernetes.node.name |
+| K8s_POD_NAME | k8s.pod.name |
+| K8s_POD_UID | k8s.pod.uid |
+| K8s_NAMESPACE_NAME | k8s.namespace.name |
+| K8s_NODE_NAME | k8s.node.name |
 
 Detection of the above semantic conventions using environment variables should not be done inside vendor-specific (e.g. GKE-specific) resource detectors.
 
@@ -95,7 +95,7 @@ Detection of the above semantic conventions using environment variables should n
 
 Similar to `OTEL_RESOURCE_ATTRIBUTES`, we could require the proposed environment variables to always be detected.  While this would further improve the ease of use, [Resource semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md) (including Kubernetes) are still experimental, and the [SDK environment variable specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration) are stable. Additionally, Kubernetes is [explicitly called out](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#detecting-resource-information-from-the-environment) as a custom resource detector.
 
-### Alternative: Using HOSTNAME for kubernetes.pod.name
+### Alternative: Using HOSTNAME for k8s.pod.name
 
 Many kubernetes detectors currently use `HOSTNAME` environment variable, which defaults to the Pod name. However, the `HOSTNAME` can be [modified in a few ways](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-hostname-and-subdomain-fields) in the pod spec. Kubernetes resource detectors may fall back to detecting the pod name using `HOSTNAME` if `KUBERNETES_POD_NAME` is not available, but this may cause user confusion in some cases.
 
--- a/text/0195-kubernetes-resource-detection.md
+++ b/text/0195-kubernetes-resource-detection.md
@@ -1,0 +1,108 @@
+# Environment Variable Resource Detection in Kubernetes
+
+This OTEP proposes to standardize environment variables used to detect resources exposed through the [Kubernetes downward API](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/) to provide a consistent experience on Kubernetes across languages and the collector.
+
+There is a mismatch between the format otel expects resources to be specified in for environment variables (a single, comma-deliniated list of key-value pairs in `OTEL_RESOURCE_ATTRIBUTES`), and the format Kubernetes exposes resource information in (one environment-variable for each attribute).  This makes correct resource detection in Kubernetes difficult, and the lack of specification has led to differing implementations across language SDKs and the collector.
+
+## Background
+
+Adding Kubernetes resources attributes to telemetry can be done one of three ways, depending on your telemetry collection pipeline:
+
+1. The application detects its own resource attributes from environment variables
+2. A "sidecar" collector container detects its own resource attributes from environment variables
+3. A centralized collector detects resource attributes for many Kubernetes workloads by querying the kube-apiserver.
+
+Method 3 is already implemented today as the `k8sattributesprocessor`, but requires running a collector, and adds additional load on the kube-apiserver. Methods 1 and 2 require substantial configuration today, which differs based on the language used, or if using a sidecar collector.
+
+Not all attributes in the [k8s semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/resource/k8s.yaml) are supported by the Kubernetes downward API.  The OpenTelemetry attributes that could be supported ([api reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#envvarsource-v1-core)) are:
+
+* `kubernetes.pod.name`
+* `kubernetes.pod.uid`
+* `kubernetes.namespace.name`
+* `kubernetes.node.name`
+
+The current state of detecting downward-api-based resources differs, depending on where detection is being done:
+
+* The collector's [`resourcedetectionprocessor`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor#resource-detection-processor) only accepts `OTEL_RESOURCE_ATTRIBUTES` for env-var-based resource detection, which does not work with the downward API.
+* The collector's [`resourceprocessor`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourceprocessor#resource-processor) can be used in conjunction with [environment variable expansion](https://opentelemetry.io/docs/collector/configuration/#configuration-environment-variables) as described in [this issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/570#issuecomment-1006885133).
+* The [golang GKE detector](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/26b22d30acb587d893b3a0fd3c8dd7895f79b352/detectors/gcp/gke.go) uses `NAMESPACE`, and `HOSTNAME` (pod name)
+* The [ruby GKE detector](https://github.com/open-telemetry/opentelemetry-ruby/blob/8fe083cdf00bfe866ae89ea8f5fc06fd981ebf6a/resource_detectors/lib/opentelemetry/resource/detectors/google_cloud_platform.rb) uses `GKE_NAMESPACE_ID`, and `HOSTNAME` (pod name)
+* The [javascript env var detector](https://github.com/open-telemetry/opentelemetry-js/blob/feea5167c15c41f0aeedc60959e36c18315c7ede/packages/opentelemetry-core/src/utils/environment.ts) uses `NAMESPACE`, and `HOSTNAME` (pod name)
+* Others not listed do not attempt to detect Kubernetes resource information through environment variables.
+
+## Motivation
+
+A common set of environment variables across languages and the collector would enable a single pod `env` configuration to work correctly across languages and collector sidecars. At minimum, this would simplify our documentation and improve day-0 user experience.  But the Kubernetes API's extension points and the surrounding ecosystem of tools allow cluster administrators to go further: **They can configure these environment variables as cluster-level policy, and make OpenTelemetry resource detection "just work" on Kubernetes.**
+
+## Explanation
+
+The snippet of environment variables added the pod spec would look like:
+
+```yaml
+env:
+- name: KUBERNETES_POD_NAME
+   valueFrom:
+     fieldRef:
+       fieldPath: metadata.name
+- name: KUBERNETES_POD_UID
+   valueFrom:
+     fieldRef:
+       fieldPath: metadata.uid
+- name: KUBERNETES_NAMESPACE_NAME
+  valueFrom:
+    fieldRef:
+      fieldPath: metadata.namespace
+- name: KUBERNETES_NODE_NAME
+   valueFrom:
+     fieldRef:
+       fieldPath: spec.nodeName
+```
+
+### User Journey: Getting Started on Kubernetes
+
+As a user new to Kubernetes and/or OpenTelemetry, I can copy-and-paste the above into my pod specification to make resource discovery work for me. I don't need to set up any complex infrastructure, know about the downward API or OpenTelemetry resource detection, or the specifics of the lauguage SDK I'm using to make it work.
+
+### User Journey: Admission Webhooks
+
+There are a number of existing Kubernetes policy engines that can enforce a consistent set of environment variables on pods (in addition to doing much more), such as [kyverno](https://github.com/kyverno/kyverno), [podpreset](https://github.com/redhat-cop/podpreset-webhook), [open policy agent](https://www.openpolicyagent.org/), etc., but they all use the same underlying Kubernetes mechanism: [Admisison Webhooks](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#admission-webhooks).
+
+As a cluster administrator, I can use my preferred policy agent or my own mutating admission webhook to add the OpenTelemetry-specified set of environment variables to all pods subsequently created in the cluster. Application developers deploying to the cluster just need to use the OpenTelemetry SDK with the Kubernetes resource detector.
+
+### User Journey: Kubernetes Config Management
+
+Kubernetes templating tools, such as [Helm templates](https://helm.sh/docs/chart_template_guide/), allow users to easily re-use snippets of yaml in many places in their configuration.  A common set of environment variables would users to define the OpenTelemetry env var config once, and use it across all of their OpenTelemetry deployments.
+
+**Note:** This pattern unfortunately isn't easy to use with kubernetes' built-in patching tools, [kustomize](https://kustomize.io/), because it doesn't [currently](https://github.com/kubernetes-sigs/kustomize/issues/1493) support patching all containers across deployments with a single patch.
+
+## Internal details
+
+Add an experimental k8s.md specification under specification/resource, which specifies how kubernetes resource detectors should detect the kubernetes semantic conventions:
+
+SDK's should add a Kubernetes environment variable-based resource detector.  The Kubernetes resource detector and the collector's `resourcedetectionprocessor` should support the following environment variables, and map them to the following semantic conventions:
+
+| Environment Variable | Semantic Convention |
+| ---- | ---- |
+| KUBERNETES_POD_NAME | kubernetes.pod.name |
+| ---- | ---- |
+| KUBERNETES_POD_UID | kubernetes.pod.uid |
+| ---- | ---- |
+| KUBERNETES_NAMESPACE_NAME | kubernetes.namespace.name |
+| ---- | ---- |
+| KUBERNETES_NODE_NAME | kubernetes.node.name |
+| ---- | ---- |
+
+Detection of the above semantic conventions using environment variables should not be done inside vendor-specific (e.g. GKE-specific) resource detectors.
+
+## Prior art and alternatives
+
+### Alternative: Detection required by SDKs
+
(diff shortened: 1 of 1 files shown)


text/0195-kubernetes-resource-detection.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Alternative: Using HOSTNAME for kubernetes.pod.name
+
+Many kubernetes detectors currently use `HOSTNAME` environment variable, which defaults to the Pod name. However, the `HOSTNAME` can be [modified in a few ways](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-hostname-and-subdomain-fields) in the pod spec. Kubernetes resource detectors may fall back to detecting the pod name using `HOSTNAME` if `KUBERNETES_POD_NAME` is not available, but this may cause user confusion in some cases.
+
+## Future possibilities
+
+As a community, we could consider introducing an OpenTelemetry mutating admisison webhook. Similar to the [Istio admission webhook](https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#automatic-sidecar-injection), it could potentially handle collector sidecar injection, downward API configuration, and exporter configuration (also through environment variables) in a central place.
```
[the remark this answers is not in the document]
dashpole: Duh, right :).  I've updated this section to suggest extending the OpenTelemetry Operator instead of developing a new webhook.

dashpole:
cc @Aneurysm9 @pmm-sumo @dmitryax
who might be interested

Aneurysm9 (approved):

text/0195-kubernetes-resource-detection.md
```diff
@@ -0,0 +1,104 @@
...
+Add an experimental k8s.md specification under specification/resource, which specifies how kubernetes resource detectors should detect the kubernetes semantic conventions:
+
+SDK's should add a Kubernetes environment variable-based resource detector.  The Kubernetes resource detector and the collector's `resourcedetectionprocessor` should support the following environment variables, and map them to the following semantic conventions:
+
+| Environment Variable | Semantic Convention |
+| ---- | ---- |
+| KUBERNETES_POD_NAME | kubernetes.pod.name |
+| KUBERNETES_POD_UID | kubernetes.pod.uid |
+| KUBERNETES_NAMESPACE_NAME | kubernetes.namespace.name |
+| KUBERNETES_NODE_NAME | kubernetes.node.name |
```
dmitryax, line 88: Why do we use `kubernetes.*` namespace for attributes while `k8s.*` is defined in the spec?

dmitryax (commented):
Thanks for putting it together!

anuraaga (commented):
Just for reference we currently use `OTEL_RESOURCE_ATTRIBUTES` in the operator

https://github.com/open-telemetry/opentelemetry-operator/blob/a94a13888165ec688a5f83d4e444e9fbc934fca2/pkg/instrumentation/sdk.go#L144

If there were different recognized variables we would probably switch to it though, as far as I can tell it's not practical to set that purely with YAML so having dedicated env vars seems worth it.

pmm-sumo:
I think that the downward API can currently also expose labels and annotations. While it's not included in the semantic conventions, `k8sattributesprocessor` can currently [extract those into resource attributes](https://pkg.go.dev/github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sattributesprocessor#ExtractConfig) (and this information is frequently very useful).

I am wondering if there's a space to include such capability via env resource detection. Currently, the labels/annotations are filtered by `k8sattributesprocessor` but perhaps we could include those in a separate namespace and let filtering happen (if needed) using e.g. [resourceprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourceprocessor)

text/0195-kubernetes-resource-detection.md
dashpole: fixed!

text/0195-kubernetes-resource-detection.md
```diff
@@ -0,0 +1,104 @@
...
+Add an experimental k8s.md specification under specification/resource, which specifies how kubernetes resource detectors should detect the kubernetes semantic conventions:
+
+SDK's should add a Kubernetes environment variable-based resource detector.  The Kubernetes resource detector and the collector's `resourcedetectionprocessor` should support the following environment variables, and map them to the following semantic conventions:
+
+| Environment Variable | Semantic Convention |
+| ---- | ---- |
+| K8s_POD_NAME | k8s.pod.name |
+| K8s_POD_UID | k8s.pod.uid |
+| K8s_NAMESPACE_NAME | k8s.namespace.name |
+| K8s_NODE_NAME | k8s.node.name |
```
dmitryax, line 88: We should probably use al capital letters in env vars? also other mentions of these env vars need to be updated  in this doc.

```suggestion
| K8S_POD_NAME | k8s.pod.name |
| K8S_POD_UID | k8s.pod.uid |
| K8S_NAMESPACE_NAME | k8s.namespace.name |
| K8S_NODE_NAME | k8s.node.name |
```

dashpole:
> If there were different recognized variables we would probably switch to it though, as far as I can tell it's not practical to set that purely with YAML so having dedicated env vars seems worth it.

Oh, wow.  I just realized you can [define dependent env vars](https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/#define-an-environment-dependent-variable-for-a-container), which might be a better solution than i'm proposing.

I could add:
```yaml
- name: OTEL_RESOURCE_ATTRIBUTES
  value: k8s.pod.name=$(K8S_POD_NAME),k8s.pod.uid=$(K8S_POD_UID),k8s.namespace.name=$(K8S_NAMESPACE_NAME),k8s.node.name=$(K8S_NODE_NAME)
```

To achieve the goals of this proposal, without requiring any detectors to be installed.

mat-rumian:
@dashpole I would like to share my opinion. What do you think about configuring additional env var holding resource attributes related to k8s in separated var like:

```yaml
- name: OTEL_RESOURCE_ATTRIBUTES_K8S
  value: k8s.pod.name=$(K8S_POD_NAME),k8s.pod.uid=$(K8S_POD_UID),k8s.namespace.name=$(K8S_NAMESPACE_NAME),k8s.node.name=$(K8S_NODE_NAME)
```

and use it this way:

```yaml
- name: OTEL_RESOURCE_ATTRIBUTES
  value: $(OTEL_RESOURCE_ATTRIBUTES_K8S),other=attribs...
```

I think it will be simpler, transparent and easier to handle and modify.

dashpole:
> I think that the downward API can currently also expose labels and annotations. ... I am wondering if there's a space to include such capability via env resource detection.

If we went with the current proposal (explicitly defined env vars that are detected with a kubernetes detector), we could also support `K8S_LABELS` and `K8S_ATTRIBUTES`.  The kubernetes detector could accept a configuration option to allow elevating parsed labels or attributes to resource attributes.

If we used dependent environment variables, as we are discussing above (no detector needed), users could just fetch labels and attributes themselves:
```
- name: MY_CUSTOM_ATTRIBUTE
   valueFrom:
     fieldRef:
       fieldPath: metadata.labels['my-custom-attribute']
- name: OTEL_RESOURCE_ATTRIBUTES
  value: other.semantic.convention=$(MY_CUSTOM_ATTRIBUTE)
```

WDYT?

dashpole:
> What do you think about configuring additional env var holding resource attributes related to k8s in separated var like ...

That seems like a perfectly reasonable thing to do. In terms of this proposal, I think the core question is whether or not we want to specify a new resource detector.  I'd prefer to keep the yaml simpler for readability, if thats acceptable.

text/0195-kubernetes-resource-detection.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Alternative: Using HOSTNAME for k8s.pod.name
+
+Many kubernetes detectors currently use `HOSTNAME` environment variable, which defaults to the Pod name. However, the `HOSTNAME` can be [modified in a few ways](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-hostname-and-subdomain-fields) in the pod spec. Kubernetes resource detectors may fall back to detecting the pod name using `HOSTNAME` if `K8S_POD_NAME` is not available, but this may cause user confusion in some cases.
+
+### Alternative: Using Dependent Environment Variables
```
dashpole, line 102: @tigrannajaryan  
@jsuereth mentioned to me that this approach may not interact well with schemas, since OTEL_RESOURCE_ATTRIBUTES is merged into resource attributes without any notion of version according to the specification.  Using this approach would seem to remove the benefits of https://github.com/open-telemetry/oteps/pull/161.  Is there a way to make OTEL_RESOURCE_ATTRIBUTES produce resource attributes associated with a particular version of the semantic conventions?

text/0195-kubernetes-resource-detection.md
```diff
@@ -0,0 +1,113 @@
...
+* The [javascript env var detector](https://github.com/open-telemetry/opentelemetry-js/blob/feea5167c15c41f0aeedc60959e36c18315c7ede/packages/opentelemetry-core/src/utils/environment.ts) uses `NAMESPACE`, and `HOSTNAME` (pod name)
+* Others not listed do not attempt to detect Kubernetes resource information through environment variables.
+
+## Motivation
+
+A common set of environment variables across languages and the collector would enable a single pod `env` configuration to work correctly across languages and collector sidecars. At minimum, this would simplify our documentation and improve day-0 user experience.  But the Kubernetes API's extension points and the surrounding ecosystem of tools allow cluster administrators to go further: **They can configure these environment variables as cluster-level policy, and make OpenTelemetry resource detection "just work" on Kubernetes.**
+
+## Explanation
+
+The snippet of environment variables added the pod spec would look like:
```
aabmass, line 39:
```suggestion
The snippet of environment variables added to the pod spec would look like:
```

text/0195-kubernetes-resource-detection.md
```diff
@@ -0,0 +1,113 @@
...
+
+**Note:** This pattern unfortunately isn't easy to use with kubernetes' built-in patching tools, [kustomize](https://kustomize.io/), because it doesn't [currently](https://github.com/kubernetes-sigs/kustomize/issues/1493) support patching all containers across deployments with a single patch.
+
+## Internal details
+
+Add an experimental k8s.md specification under specification/resource, which specifies how kubernetes resource detectors should detect the kubernetes semantic conventions:
+
+SDK's should add a Kubernetes environment variable-based resource detector.  The Kubernetes resource detector and the collector's `resourcedetectionprocessor` should support the following environment variables, and map them to the following semantic conventions:
+
+| Environment Variable | Semantic Convention |
```
aabmass, line 83: I wonder if these environment variable names will run into conflicts. Should we prefix these with `OTEL_`?

text/0195-kubernetes-resource-detection.md
aabmass, line 102: Would this approach conflict with "Kubernetes is [explicitly called out](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#detecting-resource-information-from-the-environment) as a custom resource detector" above?

I wonder if having infrastructure set `OTEL_RESOURCE_ATTRIBUTES` as a means of resource detection is generally a good pattern. I know `OTEL_RESOURCE_ATTRIBUTES` is safe to self-concatenate, but it would be easy to accidentally forget to do so and override a previously set value. @mat-rumian's suggestion would improve this, but then the detection becomes less automatic.

text/0195-kubernetes-resource-detection.md
```diff
@@ -0,0 +1,113 @@
...
+
+Adding Kubernetes resources attributes to telemetry can be done one of three ways, depending on your telemetry collection pipeline:
+
+1. The application detects its own resource attributes from environment variables
+2. A "sidecar" collector container detects its own resource attributes from environment variables
+3. A centralized collector detects resource attributes for many Kubernetes workloads by querying the kube-apiserver.
+
+Method 3 is already implemented today as the `k8sattributesprocessor`, but requires running a collector, and adds additional load on the kube-apiserver. Methods 1 and 2 require substantial configuration today, which differs based on the language used, or if using a sidecar collector.
+
+Not all attributes in the [k8s semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/resource/k8s.yaml) are supported by the Kubernetes downward API.  The OpenTelemetry attributes that could be supported ([api reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#envvarsource-v1-core)) are:
```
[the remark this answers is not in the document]
punya: Although the container name isn't accessible from the downward API, it's still possible to

* standardize on an environment variable approach for injecting it
* inject this as an environment variable in a uniform way (e.g. mutating admission webhook)

Since this OTEP is scoped to "Env variables on Kubernetes" not just "Env variables via downward API", I think it's reasonable to cover it here.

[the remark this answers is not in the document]
dashpole: There are a bunch of fields I don't cover here, since I'm trying to stick to fields available through the downward API. 
I'm going to talk to some k8s folks to see what it would take to have some of those missing conventions added to the downward API's list of allowed fieldPaths.  Today, you can just set OTEL_RESOURCE_ATTRIBUTES to: `k8s.container.name=my-container-name`, but that obviously isn't a great solution.

text/0195-kubernetes-resource-detection.md
```diff
@@ -0,0 +1,113 @@
...
+
+Detection of the above semantic conventions using environment variables should not be done inside vendor-specific (e.g. GKE-specific) resource detectors.
+
+## Prior art and alternatives
+
+### Alternative: Detection required by SDKs
+
+Similar to `OTEL_RESOURCE_ATTRIBUTES`, we could require the proposed environment variables to always be detected.  While this would further improve the ease of use, [Resource semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md) (including Kubernetes) are still experimental, and the [SDK environment variable specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration) are stable. Additionally, Kubernetes is [explicitly called out](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#detecting-resource-information-from-the-environment) as a custom resource detector.
+
+### Alternative: Using HOSTNAME for k8s.pod.name
```
punya, line 98: I see some older issues ([e.g.](https://github.com/kubernetes/kubernetes/issues/4825)) that suggest that `HOSTNAME` is truncated to 64 bytes whereas the Pod name can be up to 253 bytes long. Should we add that as another argument against this alternative?

dashpole: added as a downside of this alternative.

text/0195-kubernetes-resource-detection.md
dashpole: If we adopted this alternative, it would probably change that phrasing in the spec, which I suspect intends to use kubernetes as an example of a platform, rather than prescribe how k8s detection should be done.

I agree using the single `OTEL_RESOURCE_ATTRIBUTES` is easy to mess up, and hard to do consistently.  I'll call that out.

text/0195-kubernetes-resource-detection.md
dashpole: I don't have a strong opinion.  Is there precedent from other detectors?

text/0195-kubernetes-resource-detection.md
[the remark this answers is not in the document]
jpkrohling: Note that it might be tricky to detect the container name in a sidecar: either it will inject the sidecar's name and not the workload's, or it will have to try to determine which container contains the actual workload. While this is possible in most cases, there are cases where pods have multiple sidecars (3+ containers in the pod), so that one sidecar can't reliably tell which of the remaining two is the actual workload.

text/0195-kubernetes-resource-detection.md
aabmass: I don't think there is a precedent. Since SDK resource detectors would need to interpret them, would you consider these SDK environment variables?

dmitryax: Likely an instrumented application would need these vars from downward API as well, but it's very unlikely that they would expect any different value. So conflict here is rather useful, otherwise they would need to repeat all these definitions with `OTEL_` prefix and without.

The question here is whether we want consider and support an application that uses `K8S_POD_NAME` to represent a different value like name of deployment name or something else. I'd lean towards not supporting these weird use cases.

If we want to read these env vars from instrumentation libraries as is, we can probably support both prefixed (`OTEL_K8S_POD_NAME`) and not prefixed (`K8S_POD_NAME`) just to not block the use case provided above. If we still going to use only `OTEL_RESOURCE_ATTRIBUTES` in instrumentation libraries and compose it like `k8s.pod.name=$(K8S_POD_NAME),k8s.pod.uid=$(K8S_POD_UID)`, I think we shouldn't use any prefix, users always can rename them if really needed.

text/0195-kubernetes-resource-detection.md
```diff
@@ -0,0 +1,120 @@
...
+
+```yaml
+- name: OTEL_RESOURCE_ATTRIBUTES
+  value: k8s.pod.name=$(K8S_POD_NAME),k8s.pod.uid=$(K8S_POD_UID),k8s.namespace.name=$(K8S_NAMESPACE_NAME),k8s.node.name=$(K8S_NODE_NAME)
+```
+
+This approach has a few drawbacks:
+
+* `OTEL_RESOURCE_ATTRIBUTES` may need to include additional attibutes other than those used for kubernetes. This makes it harder to apply the same environment variables across all pods.
+* `OTEL_RESOURCE_ATTRIBUES` doesn't have a way to specify which version of the semantic conventions it should apply to.
```
dmitryax, line 116:
```suggestion
* `OTEL_RESOURCE_ATTRIBUTES` doesn't have a way to specify which version of the semantic conventions it should apply to.
```

text/0195-kubernetes-resource-detection.md
```diff
@@ -0,0 +1,120 @@
...
+Not all attributes in the [k8s semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/resource/k8s.yaml) are supported by the Kubernetes downward API.  The OpenTelemetry attributes that could be supported ([api reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#envvarsource-v1-core)) are:
+
+* `k8s.pod.name`
+* `k8s.pod.uid`
+* `k8s.namespace.name`
+* `k8s.node.name`
+
+The current state of detecting downward-api-based resources differs, depending on where detection is being done:
+
+* The collector's [`resourcedetectionprocessor`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor#resource-detection-processor) only accepts `OTEL_RESOURCE_ATTRIBUTES` for env-var-based resource detection, which does not work with the downward API.
```
pavolloffay, line 26: `OTEL_RESOURCE_ATTRIBUTES` can be set to `$(VAR_NAME)` which can substituted to the value from the downward API e.g.:

`OTEL_RESOURCE_ATTRIBUTES=k8s.container.name=myapp,k8s.pod.name=$(POD_NAME)`


See https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#envvar-v1-core

pavolloffay (commented):
This proposal is related to https://github.com/open-telemetry/opentelemetry-specification/issues/2135 which proposes support for `OTEL_RESOURCE_ATTRIBUTES_*` env var. These env vars are added to OTEL_RESOURCE_ATTRIBUTES`

e.g. 
```
OTEL_RESOURCE_ATTRIBUTES_K8S_POD_NAME=myapp
OTEL_RESOURCE_ATTRIBUTES=foo=bar
```

the final value of `OTEL_RESOURCE_ATTRIBUTES` would be `foo=bar,k8s_pod_name=myapp`

text/0195-kubernetes-resource-detection.md
dashpole: Updated the wording of this section, as it was outdated.  See the "alternative: using dependent environment variables" section below.

text/0195-kubernetes-resource-detection.md
bogdandrutu: Not resolved?

bogdandrutu:
@dashpole very nice OTEP, please consider to also refer to that proposal of supporting `OTEL_RESOURCE_ATTRIBUTES_*`

text/0195-kubernetes-resource-detection.md
dashpole: fixed

dashpole:
> please consider to also refer to that proposal of supporting OTEL_RESOURCE_ATTRIBUTES_*

Done.

text/0195-kubernetes-resource-detection.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+This approach has a few drawbacks:
+
+* `OTEL_RESOURCE_ATTRIBUTES` may need to include additional attibutes other than those used for kubernetes. This makes it harder to apply the same environment variables across all pods.
+* `OTEL_RESOURCE_ATTRIBUTES` doesn't have a way to specify the schema version it should apply to. This would conflict with the ability of [telemetry schemas](https://github.com/open-telemetry/oteps/blob/main/text/0152-telemetry-schemas.md#solution-summary) to convert from older versions of semantic conventions to newer ones.
+
+### Alternative: Support multiple OTEL_RESOURCE_ATTRIBUTES_* variables
+
+Proposed in [open-telemetry/opentelemetry-specification#2135](https://github.com/open-telemetry/opentelemetry-specification/issues/2135), SDKs would detect all environment variables starting with `OTEL_RESOURCE_ATTRIBUTES_`, and would add these to the detected resource attributes. The `OTEL_RESOURCE_ATTRIBUTES` prefix is trimmed, letters are lower-cased, and `_` are replaced with `.`.  For example, `OTEL_RESOURCE_ATTRIBUTES_K8S_NODE_NAME=foo` would become `k8s.node.name=foo`.
```
mx-psi, line 120: This, in general, is ambiguous; take `cloud.availability_zone` defined [here](https://github.com/open-telemetry/opentelemetry-specification/blob/48ea76ccf06a027278e885e774b1371b91d6454d/specification/resource/semantic_conventions/cloud.md#cloud): env variable would presumably be `OTEL_RESOURCE_ATTRIBUTES_CLOUD_AVAILABILITY_ZONE` but the wording here would result in `cloud.availability.zone`. The main proposal doesn't suffer from this (the env variable <-> semantic convention is ad-hoc), but this alternative does suffer from it (e.g. would be a problem if we add a `_` on any of the relevant k8s semantic conventions)

dashpole: Thanks.  I added this to the list of drawbacks.

text/0195-kubernetes-resource-detection.md
dashpole: Thinking about this more, I think we should switch to OTEL_ prefixed env vars.  The reason isn't that someone might want opentelemetry to attach attributes that don't match your own identity.  This is mostly relevant for container name (e.g. sidecar), but I could also imagine it for pod or even namespace.

mx-psi (approved):

text/0195-kubernetes-resource-detection.md
dmitryax: @dashpole do you mean that `OTEL_K8S_CONTAINER_NAME` env var added to one container would mean something else other than a name of the container which it's added to, like name of an otel sidecar when added to an application container?

dashpole: Yes, thats what I mean.  container name isn't part of this proposal, but would be a likely future extension of this.

dmitryax: IMO in that case meaning of `OTEL_K8S_CONTAINER_NAME` becomes a bit confusing and inconsistent with other `OTEL_K8S_*` env vars. I, as a user, would expect `OTEL_K8S_CONTAINER_NAME` to have a value of the container where env var is defined, giving that other `OTEL_K8S_*` have values of other k8s objects where the env var is defined.

I think we should use another name for the sidecar use case, `OTEL_K8S_CONTAINER_NAME`

dashpole: I don't think multiple env vars for the same semantic convention would work, as resource detectors will ultimately have to choose one of the two to set in that resource attribute.

If we named it without a prefix, I'm not sure that would stop users from treating the env vars as a way to set attribute X in OTEL, even if we would prefer they use another mechanism in that scenario. We just risk introducing naming conflicts if we don't prefix.

jsuereth (approved):

jmacd (approved):

dashpole:
Thanks @seh, i've added that to the drawbacks of using dependent environment variables.

PR #201, opened 2022-04-25
Introduce Scope Attributes

tigrannajaryan:
There are a few reasons why adding Scope attributes is a good idea:
- There are 2 known use cases where Scope attributes can solve specific problems:
  - Add support for [Meter "short_name"](https://github.com/open-telemetry/opentelemetry-specification/pull/2422),
    represented as an attribute of Meter's Scope.
  - Add support for differentiating the type of data emitted from the scopes that belong
    to different data domains, e.g. profiling data emitted as log records or client-side
    data emitted as log records needs to be differentiated so that it can be easily
    routed and processed differently in the backends. We don't have a good way to handle
    this today. The type of the data can be recorded as an attribute Logger's Scope.
- It makes Scope consistent with the other primary data types: Resource, Span, Metric,
  LogRecord.

See additional [discussion here](https://github.com/open-telemetry/opentelemetry-specification/issues/2450).


bogdandrutu (approved):

yurishkuro (approved):

text/0201-scope-attributes.md
```diff
@@ -0,0 +1,138 @@
...
+There are a few reasons why adding Scope attributes is a good idea:
+
+- There are 2 known use cases where Scope attributes can solve specific problems:
+  - Add support for [Meter "short_name"](https://github.com/open-telemetry/opentelemetry-specification/pull/2422),
+    represented as an attribute of Meter's Scope.
+  - Add support for differentiating the type of data emitted from the scopes that belong
+    to different data domains, e.g. profiling data emitted as log records or client-side
+    data emitted as log records needs to be differentiated so that it can be easily
+    routed and processed differently in the backends. We don't have a good way to handle
+    this today. The type of the data can be recorded as an attribute Logger's Scope.
```
scheler, line 16: Shouldn't the domain identifying attribute be at the resource level, since it will be repeated for each instrumentation scope anyway? Not sure about the `profiling data emitted as log records`, but the client-side is a larger vertical and so every instrumentation scope emitted from the client-side will have it repeated. I guess this will require a different example to illustrate the point.

text/0201-scope-attributes.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+`Get LogEmitter` SDK call will be altered to the following:
+
+```
+Accepts the instrumentation scope name and optional version and attributes and
+returns a LogEmitter associated with the instrumentation scope.
+```
+
+Since the attributes are optional this is a backwards compatible change.
```
scheler, line 62: There might be value in making the attributes mutable, so that a common `state` can be recorded. The scope identifying fields are separate from attributes anyway.

text/0201-scope-attributes.md
tigrannajaryan: If I have an app from which I want to emit profiling data, client side events, and regular logs all represented using OTLP LogRecord, then I need to somehow different the 3 kind of LogRecord although they are all coming from the same Resource (same app).

text/0201-scope-attributes.md
tigrannajaryan: I am not sure I understand the mutability reference and common state can be recorded. Please elaborate.

alanwest (approved):

tedsuo:
Are there prototypes available for review?

text/0201-scope-attributes.md
```diff
@@ -0,0 +1,148 @@
...
+This change is backwards compatible from OTLP's interoperability perspective. Recipients
+of old OTLP versions will not see the Scope attributes and will ignore them, which we
+consider acceptable from interoperability perspective. This is aligned with our general
+stance on what happens when telemetry sources _add_ new data which old recipients
+don't understand: we expect the new data to be safely ignored.
+
+## Attribute Value Precedence
+
+If the same attribute is specified both at the Span/Metric/LogRecord and at the Scope
+then the attribute value at Span/Metric/LogRecord takes precedence.
```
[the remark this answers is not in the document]
tsloughter: The resource is separate from the attributes in the span. But so are the ones in the Scope, as shown above in the protobuf. So I'm not sure why there is a need for precedence at all? If precedence is given to the span/metric/logrecord over the attributes that are separately kept in Scope and potentially Resource then you end up with a situation of allowing the service name to be overridden at the span level.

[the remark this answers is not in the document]
arminru: @tsloughter The paragraph just below these lines explains that this is necessary for non-OTLP exporters that "flatten" attributes since non-OTLP formats and consumers are not aware of the hierarchy we have in OTel with Resource vs Span/Metric/LogRecord (vs InstrumentationScope).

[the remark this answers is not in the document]
tsloughter: Ah, of course, thanks. Yea, then needs to define that the Scope attributes take precedence over the Resource attributes, in my opinion --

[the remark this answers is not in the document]
tsloughter: Precedence for Resource attributes and Span attributes must be defined then, right? I went to quickly find it but haven't yet.

text/0201-scope-attributes.md
```diff
@@ -0,0 +1,148 @@
...
+```
+- `attributes` (optional): Specifies the instrumentation scope attributes to associate
+  with emitted telemetry.
+```
+
+`Get a Metter` API will be extended to add the following parameter:
+
+```
+- `attributes` (optional): Specifies the instrumentation scope attributes to associate
+  with emitted telemetry.
```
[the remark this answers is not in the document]
tigrannajaryan: My intuition was that it would be cheaper to specify the scope attribute once when obtaining the Tracer/Meter/LogEmitter (I may be wrong). The other benefit is that for the use cases I had in my mind I would want to specify the scope attributes once.

You are right though that there may be use cases where scope attributes may need to be specifying for every data point. An example of these may be the client-side session id that was discussed in the SIG meeting today. I think it largely depends on how exactly the client-side SIG intends to perform the instrumentation, where the session id is available and what is the most ergonomic way to set it on the scope.

I think we need feedback from multiple language maintainers to help shape this particular API. It may be that we need more than one way to specify the scope attributes (however, I would prefer to have only one if possible).

text/0201-scope-attributes.md
scheler: Something like a client side session-id, if we decide to use at the level of instrumentation scope, will be changing over time. Hence it will need this ability for the attributes to change.

tigrannajaryan: >Hence it will need this ability for the attributes to change.

That's one possible approach. Another approach is to get a new Scope every time you have a new session-id.

I don't know what's a better way. In this OTEP I thought it is going to be immutable, but I am open to re-considering.

text/0201-scope-attributes.md
scheler: Logs and Events, when both represented using LogRecord, will be distinguishable using the event.name attribute that exists only for events. So this idea does seem only like a mechanism for compression and avoiding repetition of attributes on each signal. If that's the intention, I am wondering why we don't want to take it one level higher above InstrumentationScope, maybe, by introducing another scope layer above InstrumentationScope if we want to have these attributes in resource. 

As an example, the `page.url` attribute representing the base web page will have to be present in all signals emitted while the end-user is on a certain page, however it is not going to be in resource as it doesn't identify the resource. So it is a good candidate to be a top-level scope for all signals, across all InstrumentationScopes.

text/0201-scope-attributes.md
```diff
@@ -0,0 +1,148 @@
...
+This OTEP adds attributes to the Scope.
+
+## Motivation
+
+There are a few reasons why adding Scope attributes is a good idea:
+
+- There are 2 known use cases where Scope attributes can solve specific problems:
+  - Add support for [Meter "short_name"](https://github.com/open-telemetry/opentelemetry-specification/pull/2422),
+    represented as an attribute of Meter's Scope.
+  - Add support for differentiating the type of data emitted from the scopes that belong
```
blumamir, line 12: Do we imagine moving some of the existing span attributes to be a scope attributes?
Things like:
- `db.system` (static for the instrumentation library)
- `messaging.system`
- `messaging.protocol`
- `messaging.protocol_version`

etc.

I guess it can be handy to sometimes store "Connection level attributes" in the scope, instead of repeating them for each new span, things like: `db.connection_string`, `net.peer.name` etc. The downside is that user will need to create a different instrumentation scope for each connection, with the same name version and schema but different attributes.

text/0201-scope-attributes.md
```diff
@@ -0,0 +1,148 @@
...
+client-side events and regular logs. The Scope attribute is used for multiplexing
+and routing the LogRecords:
+
+![LogRecord Multiplexing](img/0201-scope-multiplexing.png)
+
+## Open questions
+
+- Should we allow/encourage recording Span/Metric/LogRecord attributes at the Scope level?
+  The alternate is to disallow this and have completely separate set of semantic
+  conventions that are allowed for Scope attributes only.
```
blumamir, line 140: I think that the same attribute can sometimes be placed in scope, but sometimes it cannot and must be used in the Span/Metric/LogRecord level.
For example - `messaging.protocol` can sometimes be static and placed at the scope level, but another messaging library may support multiple protocols in which case it is not static and can change from call to call.

text/0201-scope-attributes.md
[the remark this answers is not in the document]
blumamir: > You are right though that there may be use cases where scope attributes may need to be specifying for every data point. An example of these may be the client-side session id that was discussed in the SIG meeting today

I was under the impression that session id is global (at a specific point in time) to the application and affects all the telemetry data collected, no matter what scope it is coming from. Thus it made sense to me to include it in the resource.
I wonder when would someone want to store a session id on one scope and omit it in another.

tsloughter:
I assume different attributes, or no attributes, means it is a different Tracer?

My only concern with this is more related to implementation in languages without globals where sharing a Tracer variable isn't possible. Having to supply the attributes every time `get_tracer` is called would hurt the purpose as Tigran notes https://github.com/open-telemetry/oteps/pull/201#discussion_r858946170

If attributes are treated the same as name/version/schema_url then it may be useful for us to have another operation that is for "registering" a tracer or scope so the attributes can be only set once. But this may be more implementation specific and not a concern for the spec itself.

text/0201-scope-attributes.md
```diff
@@ -0,0 +1,148 @@
...
+- It makes Scope consistent with the other primary data types: Resource, Span, Metric,
+  LogRecord.
+
+See additional [discussion here](https://github.com/open-telemetry/opentelemetry-specification/issues/2450).
+
+## Summary
+
+The following is the summary of proposed changes:
+
+- We will extend OpenTelemetry API to allow specifying Scope attributes when obtaining a
```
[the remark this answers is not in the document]
tigrannajaryan: Yes, it needs to be a different Tracer/Meter/LogEmitter. It is required for the use cases we have. For example we may need to simultaneously produce LogRecords that belong to profiling or to client-side events. These means we need 2 different LogEmitters, which may be from the same library, have the same name and version, but have different attributes.

tigrannajaryan:
@bogdandrutu @yurishkuro @alanwest I made an important [change in a new commit](https://github.com/open-telemetry/oteps/pull/201/commits/e89ad2107c414c4654aefbe572a39c514107f019). I will re-request your review to make sure you saw the change.

text/0201-scope-attributes.md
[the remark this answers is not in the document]
tigrannajaryan: There is a separate discussion about session id elsewhere: https://docs.google.com/document/d/1bsS8uAXv1U-i97dj5PP7LtVxWp08mmYXlah6XDVS-xQ/edit
It may or may not be solved by this new Scope attributes.

text/0201-scope-attributes.md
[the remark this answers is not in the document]
tigrannajaryan: >Precedence for Resource attributes and Span attributes must be defined then, right? I went to quickly find it but haven't yet.

I can't find it either. The reason is it is not defined in the spec is likely because we did not anticipate that we will find the same attributes on the Resource and on the Span. They each typically have their own semantic conventions that don't overlap.

You are right though that this likely still needs to be defined because it can happen and when flattening a decision must be made one way or another (and Collector indeed does such decisions for some formats, for example [Zipkin](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/1c4dd67f7af084c467ce5378115fa151f7276547/pkg/translator/zipkin/zipkinv2/from_translator.go#L70).

I think that is outside the scope of this OTEP though and can be an independent change in the spec.

text/0201-scope-attributes.md
```diff
@@ -0,0 +1,200 @@
...
+```
+It is unspecified whether or under which conditions the same or different
+`Tracer` instances are returned from this functions.
+```
+
+and replace it by:
+
+```
+The implementation MUST NOT return the same `Tracer` when called repeatedly with
+different values of parameters.
```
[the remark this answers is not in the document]
tigrannajaryan: For 2:

I think we need feedback on this from language maintainers.

In theory It should be possible to attach an instance of a Scope to each span created and do the grouping by the Scope instance reference without making comparison of all fields (I think a concept or a map by reference or pointer exists in most languages?). In practice, I don't know if how complicated it is to alter current implementations to work like this.

pyohannes, line 63: Will no-op traces be an exception to this (in case no SDK was initialized)?

tigrannajaryan: Yes, I think for no-op it is safe to return the same Tracer.

text/0201-scope-attributes.md
tigrannajaryan: >I am wondering why we don't want to take it one level higher above InstrumentationScope, maybe, by introducing another scope layer above InstrumentationScope if we want to have these attributes in resource.

We briefly chatted in Log SIG about this today, posting for others:
- If you mean adding another layer between Resource and Scope then that is hardly doable since it is unclear how to represent this in OTLP. It would be a major breaking change for OTLP which we cannot allow.
- If you mean actually having these attributes in the resource then I am not sure I understand. We already have resource attributes, why would we have another set?

tsloughter:
> > I assume different attributes, or no attributes, means it is a different Tracer?
> 
> @tsloughter yes. I made a clarification [here](https://github.com/open-telemetry/oteps/pull/201/commits/e89ad2107c414c4654aefbe572a39c514107f019).

Ah, makes sense, a use case like distinguishing profiling logs from others would need to have name/version/schema_url the same but different attributes.

As is it is a simple enough addition to the Erlang/Elixir library. I'll worry about making it possibly more efficient and ergonomic later.

tsloughter (approved):

text/0201-scope-attributes.md
tsloughter: I like this idea. And think that creating a new scope makes sense in those cases, so it fits. The only downside I see is repeating the "parent scope's" attribute so would want the ability to merge scope attributes, allowing the creation of the per-connection scope by merging those additional attributes with the instrumentation library scope.

I'd argue this is expanding the scope of the OTEP too though :). Maybe keeping it to adding attributes to scope and the necessary things like how merging happens when exported through non-OTLP protocols is best and these use-cases can be worked on after the addition is accepted.

tigrannajaryan: I think this is one of the Open Questions I have in the OTEP.
I don't know the answer. :-)

text/0201-scope-attributes.md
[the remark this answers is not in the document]
jack-berg: > I think we need feedback from multiple language maintainers to help shape this particular API. It may be that we need more than one way to specify the scope attributes (however, I would prefer to have only one if possible).

Agreed!

jack-berg (approved):

text/0201-scope-attributes.md
tigrannajaryan: Clarified in the text.

text/0201-scope-attributes.md
tigrannajaryan: Either way, it seems like this is outside the scope of this particular OTEP. It can be discussed separately and become a new OTEP.

text/0201-scope-attributes.md
```diff
@@ -38,28 +38,80 @@ The following is the summary of proposed changes:
...
```
bogdandrutu: Why is this important? Are you trying to encourage users to create a "Tracer" for every request to add runtime properties there?

Please consider to write the considerations/use-cases for making this decision.

bogdandrutu (changes requested):

text/0201-scope-attributes.md
tigrannajaryan: I [expanded](https://github.com/open-telemetry/oteps/pull/201/commits/5e50e420286c69f7c093139820daae6e945c28b8) the code example in the OTEP to also demonstrate this.

tigrannajaryan:
> Are there prototypes available for review?

@tedsuo Not a full prototype by any means, but here is a very quick draft that shows how `WithScopeAttributes` option can be added to the Tracer without breaking existing API: https://github.com/open-telemetry/opentelemetry-go/pull/2884
The Go maintainers may have much better proposals about this should be done properly. (cc @open-telemetry/go-maintainers)

text/0201-scope-attributes.md
```diff
@@ -0,0 +1,209 @@
...
+
+### Usage in Code
+
+The following is an example usage where LogEmitter is used to emit client-side
+log records (pseudocode follows):
+
+```
+// Obtain loggers once, at startup.
+appLogger = LogEmitterProvider.GetLogEmitter("mylibrary", "1.0.0")
+loggerForUserEvents = LogEmitterProvider.GetLogEmitter("mylibrary", "1.0.0", KeyValue("otel.clientside", true))
```
scheler, line 178: btw, this isn't an agreed semantic convention to represent client-side events so this could be confusing if it goes into documentation.  @martinkuba

text/0201-scope-attributes.md
```diff
@@ -0,0 +1,209 @@
+# Introduce Scope Attributes
+
+This OTEP adds attributes to the Scope.
```
Oberon00, line 3:
```suggestion
This OTEP adds attributes to the scope of a telemetry emitter (e.g. Tracer, Meter, Logger).
```

"scope" without further qualification is too ambigous. E.g. I had hoped this might bring a solution for https://github.com/open-telemetry/opentelemetry-specification/issues/335 / https://github.com/open-telemetry/opentelemetry-specification/issues/1143 but this addresses a different problem.

tigrannajaryan: Fixed.

alanwest (approved):

dashpole (approved):

text/0201-scope-attributes.md
```diff
@@ -0,0 +1,209 @@
...
+
+If the same attribute is specified both at the Span/Metric/LogRecord and at the Scope
+then the attribute value at Span/Metric/LogRecord takes precedence.
+
+This rule applies to non-OTLP exporters in SDKs, to conversions from OTLP to non-OTLP
+formats in the Collector and to OTLP recipients of data that need to interpret the
+attributes in the received data.
+
+## Exporting to non-OTLP
+
```
jsuereth, line 149: This should call out what to do with conflicts in attributes between instrumentation scope + the "raw" telemetry.

jsuereth (approved):

text/0201-scope-attributes.md
tigrannajaryan: I think it is covered by `Attribute Value Precedence` section. Are you looking for more details?

text/0201-scope-attributes.md
```diff
@@ -0,0 +1,209 @@
...
+## Summary
+
+The following is the summary of proposed changes:
+
+- We will extend OpenTelemetry API to allow specifying Scope attributes when obtaining a
+  Tracer, Meter or LogEmitter. Scope attributes will be optional.
+- We will add `attributes` field to the [InstrumentationScope](https://github.com/open-telemetry/opentelemetry-proto/blob/88faab1197a2a105c7da659951e94bc951d37ab9/opentelemetry/proto/common/v1/common.proto#L83)
+  message of OTLP.
+- We will specify that Telemetry emitted via a Scope-ed Tracer, Meter or LogEmitter will
```
reyang, line 30: If there are multiple Meter instances under the same MeterProvider, and these Meter instances have different Scope attributes, it could affect the aggregation logic. Also the user might want to configure if they want to have the Scope attributes be reported as metrics dimensions (e.g. when exporting to Prometheus) at all or not.

Probably need to do some prototype to have a meaningful conversation. I think it won't block this OTEP 👍

reyang (approved):

tedsuo:
Hi @tigrannajaryan. 

I really like this proposal; it's clean, simple, and introduces a new context that I think we will really benefit from. Among other things, configuration attributes for things like database clients could potentially be reported as scoped attributes, which could be a very valuable input to automated analysis.

I've also reviewed this OTEP from the perspective of client-side (RUM) attributes such as SessionID, UserID, etc. I don't think these attributes could be placed within instrumentation scopes without adding additional complexity. Potential approaches for integrating scopes with something like a SessionManager:
* Scopes could be nested, and the SessionManager could have access to the parent scope.  Scope attributes would probably have to be mutable so that values in the parent scope could be changed without breaking the link between child and parent scopes.
* Scopes could remain flat, and something like a ScopeAttributeProvider could need to be handed out everywhere so that the SessionManager could trigger an update of the SessionID attribute in all Scopes. This is essentially the same as the ResourceProvider concept, except the same data would be duplicated across many scopes.

I'm reluctant to suggest adding this complexity here. I think your current proposal is elegant, and there is no need for this additional complexity when all scope attributes are supplied locally by the instrumentation. Even in the case of instrumentation which may want to update it's scope attributes, there's no need for a ScopeAttributeProvider, as the instrumentation can just recreate the tracer/logger/etc.

All of the additional complexity involved in storing a SessionID as a scope attribute comes from the need for non-local attributes to be mixed into the scope. We wouldn't want every instrumentation library to be forced to take something like a SessionManager as an input, not to mention the additional code needed to recreate tracers and loggers whenever the SessionManger updates. A ScopeAttributeProvider would centralize some of this complexity, but I don't think it's an improvement over a ResourceProvider. 

Here's why I don't think it would be an improvement: conceptually, scopes represent a localized context. Pushing non-local context like SessionID into every scope violates that concept, and creates additional complexity as a side effect. If we have to add some complexity to OTel in order to manage sessions properly, I'd prefer we do that without mixing non-local attributes into a local context. In our data model, I believe that Resources represent the correct context for things like SessionID and UserID in mobile/web clients. The fact that these attributes would not need to be duplicated if they were resources is an indicator that this is the correct context to place them in.

Having to wrangle with the concept of updatable resources is unfortunate, but I'm concerned that our data model would lose clarity while retaining much of the same complexity if we tried to use scopes for this purpose.

Hope that analysis is helpful. We can continue this conversation as part of the RUM SIG, but if you're satisfied that scopes are not the right answer, I would suggest we approve this OTEP and not the sessionID issue slow down the implementation of scope attributes.

tigrannajaryan:
@tedsuo I agree with you.

The Scopes attributes _as they are defined by this OTEP_ are not a very good fit for capturing the session id. The Scope attributes API defined by this OTEP is targeted for use-cases, which know upfront, at the time of Tracer/Meter/LogEmitter creation what the attributes values should be and sets them once and for the entire lifetime of the Tracer/Meter/LogEmitter.

The session id appears to be a different beast. It can change any time and these changes are not synchronized in any way to the lifetime of the Tracer/Meter/LogEmitter as it is understood by Otel APIs today.

I think you do need a different mechanism for session ids. Perhaps the right framing for session id is "enrichment" or "on the fly modification". All emitted telemetry is enriched by an attribute called "session_id", the value of which can be _provided_ by a user-defined function. From this perspective it appears the notion of the "Provider" that you suggest can be the right approach.

I think we can continue the session id discussion in its own thread (your doc or in a new OTEP). I am happy to keep this OTEP open for a while. I want to make sure we find a good, different solution for sessions ids, a solution which does not rely on Scope Attributes and does not require changes to this OTEP.

tedsuo:
Great, sounds good.

MrAlias (approved):

Oberon00:
@tedsuo I had similar thoughts at first: https://github.com/open-telemetry/oteps/pull/201#discussion_r863794422

While this is for sharing attributes that apply to all emitters, there are more "scopes" thinkable that could use sharing, e.g. https://github.com/open-telemetry/opentelemetry-specification/issues/335 / https://github.com/open-telemetry/opentelemetry-specification/issues/1143 would probably call for sharing attributes on some subtree of a trace, or on a subset of spans created by an telemetry producer (e.g. something like https://github.com/open-telemetry/opentelemetry-specification/issues/579)

text/0201-scope-attributes.md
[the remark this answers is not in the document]
tigrannajaryan: Added an issue to figure out the Resource precedence: https://github.com/open-telemetry/opentelemetry-specification/issues/2535

text/0201-scope-attributes.md
tigrannajaryan: @jsuereth Are you looking for more details?

tigrannajaryan:
We have a [confirmation](https://github.com/open-telemetry/opentelemetry-go/pull/2884#pullrequestreview-960540442) from Go maintainer that the OTEP is implementable in Go.

tigrannajaryan:
@jack-berg since you approved can I assume that it means you think the OTEP is implementable in Java?

tigrannajaryan:
@blumamir do you think the OTEP is implementable in a reasonably performant way in JS SDK?

tigrannajaryan:
We have a [confirmation](https://github.com/open-telemetry/opentelemetry-java/issues/4473) from Java maintainers that this OTEP is implementable.

tigrannajaryan:
We have a [confirmation](https://github.com/open-telemetry/opentelemetry-python/issues/2699) from Python maintainers that this OTEP is implementable.

tigrannajaryan:
We have a [confirmation](https://github.com/open-telemetry/opentelemetry-ruby/issues/1262) from Ruby maintainers that this OTEP is implementable.

text/0201-scope-attributes.md
Oberon00: Such attributes are not really attributes of the telemetry emitter. There was once the proposal of creating a "prepared span" API for such purposes: https://github.com/open-telemetry/opentelemetry-specification/issues/734.

tigrannajaryan:
I think we now have sufficient confirmations from many maintainers to be confident that the OTEP is implemented.

If you still have any concerns please comment.

text/0201-scope-attributes.md
tigrannajaryan: Partial Go prototype for traces is here: https://github.com/open-telemetry/opentelemetry-go/pull/2884

I am resolving this comment since you said it won't block the OTEP. Feel free to reopen if needed.

yurishkuro (approved):

text/0201-scope-attributes.md
tigrannajaryan: I think this OTEP leaves enough leeway to decide one way or another, so this decision is not a blocker for the OTEP move forward.

tigrannajaryan:
I resolved all comments. I will keep this open for a bit longer to give people time to object. Unless I here objections I will merge it.

PR #202, opened 2022-05-11
Events and Logs API

scheler:
We introduce an Events and Logs API that is based on the OpenTelemetry Log signal, backed by LogRecord data model and LogEmitter SDK.  This is meant to be used by end-users for creating Events and by the Log Appender authors to create Logs.


text/0202-events-api.md
```diff
@@ -0,0 +1,227 @@
+# Introducing Events API
+
+
+In this OTEP we introduce an Events API that is based on the OpenTelemetry Log signal. In OpenTelemetry's perspective Log Records and Events are different names for the same concept - however, there is a subtle difference in how they are represented using the underlying data model that is described below. Although every language has APIs for logs, they are not all capable of creating events. We will describe why the existing Logging APIs are not sufficient for the purpose of creating events.  It will then be evident that we will need an API in OpenTelementry for creating events.
```
yurishkuro, line 4: > In OpenTelemetry's perspective Log Records and Events are different names for the same concept

Where in the existing spec is this stated?  afaik Events are not a top-level entity, but a sub-entity within Spans.

scheler: It is mentioned here - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/overview.md#events-and-logs

text/0202-events-api.md
```diff
@@ -0,0 +1,227 @@
...
+For the above reasons we can conclude that we will need a separate API for generating Events API
+
+## Should OpenTelemetry have an API for logs?
+
+There’s a general consensus in the Otel community that we should not have a full-fledged logging API unless there is a language that doesn't already have a plethora of logging libraries & APIs to choose from where it might make sense to define one. Further, we will not be able to have the [rich set of configuration options](https://logging.apache.org/log4j/2.x/manual/configuration.html) that some popular logging frameworks provide so the logging API in Otel will only become yet another API.
+
+## Events API Interface
+
+
+For reference, a prototype of the Events API in Java is [here](https://github.com/scheler/opentelemetry-java/pull/1/files)
```
jack-berg, line 40: I propose we go in a slightly different direction that what is described in this document, and what is embodied in the prototype link. 

Here's an alternative prototype, illustrated as [PR](https://github.com/jack-berg/opentelemetry-java/pull/1/files) against the current main branch of otel java. This [test class](https://github.com/jack-berg/opentelemetry-java/pull/1/files#diff-b039e70bb83ee0c5792109dc4fa786c05617f04bebde3d8bde24ce6df26f545e) demonstrates simple usage of the API. These ideas come from conversations in the  Log SIG, which I wanted to make concrete to compare / contrast.

**Summary of differences**
- There is a single API for producing events and log records. NOTE: only the event portion of the API is intended to be used by instrumentation authors.
- Instead of an EventEmitter or Logger (and their corresponding providers) there is a LogEmitter
- There is one method to emit an event called `LogEmitter#eventBuilder(String)`, which accepts the name of the event and returns a EventBuilder for adding additional event details. The `EventBuilder#emit()` method emits the event. 
- There is one method to emit a log record called `LogEmitter#logRecordBuilder()`, which returns a LogRecordBuilder for adding details to log record. The `LogRecordBuilder#emit()` method emits the log record. Note that `EventBuilder` extends `LogRecordBuilder` so they share all the same setters. 

**Reasoning**

There's consensus that we don't want a log API to compete with existing frameworks. The log record API looks remarkably similar to a log API, so why include it instead of just an event API? Well, we need an API to bridge the existing log APIs into OpenTelemetry. That API has a lot of conceptual overlap with an event API. We could add a separate event API and log appender API, but why have all the code and conceptual duplication? Seems better to have a single API that serves both purposes, and to indicate **using documentation** that the log appender API is only meant for appenders, and **NOT** for instrumentation authors.

In otel java, we've bridged two popular logging frameworks ([Log4j](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/log4j/log4j-appender-2.17/library/src/main/java/io/opentelemetry/instrumentation/log4j/appender/v2_17/OpenTelemetryAppender.java) and  [Logback](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/logback/logback-appender-1.0/library/src/main/java/io/opentelemetry/instrumentation/logback/appender/v1_0/OpenTelemetryAppender.java)) to otel, but needed to introduce an [appender API](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation-appender-api-internal/src/main/java/io/opentelemetry/instrumentation/api/appender/internal) to decouple the appenders from taking a dependency on the otel java log SDK. The appender API is pretty silly - it just duplicates all the APIs of the otel java log SDK and delegates to a log SDK instance. 

If the API proposed in this otep had a mechanism for producing log records, the otel java appender API would be unnecessary.

Additionally, a single API for producing events as well as log records produces a API / SDK that is very symmetric to the design of tracing / logs. There is no precedent in tracing in logs for a setup where two APIs (events and log appender) delegate to one (or two SDKs) which share the same data model. This symmetry reduces maintenance and cognitive overhead.

jack-berg: Here's another prototype, again illustrated as a [PR](https://github.com/jack-berg/opentelemetry-java/pull/2/files) and demonstrated in a [test class](https://github.com/jack-berg/opentelemetry-java/pull/2/files#diff-b039e70bb83ee0c5792109dc4fa786c05617f04bebde3d8bde24ce6df26f545e).

This version has separate APIs for log appenders and events. The naming of classes and required fields reflect the two different use cases.
- On the event side, you have `EventEmitterProvider` and `EventEmitter`. To obtain a `EventEmitter`, you must specify a domain (event names are unique within a domain): `logEmitterProvider.get("my-event-scope", "my-event-domain")`. When emitting an event, you must include an event name: `eventEmitter.eventBuilder("my-event").setAttributes(...).emit()`. 
- On the log side, you have `LoggerProvider` and `Logger`. There is no requirement for an event domain, and there are no required fields when building a log record: `logger.logRecordBuilder().body().setAttributes(...).emit()`. Some subtle differences make it a bit harder to obtain a `Logger` than it is to obtain a `Tracer`, `Meter`, or `EventEmitter`, which is intentional to reflect the fact that the API is only for use by appenders, not instrumentation authors. 

These APIs can be both be implemented by a single Log SDK. This approach has a larger API surface area than my first proposal, but results in an API with more precise language which is arguably less likely to confuse users.

tigrannajaryan: I review both prototypes and I think the first approach looks nice (commented on the PR).

martinkuba: In the second prototype, the class emitting events is called Logger instead of LogEmitter like in the other prototype. Is this on purpose? Does "Logger" make more sense semantically for the appenders use case?

I am in favor of the second option - it makes it clear that the API is used for events, and it also does not expose the low-level log API to instrumentations libraries.

jack-berg: We discussed these options in the java SIG and there was a preference for the second prototype. 

cc @open-telemetry/java-maintainers @open-telemetry/java-instrumentation-maintainers

text/0202-events-api.md
```diff
@@ -0,0 +1,227 @@
...
+
+## Who requires Events API
+
+Here are a few situations that require recording of Events, there will be more.
+* RUM events (Client-side instrumentation)
+* Recording kubernetes events
+* Recording eBPF events
+* Few other event systems described in [example mappings](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#appendix-a-example-mappings) in the data model.
+
+## Can the current Log API interfaces be used for events?
```
tsloughter, line 19: This list reads to me like the issue is not all languages/runtimes having a structured logging API?

If this is the case it might make more sense to instead have the Log spec say that in those cases Otel logging should provide an API, instead of adding an Events API.

text/0202-events-api.md
scheler: I am okay with either options but since we have gone back and forth on this a couple times I will wait some time for further comments - on whether we should have one "Logs API" to create both logs and events or have Logs API and Events API separately.

In either case, I assume the API the create application logs will be restricted for use in appenders.

tigrannajaryan: We need to circulate this more widely. Tracing/metrics people may have stronger opinions about consistency argument.

text/0202-events-api.md
```diff
@@ -0,0 +1,178 @@
+# Introducing Events API
+
```
reyang, line 2: nit: remove the extra blank line here

MSNev:
Related issue, which could allow for having a common EventsApi for using Logs or Traces, if the Attributes definition is expanded to match the protocol. https://github.com/open-telemetry/opentelemetry-specification/issues/2581

bogdandrutu:
As logs approvers, I would like more opinions:

@kumoroku
@djaglowski
@zenmoto

text/0202-events-api.md
```diff
@@ -0,0 +1,178 @@
...
+## Subtle differences between Logs and Events
+In OpenTelemetry's perspective Log Records and Events are different names for the same concept. However, there are subtle differences in how they are represented in the underlying LogRecord data model. Logs have a mandatory severity level as a first-class parameter that events do not have, and events have a mandatory name that logs do not have. Further, logs typically have messages in string form and events have data in the form of key-value pairs. It is due to this that their API interface requirements are slightly different.
```
reyang, line 9: nit: add a blank line between Ln8 and Ln9.

text/0202-events-api.md
```diff
@@ -0,0 +1,178 @@
...
+* In python, logging.LogRecord’s extra field is mapped to Otel LogRecord’s attributes but this field is a hidden field and not part of the public interface.
+3. The Log APIs have a message parameter which could map to the Body field of LogRecord. However, this is restricted to String messages and does not allow for a complex structure.
+
+For the above reasons we can conclude that we will need a separate API for generating Events API
+
+## Should OpenTelemetry have an API for logs?
+
+There’s a general consensus in the Otel community that we should not have a full-fledged logging API unless there is a language that doesn't already have a plethora of logging libraries & APIs to choose from where it might make sense to define one. Further, we will not be able to have the [rich set of configuration options](https://logging.apache.org/log4j/2.x/manual/configuration.html) that some popular logging frameworks provide so the logging API in Otel will only become yet another API.
```
reyang, line 35: OTel (uppercase `T`) https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification#acronym

text/0202-events-api.md
```diff
@@ -0,0 +1,178 @@
...
+
+A LogRecord representing exception event will look like this:
+```
+{
+  time_unix_nano: 1653682410,
+  Attributes: {
+   event.name: exception,
+   exception.type:  OSError,
+   exception.message: Division by Zero,
+   exception.stacktrace: “”,
```
reyang, line 104:
```suggestion
   exception.stacktrace: "",
```

text/0202-events-api.md
martinkuba: Here is a second prototype in JS, which has a separate Events API. There is also a draft of Events SDK implementation that is using the Logs SDK
https://github.com/martinkuba/opentelemetry-js/pull/1/files

text/0202-events-api.md
```diff
@@ -0,0 +1,178 @@
...
+
+There’s a general consensus in the Otel community that we should not have a full-fledged logging API unless there is a language that doesn't already have a plethora of logging libraries & APIs to choose from where it might make sense to define one. Further, we will not be able to have the [rich set of configuration options](https://logging.apache.org/log4j/2.x/manual/configuration.html) that some popular logging frameworks provide so the logging API in Otel will only become yet another API.
+
+## Events API Interface
+
+
+For reference, a prototype of the Events API in Java is [here](https://github.com/scheler/opentelemetry-java/pull/1/files)
+
+
+Client-side telemetry is one of the initial clients that will use the Events API and so the API will be made available in JavaScript, Java and Swift first to be able to use in the SDKs for Browser, Android and iOS.  It may also be added in Go since there is a kubernetes events receiver implemented in Collector based on Logs data-model.
```
reyang, line 43:
```suggestion
Client-side telemetry is one of the initial clients that will use the Events API and so the API will be made available in JavaScript, Java and Swift first to be able to use in the SDKs for Browser, Android and iOS.  It may also be added in Go since there is a Kubernetes events receiver implemented in Collector based on Logs data-model.
```

text/0202-events-api.md
reyang, line 9: "Further, logs typically have messages in string form and events have data in the form of key-value pairs." where is this coming from?

text/0202-events-api.md
```diff
@@ -0,0 +1,178 @@
...
+* In Android, android.util.Log has methods  Log.v(), Log.d(), Log.i(), Log.w(), and Log.e() to write logs. These methods correspond to the severity level.
+* Swift on iOS has Logger interface that has methods corresponding to severity level too.
+2. The Log APIs do not have a standard way to pass event attributes. 
+* It may be possible to use the interpolation string args as the parameter to pass event attributes. However, the logging spec seems to map the human readable message (which is obtained after replacing the args in the interpolated string)  to the Body field of LogRecord. 
+* Log4j has an EventLogger interface that can be used to create structured messages with arbitrary key-value pairs, but log4j is not commonly used in Android apps as it is not officially supported on Android as per this [Stack Overflow thread](https://stackoverflow.com/questions/60398799/disable-log4j-jmx-on-android/60407849#60407849) by one of log4j’s maintainers.
+* In python, logging.LogRecord’s extra field is mapped to Otel LogRecord’s attributes but this field is a hidden field and not part of the public interface.
```
reyang, line 28:
```suggestion
1. The log level is fundamental to the Log APIs in almost all the languages; all the methods in the Log interface are named after the log level, and there is usually no generic method to submit a log entry without log level.
* In JavaScript for Web, the standard method of logging is to use console.log. Events can be created using [Event/CustomEvent](https://developer.mozilla.org/en-US/docs/Web/Events/Creating_and_triggering_events) interfaces.However, there is no option to define custom destination for these logs and events. Logs go only to console and event listeners are attached to the DOM element that dispatches it.
* In Android, android.util.Log has methods  Log.v(), Log.d(), Log.i(), Log.w(), and Log.e() to write logs. These methods correspond to the severity level.
* Swift on iOS has Logger interface that has methods corresponding to severity level too.
2. The Log APIs do not have a standard way to pass event attributes. 
* It may be possible to use the interpolation string args as the parameter to pass event attributes. However, the logging spec seems to map the human readable message (which is obtained after replacing the args in the interpolated string)  to the Body field of LogRecord. 
* Log4j has an EventLogger interface that can be used to create structured messages with arbitrary key-value pairs, but log4j is not commonly used in Android apps as it is not officially supported on Android as per this [Stack Overflow thread](https://stackoverflow.com/questions/60398799/disable-log4j-jmx-on-android/60407849#60407849) by one of log4j’s maintainers.
* In python, logging.LogRecord's extra field is mapped to Otel LogRecord's attributes but this field is a hidden field and not part of the public interface.
```

reyang, line 28:
```suggestion
1. The log level is fundamental to the Log APIs in almost all the languages; all the methods in the Log interface are named after the log level, and there is usually no generic method to submit a log entry without log level.
* In JavaScript for Web, the standard method of logging is to use console.log. Events can be created using [Event/CustomEvent](https://developer.mozilla.org/en-US/docs/Web/Events/Creating_and_triggering_events) interfaces.However, there is no option to define custom destination for these logs and events. Logs go only to console and event listeners are attached to the DOM element that dispatches it.
* In Android, android.util.Log has methods  Log.v(), Log.d(), Log.i(), Log.w(), and Log.e() to write logs. These methods correspond to the severity level.
* Swift on iOS has Logger interface that has methods corresponding to severity level too.
2. The Log APIs do not have a standard way to pass event attributes. 
* It may be possible to use the interpolation string args as the parameter to pass event attributes. However, the logging spec seems to map the human readable message (which is obtained after replacing the args in the interpolated string)  to the Body field of LogRecord. 
* Log4j has an EventLogger interface that can be used to create structured messages with arbitrary key-value pairs, but log4j is not commonly used in Android apps as it is not officially supported on Android as per this [Stack Overflow thread](https://stackoverflow.com/questions/60398799/disable-log4j-jmx-on-android/60407849#60407849) by one of log4j’s maintainers.
* In python, logging.LogRecord's extra field is mapped to Otel LogRecord's attributes but this field is a hidden field and not part of the public interface.
```

reyang, line 28:
```suggestion
* In python, logging.LogRecord's extra field is mapped to Otel LogRecord's attributes but this field is a hidden field and not part of the public interface.
```

text/0202-events-api.md
```diff
@@ -0,0 +1,178 @@
...
+3. The Log APIs have a message parameter which could map to the Body field of LogRecord. However, this is restricted to String messages and does not allow for a complex structure.
```
reyang, line 29: Would you elaborate more here? What is/are "Log APIs" referring to?

text/0202-events-api.md
scheler: @MSNev The Span Events created using the Trace API is outside the scope of this OTEP, although I included some text related to it towards the end. The Events API in this OTEP is intended primarily to create Events *outside* the scope of a Span, and so Trace API cannot be used for that purpose.

If you see the section `Comparing with Span Events`, I provided some guidance on how we can extend Trace API to create Span Events using LogRecords. However, that is not intended to be implemented via this OTEP and needs to be taken up separately.

text/0202-events-api.md
scheler: Until structured logging, which is relatively new, log messages are typically strings. Event data is more generic and hence best represented using key-value pairs.

text/0202-events-api.md
scheler: I updated the text a bit. The current Log APIs available in the language SDK do not support creating structured logs.

text/0202-events-api.md
```diff
@@ -0,0 +1,178 @@
...
+
+We have an option of adding API for both Logs and Events. However, there is a general consensus that we should not have an API in Otel for creating logs since each language already has multiple logging frameworks. Therefore we restrict the API specification below to Events and call it Events API. For logs, it is recommended that end-users continue to use existing Logging APIs and export the logs into OTLP using the  [appender API](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation-appender-api-internal/src/main/java/io/opentelemetry/instrumentation/api/appender/internal) and LogEmitter SDK. The Events API will offer a subset of the features of LogEmitter SDK and so it will be backed by LogEmitter SDK and the LogRecord data model.
+
+## Subtle differences between Logs and Events
```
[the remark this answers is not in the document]
scheler: updated the text, can you take a look?

text/0202-events-api.md
tigrannajaryan: > We discussed these options in the java SIG and there was a preference for the second prototype.

@jack-berg can you post what arguments the Java SIG has in favour of doing this?

So far personally for me the consistency argument is very strong and I have not yet seen opposing arguments that are sufficient to warrant having separate APIs.

I do understand that for someone who primarily is interested in Events API, a separate Events API sharply focused on the specific use case is very attractive. However, I think that it is not a good enough argument to warrant losing consistency across OpenTelemetry APIs.

tigrannajaryan: @jack-berg @scheler @martinkuba @MSNev @zenmoto @djaglowski 

I started a document to compare the 2 approaches (separate vs one API) side-by-side: https://docs.google.com/document/d/1Uy5L3UThFyc4PelgH51G3OCLFbvzq2xWiNDy2G8iXxU/edit#

Please help me fill the details here.

As agreed in the call we are going to try to make the choice by the next Log SIG. If we don't have a definitive choice then I as a TC member will be the tiebreaker (unless folks don't want tie-breaking and want to make the choice in some other way - I am fine with that).

tigrannajaryan:
All, we had an extensive discussion of this OTEP in Log SIG. There is one major blocker which prevents us from moving forward: whether there needs to be a single Logs/Events API or 2 separate APIs.

I started a document to compare the 2 approaches (separate vs one API) side-by-side: https://docs.google.com/document/d/1Uy5L3UThFyc4PelgH51G3OCLFbvzq2xWiNDy2G8iXxU/edit#
The document also has the links to prototypes with both approaches.

We agreed in the Log SIG we are going to make the choice by the next Log SIG (next Wed). If you have any opinion on this topic please add arguments in favour of the one of the approach to the document.

text/0202-events-api.md
```diff
@@ -0,0 +1,184 @@
...
+
+In OpenTelemetry's perspective Logs and Events are different names for the same concept. However, there are subtle differences in how they are represented using the underlying LogRecord data model. Logs have a mandatory severity level as a first-class parameter that events do not have, and events have a mandatory name that logs do not have. Further, logs typically have messages in string form and events have data in the form of key-value pairs. It is due to this that their API interface requirements are slightly different.
```
cijothomas, line 9: Events here refer to Span Events? If yes, consider linking to the span event spec.

text/0202-events-api.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Who requires Events API
+
+Here are a few situations that require recording of Events, there will be more.
+
+- RUM events (Client-side instrumentation)
+- Recording kubernetes events
+- Recording eBPF events
+- Few other event systems described in [example mappings](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#appendix-a-example-mappings) in the data model.
```
MSNev, line 18: I think Profiler events would also fit and should be optionally listed

text/0202-events-api.md
```diff
@@ -0,0 +1,184 @@
...
+  - This is a specialized variant of logEvent function, so for anything not specified here, the same requirements as for logEvent apply.
+  - The signature of the method is to be determined by each language and can be overloaded as appropriate. The method MUST record an exception as an Event with the conventions outlined in the exception semantic conventions document. The minimum required argument SHOULD be no more than only an exception object.
+  - If RecordException is provided, the method MUST accept an optional parameter to provide any additional event attributes (this SHOULD be done in the same way as for the AddEvent method). If attributes with the same name would be generated by the method already, the additional attributes take precedence.
+  - Note: RecordException may be seen as a variant of logEvent with additional exception-specific parameters and all other parameters being optional (because they have defaults from the [exception semantic convention](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/exceptions.md)).
+
+A LogRecord representing exception event will look like this:
```
MSNev, line 105: nit: `A LogRecord representing exception event may look like this:`
As we haven't defined the conventions for exceptions yet, it would be better to have the language define this as a possible view as opposed to _the_ (will) look like.

tigrannajaryan:
@scheler to clarify, you are going with this, right?
image

scheler:
Yes, but the `logEvent()` method is on the Logger and not on
`EventBuilder`.  The `EventBuilder` is obtained from Logger as
Logger.eventBuilder("myEventName"), and so the EventBuilder returned will
have the `event.name` attribute propulated in the LogRecord. Further, the
EventBuilder will only have setters for various properties of LogRecord and
will not need the logEvent method. The `logEvent` method is a convenience
method in the Logger class, to create the Event without using the builder
pattern.

I was also thinking of using the terms eventBuilder and logBuilder, since
both use LogRecords underneath. Let me know what you think about this.

text/0202-events-api.md
jack-berg: FYI, I've updated the naming of interfaces / classes in the prototype [PR](https://github.com/jack-berg/opentelemetry-java/pull/1/files) that demos a unified API to reflect the naming we discussed in the 6/8 log sig and mentioned by @tigrannajaryan in this [comment](https://github.com/open-telemetry/oteps/pull/202#issuecomment-1150291147).

text/0202-events-api.md
scheler: @MSNev the conventions for exceptions are already defined, please see - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/exceptions.md

scheler: I changed the text to say `may`.

text/0202-events-api.md
scheler: @pmm-sumo removed the redundant text.
@cijothomas The events here refer to independent events, not span events created using the Trace API. I updated the initial paragraph to clarify this.

text/0202-events-api.md
```diff
@@ -0,0 +1,184 @@
+# Introducing Events and Logs API
+
+In this OTEP we introduce an Events and Logs API that is based on the OpenTelemetry Log signal. The Events here refer to the independent Events and not to be confused with Span Events which occur only in the context of a span. In OpenTelemetry's perspective Log Records and Events are different names for the same concept - however, there is a subtle difference in how they are represented using the underlying data model that is described below. Although every language has APIs for logs, they are not all capable of creating events. We will describe why the existing Logging APIs are not sufficient for the purpose of creating events.  It will then be evident that we will need an API in OpenTelementry for creating events.
```
tigrannajaryan, line 3: >Although every language has APIs for logs, they are not all capable of creating events.

I think this sentence can be misinterpreted and is probably unnecessary. Perhaps delete it.

text/0202-events-api.md
```diff
@@ -0,0 +1,184 @@
...
+
+We have an option of adding API for both Logs and Events. However, there is a general consensus that we should not have an API in Otel for creating logs since each language already has multiple logging frameworks. Therefore we restrict the API specification below to Events and call it Events API. For logs, it is recommended that end-users continue to use existing Logging APIs and export the logs into OTLP using the  [appender API](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation-appender-api-internal/src/main/java/io/opentelemetry/instrumentation/api/appender/internal) and LogEmitter SDK. The Events API will offer a subset of the features of LogEmitter SDK and so it will be backed by LogEmitter SDK and the LogRecord data model.
```
tigrannajaryan, line 5: >Therefore we restrict the API specification below to Events and call it Events API.

This needs some rewording, since we decided to call it "Events and Logs API".

text/0202-events-api.md
scheler: Deleted.

text/0202-events-api.md
scheler: I removed this paragraph entirely and added that the API for Logs is supposed to be only for log appenders.

text/0202-events-api.md
scheler: @MSNev since the profiler events is very new, I will omit it for now. The list here is not meant to be comprehensive.

scheler:
> 

@tigrannajaryan I updated the OTEP with all the discussed changes, can you take a look?

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,193 @@
...
+- RUM events (Client-side instrumentation)
+- Recording kubernetes events
+- Recording eBPF events
+- Collector Entity events [link](https://docs.google.com/document/d/1Tg18sIck3Nakxtd3TFFcIjrmRO_0GLMdHXylVqBQmJA/edit)
+- Few other event systems described in [example mappings](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#appendix-a-example-mappings) in the data model.
+
+## Can the current Log API interfaces be used for events?
+
+- The log level is fundamental to the Log APIs in almost all the languages; all the methods in the Log interface are named after the log level, and there is usually no generic method to submit a log entry without log level.
+  - In JavaScript for Web, the standard method of logging is to use console.log. Events can be created using [Event/CustomEvent](https://developer.mozilla.org/en-US/docs/Web/Events/Creating_and_triggering_events) interfaces.However, there is no option to define custom destination for these logs and events. Logs go only to console and event listeners are attached to the DOM element that dispatches it.
```
jack-berg, line 24:
```suggestion
  - In JavaScript for Web, the standard method of logging is to use console.log. Events can be created using [Event/CustomEvent](https://developer.mozilla.org/en-US/docs/Web/Events/Creating_and_triggering_events) interfaces. However, there is no option to define custom destination for these logs and events. Logs go only to console and event listeners are attached to the DOM element that dispatches it.
```

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,193 @@
...
+  - In Android, android.util.Log has methods  Log.v(), Log.d(), Log.i(), Log.w(), and Log.e() to write logs. These methods correspond to the severity level.
+  - Swift on iOS has Logger interface that has methods corresponding to severity level too.
+- The current Log APIs do not have a standard way to pass event attributes.
+  - It may be possible to use the interpolation string args as the parameter to pass event attributes. However, the logging spec seems to map the human readable message (which is obtained after replacing the args in the interpolated string)  to the Body field of LogRecord.
```
jack-berg, line 28:
```suggestion
  - It may be possible to use the interpolation string args as the parameter to pass event attributes. However, the logging spec seems to map the human readable message (which is obtained after replacing the args in the interpolated string) to the Body field of LogRecord.
```

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,193 @@
...
+  - Log4j has an EventLogger interface that can be used to create structured messages with arbitrary key-value pairs, but log4j is not commonly used in Android apps as it is not officially supported on Android as per this [Stack Overflow thread](https://stackoverflow.com/questions/60398799/disable-log4j-jmx-on-android/60407849#60407849) by one of log4j’s maintainers.
+  - In Python, logging.LogRecord's extra field is mapped to Otel LogRecord's attributes but this field is a hidden field and not part of the public interface.
+- The current Log APIs have a message parameter which could map to the Body field of LogRecord. However, this is restricted to String messages and does not allow for structured logs.
```
jack-berg, line 31: Additionally, none of the log APIs I'm aware of support something like the `name` of otel events. The presence of `name` is the defining characteristic of events on otel and necessitates an API that puts it front and center.

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,193 @@
...
+* LoggerProvider is the entry point of the API. It provides access to Loggers.
+* Logger is the class responsible for creating events using Log records.
+
+![Events and Logs API classes](img/0202-events-and-logs-api.png)
+
+### LoggerProvider
+
+Logger can be accessed with an LoggerProvider.
+
+In implementations of the API, the LoggerProvider is expected to be the stateful object that holds any configuration.
```
jack-berg, line 56: Should note that the SDK implementation of this is what we currently call the [LogEmitterProvider](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/logging-library-sdk.md#logemitterprovider).

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,193 @@
...
+
+Normally, the LoggerProvider is expected to be accessed from a central place. Thus, the API SHOULD provide a way to set/register and access a global default LoggerProvider.
+
+Notwithstanding any global LoggerProvider, some applications may want to or have to use multiple LoggerProvider instances, e.g. to have different configuration (like LogRecordProcessors) for each (and consequently for the Loggers obtained from them), or because it's easier with dependency injection frameworks. Thus, implementations of LoggerProvider SHOULD allow creating an arbitrary number of Logger instances.
```
jack-berg, line 60:
```suggestion
Notwithstanding any global LoggerProvider, some applications may want to or have to use multiple LoggerProvider instances, e.g. to have different configuration (like LogRecordProcessors) for each (and consequently for the Loggers obtained from them), or because it's easier with dependency injection frameworks. Thus, implementations of LoggerProvider SHOULD allow creating an arbitrary number of instances.
```

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,193 @@
...
+
+- version (optional): Specifies the version of the instrumentation scope if the scope has a version (e.g. a library version). Example value: 1.0.0.
+- schema_url (optional): Specifies the Schema URL that should be recorded in the emitted telemetry
+- event_domain (optional): Specifies the domain for the events created, which should be added in the attribute `event.domain` in the instrumentation scope.
+- pass_context (optional): Specifies whether the Trace Context should automatically be passed on to the events and logs created by the Logger. This SHOULD be false by default.
```
jack-berg, line 77: I prefer to rename this to `include_trace_context`

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,193 @@
...
+
+It is unspecified whether or under which conditions the same or different Logger instances are returned from this function.
+
+Implementations MUST NOT require users to repeatedly obtain an Logger again with the same name+version+schema_url+event_domain+pass_context to pick up configuration changes. This can be achieved either by allowing to work with an outdated configuration or by ensuring that new configuration applies also to previously returned Loggers.
+
+Note: This could, for example, be implemented by storing any mutable configuration in the LoggerProvider and having Logger implementation objects have a reference to the LoggerProvider from which they were obtained. If configuration must be stored per-Logger (such as disabling a certain Logger), the Logger could, for example, do a look-up with its name+version+schema_url in a map in the LoggerProvider, or the LoggerProvider could maintain a registry of all returned Loggers and actively update their configuration if it changes.
```
jack-berg, line 83:
```suggestion
Note: This could, for example, be implemented by storing any mutable configuration in the LoggerProvider and having Logger implementation objects have a reference to the LoggerProvider from which they were obtained. If configuration must be stored per-Logger (such as disabling a certain Logger), the Logger could, for example, do a look-up with its name+version+schema_url+event_domain+pass_context in a map in the LoggerProvider, or the LoggerProvider could maintain a registry of all returned Loggers and actively update their configuration if it changes.
```

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,193 @@
...
+
+The effect of associating a Schema URL with a Logger MUST be that the telemetry emitted using the Logger will be associated with the Schema URL, provided that the emitted data format is capable of representing such association.
+
+### Logger
+
+The Logger is responsible for creating Events and Logs.
+
+Note that Loggers should usually not be responsible for configuration. This should be the responsibility of the LoggerProvider instead.
```
jack-berg, line 91:
```suggestion
Note that Loggers should not be responsible for configuration. This should be the responsibility of the LoggerProvider instead.
```

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,193 @@
...
+
+#### Logger operations
+
+The Logger MUST provide functions to:
+
+- Function named “logEvent” to create an Event with the provided event name and attributes. The event name provided should be inserted as an attribute with key “event.name”. It will override any attribute with the same key in the attributes passed.
+- Function named “recordException” to record an Exception as an Event. This is to facilitate recording an exception outside of a trace context for languages that already do not support recording an exception in a log message. This should work similar to [Record Exception](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#record-exception) in Trace API, with the following semantics
```
jack-berg, line 98: I would omit this API in the initial OTEP.

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,193 @@
...
+   exception.message: Division by Zero,
+   exception.stacktrace: "",
+   exception.escaped: false
+ }
+}
+```
+
+The Logger SHOULD additionally provide the following functions:
+
+- Create a new Event using `LogRecord` data model.
```
jack-berg, line 121: Why? Let's keep the API surface area as small as possible to start. I see the need for only two methods:
- logEvent
- logRecord

In java, we'd likely go the route of having a builder pattern for each of these.

jack-berg (commented):
Just a couple of comments but I'm mostly aligned 👍

text/0202-events-and-logs-api.md
scheler: The idea was to make the end-users use these convenience methods right from the beginning instead of using the builders, which is really only for the advanced cases.

text/0202-events-and-logs-api.md
scheler: I think there is a misunderstanding here.  For logEvent, I meant methods like [these](https://github.com/scheler/opentelemetry-java/pull/1/files#diff-52068c2656fad153f3cd91781ec5028b30f400a258964282fb9b3b967a82f840R28). I would like to use the builder pattern only as a last option when I cannot create the Event using other methods.

jack-berg: I think its unnecessary to specify the existence of builder methods and / or overloads which accept a variety of arguments. Those types of decision should be left to the languages. The spec should simply indicate that there is a method for logging events, that the method requires a name argument, and optionally accepts a variety of other arguments in line with the fields on LogRecord.

Now speaking as a maintainer for java, I think we'll likely choose to include a single API (either `logEvent(Event)` or the builder pattern `logEvent(String).setAttributes(..).build()`), but not multiple forms of the same API for convenience.

text/0202-events-and-logs-api.md
jack-berg: For this comment I was referring only to omitting `recordException`. 

Why does a builder represent a more advanced use case than the convenience methods? Builders are heavily used through the otel java API and SDK. They allow the user to convey the same information as is provided in overloads with an API surface area that is smaller and easier to evolve.

text/0202-events-and-logs-api.md
martinkuba: If there was an API such as `logEvent(Event)`, then I think there would also have to be a way to create an instance of an `Event` through the API. I originally thought that this would have to be the case following the example of the Log SDK, which has `emit(LogRecord)` method.

I think we either need to keep this statement there, or the API should be just `logEvent(name, attributes)`.

text/0202-events-and-logs-api.md
weyert: There is also the use case of recording exceptions on  non browser environments, e.g. you could log the exception and also send it to something like Google Error Reporting at the same time

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Subtle differences between Logs and Events
+
+Logs have a mandatory severity level as a first-class parameter that events do not have, and events have a mandatory name that logs do not have. Further, logs typically have messages in string form and events have data in the form of key-value pairs. It is due to this that their API interface requirements are slightly different.
+
+## Who requires Events API
+
+Here are a few situations that require recording of Events, there will be more.
+
+- RUM events (Client-side instrumentation)
```
weyert, line 15: I assume this isn't covering the case of Google Analytics tracking events?

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,172 @@
...
+### LoggerProvider
+
+Logger can be accessed with an LoggerProvider.
+
+In implementations of the API, the LoggerProvider is expected to be the stateful object that holds any configuration.
+
+Normally, the LoggerProvider is expected to be accessed from a central place. Thus, the API SHOULD provide a way to set/register and access a global default LoggerProvider.
+
+Notwithstanding any global LoggerProvider, some applications may want to or have to use multiple LoggerProvider instances, e.g. to have different configuration (like LogRecordProcessors) for each (and consequently for the Loggers obtained from them), or because it's easier with dependency injection frameworks. Thus, implementations of LoggerProvider SHOULD allow creating an arbitrary number of instances.
+
```
tigrannajaryan, line 61: Is it worth adding a sentence that says that LoggerProvider/Logger are analogous to TracerProvider/Tracer to make the intent clear?

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,172 @@
...
+- version (optional): Specifies the version of the instrumentation scope if the scope has a version (e.g. a library version). Example value: 1.0.0.
+- schema_url (optional): Specifies the Schema URL that should be recorded in the emitted telemetry
+- event_domain (optional): Specifies the domain for the events created, which should be added in the attribute `event.domain` in the instrumentation scope.
+- include_trace_context (optional): Specifies whether the Trace Context should automatically be passed on to the events and logs created by the Logger. This SHOULD be false by default.
```
tigrannajaryan, line 77: Also need:
>attributes (optional): Specifies the instrumentation scope attributes to associate with emitted telemetry.

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+It is unspecified whether or under which conditions the same or different Logger instances are returned from this function.
```
tigrannajaryan, line 79: Please make sure the wording here matches what we have in [Get Tracer section](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#get-a-tracer) (it has changed recently after attributes were added).

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,172 @@
...
+Logger logger = openTelemetry.getLogger("another-scope");
+logger.logRecordBuilder().build().setBody("I am a log message").emit();
+
+```
+
+### Usage in Client-side telemetry
+
+```java
+public void addBrowserEvent(String name, Attributes attributes) {
+   Logger logger = openTelemetry.getLogger("my-scope", "1.0", "browser");
```
tigrannajaryan, line 126: Since we have multiple string parameters I think we should come up with a different to specify them otherwise it is not clear which one is which. It doesn't affect the OTEPs proposal but we need a better way in implementations.

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,172 @@
...
+**type:** `event`
+
+**Description:** Event attributes.
+
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `event.name` | string | Name or type of the event. | `network-change`; `button-click`; `exception` | Yes |
+| `event.domain` | string | Domain or scope for the event. | `profiling`; `browser`, `db`, `k8s` | No |
+
+An `event.name` is supposed to be unique only in the context of an `event.domain`, so this allows for two events in different domains to have same `event.name`. No claim is made about the uniqueness of `event.name`s in the absence of `event.domain`.
```
tigrannajaryan, line 151:
```suggestion
An `event.name` is supposed to be unique only in the context of an `event.domain`, so this allows for two events in different domains to have same `event.name`, yet be unrelated events. No claim is made about the uniqueness of `event.name`s in the absence of `event.domain`.
```

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+Note that Scope attributes are equivalent to the attributes at Span and LogRecord level, so recording the attribute `event.domain` on the Scope is equivalent to recording it on Spans and LogRecords within the Scope.
```
tigrannajaryan, line 153: I am not against this, but we need to clarify this behavior generally in the spec, not just for logs but also for traces and metrics.

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Causality on Events
```
tigrannajaryan, line 155: This entire section is not clear to me. It seems to be specifically about browser events?

tigrannajaryan:
Generally LGTM. This aligns well with what we discussed in the Log SIG.
I left some mostly minor comments.

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+In this OTEP we introduce an Events and Logs API that is based on the OpenTelemetry Log signal. The Events here refer to the independent Events and not to be confused with Span Events which occur only in the context of a span. In OpenTelemetry's perspective Log Records and Events are different names for the same concept - however, there is a subtle difference in how they are represented using the underlying data model that is described below. We will describe why the existing Logging APIs are not sufficient for the purpose of creating events.  It will then be evident that we will need an API in OpenTelementry for creating events.
+
+The Logs part of the API introduced here is supposed to be used only by the Log Appenders and end-users should continue to use the logging APIs available in the languages.
+
+## Subtle differences between Logs and Events
+
+Logs have a mandatory severity level as a first-class parameter that events do not have, and events have a mandatory name that logs do not have. Further, logs typically have messages in string form and events have data in the form of key-value pairs. It is due to this that their API interface requirements are slightly different.
```
cijothomas, line 9: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#severity-fields Severity fields are not mandatory. You may want to remove "mandatory" from this text.

cijothomas: On another round of reading, you can ignore the above comment. I realised you are not directly referring to LogDataModel, but the typical logging API which has a mandatory severity level in most languages.

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,172 @@
...
+- The current Log APIs do not have a standard way to pass event attributes.
+  - It may be possible to use the interpolation string args as the parameter to pass event attributes. However, the logging spec seems to map the human readable message (which is obtained after replacing the args in the interpolated string) to the Body field of LogRecord.
+  - Log4j has an EventLogger interface that can be used to create structured messages with arbitrary key-value pairs, but log4j is not commonly used in Android apps as it is not officially supported on Android as per this [Stack Overflow thread](https://stackoverflow.com/questions/60398799/disable-log4j-jmx-on-android/60407849#60407849) by one of log4j’s maintainers.
+  - In Python, logging.LogRecord's extra field is mapped to Otel LogRecord's attributes but this field is a hidden field and not part of the public interface.
+- The current Log APIs have a message parameter which could map to the Body field of LogRecord. However, this is restricted to String messages and does not allow for structured logs.
+
+For the above reasons we can conclude that we will need an API for generating Events API
```
cijothomas, line 33:
```suggestion
For the above reasons we can conclude that we will need an API for generating Events.
```

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+An `event.name` is supposed to be unique only in the context of an `event.domain`, so this allows for two events in different domains to have same `event.name`. No claim is made about the uniqueness of `event.name`s in the absence of `event.domain`.
+
+Note that Scope attributes are equivalent to the attributes at Span and LogRecord level, so recording the attribute `event.domain` on the Scope is equivalent to recording it on Spans and LogRecords within the Scope.
+
+## Causality on Events
+
+For creating causality between events we can create wrapper spans that are part of the same trace. However, note that the events themselves are represented using LogRecords and not as Span Events.
+
+```java
```
cijothomas, line 159: would be nice to show "include_trace_context" usage here, as without that, there is nothing connecting the "wrapper spans" to the LogRecords.

cijothomas (approved):

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+### LoggerProvider
+
+Logger can be accessed with an LoggerProvider.
+
+In implementations of the API, the LoggerProvider is expected to be the stateful object that holds any configuration.
+
+Normally, the LoggerProvider is expected to be accessed from a central place. Thus, the API SHOULD provide a way to set/register and access a global default LoggerProvider.
+
+Notwithstanding any global LoggerProvider, some applications may want to or have to use multiple LoggerProvider instances, e.g. to have different configuration (like LogRecordProcessors) for each (and consequently for the Loggers obtained from them), or because it's easier with dependency injection frameworks. Thus, implementations of LoggerProvider SHOULD allow creating an arbitrary number of instances.
```
dyladan, line 60: It says near the top that the logging API is only for use by long appenders, not for end-users, but this paragraph seems to imply that end-users will actually use it.

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,172 @@
...
+// Using the event builder to log an event
+logger.eventBuilder("page-navigated").build().setAttribute("url", "http://foo.com/bar#new").emit();
+
+// Using the logRecord builder to log a record
+Logger logger = openTelemetry.getLogger("another-scope");
+logger.logRecordBuilder().build().setBody("I am a log message").emit();
+
+```
+
+### Usage in Client-side telemetry
```
dyladan, line 122: I see the example code and it looks fine but again, I would have hoped to see a description here of when the event API should be used in client-side telemetry and the value provided. This again looks more like a draft spec to me than an OTEP

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+public void addMobileEvent(String name, Attributes attributes) {
+   Logger logger = openTelemetry.getLogger("my-scope", "1.0", "mobile");
+   logger.logEvent(name, attributes);
+}
+```
+
+### Usage in eBPF
+
+From the eBPF [demo](https://youtu.be/F1VTRqEC8Ng?t=233), it looks like they have chosen to put the Event data in the Body field of LogRecord instead of Attributes.  They might have to make the changes to move the event data to attributes to conform to this API specification.
```
dyladan, line 138: The wording of this paragraph makes it seem like a critique of eBPF. Unless you are already very familiar with the Flowmill/otel relationship this seems like a link to a random video. Are there other components which will be changed if this OTEP is accepted?

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Semantic Convention for event attributes
```
dyladan, line 140: This section is extremely specific. At the risk of repeating myself, this reads more like a draft spec than an OTEP. I see rigid definitions but no arguments for the value or usefulness of any of the specified features/requirements.

dyladan (commented):
I would almost argue for the removal of the `Events and Logs API Interface` section here and replacing it with use cases and motivations. I see very little about what these APIs are expected to be used for, alternatives that were considered, and why they are not sufficient.

text/0202-events-and-logs-api.md
scheler: cc: @jack-berg for ideas - maybe the parameters all can go into a separate class for logger configuration.

text/0202-events-and-logs-api.md
scheler: Removed the entire section as it seems out of place here. For context, when we originally considered 0-duration spans for representing standalone events, spans served well to represent causality between the events. With LogRecords replacing Spans for standalone events with this proposal, we still cannot represent causality with just LogRecords. That is where I added this section suggesting use of wrapper spans for the purpose of show causality between standalone events.

text/0202-events-and-logs-api.md
scheler: > I would almost argue for the removal of the `Events and Logs API Interface` section here and replacing it with use cases and motivations. I see very little about what these APIs are expected to be used for, alternatives that were considered, and why they are not sufficient.

@dyladan I see your point and can add more content in the OTEP on use-cases and motivations. It was discussed in the calls but if you feel it is not represented enough here I will address that.

text/0202-events-and-logs-api.md
scheler: One other component is the [k8seventsreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/k8seventsreceiver/k8s_event_to_logdata.go#L43) that needs to add the `event.name` attribute to the events.  I am not aware of any other cases.

text/0202-events-and-logs-api.md
scheler: @dyladan I added some text in this commit to address some of what you asked but it's not elaborate. The primary motivation of this OTEP is really to distinguish between Events and Logs, both represented using the `LogRecord` data model, via semantic conventions and the API is to enforce those conventions - that's why there's not really much to elaborate on use-cases and motivation. So far what we agreed on during the discussion on this OTEP is having an `event.name` attribute to identify a `LogRecord` as being an `Event`, and that the event data is mostly in the attributes.

text/0202-events-and-logs-api.md
scheler: I added it back giving this context.

text/0202-events-and-logs-api.md
scheler: The LoggerProvider is common to both Events and Logs.

text/0202-events-and-logs-api.md
scheler: > when the event API should be used in client-side telemetry and the value provided

It is for creating standalone events that occur when there is no span in progress. I updated the text with this.

text/0202-events-and-logs-api.md
scheler: It is for recording any standalone events that may occur even when there is no span in progress.

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,187 @@
...
+- The current Log APIs have a message parameter which could map to the Body field of LogRecord. However, this is restricted to String messages and does not allow for structured logs.
+
+For the above reasons we can conclude that we will need an API for generating Events.
+
+## Should OpenTelemetry have an API for logs?
+
+A part of the OTel community thinks that we should not have a full-fledged logging API unless there is a language that doesn't already have a plethora of logging libraries & APIs to choose from where it might make sense to define one. Further, we will not be able to have the [rich set of configuration options](https://logging.apache.org/log4j/2.x/manual/configuration.html) that some popular logging frameworks provide so the logging API in OTel will only become yet another API. However, it was noted that the Log Appender API is very similar to the API for Events and so instead of having API for Events and API for Log Appenders separately it was agreed to have one API for Events and Logs, and that the API for Logs is targeted only to Log Appenders.
```
martinkuba, line 42: I think the main argument for a combined log/event API was **consistency** with other signals - single API for each signal.

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,187 @@
...
+
+- `version` (optional): Specifies the version of the instrumentation scope if the scope has a version (e.g. a library version). Example value: 1.0.0.
+- `schema_url` (optional): Specifies the Schema URL that should be recorded in the emitted telemetry
+- `event_domain` (optional): Specifies the domain for the events created, which should be added in the attribute `event.domain` in the instrumentation scope.
+- `include_trace_context` (optional): Specifies whether the Trace Context should automatically be passed on to the events and logs created by the Logger. This SHOULD be false by default.
```
martinkuba, line 84: This could be an implementation detail of the SDK. Is there a reason why this is part of the API? (some SDKs may not be able to make these links)

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,187 @@
...
+- `attributes` (optional): Specifies the instrumentation scope attributes to associate with emitted telemetry.
+
+It is unspecified whether or under which conditions the same or different Logger instances are returned from this function.
```
martinkuba, line 87: Is this statement still valid, since underneath it says that different instances must be returned?

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,187 @@
...
+The Logger is responsible for creating Events and Logs.
+
+Note that Loggers should not be responsible for configuration. This should be the responsibility of the LoggerProvider instead.
+
+#### Logger operations
+
+The Logger MUST provide functions to:
+
+- Create an `Event` and emit it to the processing pipeline.
+  - The API MUST accept an event name as a parameter. The event name provided should be inserted as an attribute with key `event.name`. Care MUST be taken to not override or delete this attribute while the `Event` is created.  This function MAY be named `logEvent`.
```
martinkuba, line 108: > Care MUST be taken to not override or delete this attribute while the `Event` is created.

Does this mean that if an attribute `event.name` is included in the list of attributes, it should be discarded in favor of the `name` provided through the API?

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,187 @@
...
+**Description:** Event attributes.
+
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `event.name` | string | Name or type of the event. | `network-change`; `button-click`; `exception` | Yes |
+| `event.domain` | string | Domain or scope for the event. | `profiling`; `browser`, `db`, `k8s` | No |
+
+An `event.name` is supposed to be unique only in the context of an `event.domain`, so this allows for two events in different domains to have same `event.name`, yet be unrelated events. No claim is made about the uniqueness of `event.name`s in the absence of `event.domain`.
+
+Note that Scope attributes are equivalent to the attributes at Span and LogRecord level, so recording the attribute `event.domain` on the Scope is equivalent to recording it on Spans and LogRecords within the Scope.
```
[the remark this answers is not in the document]
scheler: It may be added later if there is a need. We will skip this initially.

text/0202-events-and-logs-api.md
scheler: What choice will the SDK make, as there are use-cases for both? That is the reason we are leaving it to the user to make the choice using this flag.

text/0202-events-and-logs-api.md
scheler: Yes.

text/0202-events-api.md
scheler: > If this is the case it might make more sense to instead have the Log spec say that in those cases Otel logging should provide an API, instead of adding an Events API.

@tsloughter I want to leave the otep at proposing the Logs API only for log appenders initially. If the language maintainers feel the need for an end-user API for structured logs in OTel, they can add that incrementally later. Let me know what you think.

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+This API MUST accept the following parameters:
+
+- name (required): This name SHOULD uniquely identify the instrumentation scope, such as the instrumentation library (e.g. io.opentelemetry.contrib.mongodb), package, module or class name.  If an application or library has built-in OpenTelemetry instrumentation, both Instrumented library and Instrumentation library may refer to the same library. In that scenario, the name denotes a module name or component name within that library or application. In case an invalid name (null or empty string) is specified, a working Logger implementation MUST be returned as a fallback rather than returning null or throwing an exception, its name property SHOULD be set to an empty string, and a message reporting that the specified value is invalid SHOULD be logged. A library implementing the OpenTelemetry API may also ignore this name and return a default instance for all calls, if it does not support "named" functionality (e.g. an implementation which is not even observability-related). A LoggerProvider could also return a no-op Logger here if application owners configure the SDK to suppress telemetry produced by this library.
```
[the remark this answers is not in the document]
scheler: Not currently, but I added a note that most of these are common with the scope parameters for `Tracer` and `Meter`, except for `event_domain` and `include_trace_context` which is specific to `Logger`.

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+- version (optional): Specifies the version of the instrumentation scope if the scope has a version (e.g. a library version). Example value: 1.0.0.
+- schema_url (optional): Specifies the Schema URL that should be recorded in the emitted telemetry
+- event_domain (optional): Specifies the domain for the events created, which should be added in the attribute `event.domain` in the instrumentation scope.
```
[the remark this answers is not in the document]
scheler: It is defined in the `Semantic Convention for event attributes` section.

text/0202-events-and-logs-api.md
scheler: @dyladan Regarding the inclusion of `Events and Logs API Interface` section in this OTEP, we did that only to keep things together for easier understanding of the contents. Apparently, few other OTEPs have done this in the past too. Let me know if you feel strongly about its removal from the OTEP.

text/0202-events-and-logs-api.md
scheler: Removed this statement.

text/0202-events-and-logs-api.md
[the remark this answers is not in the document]
dyladan: The definition should be either at the place of first use, or in a specific definitions section you can link to. Reading the document top to bottom should not be confusing.

text/0202-events-and-logs-api.md
dyladan: I don't think the example code is providing anything here. OTEPs should focus on motivations and leave implementation detail to the spec and SDKs.

text/0202-events-and-logs-api.md
dyladan: Most of my comment is not addressed. The paragraph reads like a comment in a thread or an answer to a question or something. Nothing in the paragraph says anything about when the event API might be useful for someone from an eBPF perspective and how it is different than alternatives.

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,187 @@
...
+```java
+OpenTelemetry openTelemetry = OpenTelemetry.noop();
+Logger logger = openTelemetry.getLogger("my-scope");
+
+// Using the convenience method to log an event directly
+logger.logEvent("network-changed", 
+                 Attributes.builder().put("type", "wifi").build());
+
+// Using the event builder to log an event
+logger.eventBuilder("page-navigated").build().setAttribute("url", "http://foo.com/bar#new").emit();
```
jack-berg, line 122:
```suggestion
logger.eventBuilder("page-navigated").setAttribute("url", "http://foo.com/bar#new").emit();
```

text/0202-events-and-logs-api.md
```diff
@@ -0,0 +1,187 @@
...
+
+// Using the logRecord builder to log a record
+Logger logger = openTelemetry.getLogger("another-scope");
+logger.logRecordBuilder().build().setBody("I am a log message").emit();
```
jack-berg, line 126:
```suggestion
logger.logRecordBuilder().setBody("I am a log message").emit();
```

text/0202-events-and-logs-api.md
jack-berg: The java implementation will use a builder pattern for this, in line with [traces](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/trace/TracerBuilder.java) and [metrics](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/metrics/MeterBuilder.java), which each argument getting its own named setter.

text/0202-events-and-logs-api.md
jack-berg: Can probably just omit this paragraph entirely.

jack-berg (approved):
Some additional comments but none are blocking

text/0202-events-and-logs-api.md
martinkuba: Is there a reason why the user of the API may not want to include the trace context (if it is available)?

scheler: Consider a user clicking on a random button that results in an error, while an XHR call is in progress. You do not want to record the error event with the trace context of the Span for the XHR call, as they are not related.

scheler:
It makes sense, I will make the changes.

text/0202-events-and-logs-api.md
scheler: I've now modified the OTEP following the OTEP template and also removed the spec aspects now. 

@jack-berg I followed the example from https://opentelemetry.io/docs/instrumentation/java/manual/#tracing which is not using a builder pattern.

```
Tracer tracer =
    openTelemetry.getTracer("instrumentation-library-name", "1.0.0");
```

scheler:
@dyladan @tigrannajaryan @jack-berg @martinkuba @MSNev I've made changes to follow the OTEP template and also removed the spec aspects. The  draft spec content is at https://github.com/scheler/opentelemetry-specification/blob/events-and-logs-api/specification/logs/events-and-logs-api.md and I will submit separate PR later.

text/0202-events-and-logs-api.md
martinkuba: They would not be associated - in JS the click handler would be seen as separate code path (the async context propagation handles this), or in sync code like Java it would be on a separate thread.

text/0202-events-and-logs-api.md
scheler: Some do - 
Log4j's EventLogger supports this via `id` in the [StructuredDataMessage](https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/message/StructuredDataMessage.html).
The Web API's Event has it via [type](https://developer.mozilla.org/en-US/docs/Web/API/Event/Event)

text/0202-events-and-logs-api.md
scheler: Yes, I removed it. Apparently, Flowmill is moving to using OTel metrics to report their telemetry data and will not be Log data model.

text/0202-events-and-logs-api.md
scheler: Added this point.

dyladan:
Possibly it is covered somewhere in the conversation but why do we use the term "domain" for event instead of the term "scope" used by other signals?

dyladan:
> > Possibly it is covered somewhere in the conversation but why do we use the term "domain" for event instead of the term "scope" used by other signals?
> 
> Scope is a runtime concept. It is [defined](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md?rgh-link-date=2022-06-28T13%3A01%3A10Z) as a place in the code:
> 
> > A logical unit of the application code with which the emitted telemetry can be associated. It is typically the developer's choice to decide what denotes a reasonable instrumentation scope. The most common approach is to use the [instrumentation library](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md?rgh-link-date=2022-06-28T13%3A01%3A10Z#instrumentation-library) as the scope, however other scopes are also common, e.g. a module, a package, or a class can be chosen as the instrumentation scope.
> 
> Domain is not yet defined in the spec, we need a proper definition for it. The purpose of the domain is to group event definitions in semantic conventions such that event names only within the domain are unique, but between the domains the event names may overlap. The domain is a mechanism for namespacing event names. A domain can be allocated to a SIG and they are free to decide event names within the domain, knowing that any event names they choose are not going to be ambiguous and conflict with the event names in other domains. An event's semantic convention is essentially identified by a (domain.name, event.name) tuple.

Thanks for the description. I didn't understand that from the existing definition of domain in the OTEP so it might be helpful to add some details there. Beyond that this OTEP looks good to me.

zenmoto (approved):

djaglowski (approved):

MSNev (approved):

tigrannajaryan (approved):
The OTEP no longer contains the details that describe the API precisely and only contains the general description of the intent to add one.
I think this is fine, we do allow OTEPs which are an "intent", with a purpose to seek alignment in Otel community.
I agree with the intent and that's why I am approving the OTEP. The details of the Logs and Events API can be either a follow-up OTEP or direct PRs to the spec repo.

tigrannajaryan:
@open-telemetry/specs-logs-approvers please review.

text/0202-events-and-logs-api.md
scheler: We can discuss this further during the spec changes, resolving for now.

dyladan:
> The OTEP no longer contains the details that describe the API precisely and only contains the general description of the intent to add one. I think this is fine, we do allow OTEPs which are an "intent", with a purpose to seek alignment in Otel community. I agree with the intent and that's why I am approving the OTEP. The details of the Logs and Events API can be either a follow-up OTEP or direct PRs to the spec repo.

I think we should clarify what is meant to be in an OTEP and how far it is acceptable to diverge from the OTEP when writing spec. The specification parts of this OTEP were removed because of concerns that it was too prescriptive and didn't leave room for the spec to make any decisions if something came up in a language SIG which required a change. If we're just going to add those things back in a follow-up OTEP, then maybe they should be added back to this one (with my apologies to @scheler and others who removed those at least partially due to my comments)?

martinkuba (approved):

text/0202-events-and-logs-api.md
scheler: let's take this up later during the spec changes.

text/0202-events-and-logs-api.md
scheler: Removed the API interface section.

tigrannajaryan:
@open-telemetry/specs-approvers @open-telemetry/specs-trace-approvers @open-telemetry/specs-logs-approvers @open-telemetry/specs-metrics-approvers  We have enough approvals to merge. I am going to let this open for a couple more days, but unless I see objections I am going to merge this OTEP and proceed to defining/adding the new Events and Logs API to Otel specification.

tigrannajaryan:
No new comments, no objections for the last 5 days. Merging.

PR #204, opened 2022-05-20
Include OTLP proto version identifier in requests

MikeGoldsmith:
Adds OTLP request identifer to descrive the OTLP version used to create the binary or JSON payload.

### DIFF
--- a/text/0000-otlp-proto-version-identifier.md
+++ b/text/0000-otlp-proto-version-identifier.md
@@ -1,5 +1,3 @@
-
-
 # Include OTLP proto version identifier in requests
 
 Adds OTLP request identifer to descrive the OTLP version used to create the binary or JSON payload.
--- a/text/0000-otlp-proto-version-identifier.md
+++ b/text/0000-otlp-proto-version-identifier.md
@@ -1,0 +1,33 @@
+
+
+# Include OTLP proto version identifier in requests
+
+Adds OTLP request identifer to descrive the OTLP version used to create the binary or JSON payload.
+
+## Motivation
+
+It is useful to know the version of the OTLP protobuf definition used to create the binary or JSON.
+
+This would enable routing to different services by inspecting the header value. This is especially useful for organisations that wish to support multiple versions where the OTLP proto definitions have changed. A currently used alternative is to append the version to the path, but this is clumsy and requires additional configuration when setting up the exporter.
+
+## Explanation
+
+Currently it is not possible to know the version of the OTLP proto definition before receving it. This is challening because there have been changes to the OTLP proto specification that are breaking.
+
+This also aids organisations that receive multiple streams of OTLP data and be able to monitor and encourage users to upgrade to newer versions as appropriate.
+
+## Internal details
+
+* Add gRPC metadta entry or HTTP request header entry with the OTLP version used to create the binary or JSON payload
+
+For example, we could use the key `x-otlp-version` and the string value of the OTLP proto version in the form "0.17.0".
+
+## Trade-offs and mitigations
+
+Currently releaed SDKs and collector versions will not include these identifiers.
+
+## Prior art and alternatives
+
+## Open questions
+
+## Future possibilities


text/0000-otlp-proto-version-identifier.md
```diff
@@ -0,0 +1,31 @@
...
+
+* Add gRPC metadta entry or HTTP request header entry with the OTLP version used to create the binary or JSON payload
+
+For example, we could use the key `x-otlp-version` and the string value of the OTLP proto version in the form "0.17.0".
+
+## Trade-offs and mitigations
+
+Currently releaed SDKs and collector versions will not include these identifiers.
+
+## Prior art and alternatives
```
tigrannajaryan, line 27: Previously people asked for the version number in the payload a few times. Can you consider this alternate and why in the header and not in the payload (or vice versa)?

text/0000-otlp-proto-version-identifier.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Motivation
+
+It is useful to know the version of the OTLP protobuf definition used to create the binary or JSON.
+
+This would enable routing to different services by inspecting the header value. This is especially useful for organisations that wish to support multiple versions where the OTLP proto definitions have changed. A currently used alternative is to append the version to the path, but this is clumsy and requires additional configuration when setting up the exporter.
+
+## Explanation
+
+Currently it is not possible to know the version of the OTLP proto definition before receving it. This is challening because there have been changes to the OTLP proto specification that are breaking.
```
jmacd, line 13: As an example, *even between stable OTLP versions*, if a 0.16 OTLP handler receives data encoded for v0.9, it will receive data that is no longer recognized (IntSum, IntGauge, IntHistogram, and Labels fields have all been removed by now). The upgrade is dangerous because the payload has no metadata to indicate its version, so once I remove the v0.9 handler I'll have no way to tell data is being dropped.

jmacd (approved):

text/0000-otlp-proto-version-identifier.md
MikeGoldsmith: Add considered request body resource attribute in alternatives. Headers offer more flexibility in routing the request as they body does not need to be decoded to get the value.

MikeGoldsmith: Hey @tigrannajaryan - sorry for delay in replying.

Encoding the OTLP version within the payload doesn't feel like a good solution as we'd be forcing consumers who do care about the encoding to decode twice, first to get the OTLP version and again to decode using the appropriate OTLP handler. OTLP request object can be large and it's very inefficient the whole rquest to potentially throw away. Not only that, many gRPC handler frameworks don't even offer you the raw binary payload and handle the decoding for you.

For the use case of OTLP requests that have been serialised to disk (eg binary / JSON), do we have some way to describe the metadata associated to the request payload when it was received?

For example the structure could be something like this in JSON:

```json
{
  "metadata": {
    "otlp-version": "0.17.0",
  },
  "request": {
    ... JSON for request
  }
```

Is there a stable format request data to use when serialising to disk?

MikeGoldsmith: You're right - the original purpose of the OTLP version being available was only to be used for informational reasons (eg identify users who send which versions and promote them to upgrade as appropriate).

However, we are in a situation where there have seen some breaking changes in the 0.x versions of the proto definitions, particularly around the fairly aggressive removing of deprecated types / fields. A good example of this is the deprecated Int/Float metrics that were removed in proto version 0.12 and new fields has been added to ScopeXXX in 0.17. If deprecated fields had been left in the proto files until the 1.0 release, they likely would not have caused this problem.

I also believe the use-case originally shared where some receivers wish to route requests based on the OTLP version is still valid and would be best served via a gRPC metadata entry or HTTP header. This would be possible if the OTLP version was part of the request payload, but would require the routing process to decode the whole body to get the version and potentially route to handler which would then decode again.

Why do you feel having the version both in metadata/header and request body field be an issue?

MikeGoldsmith: To follow-up on this;

I think the OTLP version should be treated as metadata for the given request and should be stored separately from the payload data. For example, if you're storing requests for playback at a later time, you'd likely want to record other useful information such as the sender (eg user agent header) and the time it was received. I think the OTLP version also falls into the category of useful info, that helps you understand and process the request payload.

MikeGoldsmith: @tigrannajaryan @jmacd @bogdandrutu Any further feedback on this?

text/0000-otlp-proto-version-identifier.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Motivation
+
+It is useful to know the version of the OTLP protobuf definition used to create the binary or JSON, without having to decode the payload.
+
+This would enable routing to different OTLP handlers by inspecting the identifier value. This is especially useful for organisations that wish to support multiple OTLP versions where the proto definitions have changed. A currently used alternative is to append the version to the path, but this is clumsy and requires additional configuration when setting up the exporter.
+
+## Explanation
+
+Currently it is not possible to know the version of the OTLP proto definition before receiving it. This is challenging because there have been changes to the OTLP proto specification that are breaking.
```
tigrannajaryan, line 13: I think this phrasing indirectly normalizes the fact that there may be breaking changes, which may contribute to our sloppiness and indirectly cause breakage. I think we should do the opposite and emphasize even stronger that breaking changes are extremely bad and we should never do them without officially introducing a new major OTLP version (in which case the version number outside the payload may be introduced).

My opinion is still that this OTEP and the version number outside the payload moves us in the wrong direction.
Instead I think we should make a bigger effort to avoid breaking changes, for example by adding tooling that catches it, by making it clearer in our spec/docs what changes constitute breaking changes and should avoided, etc.

MikeGoldsmith: I understand and fully appreciate the intention to avoid breaking changes to the wire format. However, we've just seen another _breaking_ change be merged into the [OTLP spec](https://github.com/open-telemetry/opentelemetry-proto/pull/407).

Removing deprecated fields like these forces OTLP receivers to chose between either supporting senders who have not upgraded to a newer version of the OTLP spec (<0.12) or supporting senders who have upgraded(0.19+). Not everyone has the luxury of dictating when senders upgrade. These deprecated fields could have existed in the spec until a major release (eg 1.0) and receivers can choose to support them if they wish to.

A solution to the above is for receivers to utilise multiple handlers using different OTLP formats. However, as we've discussed previous, it's cumbersome and inefficient to decode a request payload to get the version to then pass onto one of these handlers which will likely decode again. An identifier outside of the payload allows receivers to decide how they handle the request before using resources on decoding.

robbkidd: I see a "Why not both?" solution. 

We should certainly strive for all the bigger, ideal efforts @tigrannajaryan describes: avoid proto breakage, more tooling to detect breakage. If we also include the proto version in the request and outside of the body (an `x-otlp-version` header? `otlp/1.2.3` included in [exporter user_agent](https://github.com/open-telemetry/opentelemetry-specification/pull/2684)?), receivers of OTLP have easy access to the data necessary to troubleshoot communications problems when we haven't lived up to that ideal.

tedsuo:
@MikeGoldsmith  we are marking OTEPs which are no longer being updated as `stale`. We will close this PR in one week. If you wish to resurrect this PR, that is fine.

PR #192, opened 2021-12-02
Stabilize messaging semantic conventions for tracing

pyohannes:
This OTEP aims to describe the necessary changes for bringing the [existing semantic conventions for messaging](https://github.com/open-telemetry/opentelemetry-specification/blob/a1a8676a43dce6a4e447f65518aef8e98784306c/specification/trace/semantic_conventions/messaging.md) from the current [experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#experimental) to a [stable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#stable) state.

It is based on [OTEP 0173](0173-messaging-semantic-conventions.md), which defines basic terms and describes messaging scenarios that should be supported by the semantic conventions.

**NOTE: This is an early draft document. It captures results of discussions currently going on in the messaging workgroup.**

PR #205, opened 2022-06-02
Context propagation requirements for messaging semantic conventions

pyohannes:
This PR is a result of discussions in the messaging workgroup and has been split off #192 for easier review.

The [existing messaging semantic conventions for tracing](https://github.com/open-telemetry/opentelemetry-specification/blob/a1a8676a43dce6a4e447f65518aef8e98784306c/specification/trace/semantic_conventions/messaging.md) implicitly impose certain requirements on context propagation mechanisms used: while context is usually propagated on a per-request basis, the messaging semantic conventions assume that context is propagated on a per-message basis.

This OTEP proposes an addition to the existing messaging semantic conventions to make this requirement explicit. This is very important for reasoning about design questions regarding trace and span structure. Furthermore, explicating those requirements is of crucial importance for coming up with standardized context propagation mechanisms for messaging protocols. Some drafts for such standardization exist ([AMQP](https://w3c.github.io/trace-context-amqp/), [MQTT](https://w3c.github.io/trace-context-mqtt/), [CloudEvents](https://github.com/cloudevents/spec/blob/v1.0.1/extensions/distributed-tracing.md)), and the changes in this PR should help to verify those drafts against the requirements of OpenTelemetry semantic conventions.

The addition proposed in this document does not break or invalidate any of the existing semantic conventions for messaging, but rather makes an implicit requirement explicit.


lmolkova (approved):

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,173 @@
...
+       |
+  +----------+
+  | Consumer |
+  +----------+
+```
+
+In this scenario, it is not possible to correlate producer and consumer traces
+for a given message by using HTTP context propagation mechanisms and without
+special broker instrumentation. Both HTTP requests, the one that publishes and
+the one that pulls the message, originate from different traces. There is no
```
yurishkuro, line 81: > originate from different traces

Not clear why this is. What is the different "trace" on the consumer side? What workflow does that trace represent?

pyohannes: The trace on the consumer side represents the workflow of fetching a message from the intermediary and possibly processing it.

The consumer sends an HTTP GET request to the intermediary to receive the message. This request will propagate context from the consumer to the intermediary, but it doesn't propagate any context from the intermediary (or producer) to the consumer. So, via plain HTTP context propagation, the consumer has no way of obtaining context information from the intermediary (or producer), and thus cannot create parent/child or link relationships with the producer trace.

yurishkuro: It can certainly be treated as a trace (sans the processing of the message, that makes less sense), but as you describe later those are not the same category of traces, you call it "transport" trace.

So I think this whole use case could be made much clearer by stating that the problem is with potentially overlapping workflows where more than one trace could be active during specific operations. I wanted to say "during the same tracepoints", but that does not seem to be the case, because each of the concurrent traces uses their own tracepoints, you explicitly mention that saying that message broker HTTP exchanges do not need to be recorded in the logical message processing call graph.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,173 @@
...
+   This layer helps to gain insights into details of the message transport.
+
+This would keep the existing correlation between producers and consumers intact
+while allowing intermediaries to use the transport context to correlate
+intermediary instrumentation with existing producer and consumer
+instrumentations.
+
+### Standards for context propagation
+
+Currently, instrumentors have to decide how to attach and extract context from
```
joaopgrassi, line 156:
```suggestion
Currently, instrumentations have to decide how to attach and extract context from
```

Not sure if we have `instrumentor` in our glossary or if it's a known term in OTel? Maybe we should check if this is OK to use, to avoid confusion?

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,173 @@
...
+messages to fulfil the [requirements for context propagation](#context-propagation).
+While preserving the freedom for instrumentors to choose how to propagate
+context, in the future these conventions should list recommended ways of how to
+propagate context using popular messaging protocols.
```
joaopgrassi, line 160:
```suggestion
propagate context using well-established messaging protocols.
```
Maybe? I feel `popular` is a loaded word that can mean different things for different people :)

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,173 @@
...
+
+Currently, several attempts exist to standardize context propagation for different
```
joaopgrassi, line 162:
```suggestion
There are several working-in-progress efforts to standardize context propagation for different
```
Maybe?

joaopgrassi: This way, we could also drop the  `Those standards are in draft states and/or are not widely adopted yet` part below, since it's implicit that if they are "work-in-progress" that they are drafts and not adopted.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,173 @@
...
+messaging protocols and scenarios:
+
+* [AMQP](https://w3c.github.io/trace-context-amqp/)
+* [MQTT](https://w3c.github.io/trace-context-mqtt/)
+* [CloudEvents via HTTP](https://github.com/cloudevents/spec/blob/v1.0.1/extensions/distributed-tracing.md)
+
+Those standards are in draft states and/or are not widely adopted yet. It is
+planned to drive those standards to a stable state and to make sure they cover
```
joaopgrassi, line 170: This part is a bit confusing to me. I think it's trying to say 2 things in the same paragraph. 

>  It is planned to drive those standards to a stable state and to make sure they cover requirements put forth by these semantic conventions

Who is exactly involved in "planned to drive those standards"? Is it us - OTel?  If so, then maybe making it more clear like:

> OpenTelemetry is going to help drive these standards, making sure they are compatible and cover the requirements put forth by these semantic conventions

And: 

> Finally, these semantic conventions should give a clear and stable recommendation for each protocol and scenario

I think this is a bit redundant? The paragraph above `in the future these conventions should list recommended ways..` says the same thing. I guess we can still keep this as a more "strong" assertion, but rephrase it a bit and make it a separated statement from the above:

> Once the standards reach a stable state, these semantic conventions will give a clear and stable recommendation for each protocol and scenario.

WDYT?

>

joaopgrassi (approved):
I think this overall makes sense. I left a few minor things, mostly around wording and structure of text. Feel free to address them or not.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
pyohannes: The glossary contains the term "instrumentation author". I'll use that.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,173 @@
...
+### Transport context propagation
+
+As illustrated in the [example](#example) above, a creation context can be
+attached to a message, while different contexts are propagated via HTTP requests
+publishing and fetching the message. When coming up with conventions and
+guidance for intermediary instrumentation, it will be beneficial to clearly
+outline those two layers of context propagation and build conventions for
+intermediary instrumentation on top of this outline:
+
+1. The _creation context layer_ allows correlating the producer with the
```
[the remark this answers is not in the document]
pyohannes: Some people in the workgroup expressed a preference for "creation context" (as opposed to "application context"). I don't have a strong opinion on that, if people think "application context" is clearer, we can use that name.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,173 @@
...
+  | Consumer |
+  +----------+
+```
+
+In this scenario, it is not possible to correlate producer and consumer traces
+for a given message by using HTTP context propagation mechanisms and without
+special broker instrumentation. Both HTTP requests, the one that publishes and
+the one that pulls the message, originate from different traces. There is no
+way the context attached to the HTTP POST request publishing the message can be
+forwarded by the HTTP GET request fetching a message.
```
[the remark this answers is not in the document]
pyohannes: I would need to be forwarded _to_ the consumer, but it would need to be forwarded _by_ the HTTP GET request, when only using HTTP context propagation (without messaging specific context propagation).

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,173 @@
...
+
+A producer SHOULD attach a creation context to each message. The creation context
+SHOULD be attached in a way so that it is not possible to be changed by intermediaries.
+
+## Future possibilities
+
+### Transport context propagation
+
+As illustrated in the [example](#example) above, a creation context can be
+attached to a message, while different contexts are propagated via HTTP requests
```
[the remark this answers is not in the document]
pyohannes: You are right. I removed the reference to the example.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Terminology
+
+For terms used in this document, refer to [OTEP 173](#0173-messaging-semantic-conventions.md#terminology).
+
+## Motivation
+
+The current [messaging semantic conventions for tracing](https://github.com/open-telemetry/opentelemetry-specification/blob/a1a8676a43dce6a4e447f65518aef8e98784306c/specification/trace/semantic_conventions/messaging.md)
+provide a list of [examples](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#examples),
+which illustrate how producer and consumer spans can be correlated by
```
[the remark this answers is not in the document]
pyohannes: I split the sentence up to make clear this refers to individual examples.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,173 @@
...
+
+A message may pass many different components and layers in one or more
+intermediaries when it is propagated from the producer to the consumer. It
+cannot be assumed, and in many cases, it is not even desired, that all those
+components and layers are instrumented and propagate context according to
+OpenTelemetry requirements.
+
+A _creation context_ allows correlating the producer with the consumers of a
+message, regardless of intermediary instrumentation. The creation context is
+created by the producer and must be propagated to the consumers. It must not be
```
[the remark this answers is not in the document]
pyohannes: I removed the must. The strongest we want to use is SHOULD.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
pyohannes: > Who is exactly involved in "planned to drive those standards"? Is it us - OTel?

No, those standards are outside OTel. However, people working on those are counting on OTel to validate whether the specification the came up with makes sense and can be stable.

I agree this paragraph can be confusing. I will simplify it and just go with the last suggestion you made:

> Once the standards reach a stable state, these semantic conventions will give a clear and stable recommendation for each protocol and scenario.

lmolkova: BTW, I know we're not defining the standards, and having them defined would take a lot of time. Can we still have some soft recommendations of what people should do in absence of those?

E.g. "while the standards are being defined, we suggest that instrumentations that follow this convention use [Propagator API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#propagators-api) to propagate context using default or user-configured propagator."

Maybe we can also discuss it at the Thu meeting - I think we need to understand how it can evolve after AMQP/MQTT and other standards come into play and new instrumentations will have to stay backward-compatible.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
pyohannes: > So I think this whole use case could be made much clearer by stating that the problem is with potentially overlapping workflows where more than one trace could be active during specific operations.

The problem is with providing the context information from the producer to the consumer, so the consumer can create link or parent/child relationships with the consumer.

I will try to improve the example so this becomes clearer.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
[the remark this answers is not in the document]
joaopgrassi: We could try getting a consensus during the next meeting maybe? I think this is an important concept and a clear and easy to understand name could go a long way in helping folks grasp what we are suggesting.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
pyohannes: @lmolkova We can discuss this on Thursday.

From my point of view, this PR makes requirements for context propagation clear, but it doesn't give any recommendations for context propagation on top of that. As at this point we cannot give any stable recommendations (as currently none of the standard messaging protocols has a stable specification for context propagation established), I think we should refrain from giving any recommendations.

Also, it would be hard to give generic recommendations, as those recommendations would always be specific to certain protocols and messaging systems.

However, I'm open to other suggestions here, I'll put it on the agenda for Thursday.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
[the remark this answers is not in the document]
pyohannes: We will call this _message context layer_ for now.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
pyohannes: Summarizing the discussion in today's messaging workgroup. Giving clear recommendations for context propagation requires 3 sequential steps:
1. Making requirements for context propagation clear (this PR).
2. Verifying that protocols (AMQP, MQTT, CloudEvents) satisfy those requirements, if so, drive context propagation extensions for those protocols to a stable state.
3. Recommend stable context propagation mechanisms for those protocols in OpenTelemetry semantic conventions.

Before all three steps are done, we will not really be able to give stable recommendations for context propagation, and users will be left with a certain (intended) uncertainty here.

lmolkova: Created https://github.com/w3c/trace-context-binary/issues/19 on the W3C binary protocol to prevent potential header name collisions between formats

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,175 @@
...
+underlying messaging transport mechanism and its instrumentation.
+
+Instrumentors are required to instrument producer and consumer applications
+so that context is attached to messages and extracted from messages in a
+coordinated way. Future versions of these conventions might recommend [context propagation according to certain industry standards](#standards-for-context-propagation).
+
+### Requirements
+
+A producer SHOULD attach a message creation context to each message. The message creation context
+SHOULD be attached in a way so that it is not possible to be changed by intermediaries.
```
blumamir, line 126: I would also mention that attaching a context per message is not always possible technically (redis pub sub, socket.io etc), and in those cases, there will be no correlation between producer and consumer. Although the SHOULD  word covers this case, it is so common that I think it deserves to be explicitly mentioned, and also communicate to readers that it is expected to happen and the implications.

blumamir (approved):
LGTM
Thank you Johannes

text/trace/0205-messaging-semantic-conventions-context-propagation.md
pyohannes: I added a remark about this in the section above.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,173 @@
...
+   consumers. It must not be altered by intermediaries.
+
+   This layer helps to model dependencies between producers and consumers,
+   regardless of the underlying messaging transport mechanism and its
+   instrumentation.
+2. An additional _transport context layer_ allows correlating the producer and
+   the consumer with an intermediary. It also allows correlating multiple
+   intermediaries among each other. The transport context can be changed by
+   intermediaries, according to intermediary instrumentations. Intermediaries that
+   are not instrumented might simply drop the transport context.
```
dpauls, line 152: I know this is a "future", but I think intermediaries that are not instrumented might pass through the transport context unaltered. 

If we consider a message following a path:
producer --> messaging SDK --> intermediary --> messaging SDK --> consumer

If the SDK on the producer side includes a transport context and the intermediary isn't instrumented, having the transport context available to the consumer SDK could be useful. The SDK may instrument details like "received from intermediary", "enqueued to application", "delivered to application". I think these should be considered part of the transport trace, even if the intermediary didn't generate spans and modify the context.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,173 @@
...
+
+* [AMQP](https://w3c.github.io/trace-context-amqp/)
+* [MQTT](https://w3c.github.io/trace-context-mqtt/)
+* [CloudEvents via HTTP](https://github.com/cloudevents/spec/blob/v1.0.1/extensions/distributed-tracing.md)
+
+Those standards are in draft states and/or are not widely adopted yet. It is
+planned to drive those standards to a stable state and to make sure they cover
+requirements put forth by these semantic conventions. Finally, these semantic
+conventions should give a clear and stable recommendation for each protocol and
+scenario.
```
dpauls: Specifically, is this saying that they must define how both creation context and transport context will be carried as spearate entities? I think this should be a condition and stated a bit more clearly here.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Context propagation
+
+A message may pass many different components and layers in one or more
+intermediaries when it is propagated from the producer to the consumer. It
+cannot be assumed, and in many cases, it is not even desired, that all those
+components and layers are instrumented and propagate context according to
+OpenTelemetry requirements.
+
+A _creation context_ allows correlating the producer with the consumers of a
```
dpauls: I think last Thursday a suggestion was made to use the term "message context" instead of "creation context". Is this still being considered?

dpauls (changes requested):

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,177 @@
...
+
+There are several work-in-progress efforts to standardize context propagation for different
+messaging protocols and scenarios:
+
+* [AMQP](https://w3c.github.io/trace-context-amqp/)
+* [MQTT](https://w3c.github.io/trace-context-mqtt/)
+* [CloudEvents via HTTP](https://github.com/cloudevents/spec/blob/v1.0.1/extensions/distributed-tracing.md)
+
+Once the standards reach a stable state, these semantic conventions will give a
+clear and stable recommendation for each protocol and scenario.
```
dpauls, line 177: I think we want the standards to reach a stable state **and** define how both message creation context and transport context should be carried in the message. The goal is to provide direction here what we would like the standards to achieve before being adopted by these semantic conventions.

Here's an example of what I think could go wrong without specific direction. The current MQTT draft standard defines a single context only. One path forward could be for the MQTT standard to declare that "this context is the message creation context and must not be modified by intermediaries" in the draft. This would allow it to be adopted here, but this would leave a gap for transport context propagation. 

Even if these conventions don't define transport layer tracing for messaging now, we want it to evolve to include transport-layer tracing. So I think we should provide guidance as to what is generally needed by those standards for messaging.

```suggestion
Once the standards reach a stable state and define how the message creation
context and transport context are represented, these semantic conventions will
give a clear and stable recommendation for each protocol and scenario.
```

dpauls (changes requested):

text/trace/0205-messaging-semantic-conventions-context-propagation.md
joaopgrassi: Not sure if I follow but if the intermediary is not instrumented, how would the transport trace continue? Somehow there must be some code that reads the incoming remote span and either creates a new child span or simply forwarded the same (which I'm not it's a good approach).

pyohannes: I think the main point we wanted to make here was, that we cannot make any strong assumptions regarding the transport context. By definition, intermediaries can change it. So, possibly, intermediaries might alter it, keep it as it is (that's the scenario @dpauls points out), or drop it. The main point we wanted to make is, that in contrast to the message context, the consumer cannot rely on receiving a transport context, even if the producer sent one.

I can also remove the sentence in question. I think at this point we don't want to make any assumptions or requirements about what un-instrumented intermediaries would or should do.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
pyohannes: That makes sense to me. If nobody objects, I will accept this change.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
pyohannes: I'll resolve this, because it's covered by this discussion: https://github.com/open-telemetry/oteps/pull/205#discussion_r898403880

text/trace/0205-messaging-semantic-conventions-context-propagation.md
dpauls: > Not sure if I follow but if the intermediary is not instrumented, how would the transport trace continue? 
If an intermediary doesn't even understand the concept of tracing, if it sees a message with a `traceparent` header or property, it is probably going to treat it the same as it would a `foo` or a `bar` header/property. Different intermediaries might do different things here, but a reasonable choice is to just propagate those through unchanged towards the consumer.

I agree with all three possibilities (alter, propagate, drop) that @pyohannes mentioned. While this last sentence uses a "might" which is in no way prescriptive, I think it makes sense to either mention all possibilities, or also removing the sentence would be reasonable as well. Singling out the "drop" scenario might cause some readers to interpret this as being a preferable behaviour.

pyohannes: I removed the last sentence of this paragraph.

dpauls (approved):

yurishkuro (approved):

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
+Existing semantic conventions suppose that the consumer can use context
+information from the producer trace to create links or parent/child
+relationships between consumer and producer traces. For this to be possible,
+context information from the producer needs to be propagated to the consumer.
+In the example outlined above, the consumers sends an HTTP GET request to the
+intermediary to fetch a message, the message is returned as part of the
+response. Via this HTTP request, context information can be propagated from the
+consumer to the intermediary, but not from the intermediary to the consumer.
```
carlosalberto, line 86: `producer to intermediary` instead?

text/trace/0205-messaging-semantic-conventions-context-propagation.md
```diff
@@ -0,0 +1,177 @@
...
+is propagated as part of the message itself, independent of HTTP context
+propagation.
+
+For correlating producer and consumer traces without special intermediary
+instrumentation it is thus necessary to attach a producer context to the
+message so it can be extracted and used by the consumer, regardless of the
+contexts that are propagated on HTTP requests for publishing and fetching the
+message.
+
+Although OpenTelemetry semantic conventions cannot specify the exact mechanisms
```
carlosalberto, line 97: When adding this to the Specification, concise examples (say, "this is how Kafka supports this") will be really nice to have.

text/trace/0205-messaging-semantic-conventions-context-propagation.md
pyohannes: This describes a pull-scenario, and the HTTP request referred to is the one by which the consumer fetches a message from the intermediary, so "consumer to the intermediary" is correct. 

> [...] _the consumer_ sends an HTTP GET request to the intermediary to fetch a message, the message is returned as part of the response.

carlosalberto:
@open-telemetry/specs-approvers Please review. As previously mentioned, the messaging working group approved this OTEP already, we only need one more Spec approval.

jmacd (approved):

PR #208, opened 2022-06-22
Ephemeral Resource Attributes

tedsuo:
This OTEP is part of the RUM/Client initiative.

Currently, we are missing a place to put important client information which applies to all telemetry emitted by an SDK. This information includes attributes such as session ID, language preference, locality/timezone, and other types of user data.

Normally, these attributes would be recorded as resources. However, on client processes, there are times when this information changes without the SDK re-initializing. For example:

* The browser is idle for 15 minutes, ending a session.
* The process is put to sleep, and awakened at a later date.
* The user logs in or out, or other state changes which affect both the behavior and the reporting needs for the application.

In all of these cases, the application/SDK is not restarted. Currently, the resource associated with the SDK cannot be changed after it is started. This makes it very difficult to record these needed attributes.

This OTEP proposes a mechanism for updating the SDK with a new resource, which will be applied to all future telemetry created by the SDK. The proposal attempts to do this while preserving important characteristics already defined for resources:

* The resource object itself remains immutable, and accessing the resource object when creating telemetry does not introduce a lock. The proposed ResourceProvider concept preserves these characteristics.
* Existing resource attributes have a requirement for being present at SDK start time. They are not allowed to be updated or added to the resource once the SDK has started. This OTEP proposes that resource attributes be labeled as either "permanent" or "ephemeral" in the semantic conventions. Permanent attributes may not be updated after the SDK freezes the ResourceProvider.

If there are other backwards compatibility requirements for resources that I have missed, please let me know.

Cheers,
-Ted


text/0208-ephemeral-resources.md
```diff
@@ -0,0 +1,78 @@
...
+
+This change should be fully backwards compatible, with one potential exception: fingerprinting. It is possible that an analysis tool which accepts OTLP may identify individual services by creating an identifier by hashing all of the resource attributes. 
+
+In this case, it is recommended that these systems modify their behavior, and choose a subset of permanent resources to use as a hash identifier.
+
+## Prior art and alternatives
+
+An alternative to ephemeral resources would be to create span, metrics, and log processors which attach these ephemeral attributes to every instance of every signal. This would not require a modification to the specification.
+
+There are two problems to this approach. One is that the duplication of attributes is very inefficient. This is a problem on clients, which have a limited newtwork bandwidth. This problem is compounded by a lack of support for gzip and other compression algorithms on the browser.
```
tigrannajaryan, line 66: It would be great to quantify this. How inefficient is it? A benchmark demonstrating this would be a strong argument in favour of the proposed approach.

text/0208-ephemeral-resources.md
```diff
@@ -0,0 +1,78 @@
...
+
+NewTraceProvider, NewMetricsProvider, and NewLogProvider now take a ResourceProvider as a parameter. For backwards compatibility, the Resource parameter remains functional. If both a resource and a resource provider as passed as parameters, the resource is merged into the ResourceProvider, then discarded.
+
+FreezePermanent is then called by the provider.
+
+Internally, providers hold a reference to the ResourceProvider, rather than a specific resource. When creating a signal, such as a span, metric, or log, GetResource() is called to obtain a reference to the correct resource to attach to the signal.
+
+## Trade-offs and mitigations
+
+This change should be fully backwards compatible, with one potential exception: fingerprinting. It is possible that an analysis tool which accepts OTLP may identify individual services by creating an identifier by hashing all of the resource attributes.
```
Oberon00, line 58: There is another issue: Exporters right now may be implemented to assume they only ever deal with spans with the same resource. With this proposal, they could receive a batch of mixed spans.
Such an exporter may then misbehave and e.g. use the resource from the first/last span for everything.
Implementing sorting of spans by resource can be a bit costly.

Also there may be exporters for protocols that only support a single resource per connected agent. They would then probably need to stamp the ephemeral attributes on every single telemetry item.

Similar issues may apply to span processors.

(And possibly samplers that receive a resource in their constructor, but I don't think that will be a problem in practice https://github.com/open-telemetry/opentelemetry-specification/pull/1658)

text/0208-ephemeral-resources.md
```diff
@@ -0,0 +1,78 @@
...
+
+FreezePermanent is then called by the provider.
+
+Internally, providers hold a reference to the ResourceProvider, rather than a specific resource. When creating a signal, such as a span, metric, or log, GetResource() is called to obtain a reference to the correct resource to attach to the signal.
+
+## Trade-offs and mitigations
+
+This change should be fully backwards compatible, with one potential exception: fingerprinting. It is possible that an analysis tool which accepts OTLP may identify individual services by creating an identifier by hashing all of the resource attributes. 
+
+In this case, it is recommended that these systems modify their behavior, and choose a subset of permanent resources to use as a hash identifier.
```
Oberon00, line 60: That might be a pretty big deal for some, if they only allow storing one set of resource attributes per hash.

text/0208-ephemeral-resources.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Prior art and alternatives
+
+An alternative to ephemeral resources would be to create span, metrics, and log processors which attach these ephemeral attributes to every instance of every signal. This would not require a modification to the specification.
+
+There are two problems to this approach. One is that the duplication of attributes is very inefficient. This is a problem on clients, which have a limited newtwork bandwidth. This problem is compounded by a lack of support for gzip and other compression algorithms on the browser.
+
+The second problem is that it becomes difficult to distinguish between emphemeral resources and other types pf attributes.
```
Oberon00, line 68: Is it needed to distinguish them by type? Usually the attribute keys should be all you need. E.g. if you have a `session.id` attribute, would you care whether it is an ephemeral resource or a span/event attribute?

text/0208-ephemeral-resources.md
Oberon00, line 66: In situations where at least one of the ephemeral attributes changes very often, telemetry items are created between the changes and there are lots of permanent attributes, attaching to to the telemetry items ("signal instance") could even be more efficient.

Generally, I wonder how many ephemeral attributes we expect relative to permanent ones.

text/0208-ephemeral-resources.md
```diff
@@ -0,0 +1,78 @@
+# Ephemeral Resource Attributes
+
+Define a new type of resource attribute, ephemeral resources, which are allowed to change over the lifetime of the process. Existing resources are redefined as permanent resources, which must be present at SDK initialization and cannot be changed.
```
Oberon00, line 3: I have proposed a somewhat similar OTEP https://github.com/open-telemetry/oteps/pull/207

If #207 was implemented you could store your ephemeral resource attributes on the Context, and replace the active context when they change. Please check if #207 would also cover your use case.

Oberon00 (commented):
Would the use case of ephemeral resource attributes alsoa

text/0208-ephemeral-resources.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Explanation
+
+There are two types of resource attributes, **permanent** and **ephemeral**. Attributed which are labeled as permanent in the semantic conventions must be present when the SDK is initialized. They cannot be added or updated at a later date.
+
+Resources are managed via a ResourceProvider. Setting an attribute on a ResourceProvider will cause that attribute value to be included in the resource attached to any signal generated in the future. Spans which have already been started, along with any telemetry which has already been passed to the export pipeline, will not have the new attribute value. Optionally, a check can be added to ensure that permanent resources are not modified after the SDK has started
```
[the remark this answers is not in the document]
scheler: If I understand the proposal correctly, it requires that the permanent attributes be marked so in the semantic conventions. This is the part that will not be required if we limit the special handling to only one attribute with a known name.

Consider the following resource. The `ResourceProvider` can allow anytime modifications to the key-value pairs within the `ephemeral` attribute.

```
{
  service.name: foo,
  service.instance.id: 123,
  browser.user_agent: bar,
  ephemeral: {
    session.id: 456
  }
}

```

Anyway, this is an optimization step. Let's ignore this initially until the larger proposal gets acceptance.

text/0208-ephemeral-resources.md
jack-berg: If processors can change scope attributes, they might be a good candidate to solve this as well. 

> lack of support for gzip and other compression algorithms on the browser

I'm not an expert on browser stuff but can you expand on this? On its surface it seems wrong since gzipped static resources show up everywhere on the internet and there are js implementations of gzip (like [this](https://github.com/nodeca/pako)). This [stackoverflow](https://stackoverflow.com/questions/424917/why-cant-browser-send-gzip-request) post suggests that a part of it is because a browser client can't know if the server can accept gzipped data, but [OTLP requires gzip support](https://github.com/open-telemetry/opentelemetry-specification/blame/main/specification/protocol/otlp.md#L78).

martinkuba: Never mind on the OTLP gzip support, I see it says that clients MAY gzip the content.

martinkuba: @tigrannajaryan Regarding the limited network bandwidth, the sendBeacon() API has a payload limit of 64KB. Assuming session.id attribute that looks like this when sent over the wire

```
{"key":"session.id","value":{"stringValue":"8fded6726f630a327ee3be41174a8a91"}}
```

It adds 79 bytes per each signal. The number of spans/events per export will depend on the type of application and which instrumentations are present. But assuming that 100 is plausible, this adds almost 8kB to the payload.

This will further increase if we add additional context attributes (user attributes, URL etc.).

text/0208-ephemeral-resources.md
carlosalberto: @open-telemetry/specs-approvers Please take a look - I suspect we may need a lot of eyes, in case somebody relies on this _right now_.

text/0208-ephemeral-resources.md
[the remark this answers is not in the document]
tedsuo: I agree that it would simply things if ephemeral resources were kept separate from other resources.

Validator is also something which can be run in development, but disabled in production, which would work as an optimization.

text/0208-ephemeral-resources.md
tedsuo: Actually, exporters must deal with more than one resource already, which is what made this change so simple!

text/0208-ephemeral-resources.md
tedsuo: It seems crazy to me to use a resource hash as an identifier, given that there is no requirement that the items within it would uniquely identify a service...

But I'm throwing it out there as a possibility, just to cover all the bases.

text/0208-ephemeral-resources.md
tedsuo: In the browser,the overhead of applying the `session.id` as an attribute on every span and event would be untenable.

As far as the need to differentiate, putting data in the proper envelope helps backend systems use it more effectively. 

You might ask, why have resources at all in OTLP? Why not simple apply resources as attributes on every span and event? Besides the inefficiency, it would make life very difficult for backend systems which want to apply different analysis to resources and span attributes.

text/0208-ephemeral-resources.md
tedsuo: We are not expecting large numbers of ephemeral attributes, nor are we expecting them to change with great frequency. 

The expectation is that there would be between 1 and 10 ephemeral attributes set on a client, which may update after 15 minutes of inactivity, after the application reawakens, or in response to a change in user or user settings.

text/0208-ephemeral-resources.md
tedsuo: Thanks! That looks like a good proposal, but the context scope still presumes a transactional scope within a server handling many independent transactions. 

For clients, all telemetry emitted, including logs which are not bounded by a span, are related. Which is why the resource scope appears to be the correct one for things like this.

text/0208-ephemeral-resources.md
Oberon00: There is an issue open for that: https://github.com/open-telemetry/opentelemetry-specification/issues/1690
Right now, I don't think it's clear, and Dynatrace exporters take a shortcut here and always use the resource of the first item, assuming it will be the same for every item in the batch (everything else is an absolute edge case today)

text/0208-ephemeral-resources.md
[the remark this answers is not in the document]
tedsuo: Also, one aside on nested attributes: my assumption is that attribute values wouldn't be merged, they would be replaced. 

In other words, there is still only a single string key per attribute, but  with the option of storing an object, map, or array as the value for that attribute. If you set a new value for the key, it would throw the old value away.

text/0208-ephemeral-resources.md
Oberon00: You should be using something that doesn't exist yet, instead of hashing the whole resources: https://github.com/open-telemetry/opentelemetry-specification/issues/1034

text/0208-ephemeral-resources.md
Oberon00: > In the browser,the overhead of applying the session.id as an attribute on every span and event would be untenable.

Citation needed 😃
Would these arguments also apply against #207?

text/0208-ephemeral-resources.md
jack-berg: > I think there is also a danger of an attack on the server - compressed data could be expanded to a very large content. 

I'm not familiar with this type of attack, but surely a server decompressing a payload can have logic that cuts it off after it exceeds some threshold? 

> And lastly gzip compression is not native to browsers, so there is CPU overhead, which is important to consider for impact on user experience, especially when sending data while the page is unloading.

There's always going to be CPU overhead with compression, but as a rule of thumb applications tend to be bounded by network rather than CPU, in which case the user experience is improved because it takes less time to compress and send a smaller payload than it does to send an uncompressed larger payload. Its for this reason that its standard to gzip static resources returned to browser applications - the page loads faster despite spending CPU decompressing the payload because the app network is the bound. 

I wrote a little test suite to explore some scenarios related to compression and serializing with OTLP json vs the binary format. The client instrumentation group is obviously interested in minimizing payload sizes, primarily through reducing the amount of duplicated data that is sent. I'm not necessarily opposed to this, but the real gains in reducing payload size are going to come if the group embraces compression and does the work needed to serialize using the binary format instead of json. The test suite I wrote shows this:

| Scenario | JSON Size Raw | JSON Size Compressed (% reduction) | Binary Size Raw (% reduction) | Binary Size Compressed (% reduction) |
|----------|---------------|----------------------|-----------------|------------------------|
| session id on resource - 1 record | 531 | 300 (177.00) | 162 (327.78) | 153 (347.06) |
| session id on resource - 10 records | 2934 | 325 (902.77) | 703 (417.35) | 160 (1833.75) |
| session id on resource - 100 records | 26964 | 451 (5978.71) | 6103 (441.82) | 201 (13414.93) |
| session id on span - 1 record | 530 | 305 (173.77) | 164 (323.17) | 153 (346.41) |
| session id on span - 10 records | 3977 | 335 (1187.16) | 1290 (308.29) | 166 (2395.78) |
| session id on span - 100 records | 38447 | 520 (7393.65) | 12540 (306.59) | 247 (15565.59) |

Explanation:
* Compared including session id on resource vs on span records for payloads of 1, 10, and 100 spans. Source code lives [here](https://github.com/jack-berg/opentelemetry-java/blob/gzip-test/exporters/otlp/all/src/test/java/io/opentelemetry/exporter/otlp/GzipTest.java). 
* The "JSON Size Raw" column shows the payload size of json encoded OTLP. Comparing the session id on resource rows to the session id on span rows, there's gains to be had by including the session id on the resource, especially as the number records grows.
* The "JSON Size Compressed" column shows the size of the payloads when the JSON is compressed. Notice the dramatic % reduction versus raw json, and how the difference between including session id on resource and session id on span is much more modest. For example with 10 records including the session id on spans results in only a 10 byte increase in payload size.
* The "Binary Size Raw" and "Binary Size Compressed" show the same size of raw and compressed binary encoded OTLP. Notice that the Binary Size Raw for "session id on span - 100 records" is more than 2x smaller (12540) than the JSON Size Raw for "session id on resource - 100 records" (26964). The interpretation is that setting aside compression, pursuing binary encoding yields bigger payoffs than moving session id to the resource and keeping json. Compression also yields good results on binary payloads, and the best solution would be binary + gzip. 

Again, I'm not necessarily opposed to this idea of ephemeral resource attributes (need to do some thinking of whether anything else depends on the immutability of resources before I form an opinion). Just want to make the point that compression and binary encoding are going to move the needle in a much more serious way.

tigrannajaryan: Thanks for tests @jack-berg 

To confirm, for 100 spans, putting the session id in the resource results in 451 bytes of compressed JSON payload, putting the session id in the span results in 520 bytes of compressed JSON payload. Do I read the results correctly?

scheler: @jack-berg @tigrannajaryan there's no denying that compression would beat all the contestants by a great margin. To design OTel for the long term, it would be good to design for the superset of use-cases. We want to design even for the environments that cannot afford the additional download size for the compression library and the additional CPU usage for performing the compression. 

Using scope attributes for the purpose was a good idea - iff only it was possible to manage them centrally, like a ResourceProvider in this proposal. Currently, I believe, they are managed by the individual instrumentation libraries separately. Scope attributes has an interesting property that they are equivalent to signal attributes - this allows for one to start sending the attribute on each signal instance and optimize to send once at the scope level if required.  Similarly, we could send the ephemeral attributes at the signal level now iff you can show a path to send them at the top level later.  Resource attributes do not have that property - we are seeking for it with the ephemeral attributes of Resource.

tedsuo: Thank you @t2t2 for the detailed explanation! And thank you @jack-berg and @tigrannajaryan for the research.

I do not pretend to be a browser expert, but @t2t2 explains the situation as I have heard it from others.

Basically, you can load assets into the browser using compression because the browser unpacks these for you in a multithreaded C++ environment. But when sending requests using XHR, there is no way to hand off your payload to the underlying runtime where it can be compressed using the browser's gzip implementation. So you have to bring your own gzip implementation, and then block the UI thread whenever you compress. 

As a result, no current RUM clients today use compression. Instead, they use short codes and other tricks to minimize data size, and design their protocols to avoid any unnecessary  duplication.

Longer term, it looks like we would benefit from building a new browser SDK that is separated further from the current node SDK, to minimize load times and other issues. And it may be beneficial to design a new protocol specifically tailored for use in the browser - sending data over the open internet on a low bandwidth cellular connection has enough issues that this may be beneficial for reasons beyond just compression.

All that aside: back to ephemeral resources. I present the efficiency issue here because it was raised in the Client SIG as a blocker to storing ephemeral resources as span and event attributes, even as a temporary solution.

But! While I think all of this is worth mentioning, I would prefer that the discussion of ephemeral resources not get too hung up on efficiency. I believe there is benefit in designing our protocol to reflect our intended data model. 

We could just shove these resources into other scopes which happen to be available. Heck, we could just do away with resources entirely, shove all of the resources into span attributes, and let compression take care of it... but we would lose an important chunk of out data model in the process, and that makes consuming OTLP very difficult. We benefit from creating different envelopes like resources, instrumentation scopes, and spans, and only putting attributes on those envelopes which match the scope of that specific envelope. 

In fact, even more scopes would be better. For example, we currently lack the concept of trace-level attributes; as a result, our end users abuse baggage in odd ways to fake trace-level attributes by shoving the attributes into baggage and then... you guessed it... slapping them as span attributes onto every span. And, even though it "works," they aren't too happy about it, because it feels clunky and hackish. So I think we should be open to the idea that we haven't yet finished with the design of OTLP.

And, for that reason, I would ask that we judge ephemeral attributes by this measure: how do we correctly and intentionally model attributes which apply process-wide, to all telemetry emitted by an SDK? Intriguingly, precisely this scope is already present in OTLP - resources. We just have a restriction on resources which turns out to be untenable, which we discovered when we expanded OTel from just representing servers to also representing clients.

So, that it my case for this proposal. But! I am open to other approaches! For example, if we feel that this extension breaks a compatibility guarantee (and I don't believe it does, but I could be proven wrong), then we could add ephemeral resources to OTLP as it's own separate attribute set. I didn't propose that option because I felt it was messier than extending our current definition of resources to account for resources which represent attributes that change over time. 

In any case, I ask the spec community to consider this: how do we correctly model the scope of these attributes, rather than just shove them into other scopes which happen to be lying around? These attributes appear to be a fundamental enough part of modeling clients that they should be given a proper place within our data model.

Thanks again for the discussion, and the attention to this issue. :)

text/0208-ephemeral-resources.md
tedsuo: Yes I agree! There are various attributes which could count as a unique identifier. We could clarify in the spec which ones are currently defined. 

One possibility: by default, the SDK could generate a unique ID every time it starts, which would be a reliable identifier because we generate it ourselves. However, this identifier would not be stable across restarts. So there are limits to what can be provided without user input.

text/0208-ephemeral-resources.md
tedsuo: Please see this thread (https://github.com/open-telemetry/oteps/pull/208#discussion_r904191239) for a lengthy discussion on data limitations in the browser.

I don't think these arguments apply to #207, that proposal would be helpful imho. Just not a solution for ephemeral resources, since many of the events which need these resources happen when there is no trace present.

text/0208-ephemeral-resources.md
tedsuo: Ok, I agree this should be clarified. My understanding is that a BatchSpanProcessor may be shared across multiple SDKs within the same process, and that is done in order to have different sets of resources for different sub-processes. So there is no guarantee that all spans in a batch have the same resource. I know that @MSNev has examples of this pattern. 

But, I think that this pattern is extremely rare, so it doesn't surprise me that Dynatrace and other exporters could take a shortcut without anyone noticing.

text/0208-ephemeral-resources.md
tedsuo: That might work... but it might be better to keep the concept of a "process scope" and a "context scope" separate. I see these attributes as more similar to resources and instrumentation scopes - they represent the environment the transaction is occurring within. 

Because contexts are immutable, and no rules as to when child contexts may be created, there would be synchronization issues between when ephemeral resources are updated and when they would applied, if they only change the root context and thus only affect transactions which start from a new root context.

text/0208-ephemeral-resources.md
tigrannajaryan: >JSON's data size could also gain advantages by creating an alternative of otlp/json with one-two letter object keys which could turn compression wins < extra impact of having to load a compression library.

Exactly.

The current OTLP/JSON is extremely inefficient without compression. If we believe uncompressed JSON is going to be used in production we need a much better JSON structure. It can be done via shortened keys (as mentioned above), by using dictionary encoding (eliminates duplicate strings), etc. Such targeted and comprehensive format optimization will likely have far more effect on the payload size than the single deduplication change proposed by this OTEP.

Let me reiterate: if the goal of this OTEP is to have more efficient uncompressed JSON representation then I think we can do better without the need to introduce new ephemeral attributes concept. Instead a more efficient and compact OTLP/JSON format is possible. If that's the goal now is a great time to do it since OTLP/JSON is on the verge of being declared stable (but it is not yet, so we are free to do it).

text/0208-ephemeral-resources.md
MSNev: Our examples are (currently) used using our internal (not OpenTelemetry) SDK's on clients where multiple teams provide different components to the same "view" (page etc) and need / want to report telemetry to their own backends.

And in some runtimes we have a single batching system which is shared, rather than having each component on the view creating its own SDK instance with all of the overhead and batching mechanisms. Thus reducing the runtime impact on resources for the client (CPU, Memory, etc)

text/0208-ephemeral-resources.md
scheler: > if the goal of this OTEP is to have more efficient uncompressed JSON representation

I consider session and page url as the context for events/spans emitted and I thought it is a reasonable ask to have a normalized representation and not repeat them. It will really be a good design that does not rely on compression.

I believe the hesitation for this otep is due to the complexity of the changes proposed and the unknown implications. I propose adding an entirely new field called `ephemeral_resource` as a sibling to `resource` in `ScopeSpans` and `ScopeLogs` - this way, the original `resource` remains immutable and the new field can be use for the ephemeral attributes of the resource. @tedsuo

Aneurysm9: > While decompressing is common in browsers (be it gzip or brotli), none of the current request APIs expose a way to have browser compress the request (MDN: [XHR](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest), [fetch](https://developer.mozilla.org/en-US/docs/Web/API/fetch))

This is incorrect.  The [CompressionStreams API](https://wicg.github.io/compression/) provides a native solution for this and is [supported in Chromium-based browsers](https://developer.mozilla.org/en-US/docs/Web/API/Compression_Streams_API#browser_compatibility) already.

> This does mean that indeed you have to bring your own compression methods. More code = larger bundle that the browser needs to download, parse and execute. First phase is network bound (but does benefit from compression itself), while second and third are CPU bound. 

Also benefits from caching.  In a network-constrained situation the cost of retrieving the additional code is paid once and the result cached.  Conditional requests and etags are your friend.

> Also in most cases instrumentation is required to be loaded ASAP (sometimes even before rest of the content on the page), causing site loading to be blocked until code is downloaded (should it be in the <head> of the page)

The additional code for compression is only needed to export telemetry and does not need to be loaded at the same time as the code enabling instrumentation.  Deferring until an export is required can increase the time-to-export but would not impact time-to-interaction or any other user-focused timing.

>  I propose adding an entirely new field called ephemeral_resource as a sibling to resource in ScopeSpans and ScopeLogs - this way, the original resource remains immutable and the new field can be use for the ephemeral attributes of the resource.

This is inverted.  `ResourceSpans` contain `ScopeSpans`, not the other way around.  Ephemeral resource attributes could be added as `Scope*` attributes on each `Scope*` produced during the time when the ephemeral resource attributes are active, but I'm not sure I see how changing the OTLP data structure advances the conversation in a safe way.  That is the most invasive way of going about this that I could think of.

scheler: @Aneurysm9 sorry my bad, I meant `ResourceSpans` and `ResourceLogs` and not `ScopeSpans` and `ScopeLogs`. I corrected this in my previous comment, can you check if it makes sense this time?

dyladan: The cost to generate, serialize, and compress that many spans is also not a synchronous process that takes x milliseconds, but many small processes which each take a small fraction of X. It is most important to ensure that each individual step doesn't impact user experience. With the example of 100 spans otlp -> protobuf -> pako on the pixel 4a given, the whole process is 4.393ms but you have 2 chances to yield to the event loop to ensure user experience is not affected.

t2t2: > This is incorrect. The [CompressionStreams API](https://wicg.github.io/compression/) provides a native solution for this and is [supported in Chromium-based browsers](https://developer.mozilla.org/en-US/docs/Web/API/Compression_Streams_API#browser_compatibility) already.

Have missed it but I generally don't consider new browser features as a solution unless usage% is >90% ([and well, safari has a monopoly on ios so....](https://caniuse.com/?search=CompressionStream)) (also 90% is probably low considering how much RUM products are asked for IE11 support but they already have a miserable experience due to using IE in current year so making it optional is worth consideration)

> but you have 2 chances to yield to the event loop to ensure user experience is not affected.

There is one but - not when user is leaving the page, tho generally you don't have 100 spans then, making it a question of how much do you want to maintain 2 different code paths (a sync one and an "async" one)

tedsuo:
Added an example implementation and example use case.

text/0208-ephemeral-resources.md
tedsuo: @tigrannajaryan i just want to emphasize what @scheler said, that the purpose of this OTEP is not to avoid compression or gain efficiency, but to extend our data model in a way that correctly represents these attributes. 

If we don't want to extend the current Resource concept, we could add a new concept, call it ProccessScope or something similar, and have it work in effectively the same manner. 

Personally, I'd prefer we extend resources over adding a new scope. But I prefer both over an approach that makes it impossible to cleanly implement RUM using OpenTelemetry.

In other words, I'm against "just tack on the process scope as span/event attributes" the same way I'd be opposed to "just tack on the instrumentation scope as span/event attributes." In both cases, yes it would "work." But it would create a headache for implementers and confusion for users. 

We should strive for a clean data model, where everything is explained just by looking at the data structure.

tigrannajaryan:
@tedsuo the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md) defined Resource like this:
>A [Resource](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#resources) is an immutable representation of the entity producing telemetry as [Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute).

This is a Stable document. How do we reconcile this OTEP with the spec's stance on immutability of the Resource? Are you suggesting that we break a Stable spec document? Or you do not think this is a breaking change?

t2t2:
> How do we reconcile this OTEP with the spec's stance on immutability of the Resource? 

This doesn't change anything about current resource immutability - an update on the resource provider would end up in a new resource instance. To speak in code:

```js
const resourceProvider = new ResourceProvider({
    // Initial set of attributes, internally does a new Resource(attrs) and stores it as current value
    'session.id': '1',
});
const tracerProvider = new TracerProvider({ resourceProvider });
const tracer = tracerProvider.getTracer(/* irrelevant */);

const span1 = tracer.getSpan(/* ... */);
// internally span.resource = tracer.tracerProvider.resourceProvider.getResource()

// Some time later, user logs in and their identity is known
resourceProvider.setAttribute('enduser.id', 'superadmin');
// internally currentResource = currentResource.merge(new Resource(newAttrs)), which as per the current spec
// returns a new Resource with merged attrs
// That new Resource is set as the current value in ResourceProvider

// Or session expires and a new one is set
resourceProvider.setAttribute('session.id', '2')

const span2 = tracer.getSpan(/* ... */);


span1.resource !== span2.resource

assert.deepEquals(span1.resource.attributes, { 'session.id': '1' });
assert.deepEquals(span2.resource.attributes, { 'session.id': '2', 'enduser.id': 'superadmin' });
```

tigrannajaryan:
>This doesn't change anything about current resource immutability - an update on the resource provider would end up in a new resource instance.

I disagree. This is not just about a Resource instance in memory. It is about the Resource that is emitted by the instrumented application. The recipients of telemetry expect that the resource is immutable, i.e. its attributes do not change over time.

The OTEP talk about this in the "Trade-offs and mitigations" section. I think this is a breaking change. It breaks the contract between Otel sources and telemetry destinations. The OTEP text even recommends this:
>In this case, it is recommended that these systems modify their behavior

I don't think this is acceptable. We are saying that "yes, we broke the contract, deal with it". IMO, we cannot do that.

tigrannajaryan:
I thought a bit more about this, I want to find a solution.

I don't think we can delete the requirement which says the Resource is immutable. I think this needs to stay otherwise we are breaking the contract. Additionally, unfortunately the spec says we are not allowed to change the association of the Resource and TracerProvider once that association is established:
>a resource can be associated with the [TracerProvider](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#tracerprovider) when the TracerProvider is created.

However, let's step back for a moment. I don't think recipients of telemetry care about the association inside the SDK. The recipients care about the data model and data model certainly allows the SDK to emit telemetry associated with different Resources. A new TracerProvider can be created with a new Resource and can be used to emit telemetry that was previously emitted using a different TracerProvider and this is completely legal.

Given the above, I do not see any clause in the spec that directly prohibits us from introducing a new way for TracerProvider to be associated with some proxy object which itself is associated with a Resource and allow that association to change over time. Yes, this is in a sense cheating, but it allows to introduce this new way such that it is not a breaking change for the SDK. That's what the proxy ResourceProvider here does. 

To me the following questions remain:

1. Is it right that session id is part of the Resource? It doesn't feel right but I can't put my finger on it, so I will refrain from objecting to this for now.
2. Why do we need to introduce anything called "ephemeral attributes"? I think this is not needed. They are regular attributes just like any other. Nothing ephemeral here. We only introduce a new way to specify the Resource that must be associated with the produced telemetry. That's all it is. It is a regular Resource, an immutable one. Attributes are all regular.
3. Is it really possible to introduce ResourceProvider with the ability to attach it to TracerProvider in a way that does not break any existing code? We need to see prototypes that demonstrate this.

t2t2:
Note: I originally started this as part of response to https://github.com/open-telemetry/opentelemetry-specification/issues/2500#issuecomment-1249387117 but the first section ended up being more related to this otep being stuck, so here it is!

---

Let's eliminate the confusion of what a session means for a bit. There are some other attributes that are

1) good candidates for resource level 
2) value can change over time
3) as a concept is more familiar to backend service / APM kind of usage that current otel contributors are a lot more familiar with

Let's bring in [`enduser.id`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-identity-attributes)

Currently it's defined as a span level identifying attribute. Which hey, makes total sense in a server side environment. You've got a server side service that you can have one server serve all of the users of the application. If you'd want to have enduser that caused a request set on all of the child spans, yes context makes a lot of sense since the entire server isn't dedicated to just one enduser. Anyways got my 3rd condition



Let's jump to client side. I open up _local food delivery app_, and it's instrumented to generate telemetry. Alright, what's the resource attributes. Well you've got
* the ones describing the app (service.name = app name, service.version = app version, probably also including installation source, build type, ....)
* but also the runtime (so like os, os version, ...; but when running in browser also browser user agent related info)
* but also I'm logged in and interacting with the app so most of the telemetry comes out of my interactions with the app or server side updates pushed based on me being logged in. I'd consider this a reason to have `enduser.id` on resource, fulfilling condition 1

I try to order something but suddenly app runs into a bug and crashes. Smash cut, support person is messaging devop team "hey got this guy going crazy over not being able to order, can you figure out what's going on there, why his app keeps crashing". Devops looks up data based on my name, sees an attempt to order 100 kebabs in tracing spans that caused `KebabStackOverflow` in logs. Really this paragraph is only here to be referenced later while still having a linear timeline for the domain knowledge story

Somehow you're next to me and mention you uninstalled the app due to constant crashing a while ago and now have a please come back discount on your account. I hand my phone to you, you log my account out and log in with your account. And manage to successfully order after a more reasonable order size.

Now the logged in account has changed, so if the logged in account is in resource, the above telemetry should be over 3 resources: Data from me, data from anonymous user, data from you. (so now we've fulfilled condition 2)



---

Other than _local food delivery app_, some other examples:

* The user of a self-service kiosk at an airport
* The logged in cashier at a POS system
* A rental scooter user

But other potential attributes:

* Session id (the logic may be defined in otel but it's caused to change by external factor - time, user (in)activity)
* The IP or ISP of the device the data is coming from (Roaming over the border of countries, changing the internet provider; or from wifi to mobile data)
* A case could also be made for document location or the active screen in mobile app open-telemetry/opentelemetry-js#3222

So I think something people who haven't built a RUM need to consider is that a major difference between backend services and client side apps is that **apps have (a lot more) state**. A lot of this state is global (not scoped to parts of the app like within one request that is forgotten once the request is fulfilled), it changes over time (due to time, user interactions, or completely external actions) and in a lot of the cases it's useful or needed to assist in debugging using gathered telemetry (who, what device, what screen/url, what isp, geolocation)

---

A lot of these attributes are also what you'd want to query data by. Already mentioned looking up data based on app user info, but let's consider some of the RUM use cases:

* Viewing the flow of an user during the session (querying spans/logs based on session.id)
* Which page URL an error occurred the most (querying logs based on document.location/whatever it will get spec-ed to)
* How's the webvitals score for people in a specific country (metrics(?) with geodata)
* Comparing request spans based on ISP

These add considerations for efficient data ingest and storage. Now every vendor will probably have different opinions on this based on how they use and store data. In July I got some knowledge from @mdubbyap on splunk/signalfx ingest side about our use cases (and probably should have used this knowledge earlier so I don't accidentally misremember it but oh well Ted's been on vacation anyway so it wouldn't have helped move this forward):

For ingesting best is to minimise the amount of bytes that needs to be read in order to determine where to pipe the data to (be it partitioning, buckets or whatever optimises your infra). Worst is having to read deep enough to get into each span/log/metric and check it's attributes for the value. If it's a value on the resource, ingest only needs to read resource's bytes before determining where to send the data, not needing to parse the rest of the payload. (Since we focus on showing session experience, then obviously for us session.id attribute is 👀👀)

---

There also can be ways fulfilling legal requirements can be easier if these attributes are more easily readable, eg. indexing data based on enduser info to make deleting data on user request (such as GDPR) to be easier

Also linking https://github.com/open-telemetry/opentelemetry-specification/issues/2775 as it's gone into topic of descriptive or identifying attributes, which has been to be one of the reasons against this otep so far

Oberon00:
@scheler
> We think you should pursue a strategy where these attributes are set in [context](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/context), and lifted out of context onto the individual records in a custom SpanProcessor / LogRecordProcessor

This is what I proposed in OTEP #207 to be a blessed concept with its own API, by the way. But as you said, it is in principle implementable today.

tedsuo:
Closing this in favor of a new proposal coming from the RUM/Client group.

PR #210, opened 2022-08-05
Propose a high-level process for migrating from OpenCensus to OpenTelemetry

dashpole:
Part of https://github.com/open-telemetry/opentelemetry-specification/issues/1175

We've talked about bridges for OpenCensus in the past, but I thought it would be a good idea to write down exactly how we expect them to be used, and consider the implications for library maintainers.

If the OTEP is accepted, I would add a migration guide to the current OpenCensus compatibility spec with the contents of the "migration plans" section.

In particular, please consider out the goals, alternatives considered, and open questions sections.

### DIFF
--- a/text/0000-opencensus-migration-plan.md
+++ b/text/0000-opencensus-migration-plan.md
@@ -71,13 +71,13 @@ Starting with an application using entirely OpenCensus instrumention for traces 
 1. Migrate the exporters (SDK)
     1. Install the OpenTelemetry SDK, with an equivalent exporter
     2. Install equivalent OpenTelemetry resource detectors
-    3. Install OpenTelemetry propagators for OpenCensus' `TextFormat` and `BinaryPropogator` formats.
+    3. Install OpenTelemetry propagators for OpenCensus' `TextFormat` and `BinaryPropagator` formats.
     4. **breaking**: Install the metrics and trace bridges
     5. Remove initialization of OpenCensus exporters
 2. Migrate the instrumentation (API)
     1. **breaking**: For OpenCensus instrumentation packages, migrate to the OpenTelemetry equivalent.
     2. For external dependencies, wait for it to migrate to OpenTelemetry, and update the dependency.
-    3. For instrumentation which is part of the application, migrate it following the guidance below.
+    3. For instrumentation which is part of the application, migrate it following the "library" guidance below.
 3. Clean up: Remove the metrics and trace bridges
 
 ### Migrating Libraries using OC Instrumentation
--- a/text/0000-opencensus-migration-plan.md
+++ b/text/0000-opencensus-migration-plan.md
@@ -1,0 +1,141 @@
+# OpenCensus To OpenTelemetry Migration Plan
+
+Define the process of migrating applications, libraries, and agents from [OpenCensus](https://opencensus.io/) to
+OpenTelemetry.
+
+## Motivation
+
+While OpenTelemetry has a specification for an
+[OpenCensus trace bridge](https://github.com/open-telemetry/opentelemetry-specification/blob/87a5ed7f0d4c403e2b336f275ce3e7fd66a8041b/specification/compatibility/opencensus.md),
+it lacks a broader context on how such bridges should be used to perform a migration from
+OpenCensus to OpenTelemetry. Additionally, OpenTelemetry should provide guidance to libraries
+currently using OpenCensus on how to migrate instrumentation to OpenTelemetry.
+
+## Explanation
+
+Migrating from OpenCensus to OpenTelemetry may require breaking changes to the telemetry produced
+because of:
+
+* Differences in semantic conventions for names and attributes
+* Data model differences
+* Instrumentation API feature differences
+* Differences between equivalent OC and OTel exporters
+
+Requirements:
+
+* Give application developers control over _when_ breaking changes are introduced.
+* No "long tail" of breaking changes to users when migrating.
+* Make it possible to migrate instrumentation without breaking the resulting telemetry.
+* Library maintainers should not need to maintain OpenCensus and OpenTelemetry instrumentation in parallel.
+
+This proposed process groups most breaking changes into the installation of the bridge. This gives
+users control over introducing the initial breaking change, and makes subsequent migrations of
+instrumentation (including in third-party libraries) non-breaking.
+
+## Migration Plans
+
+### Migrating from the OC Agent and Protocol
+
+Starting with a deployment of the OC agent, using the OC protocol, migrate by:
+
+1. Deploy the OpenTelemetry collector with OpenCensus and OTLP receivers and equivalent processors and exporters.
+2. **breaking**: For each workload sending the OC protocol, change to sending to the OpenTelemetry collector's OpenCensus receiver.
+3. Remove the deployment of the OC Agent
+4. For each workload, migrate the application from OpenCensus to OpenTelemetry, following the guidance below, and use the OTLP exporter
+
+### Requirements for Metrics and Trace bridges
+
+Bridges:
+
+* MUST NOT require OpenCensus to depend on OpenTelemetry
+* MUST require few or no changes to OpenCensus
+* MUST convert OpenCensus semantic conventions to OpenTelemetry semantic conventions for metric and span names and attributes
+* are NOT REQUIRED to support the entire OpenCensus instrumentation API Surface.
+
+Trace Bridges:
+
+* MUST support context propagation between OpenCensus and OpenTelemetry instrumentation
+* MUST preseve span contents where an OpenTelemetry equivalent exists
+
+Metric Bridges (to be proposed separately):
+
+* MUST be compatible with push and pull exporters
+* SHOULD support Gauges, Counters, Cumulative Histograms, and Summaries
+* are NOT REQUIRED to support Gauge Histograms
+* SHOULD support exemplars
+
+### Migrating an Application using Bridges
+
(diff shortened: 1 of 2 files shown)


text/0000-opencensus-migration-plan.md
```diff
@@ -0,0 +1,141 @@
...
+instrumentation (including in third-party libraries) non-breaking.
+
+## Migration Plans
+
+### Migrating from the OC Agent and Protocol
+
+Starting with a deployment of the OC agent, using the OC protocol, migrate by:
+
+1. Deploy the OpenTelemetry collector with OpenCensus and OTLP receivers and equivalent processors and exporters.
+2. **breaking**: For each workload sending the OC protocol, change to sending to the OpenTelemetry collector's OpenCensus receiver.
```
yurishkuro, line 42: why is this a breaking change? The app still sends the same data. Is OTEL's OC receiver going to rewrite the data, e.g. rename the attributes?

dashpole: Since I haven't defined what constitutes a breaking change, i've taken a fairly strict definition. The OTel collector's OpenCensus receiver translates OpenCensus metrics + spans to OTLP internally, which is a lossy translation.  For example, child span count for spans and GaugeDistribution-typed or the sum of squared deviation for metrics are lost when translating from OpenCensus to OTLP.  An OpenCensus exporter might've been able to use those fields, but the OpenTelemetry version of the exporter can't.

text/0000-opencensus-migration-plan.md
```diff
@@ -0,0 +1,141 @@
...
+it lacks a broader context on how such bridges should be used to perform a migration from
+OpenCensus to OpenTelemetry. Additionally, OpenTelemetry should provide guidance to libraries
+currently using OpenCensus on how to migrate instrumentation to OpenTelemetry.
+
+## Explanation
+
+Migrating from OpenCensus to OpenTelemetry may require breaking changes to the telemetry produced
+because of:
+
+* Differences in semantic conventions for names and attributes
```
punya, line 19: Consider adding a motivating example of each of these.

text/0000-opencensus-migration-plan.md
```diff
@@ -0,0 +1,141 @@
...
+* SHOULD support exemplars
+
+### Migrating an Application using Bridges
+
+Starting with an application using entirely OpenCensus instrumention for traces and metrics, it can be migrated by:
+
+1. Migrate the exporters (SDK)
+    1. Install the OpenTelemetry SDK, with an equivalent exporter
+    2. Install equivalent OpenTelemetry resource detectors
+    3. Install OpenTelemetry propagators for OpenCensus' `TextFormat` and `BinaryPropogator` formats.
```
punya, line 74:
```suggestion
    3. Install OpenTelemetry propagators for OpenCensus' `TextFormat` and `BinaryPropagator` formats.
```

text/0000-opencensus-migration-plan.md
```diff
@@ -0,0 +1,141 @@
...
+* SHOULD support Gauges, Counters, Cumulative Histograms, and Summaries
+* are NOT REQUIRED to support Gauge Histograms
+* SHOULD support exemplars
+
+### Migrating an Application using Bridges
+
+Starting with an application using entirely OpenCensus instrumention for traces and metrics, it can be migrated by:
+
+1. Migrate the exporters (SDK)
+    1. Install the OpenTelemetry SDK, with an equivalent exporter
```
punya, line 72: The OpenCensus stackdriver exporter and the OTel Cloud Ops exporters have different observable behavior (e.g. metric names). Should we declare that out of scope, e.g., by documenting that OTel SDK exporters need to provide a configuration that's equivalent to the defaults of the corresponding OC exporter?

text/0000-opencensus-migration-plan.md
```diff
@@ -0,0 +1,141 @@
...
+    2. Install equivalent OpenTelemetry resource detectors
+    3. Install OpenTelemetry propagators for OpenCensus' `TextFormat` and `BinaryPropogator` formats.
+    4. **breaking**: Install the metrics and trace bridges
+    5. Remove initialization of OpenCensus exporters
+2. Migrate the instrumentation (API)
+    1. **breaking**: For OpenCensus instrumentation packages, migrate to the OpenTelemetry equivalent.
+    2. For external dependencies, wait for it to migrate to OpenTelemetry, and update the dependency.
+    3. For instrumentation which is part of the application, migrate it following the guidance below.
```
punya, line 80:
```suggestion
    3. For instrumentation which is part of the application, migrate it following the "library" guidance below.
```

text/0000-opencensus-migration-plan.md
```diff
@@ -0,0 +1,141 @@
...
+3. Clean up: Remove the metrics and trace bridges
+
+### Migrating Libraries using OC Instrumentation
+
+#### In-place Migration
+
+Libraries which want a simple migration can choose to replace instrumentation in-place.
+
+Starting with a library using OpenCensus Instrumentation:
```
punya, line 89: OC doesn't have an API/SDK separation, so libraries generally had to keep their instrumentation in an optional addon to avoid pulling in the whole SDK. When transitioning to OTel, do we want libraries to take a core dependency on the OTel API?

text/0000-opencensus-migration-plan.md
```diff
@@ -0,0 +1,141 @@
...
+#### Migration via Config
+
+Libraries which are eager to add native OpenTelemetry instrumentation sooner, and/or want to
+provide extended support for OpenCensus may choose to provide users the option to use OpenCensus
+instrumentation _or_ OpenTelemetry instrumentation.
+
+Starting with a library using OpenCensus Instrumentation:
+
+1. Add configuration allowing users to enable OpenTelemetry instrumentation and disable OpenCensus instrumentation.
+2. Add OpenTelemetry instrumentation gated by the configuration.
```
punya, line 105: Should we document the risk of discrepancy/drift between the OC/OTel instrumentation in this approach?

text/0000-opencensus-migration-plan.md
```diff
@@ -0,0 +1,141 @@
...
+3. Remove the deployment of the OC Agent
+4. For each workload, migrate the application from OpenCensus to OpenTelemetry, following the guidance below, and use the OTLP exporter
+
+### Requirements for Metrics and Trace bridges
+
+Bridges:
+
+* MUST NOT require OpenCensus to depend on OpenTelemetry
+* MUST require few or no changes to OpenCensus
+* MUST convert OpenCensus semantic conventions to OpenTelemetry semantic conventions for metric and span names and attributes
```
punya, line 52: Which semantic conventions are in scope? Are OTel semantic conventions stable enough to target? (e.g., would a bridge need to keep adding or changing based on new versions of semantic conventions?)

text/0000-opencensus-migration-plan.md
```diff
@@ -0,0 +1,141 @@
...
+* are NOT REQUIRED to support the entire OpenCensus instrumentation API Surface.
+
+Trace Bridges:
+
+* MUST support context propagation between OpenCensus and OpenTelemetry instrumentation
```
punya, line 57: I know it's covered in the trace bridge spec but is it worth calling out that this means both directions (trace sandwich scenarios)?

text/0000-opencensus-migration-plan.md
dashpole: I intended for those to be in-scope (switching from OpenCensus exporters to OpenTelemetry exporters) is one of the categories of breaking changes above.  It is possible some exporters may make this transition non-breaking, but as a baseline, but i'm assuming it is breaking for this proposal's purposes.

The reason why this step isn't marked breaking is because the breaking change does not occur until the bridge is installed.

text/0000-opencensus-migration-plan.md
dashpole: We do.  Do you think that is important to call out here, or did you think I was implying otherwise anywhere?

text/0000-opencensus-migration-plan.md
dashpole: done

text/0000-opencensus-migration-plan.md
dashpole: All documented OpenCensus semantic conventions are in-scope, which includes HTTP and gRPC.  I've specifically included those.  I agree a bridge cannot be marked stable until the conventions it depends on are marked stable.

text/0000-opencensus-migration-plan.md
dashpole: I've updated the wording to "between **interleaved** OpenCensus and OpenTelemetry instrumentation".  Does that make it clearer?

text/0000-opencensus-migration-plan.md
dashpole: Good point.  I've updated this to suggest that the two should share unit tests, as recommended in the "in-place" migration.  That should resolve this issue.

text/0000-opencensus-migration-plan.md
punya: I thought you weren't making a recommendation either way (otelapi as a core dependency vs otelapi as an addon, using agnostic hooks).

dashpole: I agree that we should publicize the fact that it is easy to take a dependency on the OTel API compared with OC, but i'd prefer to keep this proposal centered on the migration.  Talked offline.

dashpole:
I've realized this probably doesn't need to be an OTep, since it is already an agreed upon direction by the community (part of the founding!).  I'll convert it to a PR to the spec repo.

PR #211, opened 2022-08-17
Add app-namespacing-semantic-convention

cartermp:
Proposal to enhance semantic conventions for manual instrumentation with by suggesting namespacing as a means to organize data, with the `app.` prefix.

### DIFF
--- a/text/0000-app-namespacing-semantic-convention.md
+++ b/text/0000-app-namespacing-semantic-convention.md
@@ -46,6 +46,8 @@ Finally, like any standard convention, it's not much use if it's not documented.
 
 Namespacing attributes is a documented best practice for [some vendors](https://docs.honeycomb.io/getting-data-in/data-best-practices/#namespace-custom-fields). These practices emerged before OpenTelemetry, but are equally applicable to OpenTelemetry.
 
+This convention is also being used by the [OpenTelemetry Community Demo](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/checkoutservice/main.go#L265-L267).
+
 ## Open questions
 
 Should the prefix `app.` be used, or should it be something else?
--- a/text/0000-app-namespacing-semantic-convention.md
+++ b/text/0000-app-namespacing-semantic-convention.md
@@ -1,0 +1,59 @@
+# Application namespacing semantic convention
+
+Adds a semantic convention for manual instrumentation that defines how to namespace custom attributes, starting with the `app` prefix.
+
+## Motivation
+
+Today, custom attributes from an app created through manual instrumentation have all kinds of names that are specific to the domain of that app. There are no conventions for how to organize this data, though.
+
+Data organization is critical to being able to effectively analyze observability data and avoid collisions in names. Without a well-defined convention for developers to follow, their own means of organization can differ from developer to developer, leading to a need to do things like coalesce field names using `.`/`-`/`_`/etc, not to mention having a real possibility to have name collisions. to separate clauses in an attribute name. Instead, a well-documented convention for how to organize data and in what format removes a "need to think" like code formatting conventions can do.
+
+## Explanation
+
+Attributes created through manual instrumentation should use namespacing for organizational purposes, using `app.` as a prefix for the name. Sub-namespaces should use a period (`.`) to separate sub-namespaces.
+
+Use as many layers of subnamespaces as needed to represent your app's hierarchy. For example, an eCommerce app may define attributes like this:
+
+* `app.order.id`
+* `app.cart.items.count`
+* `app.shipping.cost.total`
+
+The more specific the namespace, the less likely specific names can collide in your app.
+
+DO use a period (`.`) to separate sub-namespaces.
+DO NOT use characters other than a period to separate sub-namespaces.
+DO NOT mix different characters to separate sub-namespaces.
+
+## Internal details
+
+This would be accomplished by defining the `app.` prefix as reserved for end-user apps in the specification. Additionally, a specification entry in relevant places (trace, metrics, logs, and resource semantic conventions).
+
+Additionally, official documentation would be added:
+
+* Code samples in documentation updated to follow the convention as needed
+* Subsections in language SDK docs that explain how to add attributes will explain or link to this convention
+* Dedicated section in documentation about best practices for organizing data can be written, linked to, and link to the specification
+
+## Trade-offs and mitigations
+
+The biggest negative to introducing this convention now is that anyone using a _different_ convention will be considered out of step with best practices. For them to follow best practices, they would need to rewrite their instrumentation names, which would likely break or complicate their analyses for a period of time. To mitigate this, several analysis tools offer ways to coalesce two field names into one, but this kind of functionality is not guaranteed to be present for everyone.
+
+Another negative is that some developers may not like the idea of defining a convention for their own systems. It's not uncommon for developers to dislike standard coding conventions, for example, and we could expect to see a similar kind of backlash around defining this kind of convetions. There isn't a really a way to mitigate this kind of feedback; however, there is precedent that most developers tend to like conventions to follow rather than having to define their own all of the time. And so it's likely that most developers would appreciate following this convention.
+
+Finally, like any standard convention, it's not much use if it's not documented. The proposed additions to documentation (even just having the right link the right places) are nontrivial. One potential way to mitigate this could be to use tooling to somehow suggest or enforce this convention (e.g., Analyzers in C#). However, that is a much bigger lift than defining a spec convention, and not all languages are guaranteed to have that kind of tooling available.
+
+## Prior art and alternatives
+
+Namespacing attributes is a documented best practice for [some vendors](https://docs.honeycomb.io/getting-data-in/data-best-practices/#namespace-custom-fields). These practices emerged before OpenTelemetry, but are equally applicable to OpenTelemetry.
+
+## Open questions
+
+Should the prefix `app.` be used, or should it be something else?
+
+Should there be a limit places on number of sub-namespaces?
+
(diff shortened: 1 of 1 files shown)


text/0000-app-namespacing-semantic-convention.md
```diff
@@ -0,0 +1,61 @@
+# Application namespacing semantic convention
+
+Adds a semantic convention for manual instrumentation that defines how to namespace custom attributes, starting with the `app` prefix.
+
+## Motivation
+
+Today, custom attributes from an app created through manual instrumentation have all kinds of names that are specific to the domain of that app. There are no conventions for how to organize this data, though.
+
+Data organization is critical to being able to effectively analyze observability data and avoid collisions in names. Without a well-defined convention for developers to follow, their own means of organization can differ from developer to developer, leading to a need to do things like coalesce field names using `.`/`-`/`_`/etc, not to mention having a real possibility to have name collisions. to separate clauses in an attribute name. Instead, a well-documented convention for how to organize data and in what format removes a "need to think" like code formatting conventions can do.
```
Oberon00, line 9: This kind of high-level naming convention is already defined at https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md

text/0000-app-namespacing-semantic-convention.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Explanation
+
+Attributes created through manual instrumentation should use namespacing for organizational purposes, using `app.` as a prefix for the name. Sub-namespaces should use a period (`.`) to separate sub-namespaces.
```
Oberon00, line 13: One of the main purposes for namespaces is to avoid nameclashes. If everybody uses the same namespace, that kind of defeats the purpose. Also, something that is an "app" today might be a common library tomorrow.

cartermp: I don't think I understand this comment. You can get name clashes today without prefixing anything. But this is a way to better distinguish things that didn't come from an SDK or library or other component. How does this make name clashes worse?

Oberon00: Not sure if this makes it worse, but I think it would be better to suggest that each app chooses their own root namespace for its attributes.

cartermp: I could see that being positive for folks to have flexibility. But I'm also leaning a bit into the dynamic of automated code formatters, which take certain decisions away and optionally give off-ramps from the opinionated default. What if instead of what's proposed today, the guidance is to choose your own root namespace, with `app` as a "blessed" example?

plantfansam: +1 to reserving `app` as a default, protected-within-OTEL namespace, but encouraging users select their own prefix. This also implies that we ought to have a blacklist of protected attribute namespaces (e.g. `http.*`, `exception.*`), which can be inferred from [various](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions) [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions).

svrnm: Note: `app.*` was also suggested by the otel-client-side-telemetry SIG for semantic conventions of browser/mobile/desktop apps, not sure if this is still a thing (cc: @scheler, @martinkuba)

Beyond that +1:  I agree that it would be helpful to have namespaces that say "this is outside of the _offical_ semantic conventions". I assume that the potential is high, that without such a namespace a collision between future _official_ semantic conventions and what people are doing _custom_, is high: 

* My note above on dual use of `app` is actually an instance of a potential collision where the community wants to use something for a semantic convention while the same namespace is already used in the demo app (and others) for "custom attributes". 
* Think about any over-used word, turn it into a namespace and you might face the same problem in the future (`iot`, `mobile`,`mesh`,`runtime`,...)

I wrote "namespaces that say 'this is outside of the _offical_ semantic conventions'" above because I argue that there might be need for more than one:

* The one used by the end-user via manual instrumentations @cartermp is writing about in this OTEP (`app`, or some alternative, maybe `my`?)
* [There are already "protected" spaces for cloud vendors](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#cloud-provider-specific-attributes)
* There might be a similar use case for "vendors"/"distributions"

(Maybe that's overengineering that problem...)

> This also implies that we ought to have a blacklist of protected attribute namespaces (e.g. http.*, exception.*), which can be inferred from [various](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions) [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions).

💯 -- having some (soft/debug) checks that tell the user that they are using attributes out of scope (e.g. http.verb instead of http.method or db.order.id) would be a fantastic way to help people make good use of the semantic conventions.

text/0000-app-namespacing-semantic-convention.md
```diff
@@ -0,0 +1,61 @@
...
+
+Use as many layers of subnamespaces as needed to represent your app's hierarchy. For example, an eCommerce app may define attributes like this:
+
+* `app.order.id`
+* `app.cart.items.count`
+* `app.shipping.cost.total`
```
scheler, line 19: Not sure if I am missing something but it is not clear to me why these attributes are not prefixed with `com.<company>.`. If a customer chooses to do manual instrumentation and introduce attributes that the telemetry backends do not have a _standard_ interpretation for, then they fit the definition of custom attributes.  That is, they need to tell the telemetry backend how to interpret these attributes, say through configuration.

In this example of eCommerce app, since it is common enough I believe we can add them to the semantic conventions in the spec and still use them from manual instrumentation.

text/0000-app-namespacing-semantic-convention.md
scheler: > Note: app.* was also suggested by the otel-client-side-telemetry SIG for semantic conventions of browser/mobile/desktop apps, not sure if this is still a thing

We tried introducing `app.name` in the past as a replacement to `service.name` in the context of client apps, since the term service doesn't make sense for client side apps. It was not accepted, since making `service.name` optional has other implications.

text/0000-app-namespacing-semantic-convention.md
```diff
@@ -0,0 +1,61 @@
...
+
+Finally, like any standard convention, it's not much use if it's not documented. The proposed additions to documentation (even just having the right link the right places) are nontrivial. One potential way to mitigate this could be to use tooling to somehow suggest or enforce this convention (e.g., Analyzers in C#). However, that is a much bigger lift than defining a spec convention, and not all languages are guaranteed to have that kind of tooling available.
+
+## Prior art and alternatives
+
+Namespacing attributes is a documented best practice for [some vendors](https://docs.honeycomb.io/getting-data-in/data-best-practices/#namespace-custom-fields). These practices emerged before OpenTelemetry, but are equally applicable to OpenTelemetry.
+
+This convention is also being used by the [OpenTelemetry Community Demo](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/checkoutservice/main.go#L265-L267).
```
scheler, line 49: It would be risky to use attributes names not part of the spec.  IMHO, as a best practice, the community demo should also not be using the app namespace as it's not in the spec.

svrnm: @scheler, maybe worth raising this as an issue with the community demo?

cartermp: Best to raise it there. They were intentional about doing it but I don't know exactly what the decision process was. FWIW I think the demo should establish conventions for things like attribute naming even if they're not in the spec.

text/0000-app-namespacing-semantic-convention.md
svrnm: @Oberon00 made a similar comment [above](https://github.com/open-telemetry/oteps/pull/211#discussion_r948154299). Copying from the spec:

> * The name is specific to your company and may be possibly used outside the company as well. To avoid clashes with names introduced by other companies (in a distributed system that uses applications from multiple vendors) it is recommended to prefix the new name by your company's reverse domain name, e.g. com.acme.shopname.
> * The name is specific to your application that will be used internally only. If you already have an internal company process that helps you to ensure no name clashes happen then feel free to follow it. Otherwise it is recommended to prefix the attribute name by your application name, provided that the application name is reasonably unique within your organization (e.g. myuniquemapapp.longitude is likely fine). Make sure the application name does not clash with an existing semantic convention namespace.
> * The name may be generally applicable to applications in the industry. In that case consider submitting a proposal to this specification to add a new name to the semantic conventions, and if necessary also to add a new namespace.

All of those are suggestions (at least I don't read any enforcing language there). I read some of the OTEP and some of the comments here, that it would be good to **enforce** a namespace for applications, e.g. to avoid a clash not only with existing but also with future namespaces.

I think that the `com.companyname.whatever` is already a good approach, but I see one issue with that: imagine you have acquired two pieces of off-the-shelf software, both have opentelemetry and send traces&metrics. Now both send you some information like the `order.id` with their spans/traces. Now the data emitted from software 1 is `com.vendor1.order.id` and the data emitted from software 2 is `com.vendor2.order.id`, although both might be the same. So what you might want to have is a `heresmymagixprefix.order.id` in both cases.

scheler: @svrnm what do you have in mind on how the backend should interpret the custom attributes. My understanding is that the customers should convey that to the backends. So in this case, they would have to convey `com.vendor1.order.id` and `com.vendor1.order.id` as being for the same purpose.

tedsuo:
@cartermp we are marking OTEPs which are no longer being updated as `stale`. We will close this PR in one week. If you wish to resurrect this PR, that is fine.

Issue #193, opened 2021-12-09
Proposal: "Plugable backend" Tracing Client/Query library


lucasponce:
Hello all,

Apologizes in advance as I am not sure if this repo is the proper one to start a proposal, please, feel free to redirect me to the right channel.

I'd like to ask if the OpenTelemetry specification is considering to create an "agnostic" backend library to consume traces stored in any  tracing platform (Jaeger, Tempo, Zipkin).

Today, applications can use opentelemetry sdk to ingest traces and spans in a common format. Projects like Kiali https://kiali.io/ consume metrics, traces, logs, configuration to correlate and combine to offer to the user added value in the Service Mesh domain.

One of the request from users is the possibility to change the tracing platform i.e. https://github.com/kiali/kiali/issues/4278 but unfortunately if you want to query traces for a specific platform you'd require some technical dependency on that platform (in the Kiali case, this one is Jaeger).

There are proto definitions for a gRPC service but I think on this side it's not obvious to change from one platform or other (or at least I'm not aware of an effort in the open-telemetry group around this).

So, one possibility is to wait that the specific implementations (Jaeger, Zipkin, Tempo, others) would implement a common API for backend queries, but for backend applications it would be nice to have a single dependency (i.e. opentelemetry-backend) that can enable to query one or another backend in a common format.

In the Kiali project this is something we'd like to offer to the users and we'd like to explore the possibilites to create a proxy client [2] that can abstract the consumer of the traces/spans about the details of specific implementations (Jaeger, Tempo, others) for easy change from one platform to another.

We think this effort could be interesting for the OpenTelemetry community and it would be nice if it's fostered under the OpenTelemetry umbrella allowing users to participate more actively.

My goal here would be to get feedback about:
- Is this idea interesting for the OpenTelemetry community ?
- If there is some similar project started, we wouldn't like to start anything new but join forces and collaborate with it.
- If not we could volunteer to start some PoC around, in similar fashion like [2] but trying to help to a wider audience, not only to meet some specific requeriments of the Kiali project.

Any feedback would be welcome.

Thank you !
Lucas

[2] https://github.com/lucasponce/jaeger-proto-client

zirain:
@lucasponce any update about this? we're facing same request.

lucasponce:
Hi @zirain, I was waiting to have some feedback.
But I think we could work in two directions:
[...]

PR #218, opened 2022-09-14
Standard Telemetry Data Query API

itaykat:
## Standard Telemetry Data Query API

This change suggest a standard API definition to query telemetry data stored in any observability backend platform, both open source (Jaeger, Tempo, Zipkin, etc.) and proprietary (Datadog, appdynamics, etc.)

This idea of "having a standard for trace query API " has also been discussed by @lucasponce and others at length in https://github.com/open-telemetry/oteps/issues/193.

Collaborators and Reviewers: @ronyis @lucasponce

Comments, ideas, feedback, etc. are all very welcome and highly appreciated!


text/0218-standard-telemetry-query-api.md
```diff
@@ -0,0 +1,130 @@
...
+
+**Sooner than later, at the same rhythm that the adoption of OTel producers (instrument, collect and export) grows, the need for a standard backend to consume these telemetry signals will grow as well.**
+
+## Explanation
+
+- OTel focus on a **vendor-agnostic** standard for producing telemetry signals, but there is a gap (or no standard) in a common API to query the stored telemetry signals.
+
+  Unfortunately, if you want to query traces for a specific platform you'd require some technical dependency on that platform (in the Kiali case, this one is the Jaeger API) which results in **vendor-specific** solution.
+
+- This suggested specification should prioritize Traces API definition as this is the most mature capability of OTel and consist of great variety of backend platforms.
```
galbash, line 24: I think this is a good idea for two reasons:
1. Starting with just data type will limit the blast radius.
2. Compared to metrics which have PromQL, traces has no standardization whatsoever.

text/0218-standard-telemetry-query-api.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Internal details
+
+This change will be an additional API and SDK where a specific set of common Telemetry Data query and search capabilities will be defined.
```
galbash, line 28: IMO it's fine to also start with just an API, and build this incrementally.

text/0218-standard-telemetry-query-api.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Explanation
+
+- OTel focus on a **vendor-agnostic** standard for producing telemetry signals, but there is a gap (or no standard) in a common API to query the stored telemetry signals.
+
+  Unfortunately, if you want to query traces for a specific platform you'd require some technical dependency on that platform (in the Kiali case, this one is the Jaeger API) which results in **vendor-specific** solution.
+
+- This suggested specification should prioritize Traces API definition as this is the most mature capability of OTel and consist of great variety of backend platforms.
+
+## Internal details
```
galbash, line 26: I suggest that before getting to specing the API capabilities, we agree on the guiding principles.
One guiding principle I would suggest is "extensible": The API should be good enough for most use cases, but it can also be extended by a vendor to support it's specific use cases in an easy manner.
Another Guiding principle I would discuss would be "is it machine or human first" - efficiency vs readability.

Once we as a community agree on the goal (stated above) and the guiding principles, I believe it would be easier to agree on the actual API, because we are aligned on what's important to us on the higher-level.

text/0218-standard-telemetry-query-api.md
```diff
@@ -0,0 +1,130 @@
...
+
+This change will be an additional API and SDK where a specific set of common Telemetry Data query and search capabilities will be defined.
+Once a set of these capabilities will defined, a technical specification describing the data exchange between the telemetry backends and the consumers will be defined along with its delivery protocols (gRPS/HTTP) and schema (Protbuf i.e).
```
galbash, line 29:
```suggestion
Once a set of these capabilities will defined, a technical specification describing the data exchange between the telemetry backends and the consumers will be defined along with its delivery protocols (gRPC/HTTP) and schema (Protbuf i.e).
```

text/0218-standard-telemetry-query-api.md
```diff
@@ -0,0 +1,130 @@
...
+
+- This API could leverage the existing protbuf schema, in the case of a Trace - the response object can be derived from OTel [Trace Definition](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto)
+- The API route can complement the existing [OTLP Exporter API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md)
+
+### Example of existing Trace Query APIs
```
galbash, line 34:
```suggestion
### Examples of existing Trace Query APIs
```

text/0218-standard-telemetry-query-api.md
```diff
@@ -0,0 +1,130 @@
...
+### Standard API vs. Backend Proxy Library
+
+- In order to make the shift toward this standard more pleasant from the telemetry consumers side, while reducing the dependency on external platforms, a dedicated contrib repo should be created with a proxy client that will abstract the details of specific implementations (Jaeger, Tempo, etc.) so that the user will be able to switch one observability backend to another while enabling telemetry consumers to query any supported observability backend in a common format.
+
+## Prior art and alternatives
+
+- The proxy library mentioned above can be inspired by this [POC](https://github.com/lucasponce/jaeger-proto-client) made by @lucasponce that can plug any Jaeger solution.
+- [Tempo TraceQL](https://github.com/grafana/tempo/blob/main/docs/design-proposals/2022-04%20TraceQL%20Concepts.md), a language for selecting traces that Tempo will implement. This language is currently only focused on trace selection. This is an example for a specific implementation for querying syntax.
+
+## Open questions
```
galbash, line 114: I'd add that the actual mandatory and optional API capabilities is an open question - @itaykat I'm sure you have an initial idea of what capabilities you would like to see in the API. What is the process of making the next step in this discussion?

galbash (commented):
This is a great initiative IMO. I can see the advantage of the OTel community guiding this effort as well. The only thing that I would note is that in my opinion we should describe a base and leave room for vendor extensibility and flexibility - the goal should be to enable the vendors to move quicker and better, and not force them into an API (much like the current OTel APIs where a vendor can choose to support only the SDK and create and exporter, or to adopt OTLP for example. The technical details are different of course, but the guiding principle should be the same)

jpkrohling:
Are there products and projects committed to implementing such an API already?

tigrannajaryan:
From what I see this OTEP does not define specifically any Query API. Instead this reads more as a declaration of an intent and a request for comments and opinions about whether OpenTelemetry wants to have such query API.

It is somewhat unusual but OK to have OTEPs of this kind to gauge the initial interest around a topic. I would suggest to circulate this widely on Otel Slack and in Otel Specification SIG meeting.

I already commented on https://github.com/open-telemetry/oteps/issues/193 that an initiative like this may look more like a project than a single OTEP (unless you manage to constraint the scope such that it can be reviewed and approved as one atomic piece of work). Given that this OTEP has to be followed by at least another one that defines the actual Query API, it seems like it is already is closer to being a project :-)

My personal opinion is that this somewhat deviates from how we define OpenTelemetry's charter today. However I can also see the value of such Query API so I don't want to rule out the possibility of modifying the charter to include this. It will need to be widely supported though for this to be possible, including by vendors expressing interest in implementing this API.

reyang (changes requested):
https://github.com/open-telemetry/oteps/pull/218#issuecomment-1247081950

itaykat:
I agree with @tigrannajaryan about the initial need to "feel" the OTel community around this topic.

This need also relates to @jpkrohling question about the intent of telemetry platforms (Jaeger, Tempo, SigNoz, etc.) to implement such a standard.

I'll add another required validation which is the telemetry consumers, the platforms which consume telemetry data for internal usage.

So currently required validations:
1. The intent of observability backends to implement such an API
2. The real need for such an API from the telemetry consumers side, more examples for such platforms.
3. The intent of the OTel community to have such a project (generally speaking this is not a blocker for such initiative)

text/0218-standard-telemetry-query-api.md
itaykat: Please see https://github.com/open-telemetry/oteps/pull/218#issuecomment-1247114097

jpkrohling:
I probably can't join next week's, but I can add this to the agenda and am happy to introduce the topic on the call from 29 September 2022 if it's not addressed sooner.

I do "feel" like there's a community interested in having this and I did hear from users quite often about this, including from @lponce and, more recently, from @VineethReddy02. My only concern right now is investing energy in a proposal without anyone committing any engineering resources to implement this, where we could be investing this energy elsewhere.

reyang:
> > I think the Governance Committee needs to be involved if there is desire to change the charter.
> 
> Thanks @reyang this is a concern I also mentioned in the open questions section and would love to get some feedback about.
> 
> Can you pleas tag someone from the Governance Committee?

@open-telemetry/governance-committee please advise here, thanks!

itaykat:
> I do "feel" like there's a community interested in having this and I did hear from users quite often about this, including from @lponce and, more recently, from @VineethReddy02.

Hi @lponce and @VineethReddy02, can you please share your use-cases in that context (the need for a vendor-agnostic standard telemetry data query api), it will highly benefit the discussion. Thanks.

itaykat:
> My only concern right now is investing energy in a proposal without anyone committing any engineering resources to implement this, where we could be investing this energy elsewhere.

Our group at Cisco has the intent, will and engineering resources to back this initiative, we just need sufficient validation from the community around the need for such a standard.

pilhuhn:
The charter already says "export". This request is about exporting from a backend. Q.e.D. :-)

itaykat:
> The charter already says "export". This request is about exporting from a backend. Q.e.D. :-)

Actually this may be a good point, telemetry data query can be seen as exporting, @jpkrohling this nuance worth a mention in the GC meeting today.

jpkrohling:
I discussed this today with the other GC members, and while you can [watch the recording](https://www.youtube.com/playlist?list=PLVYDBkQ1Tdyzg1CuQgd9mdjwOUYg7ECYR) once it's uploaded and read the [meeting notes](https://docs.google.com/document/d/1-23Sf7-xZK3OL5Ogv2pK0NP9YotlSa0PKU9bvvtQwp8/edit#) yourselves, I thought it would be worth summarizing what was discussed there.

- the most important point is that even though the sentiment around this isn't positive (see points below), this isn't a binding decision
- 4 GC members weren't present at the call today, so we can't make a statement as a group, although we can share the sentiment that seems to be the consensus among the five members that were present
- without projects and vendors willing to implement this query API themselves, there's no point in having this as a standard
- we have work on many fronts right now; without strong arguments, we can't add this new one
- the argument that the word "export" could be interpreted as "export from a backend" didn't fly (sorry @pilhuhn); that wasn't the spirit of the word in that context. Additionally, it could be argued that export isn't querying: vendors and projects could very well just add an "export as OTLP" button on their interfaces without providing a query API
- OpenTelemetry's vision does not currently have a space for this -- though if we decide in the future that this is a good thing to have, we can adjust the vision and charter
- OpenTelemetry isn't a standards body, and it's even harder to sell the idea of defining a standard if we don't plan on having an implementation ourselves
- this isn't to say that this wouldn't be welcomed outside of OpenTelemetry: nothing prevents the community from establishing such an API and recruiting commercial vendors and open source projects to adopt this; if this gets enough traction, the case for having this hosted under the OpenTelemetry's project would be far stronger than it is today

My _personal_ opinions are:

- something like this would be good for the industry, as @lucasponce's use case has evidenced. Just like SQL standardizes data retrieval from databases without preventing competition, a common set of operations would certainly make sense here.
- so far, I haven't seen anyone from distributed tracing commercial vendors or open-source projects backing this, which is kind of telling. Without buy-in from possible implementors, I wouldn't consider having this as part of OpenTelemetry

bhs:
There are really two questions here...

1. Is a common effort around a multi-telemetry, analytical query language worthwhile?
2. Should such an effort be part of OpenTelemetry?

Re question 1: quite possibly!

Re question 2: OpenTelemetry is already a project that's much wider than it is deep... and even for core signal types, the core mission ("to enable effective observability by making high-quality, portable telemetry ubiquitous") is still very much a WIP. As such, I am skeptical that such an analytical query language be part of OTel per se.

My two cents. (And apologies that I had an immovable conflict with today's GC meeting – this is definitely an important topic!)

lucasponce:
I think this comment is still valid in this context https://github.com/open-telemetry/oteps/issues/193#issuecomment-1241875146.

The observability space is getting mature, and the collection of signals is just a part of the flow.

On the other side, there are "consumers" of this information to add value to the end-users.

Starting to propose some "standard" API would help; otherwise, consumers would need to deal with different products and technology.

As this proposal shows, there is a need; I think OpenTelemetry is the best organization that can lead, give visibility and shape this effort.

I like the comments that point to start defining a "common set of operations" that perhaps can be mapped from existing solutions.

I understand this triggers more questions related to "the vision" of OpenTelemetry. Still, I think this request represents valid "actors" in this business, and increasing the scope beyond the "collection" of the signals is a reasonable evolution.

cartermp:
FWIW from a commercial vendor's standpoint (Honeycomb), we have had people ask for this kind of capability (we don't have an API to get traces like jaeger does). But it's not been many people. Most asks around APIs tend to be around CRUD for product nouns and aggregate data for reporting purposes, neither of which would fit within the OpenTelemetry vision IMO.

bhs:
I would kinda like to table the question of whether this is a valid *idea* (some sort of new query language) and instead focus on whether an analytical query language belongs in OTel, regardless of independent merit.

jmacd:
I agree that OpenTelemetry is not the right community or place to try to standardize a data query language or API.

However, I'm also excited to see OpenTelemetry data used with existing data platforms for streaming data analysis. In particular, I would like to focus attention on [this proposal to integrate Apache Arrow with OpenTelemetry](https://github.com/open-telemetry/oteps/pull/171) which will enable us to apply the data analysis tools in the Apache Arrow ecosystem to telemetry data.

itaykat:
@jmacd thank you for sharing your OTEP, this one may relate also to your suggestion:
Taken from the OpenTelemetry [Mission Vision page](https://github.com/open-telemetry/community/blob/main/mission-vision-values.md#telemetry-should-be-vendor-neutral):

> **Telemetry should be vendor-neutral**
For decades, proprietary drop-in agents from monitoring and observability vendors have been the primary source for useful telemetry from across the application stack. Unfortunately, the lack of common standards or APIs across these agents has led to vendor lock-in for customers, and inhibited innovation by tightly coupling telemetry collection with telemetry storage and analysis. With OpenTelemetry, we strive to provide a level playing field for all observability providers, avoid lock-in to any vendor, and interoperate with other OSS projects in the telemetry and observability ecosystem.

**_we strive to provide a level playing field for all observability providers, avoid lock-in to any vendor, and interoperate with other OSS projects in the telemetry and observability ecosystem._**

As I see it, unless we remove the lock-in to specific observability backend (vendor) by providing a standard api for querying telemetry data and enable vendor agnostic telemetry consumption - we won't align with this section in the vision, please correct me if I miss something. Thanks 🤍

pilhuhn:
I am with @itaykat 
And an API does not need to provide for every use case, but at least define some very basic use cases like 
* get trace by id
* List traces with filters (start time, end time, some set of attributes)
The data types are already known (Trace, Span) only a List-of-traces would need to be added.
And then allow for vendors to augment those basic building blocks.
I am sure @lucasponce has some more requirements, that could be added to the basic set.

bhs:
@itaykat, apologies if I'm sounding like a broken record, but the "mission and vision" excerpt you quoted was about removing vendor **agent** lock-in, with the header (which you've retained) pertaining to vendor-neutral **telemetry** – a narrower thing than vendor-neutral analytics and full-spectrum observability. Hope that is clear.

pilhuhn:
@bhs I think no one asks for vendor-neutral *analytics*, but for a way to get your own data back out from the vendors system in a vendor-neutral way.

itaykat:
> with the header (which you've retained) pertaining to vendor-neutral telemetry – a narrower thing than vendor-neutral analytics and full-spectrum observability

Thanks @bhs for emphasizing, this is correct, I mentioned analysis as a superset which includes the telemetry query.

Adding to @pilhuhn here, this initiative talks about vendor-neutral telemetry.

If we look at the downstream Telemetry consumption path:

1. Instrumentation
2. Collection
3. Exportation & Ingestion
4. Storage
5. Query (either pure extraction of the data for enrichment purposes of further analysis of it)

We can all agree (hopefully) that stages 1-3 are indeed vendor neutral and maintained by OTel, though once we go down the path, stages 4 and 5 are definitely not.

This conflicts with the vision tittle I quoted:

> **Telemetry should be vendor-neutral**

reyang:
We've discussed this during the Sep. 21st, 2022 Technical Committee Meeting.
The @open-telemetry/technical-committee agrees with the @open-telemetry/governance-committee that a data query language/API is not belonging to the OpenTelemetry charter/scope.

In addition, with the current workstreams (e.g. Prometheus, Semantic Conventions, Logging), the Technical Committee is already spreading thin and won't be able to effectively take any new work stream.

reyang:
The OpenTelemetry Collector project still needs a solution for its data transformation, so it's likely that the Collector SIG would be interested in providing requirements/feedback regarding a data transformation language (which might be a subset of a query language, or a completely different language).

itaykat:
This conversation may now continue on the CNCF Slack channel #otel-telemetry-consumers, you are all welcomed.
Thank you for sharing your thoughts.

itaykat:
> The OpenTelemetry Collector project still needs a solution for its data transformation, so it's likely that the Collector SIG would be interested in providing requirements/feedback regarding a data transformation language (which might be a subset of a query language, or a completely different language).

Thank you for sharing this insight, also watched @bhs talked about that need in the last GC meeting. Can you please share more resources/issues/users around this topic so I can continue that conversation?

itaykat:
> FYI all, Observability TAG (not Otel) plans to work on this, see https://docs.google.com/document/d/1JRQ4hoLtvWl6NqBu_RN8T7tFaFY5jkzdzsB9H-V370A/edit

Thanks @tigrannajaryan for mentioning the charter by @manolama and also mentioning this OTEP in his charter, this is highly relevant.

PR #224, opened 2023-01-13
Experimental Semantic Convention Attribute Registration

dyladan:
TODOs

- [ ] Add lifecycle for attributes
- [ ] Specify what the dictionary itself actually contains


text/0224-experimental-semantic-conventions.md
```diff
@@ -0,0 +1,115 @@
...
+
+**Registered Attribute**: A registered attribute is any attribute which is in the provisional registry outlined in the proposal or the semantic conventions.
+
+## Proposal
+
+For the purposes of this document only attributes are discussed, but it should be understood that the TC may decide to make other telemetry components such as metric names, event names, or other components registerable.
+This proposal SHOULD also apply to any such components at the discretion of the TC.
+
+A procedure will be established by the technical committee (TC), or an entity designated by the TC, for registration of provisional semantic convention attributes.
+Any non-private attributes which can be registered SHOULD be registered in the provisional registry or an official OpenTelemetry semantic convention.
```
yurishkuro, line 34: I am somewhat uncomfortable with the vagueness here. What, realistically, is going to be an implementation of a registry? Are we going to maintain an online database with an API? Are we going to bifurcate the "registration" process of the official and provisional attributes?

Why not just make a call right now and say that provisional attributes (which I think is a better name than "registered") are going to use the exact same mechanism of registration as the official attributes, i.e. the semantic conventions yaml file in OTEL repo. And we need to decide if it's going to be a separate repo or a separate directory. Or should this be even the same directory and just an extra label in the yam definition?

yurishkuro: A related topic is that provisional attributes should also be subject to the OpenTelemetry Schemas versioning mechanism (which also suggests we should not bifurcate the registration process).

text/0224-experimental-semantic-conventions.md
```diff
@@ -0,0 +1,115 @@
+# Allowing experimental extensions to Semantic Conventions
+
+A process that allows experimental / non-stable semantic conventions to exist within stable versions and how they evolve into stable semantic conventions.
```
yurishkuro, line 3: Nit: in the below comments I used the term "provisional arguments", but I think it would be better to pick a single term and use it consistently. Is there a semantic difference between provisional and experimental?

yurishkuro (commented):
+1 overall

text/0224-experimental-semantic-conventions.md
dyladan: We already have the concept of "experimental" and "stable" semconv. It was my intention that if the TC wanted to, this OTEP could be interpreted to mean simply that the bar is drastically lowered for "experimental" semantic convention. In that interpretation there would be no difference between experimental and provisional.

I didn't want to specify that here, because I didn't know if the TC would prefer to keep the provisional registry separate as it may grow very large very quickly.

text/0224-experimental-semantic-conventions.md
```diff
@@ -0,0 +1,115 @@
...
+**Attribute**: For the purposes of this document, an attribute may refer to any component of any signal emitted by OpenTelemetry instrumentation.
+This may include span or metric attribute keys and values, span or metric names, log or event attribute names and values, or any other value emitted by OTel instrumentation.
+
+**Private Attribute**: An attribute is private if it is meant only for the exclusive private use of the instrumentation author, and is not expected to be emitted by any public library or understood by any public backend tools.
+
+**Registered Attribute**: A registered attribute is any attribute which is in the provisional registry outlined in the proposal or the semantic conventions.
+
+## Proposal
+
+For the purposes of this document only attributes are discussed, but it should be understood that the TC may decide to make other telemetry components such as metric names, event names, or other components registerable.
```
lmolkova, line 30: nit: relationships between spans

text/0224-experimental-semantic-conventions.md
```diff
@@ -0,0 +1,115 @@
+# Allowing experimental extensions to Semantic Conventions
```
lmolkova, line 1: Perhaps `experimental attributes` would be more precise? 
Conventions describe relationships between spans, how attributes depend on other attributes, span kind, name, events populated under spans, context. etc. It'd be hard to apply process outlined here for things like that.

text/0224-experimental-semantic-conventions.md
```diff
@@ -0,0 +1,115 @@
...
+Consumers SHOULD consider any security and privacy implications of displaying unhidden attribute values.
+
+## Alternatives Considered
+
+### "x." prefix
+
+This alternative involves establishing an “x.” prefix convention for experimental conventions.
+For example, "x.mydatabase.attribute" may be an experimental version of the eventual "mydatabase.attribute" attribute key.
+See [X- convention for HTTP message headers](#x--http-header-convention) below for more details about why this alternative was discarded.
+
```
lmolkova, line 66: I'd suggest the following approach as an  alternative:
1. having an otel-specification-contrib repo where semantic conventions can mature and evolve as a whole with less scrutiny or stability expectations. 
  - Otel-owned instrumentation might exist for them (in corresponding -contrib) repos, but would be in a preview state
3. tech-specific specs can live there as well
4. mature general-purpose semantic conventions can be sent to -spec repo

lmolkova (changes requested):
Attribute registration should help with the discoverability and reuse of attributes between conventions, but it is a separate concern from semantic convention as a whole.

I suggest to reflect and clarify it in the document

text/0224-experimental-semantic-conventions.md
lmolkova: An example: 
messaging WG discusses that attributes can be populated on spans or on links. E.g. span that represents batch publish should have `msg.batch.message_count` attribute AND links to all individual messages. This is equally important to the backend as an attribute name.

You can't standardize and regulate attributes without also regulating the stability of the behavior. But I don't think this document addresses the evolution of behavior. 

Hence, I think this document should be limited to attributes only and don't attempt to tackle semantic convention as a whole

trask: > I think this document should be limited to attributes only and don't attempt to tackle semantic convention as a whole

I like this idea, something like

```suggestion
# Supporting experimental attributes in OpenTelemetry instrumentation
```

dyladan: @jsuereth what do you think about this suggestion? it is a bit of a change in scope but does more closely match what we've laid out.

text/0224-experimental-semantic-conventions.md
jsuereth: Practically, there would need to be a YAML file of sorts, the same we have for semantic conventions.

I wanted to chime on this one, although I haven't had a chance to solidify my thoughts.  Effectively, such a registry of "provisional" conventions is an inevitability.  The question is whether we limit this to an exclusive group of "expert committees" like we're doing for HTTP, or if we want to allow more organic growth (like what @dyladan and I propose here).

I think currently, we're using expert committees for OTEL because of the following reasons:

- There's a lot of high-value, high-impact conventions for (HTTP, Databases, Message-Passing, e.g.) observability.
- We CAN draw an expert committee 
- The cost of failure is high.  

Once we whittle down the big hitters, we have the issue of the "long-tail" of observability and our ability to attract enough experts and *bandwidth*.   A process like this allows us to rapidly expand to help o11y users.

I guess I'm suggesting I think the need for this registry is inevitable, but it could also be "not now".  I.e. we may want to first ensure we have a solid and rich core offering of conventions before we solve long-tail issues.

This registry *does* solve the issue of "How do OpenTelemetry instrumentation authors create instrumentation that's not part of the semantic conventions".   However, this registry is also, in my view, the only possible path for semconv stability and also one that relies on having semconv be widely adopted, successful standard already, with meaningful starting point (like HTTP).

I.e. my current thinking is:

- We should enable this registry, after we've solidified semantic conventions for a few key areas.
- We probably need a different solution for "How do OpenTelemetry instrumentation authors create instrumentation that's not part of the semantic conventions"

dyladan: If the answer is "not now" we could easily simplify this to a policy change where we allow/encourage contrib and community authors to experiment with instrumentation even if it isn't covered by semantic conventions. The end result should hopefully be the same with the best solutions bubbling to the top and becoming the most popular. It would mean different instrumentations for the same module might not export data the same way, but maybe that sort of competition is healthy.

text/0224-experimental-semantic-conventions.md
```diff
@@ -0,0 +1,115 @@
...
+Instrumentation authors SHOULD assume that any attribute they create may become standardized, public, commonly deployed, or usable across multiple OpenTelemetry clients, distributions, or receivers.
+Instrumentation authors SHOULD choose attribute keys that are descriptive, relevant, and that they have reason to believe are not currently in use or reserved by the TC for future use.
+Before creating a new attribute, the attribute author SHOULD review the semantic conventions and provisional registry to see if an attribute already exists which fits the intended use case.
+If such a registered attribute exists, then it SHOULD be preferred over the creation of new attributes.
+
+An entry MUST NOT be refused without credible reason to believe that registration would be harmful.
+Some reasonable grounds for refusal might be frivolous use of the registry, TC consensus that the attribute may be a security or privacy risk, conflict with another attribute or a name reserved by the TC for future use, or that the proposal is sufficiently lacking in purpose, or misleading about its purpose, that it can be held to be a waste of time and effort.
```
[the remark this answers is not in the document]
dyladan: @yurishkuro do you want me to clarify the wording or add examples in the OTEP? If so, is the above explanation sufficient?

text/0224-experimental-semantic-conventions.md
```diff
@@ -0,0 +1,115 @@
...
+Note that a disagreement about technical details are not considered grounds to refuse listing in the provisional registry.
+
+A provisional registry entry which has gained enough support, use, and market penetration SHOULD be promoted to the semantic conventions following existing procedures for review and inclusion.
```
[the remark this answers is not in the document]
dyladan: Is this a sufficient answer? If so, should I make that explicit in the OTEP?

text/0224-experimental-semantic-conventions.md
dyladan: > I'd suggest the following approach as an alternative:
> 
> 1. having an otel-specification-contrib repo where semantic conventions can mature and evolve as a whole with less scrutiny or stability expectations.
>    
>    * Otel-owned instrumentations might exist for them (in the corresponding -contrib) repos

I'm not really trying to encourage instability of attributes as part of the experimentation process. Instead of a breaking change for a particular attribute I would rather see a new attribute registered with a different name if the need arises.

> 2. semantic conventions are versioned individually with semver and it reflects their status (stable, beta, alpha, etc). version is populated on the telemetry as part of schema URL that makes status explicit
>    
>    * otel-owned instrumentations  share the same status (stable, beta, alpha) as the conventions they implement

I think there are 2 things here which need to be versioned: the individual attributes and the semantic conventions which use them.

> 3. tech-specific specs can live there as well
> 4. mature general-purpose semantic conventions can be sent to -spec repo

I see value in your proposal here, but I think it covers a slightly different target than I was going for. I could see the two solutions both being implemented eventually.

dyladan: I think 99% of the value would be covered by simply registering attribute names (agnostic of signal type). As JS maintainer most of the instrumentation requests I get are "please add `x` data to the span," or "please add a hook/configuration to `y` instrumentation so I can add arbitrary attributes to the span." Usually if the requested data is not covered by semconv we don't know what to do. There are already many instances "in the wild" of users adding custom attributes using whatever key made sense to them because it wasn't covered by semconv. If there was a registry of these keys at least users who add custom attributes might do so in a consistent manner.

In terms of span kind and parent/child/link structure, we really don't see many requests (at least in JS)

dyladan:
I had been waiting for the spec->semconv transition to get back on this. I'll update this OTEP this week with feedback received.

text/0224-experimental-semantic-conventions.md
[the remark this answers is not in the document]
dyladan: Added more examples and attempted to clarify wording.

text/0224-experimental-semantic-conventions.md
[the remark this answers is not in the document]
dyladan: This part is now removed

text/0224-experimental-semantic-conventions.md
dyladan: I believe the newly limited scope of the proposal to an attribute dictionary should address this concern. Please resolve this conversation if you agree.

text/0224-experimental-semantic-conventions.md
dyladan: I believe the newly limited scope of the proposal to an attribute dictionary should address this concern. Please resolve this conversation if you agree.

text/0224-experimental-semantic-conventions.md
dyladan: Limited scope of the proposal to an attribute dictionary

text/0224-experimental-semantic-conventions.md
dyladan: I limited the scope of the proposal to just an attribute dictionary specifically to address this concern. If it goes well, maybe the idea will be extended to other areas of semantic conventions in the future.

text/0224-experimental-semantic-conventions.md
dyladan: I believe the newly limited scope of the proposal to an attribute dictionary should address this concern. Please resolve this conversation if you agree.

dyladan:
I limited the scope of the proposal to just an attribute dictionary to address concerns raised by @lmolkova and others. I believe 99% of the value of the proposal is retained simply by allowing instrumentation authors to use attributes that are not already included in a semantic convention. If it goes well, maybe the idea will be extended to other areas of semantic conventions in the future.

text/0224-experimental-semantic-conventions.md
```diff
@@ -0,0 +1,117 @@
...
+
+Instrumentation authors SHOULD assume that any attribute they create may become standardized, public, commonly deployed, or usable across multiple OpenTelemetry clients, distributions, or receivers.
+They SHOULD choose attribute keys that are descriptive, relevant, and that they have reason to believe are not currently in use or already proposed.
+Before creating a new attribute, the attribute author SHOULD review the attribute dictionary to see if an attribute already exists which fits the intended use case.
+If such a registered attribute exists, then it SHOULD be preferred over the creation of new attributes unless there is sufficient reason to create a new attribute.
+Any new attributes SHOULD be registered in the attribute dictionary before an instrumentation using the attribute is released.
+If an instrumentation uses an attribute not included in the semantic conventions it is recommended that the attribute not be collected by default unless it is fundamental to the instrumentation.
+For example, an instrumentation which instruments a component not yet addressed by semantic conventions.
```
lmolkova, line 43:
```suggestion
```
It probably makes sense to turn the whole instrumentation on/off for such a component, not individual attributes, but then it goes out of the scope of this OTEP

text/0224-experimental-semantic-conventions.md
```diff
@@ -0,0 +1,117 @@
...
+
+An entry MUST NOT be refused without credible reason to believe that registration would be harmful.
+Note that a disagreement about technical details are not considered grounds to refuse listing in the attribute dictionary.
+Some examples of reasonable grounds for refusal: frivolous use of the dictionary such as registering nonsense words or inside jokes, a credible security or privacy risk, conflict with another attribute or proposal for a new attribute, or that the proposal is sufficiently lacking in purpose, or misleading about its purpose, that it can be held to be a waste of time and effort.
+This list is non-exhaustive and final decision for refusal is at the discretion of the TC.
+
+Once an attribute is registered it MUST NOT receive any breaking changes.
```
lmolkova, line 50: We need some time to experiment with attributes before they stabilize and we already have `experimental` and `stable` maturity levels defined in the build tools. 
Can we define them explicitly here and allow experimental attributes to get breaking changes?

text/0224-experimental-semantic-conventions.md
```diff
@@ -0,0 +1,117 @@
...
+This document explicitly does NOT define a process for removal of a registered attribute from the attribute dictionary.
+Because absence of evidence does not constitute evidence of absence, there is no way to know if any registered attribute is in active use.
+If a registered attribute is deemed by the TC to be harmful in some way, it SHOULD be marked deprecated, sufficient reasoning for the deprecation should be included in the dictionary, and, if appropriate, a new attribute SHOULD be registered with a different name.
+A registered attribute MAY ONLY be removed if, at the discretion of the TC, it is determined to be harmful or offensive to the project, its contributors, users, or some other group.
```
lmolkova, line 54: I suggest applying this requirement to stable attributes only

text/0224-experimental-semantic-conventions.md
dyladan: This is meant to be an example of when you would _not_ turn off the attribute. I'll try to rework the wording to make that more clear.

text/0224-experimental-semantic-conventions.md
```diff
@@ -0,0 +1,117 @@
...
+A procedure will be established by the technical committee (TC), or an entity designated by the TC, for registration of new semantic convention attributes.
+Any attributes not intended strictly for private use SHOULD be registered in the attribute dictionary.
+
+Instrumentation authors SHOULD assume that any attribute they create may become standardized, public, commonly deployed, or usable across multiple OpenTelemetry clients, distributions, or receivers.
+They SHOULD choose attribute keys that are descriptive, relevant, and that they have reason to believe are not currently in use or already proposed.
+Before creating a new attribute, the attribute author SHOULD review the attribute dictionary to see if an attribute already exists which fits the intended use case.
+If such a registered attribute exists, then it SHOULD be preferred over the creation of new attributes unless there is sufficient reason to create a new attribute.
+Any new attributes SHOULD be registered in the attribute dictionary before an instrumentation using the attribute is released.
```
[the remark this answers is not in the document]
dyladan: The idea is to prevent collisions. If some instrumentation starts emitting `db.my_attribute` but doesn't register it, another instrumentation may emit the same attribute. I think the blanket normative language is fine. Carving exemptions into the language for things we don't control seems unnecessary to me.

text/0224-experimental-semantic-conventions.md
dyladan: I really don't want that. I would much prefer a new attribute to be registered with a new name if a breaking change really needs to be made. Once an attribute is "in the wild" it should be treated as stable. If we register experimental attributes we're going to end up in the same situation we are now with the semconv where they sit for years in experimental then one day get broken and users are broken.

What if instead I update the language such that an attribute in early experimentation phase should _not_ be registered until the author is confident the format will not have a breaking change?

text/0224-experimental-semantic-conventions.md
dyladan: I don't want to distinguish between stable and experimental attributes. Any attribute in use should be stable or we will be breaking users, and any registered attribute should be assumed to be in use as we have no way to prove otherwise.

lmolkova: based on our experience in HTTP semconv, we renamed certain attributes a couple of times before we were happy with them.

having `http.method`, and `http.request.method`, and `http.request_method_canonical` meaning the same thing with slight variations in definition would be confusing and harmful. 

Also, based on this proposal semantic conventions cannot use attributes outside of the dictionary, which means that, however premature new semconv is, attributes get into the dictionary first, before prototyping/development of semconv is complete. It's unrealistic to expect the first version of an attribute to be the final one. 

So we either need
- a separate experimental dictionary
- an experimental level for attributes
- allow experimental semconv define their own attributes within that semconv and only move attributes into the dictionary once semconv is in the feature-freeze/stable

text/0224-experimental-semantic-conventions.md
lmolkova: see my other comment below - we renamed attributes a lot in the past (and made other breaking changes to them), there is no reason to believe we would not continue this.

> What if instead I update the language such that an attribute in early experimentation phase should not be registered until the author is confident the format will not have a breaking change?

then we need to allow experimental semconv to use their own, non-registered attributes until they are happy with them.

text/0224-experimental-attrribute-registration.md
```diff
@@ -0,0 +1,117 @@
...
+If an instrumentation uses an attribute not included in the semantic conventions it is recommended that the attribute not be collected by default unless it is fundamental to the instrumentation.
+For example, an instrumentation which instruments a component not yet addressed by semantic conventions.
+
+An entry MUST NOT be refused without credible reason to believe that registration would be harmful.
+Note that a disagreement about technical details are not considered grounds to refuse listing in the attribute dictionary.
+Some examples of reasonable grounds for refusal: frivolous use of the dictionary such as registering nonsense words or inside jokes, a credible security or privacy risk, conflict with another attribute or proposal for a new attribute, or that the proposal is sufficiently lacking in purpose, or misleading about its purpose, that it can be held to be a waste of time and effort.
+This list is non-exhaustive and final decision for refusal is at the discretion of the TC.
+
+Once an attribute is registered it MUST NOT receive any breaking changes.
```
AlexanderWert, line 50: Isn't the opposite the goal here? The motivation states that with this proposal we want to make it easier to experiment. But, with forbidding breaking changes on attributes registered in the dictionary we wouldn't it have the opposite effect?

At the same time I see the need for users of the attributes (e.g. semconv authors, instrumentation authors, etc.) to know how much they can rely on the stability of attributes they would use. 

How about having stability levels on attributes (similar to what we have with semantic conventions today)?
This is also how we handle it in ECS.

text/0224-experimental-attrribute-registration.md
```diff
@@ -0,0 +1,117 @@
...
+This document explicitly does NOT define a process for removal of a registered attribute from the attribute dictionary.
+Because absence of evidence does not constitute evidence of absence, there is no way to know if any registered attribute is in active use.
+If a registered attribute is deemed by the TC to be harmful in some way, it SHOULD be marked deprecated, sufficient reasoning for the deprecation should be included in the dictionary, and, if appropriate, a new attribute SHOULD be registered with a different name.
+A registered attribute MAY ONLY be removed if, at the discretion of the TC, it is determined to be harmful or offensive to the project, its contributors, users, or some other group.
+For example, if an attribute `http.headers` is registered and it is later determined by the TC to be harmful, it should be deprecated and a new attribute such as `http.message_headers` or similar created to replace it.
```
AlexanderWert, line 55: OK, I see, this is sort of the answer to my question above.

AlexanderWert (approved):
LGTM.

@dyladan WDYT about including a proposal on the process HOW we want to achieve that?

text/0224-experimental-semantic-conventions.md
AlexanderWert: > then we need to allow experimental semconv to use their own, non-registered attributes.

My concern with that is that we'll end up with a sort of chaotic state where parts of attributes are registered and others remain with their definition in the semconv for a long time. So the dictionary is then not really a true dictionary because it will miss many attributes.

text/0224-experimental-semantic-conventions.md
AlexanderWert: > I don't want to distinguish between stable and experimental attributes. Any attribute in use should be stable or we will be breaking users, and any registered attribute should be assumed to be in use as we have no way to prove otherwise.

I agree with @lmolkova here. Regarding breaking users by breaking attributes: how would it be different if we deprecate attributes in favor of introducing new ones? Effectively, this is a breaking change as well, because instrumentations will stop populating the deprecated attributes at some point in time (otherwise they would need to populate redundant attributes forever). 

Also, as I wrote below, I think we need a way to indicate for user how much they can rely on the stability of attributes. Some attributes can be rather stable (where breaking changes are very unlikely) other are intentionally not stable as they have been created for "experimentation purposes". But, how would (non-expert) semconv users know what applies to which attribute?

That's why I'm in favor of a stability level for attributes, too. 

In ECS we have a stability level per namespace and per field level. [See this example](https://github.com/elastic/ecs/blob/0525ea01e36ddd7315ef1153627dfba7eb3fdf70/schemas/process.yml#L341).

text/0224-experimental-attrribute-registration.md
```diff
@@ -0,0 +1,117 @@
...
+This proposal generally follows the recommendations of [RFC6648 section 4 “Recommendations for Protocol Designers.”](https://www.rfc-editor.org/rfc/rfc6648#section-4)
+RFC6648 recommends a registry of names be established which standardizes usage and prevents collisions.
+This OTEP separates a dictionary of semantic convention attributes from the semantic conventions, to be referenced by semantic conventions when necessary, and establishes a procedure for registering new and experimental semantic convention attributes.
+
+RFC6648 further recommends that X- prefix (or similar structure) MUST NOT be defined as experimental or nonstandard, and that names missing the X- prefix MUST NOT be defined as stable or standard.
+This is addressed below in [alternatives considered](#alternatives-considered).
+
+## Definitions
+
+**Attribute Dictionary**: The attribute dictionary contains all semantic convention attributes, including their names and value definitions.
```
pyohannes, line 25: Currently, besides the name and value definitions, semantic conventions define a description, notes, examples, and requirement levels for attributes.

Is it intended that any of those are part of the dictionary? Or is this some enrichment that takes place in the semantic conventions?

text/0224-experimental-attrribute-registration.md
```diff
@@ -0,0 +1,117 @@
...
+The lack of experimentation means there is often limited experience and evidence to guide  convention authors.
+It also often leaves instrumentation authors with little or no guidance if they wish to capture telemetry not yet covered by semantic conventions.
+This [Slack thread on #otel-specification](https://cloud-native.slack.com/archives/C01N7PP1THC/p1667997213067169?thread_ts=1667933561.313819&cid=C01N7PP1THC) highlights the problem in context of a specific semantic convention pull request.
+What should OpenTelemetry do with semantic convention proposals when there's unclear evidence that a particular signal is necessary or ubiquitous?
+
+## Explanation
+
+This proposal generally follows the recommendations of [RFC6648 section 4 “Recommendations for Protocol Designers.”](https://www.rfc-editor.org/rfc/rfc6648#section-4)
+RFC6648 recommends a registry of names be established which standardizes usage and prevents collisions.
+This OTEP separates a dictionary of semantic convention attributes from the semantic conventions, to be referenced by semantic conventions when necessary, and establishes a procedure for registering new and experimental semantic convention attributes.
```
pyohannes, line 18: Just for my understanding: the idea is, that it's easier to add attributes to the dictionary than to add them to the semantic conventions itself. A user can then add an attribute to the dictionary easily, and then experiment without fearing collisions, and eventually write semantic conventions referring the attributes in the dictionary?

text/0224-experimental-attrribute-registration.md
AlexanderWert: I think the attribute definition in the dictionary should cover the following:

- `name` (required)
- `type` (required)
- `description` (required)
- `examples` (required)
- `notes` (optional)
- `stability` (depending on what we decide on the above discussion)

`requirement_level`, `sampling relevance` and `tag` must not be part of the definition but will be specified on the usage of the attributes.

text/0224-experimental-attrribute-registration.md
AlexanderWert: Exactly. That also means that adding new attributes does not require discussion on where and how the attribute is intended to be used.

text/0224-experimental-semantic-conventions.md
lmolkova: > Also, as I wrote below, I think we need a way to indicate for user how much they can rely on the stability of attributes.

Great point! It means that stable -> deprecated is not a good lifecycle for the attribute, we still need the experimentation stage to communicate to users if they can really depend on the attribute

text/0224-experimental-attrribute-registration.md
pyohannes: Makes sense.

I think the OTEP should me more specific about this and give a definition of what the dictionary contains. It's important for understanding and working out how the attribute dictionary and semantic conventions interact, which I think is crucial.

text/0224-experimental-semantic-conventions.md
dyladan: >My concern with that is that we'll end up with a sort of chaotic state where parts of attributes are registered and others remain with their definition in the semconv for a long time. 

That is exactly what the registry is to prevent and why the barrier to entry is so low.

>then we need to allow experimental semconv to use their own, non-registered attributes.

I'd like to prevent that as much as we can.

> see my other comment below - we renamed attributes a lot in the past (and made other breaking changes to them), there is no reason to believe we would not continue this.

If an attribute needs to be renamed, a new attribute can simply be registered and the old one deprecated. The new attribute is then used in the semantic convention. I'm not sure what the concern is.

text/0224-experimental-semantic-conventions.md
dyladan: >I agree with @lmolkova here. Regarding breaking users by breaking attributes: how would it be different if we deprecate attributes in favor of introducing new ones? Effectively, this is a breaking change as well, because instrumentations will stop populating the deprecated attributes at some point in time (otherwise they would need to populate redundant attributes forever).

It is different because when you are interpreting your data you can still look up an old attribute in the dictionary and find out its meaning. If the definition is changed or it is renamed without keeping the old name around somewhere, this information is simply lost.

I think an attribute-specific experimental phase is reasonable, but I think it would be better if previous versions were kept around in some way so that older data can be interpreted. What would you think if we had an attribute-specific version? Anyone can register an experimental `0.x` attribute, but only semconv maintainers can promote attributes to `1.x`?

text/0224-experimental-attrribute-registration.md
dyladan: I want instrumentation authors to easily answer the question "how do I capture this piece of data that isn't covered by semconv?" As JS maintainer it is one of the most common questions I get and we are constantly being asked to add attributes which are not in semconv. Those users need telemetry stability just as much as any other. I consider their entries to be experiments, but if they are failed experiments it doesn't mean the attribute should go away or be broken, but a new attribute created taking into account the learnings of experience.

text/0224-experimental-attrribute-registration.md
dyladan: i'll update the otep to add more specificity

AlexanderWert:
@dyladan Here's a draft PR, proposing initial thoughts on how the yaml syntax for attribute definitions could look like.

https://github.com/open-telemetry/build-tools/pull/191

I'd appreciate your thoughts on that! Maybe we could even add these / similar yaml syntax spec as a potential example how it could look like into this proposal?

text/0224-experimental-attrribute-registration.md
dyladan: If you feel your issue is resolved can you please mark it resolved (or if you don't have permission just say "resolved" so I can).

text/0224-experimental-semantic-conventions.md
dyladan: yes that is the idea. if that isn't clear from the text I can update it?

text/0224-experimental-attrribute-registration.md
dyladan: I'm going to add this but I'm going to make it a "suggested" set of fields. If there is some concern during implementation that comes up I don't want the OTEP to disallow some flexibility to solve problems.

dyladan:
After discussion with @jsuereth we have decided to close this OTEP. A lot of what this OTEP was asking for is already implemented. Specifically, the [Attributes Registry](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/attributes-registry) replaces the Attribute Dictionary mechanism described here.

We still believe that the core issue of lowering the barrier to create a new entry in the Attribute Dictionary is important, but believe that this OTEP is no longer the way to move that idea forward.

PR #226, opened 2023-03-02
Non-power-of-two consistent tail probability sampling in TraceState

jmacd:
This is meant to address https://github.com/open-telemetry/opentelemetry-specification/issues/1413.
Follows https://github.com/open-telemetry/oteps/blob/main/text/trace/0170-sampling-probability.md and https://github.com/open-telemetry/oteps/blob/main/text/trace/0168-sampling-propagation.md.

Cc: @oertl @kalyanaj @PeterF778


text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,185 @@
...
+power-of-two sampling probability is used.  P-value also remains the
+specified way to encode zero adjusted count (i.e., p=63).
+
+### T-Value encoding
+
+Since we have 7 bytes, or 56 bits of randomness available, there are
+2^56 non-zero sampling probabilities that can be encoded.  These
+probabilities can be expressed as a 56-bit number in the range [0,
+0xffffffffffffff], where 0 corresponds with sampling 1 span out of
+2^56 and 0xffffffffffffff corresponds with sampling 100% of spans.
```
oertl, line 67: I would prefer to define the t-value as the number of dropped spans out of 2^56. In this way the value 0 would correspond to 100% sampling and 0xffffffffffffffff would correspond to 1 out of 2^56. This would have the benefit that the hex-encoding (as described below) of probabilites which are powers of 2 would be much more efficient. For example, 100% would be "t:0", 50% would be "t:8", "25%" would be "t:c", "12.5%" would be "t:e", 6.25% would be "t:f0", and so on.

Another advantage is that with this definition the t-value increases with decreasing sampling probability, which is similar to the p-value.

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,185 @@
...
+
+The proposal is summarized as follows.  T-value is encoded as a
+hexadecimal string containing between 1 and 14 hex digits.  When the
+T-value is less than 14 hex digits, it is extended to 14 bytes using
+by padding with 0s.  For example, the t-value string "003f"
+corresponds with a the 14-hex-digit string "003f0000000000".
+
+Head samplers and tail samplers alike can be implemented simply by
+tersting whether the least-significant 7 bytes of the TraceID are
+lexicgraphically less-than-or-equal to the sampling threshold.  Note
```
oertl, line 77: would need to be "greater" instead of "less-than-or-equal" if the t-value is defined as I have proposed above

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,185 @@
...
+
+```
+func ThresholdToProbability(thresh string) float64 {
+    parsed, _ := strconv.ParseFloat("0x1."+thresh[:13]+"p+00", 64)
+	return math.Nextafter(parsed, 2) - 1
+}
+```
+
+Note that these transformations are not always reversible, since
+floating point numbers have less precision.  Note that only 13 bytes
```
oertl, line 144: double-precision has effectively 53bits of precision (the first 1bit is implicitly stored + 52bits mantissa), so the transformation could potentially use 13bytes + 1bit

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,185 @@
...
+
+```
+tracestate: ot=t:00da7
+```
+
+corresponds with 0.33333% sampling.
+
+## Trade-offs and mitigations
+
+Note that the t-value encoding is not efficient for encoding
```
oertl, line 178: with the proposed encoding above we could think of replacing the p-value

jmacd:
@oertl Thank you. I accept your suggestions and wonder, would you be interested in drafting blocks of replacement text for your scheme? I'll be glad to work out the pseudocode snippets from there, but your words for "hex/binary threshold equals number of spans dropped out of 2^56", for explaining the 52-bits vs 56 bits issue, and the use of greater-or-equal, I take it, for filtering. I would be glad to then work on writing a spec to deprecate p-value in favor of t-value, where as you say the threshold "0" equals "00000000000000" indicating all spans with 7-bytes >= "00000000000000".

text/trace/0226-sampling-random-traceids.md
PeterF778, line 77: typo: lexicgraphically

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,185 @@
...
+Since this Sampler is expected to behave consistently with or without
+the `random` flag, we assumes the bits are random and do not actually
+check the W3C random flag.
+
+This document propose extending the existing p-value, r-value
+mechanism with support for a new indicator for non-power-of-two
+probability sampling known as "t-value", where "t" is chosen because
+it signifies a threshold.  If widely adopted, the tracestate r-value
+can be deprecated, as it is not needed when randomness is provided in
+the TraceID.
```
PeterF778, line 54: As already discussed before, having the r-value independent from traceId has its benefits, as it allows for the same sampling decisions for groups of related traces which the user may want to keep - as long as the correct probabilistic distribution of the r-values is maintained.

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,185 @@
...
+2^56 non-zero sampling probabilities that can be encoded.  These
+probabilities can be expressed as a 56-bit number in the range [0,
+0xffffffffffffff], where 0 corresponds with sampling 1 span out of
+2^56 and 0xffffffffffffff corresponds with sampling 100% of spans.
+
+The proposal is summarized as follows.  T-value is encoded as a
+hexadecimal string containing between 1 and 14 hex digits.  When the
+T-value is less than 14 hex digits, it is extended to 14 bytes using
+by padding with 0s.  For example, the t-value string "003f"
+corresponds with a the 14-hex-digit string "003f0000000000".
```
PeterF778, line 73: typo: "a the"

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,185 @@
...
+using interpolation.  For example, a neffective sampling probability
+of 1-in-3 can be achieved by alternating between 25% and 50% sampling.
+However, interpolation only works for trace roots, otherwise
+"consistent" sampling can only be achieved at the next smaller power
+of two.  In the example, sampling at 1-in-3 using interpolation means
+traces are only guaranteed **consistent** at 25% and smaller sampling
+probabilities.
+
+The major downside of the r-value, p-value approach is that r-value
+must be encoded even for unsampled contexts.  Ideally, building
```
PeterF778, line 38: It would be nice to get this claim supported by some numbers. Assuming that the concern is about the wire protocol, it is true that passing the r-value will typically require a separate http header, but this should be seen in context of already necessary trace-id and parent span id which have to be there as well (for unsampled contexts). IMHO the overhead added by the r-value is fairly small and is outweighed by the benefits.

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,185 @@
...
+it signifies a threshold.  If widely adopted, the tracestate r-value
+can be deprecated, as it is not needed when randomness is provided in
+the TraceID.
+
+As proposed, t-value and p-value are mutually exclusive; p-value
+remains the preferred encoding for probability sampling when a
+power-of-two sampling probability is used.  P-value also remains the
+specified way to encode zero adjusted count (i.e., p=63).
+
+### T-Value encoding
```
yurishkuro, line 61: I'm confused why we're talking about encoding of t-value. Isn't the whole point of it in not requiring any additional propagated data? We can be talking about interpreting trace-id as t-value (a one way transformation).

oertl: If the trace ID is random, it could be interpreted as r-value, which would then become obsolete. Nevertheless, the information about the actual sampling probability must be propagated, which is either the p-value (if only powers of two are used) or the t-value, which supports arbitrary sampling probabilities.

oertl:
As alternative to the t-value, the p-value definition could be extended. For example, `p:11.a3d` could be interpreted as 8 leading zeros followed by a 1 and the bits defined by the extra hex code "a3d". Remaining bits are filled by ones. In this case, the value would be 
```
00000000000    1  1010      0011      1110       111111111111111111111111111111111111     = 0x001a3dffffffff = 28853590294527 

11 leading        hex "a"   hex "3"   hex "d"    32 one bits  to fill up to 56 bits    
zeros
```

The resulting value corresponds be the number of kept spans out of 2^56 subtracted by 1.

The example above would correspond to a sampling probability of (28853590294527 + 1)/2^56 = 0.000400424

With this definition `p:0`, `p:1`, `p:2`, etc. would correspond to power of two sampling probabilities as already defined.

text/trace/0226-sampling-random-traceids.md
jmacd: The TraceID contains a source of randomness but does not tell us the effective sampling rate that was used to decide to sample it. T-value additional information about how to count the span in a Span-to-Metrics pipeline that is not otherwise available.

jmacd:
> p:11.a3d

By the way, in my [prototype](https://github.com/jmacd/opentelemetry-collector-contrib/pull/2925) I used the syntax `a3d-11` where you've written `11.a3d` (b/c I read this as "a3d shifted 11"). In this draft, to simplify, I proposed to encode the 11 zeros (i.e., `00000000000a3d`), but I agree that encoding the zero count separately is more compact for most probabilities.

As for whether we overload how to parse `p` of not, that's a separate question-- it would break existing uses of p-value. That's why I prefer to create a new variable, but since `p` is marked experimental I'm open to this change.

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,185 @@
...
+## Motivation
+
+The existing, experimental [specification for probability sampling using TraceState](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md)
+supporting Span-to-Metrics pipelines is limited to powers-of-two
+probabilities and is designed to work without making assumptions about
+TraceID randomness.
+
+This proposes to extend that specification with support for 56-bit
+precision sampling probability.  This is seen as particularly
+important for implementation of probabilistic tail samplers (e.g., in
```
kalyanaj, line 12: It may be good to elaborate a bit more on the motivation for this requirement of higher precision sampling probability.

text/trace/0226-sampling-random-traceids.md
kalyanaj: Agree. It will be good to describe the situations where r-value would still be needed. Is session-based sampling (e.g., to have consistent sampling across a group of different traces within a single "session") the only one or are there additional use cases? If this is needed, then would we still have r-value as an optional parameter that overrides the TraceID based mechanism?

kalyanaj:
> fer to create a new variable, but since `p` is marked experimental I'm open to this change

If p-value is not used currently in any implementations (or if they are okay with a breaking change since it is still experimental as you call out), yes it looks like extending p-value to also encode non-powers-of-two sampling probabilities is a good idea by @oertl. 

Reasoning: Since p-value and t-value are conceptually for the same purpose (to encode sampling probabilities), ideally it will be good to just extend the p-value concept - it will make it a bit easier to understand for folks who are already familiar with the purpose of p-value, rather than trying to think of t-value as a new concept.

PeterF778:
With respect to obsoleting the r-value, I reported a [new issue 3307](https://github.com/open-telemetry/opentelemetry-specification/issues/3307) today. If we agree that the issue is real (I'm not entirely sure), the solution I proposed would make consistent probability samplers work differently if there's no r-value.

jmacd:
This was discussed in yesterday's Sampling SIG. Since the initial feedback, I had come to the following rough idea to use all the bits of the TraceID, to consistently decide how to interpolate between powers of two.

Considering an 8-bit example with 75% sampling:

- t-value indicates the trace ID at which drops take place
- 75% sampling corresponds with t=0x40, (i.e., decimal 64) indicating to drop 64 out of 256 traces
- t=0x40 corresponds with adjusted count of 4/3, calculated as `0x100 / (0x100 - 0x40)`
- t=0x40 also can be deterministically mapped to p=1 or p=2 using trace-id randomness to interpolate, in the example there are 192 out of 256 traces being accepted, they just need to be split into p=1 and p=2 to preserve the expected value of adjusted count (using a formula I admit I haven't worked out).

cc @oertl @PeterF778

oertl:
@jmacd 
> This was discussed in yesterday's Sampling SIG. Since the initial feedback, I had come to the following rough idea to use all the bits of the TraceID, to consistently decide how to interpolate between powers of two.

Unfortunately, after rethinking this proposal, I have concluded that it would not allow correct estimation of trace quantities (e.g. estimating the number of traces touching one service A and another service B) as described in [my paper](https://arxiv.org/abs/2107.07703). It would only work for span estimates (e.g. estimating the number of spans of service A). The proposal violates the basic assumption that the choice of the sampling probability is independent of the shared randomness (trace ID or p-value).

PeterF778:
Let's take one step back.

The Consistent Probability Sampling schema already allows consistent head and tail sampling if the sampling probability is a power of 2. Now we want to extend it so that consistency (head and tail) is preserved even for non-power-of-2 probabilities.

It is obvious, I think, that we need to use the random bits of trace-id somehow, if we want to get consistent sampling with any given probability, meaning multiple instances of (head or tail) samplers making the same sampling decisions wrt spans belonging to the same trace. From the past discussions, it looks like we have two categories of possible extensions:

**A**. Allow (approximation of) any non-power-of-2 sampling probabilities in TraceState.

**B**. Continue to restrict the sampling probabilities in TraceState to power-of-2 values.

and we can always have

**C**. Drop Consistent Probability Sampling as it has been proposed and replace it with something else.

Now we could try to summarize the benefits and disadvantages of these approaches.
The way I see it, the benefits of **A** are:
-  with sampling probabilities like 75% or 5%, the metric estimates (span-to-metric pipeline) are a bit better (a smaller expected error) than when switching between power-of-2 probabilities
 - makes some folks happy that they can specify any probability for a single span

while the potential issues are:
 - possible non-integer adjusted counts
 - subsequent sampling steps (head+tail, for example) for the same span force us to multiply probabilities - this puts pressure on providing more precision to represent probabilities, or entails loss of precision
 - need to deal with a potentially large number of sampling probabilities, complicating metric calculation
 - longer strings required to encode probability encoding in TraceState (if we care about such things)
 
Respectively, approach **B** is the reverse of that, with the drawbacks being:
 - slightly larger expected error for metric estimates if the requested sampling rate was not a power of 2

and the advantages are:
 - adjusted counts are full integer numbers
 - sets of spans/traces can be re-sampled multiple times using the same mechanism and still encode the resulting adjusted count accurately
 - small number of possible span probabilities make processing easier

With respect to **C**, it is a very wide open field, but let's keep in mind that for expressing sampling probabilities we do not need high precision, we need wide range of values. That's why the logarithmic scale of r-values and p-values is so powerful and efficient.

If we try to replace that with a linear model based on 56 random bits in trace-id, we quickly run out of bits.

kalyanaj:
Thanks @PeterF778 for consolidating the tradeoffs - overall the summary of the tradeoffs makes sense to me.

I didn't quite understand the below two points - can you please clarify/elaborate?

> sets of spans/traces can be re-sampled multiple times using the same mechanism and still encode the resulting adjusted count accurately

and this final point about running out of bits:

> If we try to replace that with a linear model based on 56 random bits in trace-id, we quickly run out of bits.

jmacd:
03f693c Contains an update based on recent discussions.

jmacd:
@kentquirk new draft: 
df6b1d0

PeterF778:
A very nice and clean proposal!
A few questions though: it says it is extending the r-value and p-value proposal - the p-value will be used to calculate the final adjusted count, but is the r-value used anywhere?
And an old question remains: why tail sampling requires explicit non-power-of-two sampling probabilities as opposed to (deterministic) interpolation between two adjacent power-of-twos? What do we gain or lose with either approach?

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,205 @@
...
+The existing, experimental TraceState probability sampling
+specification relies on two variables known as **r-value** and
+**p-value**.  The r-value carries the source of randomness and the
+p-value carries the effective sampling probability.  The preceding
+specification recommends the use of interpolation to achieve
+non-power-of-two sampling probabilities.
+
+This specification is proposed that aims to offer an alternative to
+that r-value, p-value specification, one that is simpler to implement,
+can be used in both head- and tail-samplers, and that naturally
```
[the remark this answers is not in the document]
kentquirk: That's a significant advantage in terms of usability and the way that real customers think about sampling. My company allows integer sampling values to be specified in a tail-sampling proxy and I have many examples of  sampling configurations using values like 10, 30, 75, 100, 500, 10000. I've literally never seen anyone write a power of 2 greater than 10.

I regularly support customers who then look at the results as expressed in their data and they expect to see those numbers. They're not going to be happy to see 8 and 16 when they expected 10. 

Yes, it's possible to express it this way in aggregate, but it's not obvious and not that easy to reason about.

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,205 @@
...
+```
+
+corresponds with 1-in-3 sampling.
+
+### 25% head sampling, 1-in-10 tail sampling
+
+The following header
+
+```
+tracestate: ot=p:2;t:10
```
kentquirk, line 185: Just for clarity, I'd recommend that we use 3 here and a 1-in-8 / 80 example below, so it's more obvious it's not a mistake.

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,205 @@
...
+## Motivation
+
+The existing, experimental [specification for probability sampling using TraceState](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md)
+supporting Span-to-Metrics pipelines is limited to powers-of-two
+probabilities and is designed to work without making assumptions about 
+TraceID randomness.
+
+Head sampling requires the use of TraceState to propagate context from
+the parent for recording in child spans, in support of Span-to-Metrics
+pipelines.  Tail sampling does not require context propagation
```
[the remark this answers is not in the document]
jmacd: I introduced the term "Intermediate Span Sampling" to describe sampling of individual spans before they arrive at a location where the complete trace is available.

text/trace/0226-sampling-random-traceids.md
jmacd: I now think it would be best to deprecate the p-value, r-value mechanism, so I removed this example that combined them.

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,205 @@
...
+these bits, a simple threshold test is defined to allow sampling based
+on TraceID randomness.
+
+This document proposes extending the p-value, r-value mechanism with
+support for a new indicator for non-power-of-two probability sampling
+known as "t-value", where "t" is chosen because it signifies a
+threshold.  Tail-based sampling encoded by t-value can be combined
+with p-value, in which case the adjusted count implied by t-value is
+**multiplied** with the adjusted count implied by p-value because they
+are independent mechanisms.
```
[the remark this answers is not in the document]
jmacd: Yes. At this point, I'm not seeing a lot of reason to keep the p-value, r-value mechanism if we adopt the t-value mechanism.

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,205 @@
...
+TraceID randomness.
+
+Head sampling requires the use of TraceState to propagate context from
+the parent for recording in child spans, in support of Span-to-Metrics
+pipelines.  Tail sampling does not require context propagation
+support, but it has many similar requirements:
+
+1. Sampling should be "consistent", so that independent collection
+   paths make identical sampling decisions.
+2. Spans should be countable in a Span-to-Metrics pipeline, which
```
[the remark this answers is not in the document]
jmacd: :shrug: I believe in fractional counting :)

text/trace/0226-sampling-random-traceids.md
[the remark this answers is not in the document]
oertl: Agree, there is no reason to support both.

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,205 @@
...
+Whether to encode sampling probabilty or adjusted count is a choice.
+In both cases, the interpreted value translates into an exact
+threshold, which determines the exact inclusion probability.  From the
+exact inclusion probability, we can determine the adjusted count to
+use in a span-to-metrics pipeline.  When the t-value is _stated_ as an
+adjusted count (as opposed to a sampling probabilty), implementations
+can use the integer value in a span-to-metrics pipeline.  Otherwise,
+implementations should use an adjusted count of 1 divided by the
+sampling probability.
+
```
oertl, line 133: This is a minor thing, but perhaps a section describing how to encode powers of two sample probabilities would be helpful. Since I am not 100% familiar with the POSIX-supported printf format, I wonder what would be the most efficient way. For example, if the sampling probability is 2^(-20) (corresponding to p=20), we could write `t=0x1p-20` or `t=1048576`, but would `t=0x1p+20` or even `t=0x1p20` be allowed?

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,214 @@
...
+OpenTelemetry Log Records on the collection path too.  This proposal
+recommends the creation of a new field in the OpenTelemetry Log Record
+with equivalent use and interpretation as the (W3C trace-context)
+TraceState field.  It would be appropriate to name this field
+`LogState`.
+
+## Explanation
+
+This document recommends deprecating the experimental p-value, r-value
+specification.
```
kalyanaj, line 61:
```suggestion
specification. Instead, it proposes a new OpenTelemetry specific tracestate value called t-value. This t-value encodes either the sampling probability (a floating point value) directly or the "adjusted count" of a span (an integer). The letter "t" here is a shorthand for "threshold". The value encoded here can be mapped to a threshold value that a sampler can compare to a value formed using the rightmost 7 bytes of the traceID. Let's look at the details of how this threshold can be calculated.
```

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,214 @@
+# Non-power-of-two Probability Sampling using 56 random TraceID bits
+
+## Motivation
+
+The existing, experimental [specification for probability sampling using TraceState](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md)
```
kalyanaj, line 5: There is lot of good information in the existing experimental spec (e.g., on what is consistent probability sampling, why it is needed, what does "adjusted count" mean, what does "0" adjusted count mean etc.). Do you plan to bring forward some of that content over here? Want to make sure we can still preserve/standardize the parts of content (that are still applicable even with this proposal) even if/when that spec gets deprecated.

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,214 @@
...
+sampling, the t-value encoding scheme could also be used to convey
+Logs sampling.  This document proposes to add an optional `LogState`
+string to the OTLP LogRecord, defined identically to the W3C
+tracecontext `TraceState` field.
+
+## Re-sampling with t-value
+
+It is possible to re-sample spans that have already been sampled,
+according to their t-value.  This allows a processor to further reduce
+the volume of data it is sending by lowering the sampling threshold.
```
kalyanaj, line 151: I want to understand the expected user behavior here. Let's take this example:

- A head sampler is used with a sampling rate of say 10%, 
- Let's say an intermediate sampler is also specified with a sampling rate of say 5%.

Let's say 1000 spans are emitted. Let's say ~100 spans are sampled by the head sampler (based on the above 10% rate). Now when they come to the intermediate sampler, isn't the user expectation that only ~5 out of these ~100 spans are sampled (and the rest 95 discarded)? If my understanding is correct, I was thinking that the t-value would need to be updated for those non-discarded spans, however the below proposal seems to be based on a different user behavior.

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,214 @@
...
+count is meant to give the user control over loss of precision.  At
+the same time, it can be read by humans.
+
+Floating point numbers can be encoded exactly to avoid ambiguity, for
+example, using hexadecimal floating point representation.  Likewise,
+adjusted counts can be encoded exactly as integers to convey the
+user's intended sampling probability without floating point conversion
+loss.
+
+## Prior art and alternatives
```
kalyanaj, line 201: Towards the end, we may want to call out that one benefit of the r-value based randomness was that it could be used to get consistent sampling across multiple traces (e.g., all traces started within a time window by a participant) - it would be good to call out that it should be possible to support it in the future as a complement to the current proposal.

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,214 @@
...
+  already been sampled with a less-than-or-equal probability compared
+  with the current sampler, so for consistency the span simply passes
+  through.
+- If the Threshold calculated from the incoming t-value is larger than
+  the current sampler's Threshold, the current sampler's Threshold is
+  re-applied; if the TraceID random value is less than the current
+  Sampler's threshold, the span passes through with the current
+  sampler's t-value, otherwise the span is discarded.
+
+## Examples
```
kalyanaj, line 169: It would be good to add two more examples that shows how consistent probability sampling can be achieved across multiple participants.

Example 1: 
- Upstream participant samples at 10% probability (ot=t:0.1  is sent as part of tracestate)
- Downstream participant does parent-based sampling. It uses the sampled flag to make the decision, gets the t-value from the parent context and emits it as part of its context (ot=t:0.1 is sent as part of tracestate to further downstream participants)

Example 2: 
- Upstream participant samples at 10% probability (ot=t:0.1 is sent as part of tracestate)
- Downstream participant samples at 5% probability - it calculates a threshold based on its sampling rate and compares with the traceID last 7 bytes to make the sampling decision (ot=t:20 is sent as part of tracestate).
- Downstream participant does parent-based sampling (uses the sampled flag to make the decision, gets the t-value from the parent context and emits it as part of its context)

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,214 @@
...
+given a sampling probability encoded as "0.1", we first compute the
+nearest base-2 floating point, which is exactly 0x1.999999999999ap-04,
+which is approximately 0.10000000000000000555.  The exact quantity in
+this example, 0x1.999999999999ap-04, is multipled by `0x1p+56` and
+rounded to an unsigned integer (7205759403792794).  This specification
+says that to carry out sampling probability "0.1", we should keep
+Traces whose least-significant 56 bits form an unsigned value less
+than 7205759403792794.
+
+## T-value encoding for adjusted counts
```
kalyanaj, line 104: It will be good to define the mutation rules and propagation rules for t-value. E.g., something on the lines of:

- if a participant is doing parent-based sampling, it should propagate the t-value from its parent.
- if a participant is doing consistent probability sampling using its own sampling rate, it should mutate the t-value to set the new adjusted count / sampling rate.

text/trace/0226-sampling-random-traceids.md
oertl: If we decide to use arbitrary sampling probabilities, we should not use the current definition of the r-value. It makes no sense to have different discretizations for the r-value (powers of two) and for the t-value (56-bit values). Therefore, the r-value should rather be a 14-digit hex value that overrides the random bits of the trace ID, if present. This way we could also handle traces where the random flag is not set in the trace context. If the flag is not set and there is also no r-value, we could require consistent samplers to set the r-value by generating a 56-bit random value.

PeterF778:
I would like to see how to provide algorithms for tail sampling based on the t-value, for the following use cases:
1. (static sampling) Given a set of N1 spans, I want to re-sample it to arrive at approximately N2 spans.
2. (dynamic sampling) Given a stream of incoming spans, I want to re-sample it so I get at most X spans/s (approximately).

The potential challenge I see is that the algorithms may require sorting the spans according to their t-value.

PeterF778:
I'm not sure I understood your algorithm for static sampling correctly, @oertl. Correct me, if I'm wrong, but it looks like the selection of spans to survive is affected heavily by their t-value, rather than the source of randomness (trace-id), which introduces a bias.
Suppose that the original set of N1 spans contains a subset S of spans that have been already heavily sampled, so their t-values are very low - much lower than the spans not belonging to set S. In that case, if the size of set S is smaller than N2, all elements of set S will survive the re-sampling process. IMHO, an unbiased algorithm should sample spans from set S roughly with the same chances of survival as those from outside set S.

oertl:
> IMHO, an unbiased algorithm should sample spans from set S roughly with the same chances of survival as those from outside set S.

It depends on what is meant by bias. My understanding is the following: The expected adjusted count is equal to 1. This is ensured by setting the adjusted count to the inverse of the sampling probability. This should be the case with the algorithm described above.

@PeterF778, I think your concern is that the algorithm balances the sampling probabilities of the previous sampling stages. Depending on what you want to estimate, this sampling strategy may or may not be beneficial. I haven't checked, but I think the algorithm is similar to VarOpt sampling (see [here](https://arxiv.org/pdf/0803.0473.pdf)), which minimizes variance when estimating arbitrary subset sums. This algorithm makes sense if you have a sampling stage that combines samples collected in earlier sampling stages with different probabilities for no good reason, e.g., due to unbalanced load or short-term load fluctuations. 

The situation is different if you have already identified certain classes of spans (e.g., spans with errors) that should be sampled with a higher probability. In this case, you want to sample these spans more frequently than others by purpose. I think the correct term for this is stratified sampling, where weights are defined for different classes of spans and the sampling algorithm tries to sample them so that the ratios of the weights are reflected by the corresponding ratios of the t-values. To complicate matters further, different stages of sampling would define different weights for the same span. Early stages may not be able to assess the importance of a span while later sampling stages have a more holistic view of the trace and therefore might assign a different weight to a span. For example, if the child span has an error.

Anyway, discussion of these sampling strategies takes us a bit off topic. The same issues must be solved when using power-of-two sampling probabilities.

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,214 @@
...
+arrive at their destination in real time, meaning before assembling
+complete traces.
+
+Here, the term Intermediate Span Sampling is used to describe sampling
+performed on individual spans at any point in their collection path.
+Like Head sampling, Intermediate Span Sampling benefits from being
+consistent, because it makes recovery of complete traces possible
+after spans have independently sampled.  On the other hand, when "Tail
+sampling" refers to sampling of complete traces, sampling consistency
+is not an important property.
```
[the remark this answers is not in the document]
kentquirk: I think the point is that when "tail sampling" means "making the decision when you have the entire trace available", you're by definition making a self-consistent decision since there's only one instance making the decision. You either get all the spans or none of them. It doesn't really matter very much in this context whether a different instance would have come to the same decision.

It's also the case that we are sometimes talking about non-probabalistic decisions like "is there an exception span event in this trace?"

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,214 @@
...
+
+Intermediate Span Sampling is exemplified by the
+[OpenTelemetry-Collector-Contrib's `probabilisticsampler`
+processor](https://pkg.go.dev/github.com/open-telemetry/opentelemetry-collector-contrib/processor/probabilisticsamplerprocessor).
+This proposal is motivated by wanting to compute Span-to-Metrics from
+spans that have been sampled by such a processor.
```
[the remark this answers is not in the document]
kentquirk: If you are doing purely probabalistic sampling at both head and tail, based on TraceID, you should update the adjusted count to the minimum of the two probabilities.

If you are doing probabalistic head sampling and some sort of non-traceID-based sampling in the tail (keep 10% of spans with http 400 errors) then you need to multiply the probabilities.

We should probably note this as well.

PeterF778:
> 
> @PeterF778, I think your concern is that the algorithm balances the sampling probabilities of the previous sampling stages. Depending on what you want to estimate, this sampling strategy may or may not be beneficial. 

Yes, I imagine "static sampling" to be applied to aging data, after one or many stratified sampling steps were performed .

> 
> Anyway, discussion of these sampling strategies takes us a bit off topic. The same issues must be solved when using power-of-two sampling probabilities.

If we want to compare the two competing consistent probability sampling mechanisms, we have to understand what they will entail in all processing stages. I believe the re-sampling algorithms will be very similar in principle, but there could be differences in complexity or accuracy of the results.

oertl:
> If we want to compare the two competing consistent probability sampling mechanisms, we have to understand what they will entail in all processing stages. I believe the re-sampling algorithms will be very similar in principle, but there could be differences in complexity or accuracy of the results.

Yes, this is true. For example, for "dynamic sampling" it is probably much simpler to estimate quantiles if there is a power-of-two discretization as you could simply aggregate into a histogram as there are only a small number of relevant values. However, any sampling stage is free to use just power-of-two sampling probabilities, if it is more efficient. Even, if spans come with non-power-of-two t-values, they could be easily downsampled to the next power of two in a first step.

text/trace/0226-sampling-random-traceids.md
oertl: The reason I ask is that powers of two sampling probabilities are a natural discretization for me, since this is the only discretization that results in integer adjusted counts while the relative spacing is constant. Thus, I believe we will often see t-values that are powers of two. Therefore, it might be useful to define a more compact representation of the t-value if it is a power of two. Possibly it makes sense to keep the p-value?

text/trace/0226-sampling-random-traceids.md
jmacd: Not quite answering your question, but I've prototyped https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/22058 with a different sort of answer to your question. 

In this case referring to span data records, where there are multiple collectors in a pipeline. The first collector may sample at 1/10; when a subsequent collector samples at 1/20, the t-value of the selected spans will be updated. If the subsequent collector samples at 1/2, however, it is being less selective than the first collector, so it should not modify the t-value. That is to say that t-value adjusted counts should not fall and t-valued probabilities should not rise.

See the logic here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/22058/files#diff-33f10350e2875f926dd2be6fc4c6bb88cfd8043cf6ac6d100295cf654771d90dR210-R219

text/trace/0226-sampling-random-traceids.md
jmacd: These examples sound good to me! Will do.

text/trace/0226-sampling-random-traceids.md
PeterF778: I think there's a problem with such sampling behavior. Let's assume that the previous collector in chain sampled all traces with errors with probability 1, and all remaining  traces with 1/100. If the next collector in chain is configured with 1/10, it will not touch the healthy traces, but will decimate the traces with errors. So any stratified sampling logic must be known and repeated by all collectors in the pipeline. Even if we prohibit stratified sampling, to set up a collector sampling probability in any meaningful way we have to know the minimum sampling probability of all the preceding collectors.

jmacd:
New developments discussed in the Sampling SIG today.

We proposed "s-value" as a mechanism to encode the accumulation of independent non-consistent sampling stage adjusted counts. t-value and s-value would be separate fields, both included in tracestate for consistency.

text/trace/0226-sampling-random-traceids.md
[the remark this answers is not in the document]
jmacd: Fixed, anyway.

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,214 @@
...
+This proposes to create a specification with support for 56-bit
+precision consistent Head and Intermediate Span sampling.  Because
+this proposal is also meant for use with Head sampling, a new member
+of the OpenTelemetry TraceState field will be defined.  Intermediate
+Span Samplers will modify the TraceState field of spans they sample.
+
+Note also there is an interest in probabilistic sampling of
+OpenTelemetry Log Records on the collection path too.  This proposal
+recommends the creation of a new field in the OpenTelemetry Log Record
+with equivalent use and interpretation as the (W3C trace-context)
```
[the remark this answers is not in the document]
jmacd: OK, probably should be reworded. But I do think that if you were not recording spans and you were recording logs, you'd record the TraceState in your log record. At that point, maybe it should still be called TraceState.

text/trace/0226-sampling-random-traceids.md
jmacd: Applied.

text/trace/0226-sampling-random-traceids.md
jmacd: I was planning to start with the existing specification and modify it (i.e., https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md) so it should start with more background. I appreciate your pointing out that this information is lost in this OTEP.

text/trace/0226-sampling-random-traceids.md
jmacd: By my understanding, the head sampler with a rate of 10% will export 100 spans (of 1000), and then the 5% sampler will export 50 of those spans if it's a consistent sampler.

My latest update to this OTEP includes an "s-value" discussed in last week's SIG which would allow non-consistent independent sampling to be reported, so the 5% simple probability sampler could attach `s:20` or `s:0.05`.

jmacd:
I've updated the document making "r-value" optional (and 56-bits) and making "s-value" for independent non-consistent sampling. I've incorporated some helpful feedback from @kalyanaj. Thanks.

text/trace/0226-sampling-random-traceids.md
PeterF778: For the record, I agree with @kalyanaj that 5% sampling rate for intermediate or tail sampling should mean that the volume of spans after sampling is about 20 times smaller than the input volume. Otherwise, assuming possible different sampling rates for the spans from the input stream, the user cannot really predict what the effect of 5% sampling would be.

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,295 @@
...
+TraceState field.  It would be appropriate to name this field
+`LogState`.
+
+This proposal does makes r-value an optional 56-bit number as opposed
```
oertl, line 78: typo: This proposal makes...

text/trace/0226-sampling-random-traceids.md
```diff
@@ -0,0 +1,295 @@
...
+
+For head samplers, there is an opportunity to synthesize a new r-value
+when the tracecontext does not set the `random` bit (as the existing
+specification recommends synthesizing r-values for head samplers
+whenever there is none).  However, this opportunity is not available
+to tail samplers.
+
+To calculate the Sampling threshold, we began with an IEEE-754
+standard double-precision floating point number.  With 52-bits of
+significand and a floating exponent, the probability value used to
```
oertl, line 137: > With 52-bits of significand...

 Double-precision floating-point values have a 52-bit mantissa but are able to rerpesent 53-bit significands (except for subnormal values). See https://cs.stackexchange.com/a/152267/102560.

text/trace/0226-sampling-random-traceids.md
```diff
@@ -166,26 +224,49 @@ threshold and compared against the new threshold.  These are two cases:
...
+### 30% simple probability sampling
+
+The tracestate value `ot=s:0.3` corresponds with 30% sampling by one
+or more sampling stages.  This would be the tracestate recorded by
+`probabilisticsampler` when using a `HashSeed` configuration instead
+of the consistent approach.
+
+### 10% probability sampling twice
+
+The tracestate value `ot=s:0.01` corresponds with 10% sampling by one
```
kentquirk, line 267: Maybe expand this to show how the tracestate would be modified at each stage?

kentquirk:
@jmacd , @kalyanaj , @oertl , @PeterF778 

As promised in the SIG meeting, I have [implemented a proposal](https://github.com/kentquirk/tvalue) for how to calculate threshold, sampling rate, and sampling probability in several different languages (Go, JavaScript, and Python). 

The proposal is simple -- here's the Go version:

```go
func threshold(tValue float64) int64 {
	const k = 0x1p+56
	if tValue < 1.0 {
		return int64(k*tValue + 0.5)
	}
	return int64(k/tValue + 0.5)
}

func samplingRate(tValue float64) int64 {
	if tValue < 1.0 {
		return int64(1.0/tValue + 0.5)
	}
	return int64(tValue + 0.5)
}

func samplingProbability(tValue float64) float64 {
	if tValue < 1.0 {
		return tValue
	}
	return 1.0 / tValue
}
```

All 3 languages seem to deliver identical results in the first 2 functions.

I was unable to convince all 3 languages to format floating point values identically (I probably could have by installing various libraries but I wanted to use the standard libraries). But they seemed to be identical through the first 14 digits.

Thoughts?

kentquirk:
I understand your point, but I'm personally much more concerned about human usability. What I see from users in real sampling configurations are sampling rates like 1 in 10, 1 in 6, and 1 in 1000.

| rate | hex value < 1 |
|----|------|
| 10 | `0x1.999999999999ap-4` |
| 6 | `0x1.5555555555555p-03` |
| 1000 | `0x1.0624dd2f1a9fcp-10` |

All of those representations lose the user's intent, while `10`, `6`, and `1000` do not.

There is also the problem that as far as I have found, many languages do not support hex representation in floating point. Or if they do, they do so differently.

I strongly feel that we should bias toward comprehensible and easily implementable human representations.

oertl:
> I understand your point, but I'm personally much more concerned about human usability. What I see from users in real sampling configurations are sampling rates like 1 in 10, 1 in 6, and 1 in 1000.

I also understand your point. But the limited number of bits simply does not allow to sample exactly 1 out of 3 consistently. In my opinion, it is more important to forward the actual sampling probability applied, not the user's intent. If you know that all your sampling configurations are of kind "1 out of X", it is relatively easy to reconstruct X from the applied sampling probability to make the user believe that it really was sampled as originally intended. Maybe an additional flag indicating that the reported t-value comes from a "1 out of X" rule could be a compromise.

In large systems, sampling probabilities are typically automatically chosen (e.g. based on rate limits), and it is more important that the parsing/encoding overhead is small. 

> There is also the problem that as far as I have found, many languages do not support hex representation in floating point. Or if they do, they do so differently.

It is easy to specify a lossless hex representation for the t-value. It could be defined as the integer threshold value used when comparing with the 56 random bits of the trace ID (or the optional r-value). If the t-value is an integer, it is straightforward to find a platform/language-independent hex representation. This definition would also reduce floating point operations as the sampling decision is simply the result of comparing the t-value with the random bits.

jmacd:
I have drafted the [introductory text of the changes I would propose in the tracestate-probability-sampling.md spec](https://github.com/open-telemetry/opentelemetry-specification/compare/main...jmacd:opentelemetry-specification:jmacd/tvalue).

jmacd:
This OTEP has been well-replaced by #235. Thanks @kentquirk!

PR #227, opened 2023-03-20
Separate semantic conventions from the specification.

jsuereth:
This should enable us to move quicker and provide better fine-grained ownership of semantic conventions going forward.


text/????-separate-semantic-conventions.md
```diff
@@ -0,0 +1,115 @@
...
+    `https://opentelemetry.io/semantic_conventions/schemas/`
+
+Existing semantic conventions in the specification would be marked as
+deprecated, with documentation denoting the move.
+
+There will also be the following exceptions in the specification:
+
+- Semantic convetions used to implement API/SDK details will be fully specified
+  and will not be allowed to change in the Semantic Convention directory.
+  - Error/Exception handling will remain in the specification
```
Oberon00, line 41: Please provide a justification in the OTEP.

text/????-separate-semantic-conventions.md
```diff
@@ -0,0 +1,115 @@
...
+## Motivation
+
+We need to allow semantic conventions to evolve independently of the overall
+OpenTelemetry specification. Today, any breaking change in a semantic convention
+would require bumping the version number of the entirety of the OpenTelemetry
+specification.
+
+## Explanation
+
+A new github repository called `semantic-conventions` would be created in the
```
Oberon00, line 15: Since you stressed recently that some compatibility requirements for instrumentations are not semantic conventions: How would you deal with these? https://github.com/open-telemetry/opentelemetry-specification/pull/3212#issuecomment-1453934136

IMHO they belong more in this new repository than the spec, but maybe the name is then not right?

My own opinion is that semantic conventions may encompass a broader scope (all of which is currently in the spec): The attribute "data dictionary", conventions on when to use which attributes group, how to name particular spans, how to model particular scenarios in span structures, and in some cases also some "How To" / "Cookbook" / "Best-practice" instructions on how to implement particular (cross-language) instrumentations that go a bit beyond that.

text/????-separate-semantic-conventions.md
```diff
@@ -0,0 +1,115 @@
...
+This would have the following structure:
+
+- Boilerplate files, e.g. `README.md`, `LICENSE`, `CODEOWNERS`, `CONTRIBUTING.md`
+- `Makefile` that allows automatic generation of documentation from model.
+- `model/` The set of YAML files that exist in
+  `{specification}/semantic_conventions` today.
+- `docs/` A new directory that contains human readable documentation for how to
+  create instrumentation compliant with semantic conventions.
+  - `resource/` - Contents of `{specification}/resource/semantic_conventions`
+  - `trace/` - Contents of `{specification}/trace/semantic_conventions`
```
tedsuo, line 27: We have already discussed abandoning the trace/metric/log folder structure, and replacing it with a folder per domain. This structure would keep all of the relevant information grouped together for anyone attempting to instrument their code.

Oberon00: I think there is a merit to keeping the scope smaller here as well.

text/????-separate-semantic-conventions.md
```diff
@@ -0,0 +1,115 @@
...
+    now be split into two (hopefully smaller) efforts.
+  - We expect changes from Semantic Conventions and Specification to be
+    orthogonal, so this should not add significant wall-clock time.
+
+
+## Prior art and alternatives
+
+When we evaluate equivalent communities and efforts we see the following:
+
+- `OpenCensus` - had specification and semantics merged. However, OpenCensus
```
yurishkuro, line 90: OpenTracing also had SemConv as part of its Spec (and both existed before OC, so if you're citing prior art...)

text/????-separate-semantic-conventions.md
```diff
@@ -0,0 +1,115 @@
...
+- The specification may elevate some semantic conventions as necessary for
+  compatibility requirements, e.g. `service.instance.id` and 
+  [Prometheus Compatibility](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md).
+
+## Internal details
+
+The following process would be used to ensure semantic conventions are
+seamlessly moved to their new location.
+
+- A new repository `opentelemetry/semantic_conventions` will be constructed with
```
yurishkuro, line 53: I suggest making a proposal for the approvals structure as part of this PR, since you can't setup a repo without approvers, and I see cleaning up of approvers as one of the main benefits of this OTEP.

text/????-separate-semantic-conventions.md
```diff
@@ -0,0 +1,115 @@
...
+specification.
+
+## Explanation
+
+A new github repository called `semantic-conventions` would be created in the
+OpenTelemetry organization.
+
+This would have the following structure:
+
+- Boilerplate files, e.g. `README.md`, `LICENSE`, `CODEOWNERS`, `CONTRIBUTING.md`
```
pyohannes, line 20: Will this repository also include:
* [Definitions of stability for semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#semantic-conventions-stability)
* [Attribute requirement levels](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-requirement-level.md)

yurishkuro: +1

text/????-separate-semantic-conventions.md
Oberon00, line 1: In the filename `????` should be replacec by `227` (the PR#)

text/????-separate-semantic-conventions.md
jsuereth: Great call out.

SchemaURL is defined in the specification and interacts with key aspects of the specification, like Resource merge. I am not planning to "rip" that out.  Given the intertangled nature of SchemaURL, It's not clear to me yet HOW to tease those apart.

I'm adding this to open questions, as I think we can figure out the details as we do the merge.   My current thinking is that we need to check specification dependencies on these sections and we move everything that doesn't have a direct specification dependency.

Does this sound reasonable?

text/????-separate-semantic-conventions.md
jsuereth: Let me add that.  If you have other examples, I had trouble googling for them.

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,158 @@
+# Separate Semantic Conventions
+
+Move Semantic Conventions outside of the main Specificaiton and version them
```
djaglowski, line 3:
```suggestion
Move Semantic Conventions outside of the main Specification and version them
```

text/????-separate-semantic-conventions.md
jsuereth: Added

text/????-separate-semantic-conventions.md
jsuereth: Added callouts to this as future direction.

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,155 @@
...
+
+As such, the Specification should define the absolute minimum of reserved or
+required attribute names and their interaction with the SDK.
+
+## Internal details
+
+The following process would be used to ensure semantic conventions are
+seamlessly moved to their new location.
+
+- A new repository `opentelemetry/semantic_conventions` will be constructed with
```
arminru, line 66:
```suggestion
- A new repository `open-telemetry/semantic-conventions` will be constructed with
```

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,155 @@
...
+- All semantic conventions in the Specification will be copied to the new
+  repository.
+- Semantic conventions in the Specification will be marked as deprecated with
+  links to the new location.
+- The Specification will be updated to require "special" conventions, like
+  `service.name` and configuration interaction.
+- Instrumentation authors will update their `SchemaURL` to
+   `https://opentelemetry.io/semantic_conventions/schemas/{semconv_version}`
+   from
+   `https://opentelemetry.io/schemas/{spec_version}`
```
arminru, line 79: Is this effort/pain worth it or should we just keep the original URL form?

arminru: Or would this only apply to users updating to new versions post this move?

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,155 @@
...
+
```
arminru, line 80: Currently missing: Which semantic conventions version will we start with in the new repo?
`v0.0.1`, `v1.0.0`, `v1.21.0` (the spec is currently at `v1.20.0`), ...?

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,155 @@
...
+- Semantic Conventions can decide to bump major version numbers to accommodate
+  new signals or hard-to-resolve new domains without breaking the Specification
+  version number.
+- Semantic Conventions can have dedicated maintainers and approvers.
+- Semantic Conventions can restructure to better enable subject matter experts
+  (SMEs) to have approver/CODEOWNER status on relevant directories.
+
+There is a desire to move semantic conventions to domain-specific directories
+instead of signal-specific. This can occur after a separation of the repository.
```
arminru, line 145:
```suggestion
instead of signal-specific. This can occur after a separation of the repository
and will be proposed and discussed separately from this OTEP.
```

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,155 @@
...
+## Trade-offs and mitigations
+
+This proposal has a few drawbacks:
+
+- The semantic conventions will no longer be easily referencable form the specification.
+  - This is actually a benefit. We can ensure isolation of convention from
+    specification and require the Specification to use firm language for
+    attributes it requires, like `service.name`.
+  - We will provide x-links from existing location to the new location.
+- Semantic Convention version will no longer match the specification version.
```
Aneurysm9, line 90: Will this be a new version starting at `v0.0.0` or will it carry on from the version at which the spec and conventions diverged?  The Go API and Collector both use the schema version in package identifiers for semantic convention constants.

text/0227-separate-semantic-conventions.md
jsuereth: This is because there will be a version split from the specification.  So the pain of moving is ONLY a  onetime thing for existing users, but also the new repository can start at a non-1.0 version number to denote the current instability.

text/0227-separate-semantic-conventions.md
jsuereth: Good callout, I can add this. I was planning to drop back to a `v0.0.1` or `v0.1` while we ramp-up to HTTP stability.

text/0227-separate-semantic-conventions.md
jsuereth: This will be a new version number.

I'd suggest you'll want a new package name to avoid breakage as well.  This is why we plan to move the entire SchemaURL.  Effectively this will look/feel like a "new" thing to existing users to migrate to, hopefully alleviating breakages.

text/0227-separate-semantic-conventions.md
arminru: Alright if we go with this approach, we would need to use a new schema URL format anyway (https://github.com/open-telemetry/oteps/pull/227#discussion_r1143577641) or we'd run into conflicts at some point.

text/0227-separate-semantic-conventions.md
arminru: > the new repository can start at a non-1.0 version number to denote the current instability

If we intend to indicate that using the single global version number, we'd never be able to go semconv `v1.0` though as there will always be some (new) unstable convention. (A subset of) HTTP that's considered stable would still be part of the `0.x` conventions. Maybe we should just go with `1.0.0` right away and (based on your "what stability actually enforces" proposal and other wording) define how to interpret versioning in the new repo. I assume we're keeping the stable/unstable markers for documents and sections to distinguish stability within the overall semantic-conventions repo.

Oberon00: Maybe we should not use semver for the repository as a whole, but version conventions, or sets thereof, independently?

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,155 @@
...
+The following process would be used to ensure semantic conventions are
+seamlessly moved to their new location.
+
+- A new repository `opentelemetry/semantic_conventions` will be constructed with
+  the proposed format and necessary `Makefile` / tooling.
+- A moratorium will be placed on Semantic Convention PRs to the specififcation
+  repository. (Caveat that PRs related to this proposal would be allowed).
+- All semantic conventions in the Specification will be copied to the new
+  repository.
+- Semantic conventions in the Specification will be marked as deprecated with
```
reyang, line 72: I suggest that we remove the semantic convention docs from the spec repo (for short-term, maybe just delete the contents and put a banner saying "things moved, here is the new location"), mainly due to the wording "deprecated" could introduce confusion.

jsuereth: Maybe we do that eventually, but we need to deal with two realities:

1. We have in-flight data against the current SchemaURL of this repo.
2. There will be a transition period.

dyladan:
Given our discussion in the SIG meeting today, maybe we should consider instead of a Semantic Conventions repository, an _Instrumentation_ repository which contains semantic conventions as well as guidelines for instrumentation authors. This would create a single place where instrumentation authors can look to make sure they're following all recommendations without bloating the spec with non-normative guidance for every cross-language instrumentation.

---

I also don't see addressed here:

1. ECS is not signal-specific but our semantic conventions are. This might be a good time to bridge this gap by making a dictionary of signal-agnostic attribute definitions which is referenced by signal-specific semantic conventions for their use.
2. Will the tooling be moved from the build-utils repository into this one? Since the tooling is semconv-specific it might make sense for it to be maintained by the semconv authors and live in the same place.

jsuereth:
@dyladan  A few responses:

> ECS is not signal-specific but our semantic conventions are. This might be a good time to bridge this gap by making a dictionary of signal-agnostic attribute definitions which is referenced by signal-specific semantic conventions for their use.

This is somewhat called out in future work.  I think directionally we're moving our notion of horizontal/vertical to be more use-case based (e.g. HTTP, networking, etc.) conventions vs. signal specific.  HOWEVER, for initial transitiion (to preserve git history) we would move that way as a follow on, not the initial move.

> Will the tooling be moved from the build-utils repository into this one? Since the tooling is semconv-specific it might make sense for it to be maintained by the semconv authors and live in the same place.

I can directly address this.  It's related to a concern @Aneurysm9 raised regarding semconv exports today.

The new directory will come with automated tooling and build-tools will be updated to work with this new directory.   There are some details I'd consider OUTSIDE the OTEP (but MUST be figured out as part of the transition) for how codegen should work on a language specific basis.

If you want to see that in this OTEP, then I can move it back to draft form and meet with each language SiG to figure out details for non-breaking migration of this codegen.

## On instrumentation vs. semantic conventions

I just don't think we want to expand scope of semantic conventions that far right now.  That' worth a discussion of its own, but well outside the scope of this PR.   I'm personally against expanding the scope of semantic conventions, but I'd be ok if we have supplementary & non-normative aides for instrumentation authors if we feel it's necessary.

Additionally, if we call the repository "Instrumentation" people may actually look for implementation, when in fact this is just a specification of sorts.

text/0227-separate-semantic-conventions.md
reyang: > 1. We have in-flight data against the current SchemaURL of this repo.

I assume permanent links are used?

> 2. There will be a transition period.

100%

text/0227-separate-semantic-conventions.md
jsuereth: Exactly.  By using a different URL we also split into a new "schema family" so compliance with this new repo is not the same as compliance with the current spec "schema family".   This gives us a chance to migrate and understand the progress of that migration.

text/0227-separate-semantic-conventions.md
jsuereth: Yes, we should use permanent links.

jsuereth:
> I can accept most of that.
> 
> > ## On instrumentation vs. semantic conventions
> > I just don't think we want to expand scope of semantic conventions that far right now. That' worth a discussion of its own, but well outside the scope of this PR. I'm personally against expanding the scope of semantic conventions, but I'd be ok if we have supplementary & non-normative aides for instrumentation authors if we feel it's necessary.
> 
> I want to push back a little bit on this point. Mostly I want to ask: if not here, where? We all agreed today that there should be a repository for this type of knowledge, but nobody seems to know where it should go. The options I can see are this repo, the specification repo, the website, or a new repo.
> 
> Between this repo and the spec repo, I think this one would be a much more obvious fit. The arguments not to include that guidance here applies to an even greater degree in the spec repo.



I think perhaps the nuance in what I'm saying didn't make it through.

- I'm against *normative* instructions for instrumentation in the semantic convention repository.  That's outside the scope.
- I'm *ok* with non-normative guidelines and aides for how to instrument.  In this, I think we're in alignment.

> I think putting it in a new repo is off the table for most people, but I'll document the reasons here anyway. A new repo I think further splits the community and increases the number of places where instrumentation authors need to look in order to write good effective instrumentation. Especially after this OTEP, they would need to look here, in the spec, and in whatever guidance repo.

+1.  I agree that would make it harder.

> > Additionally, if we call the repository "Instrumentation" people may actually look for implementation, when in fact this is just a specification of sorts.
> 
> I'm open to a different name, but calling it "Semantic Conventions" seems to exclude the idea of instrumentation guidance in the future.

I don't really want to bikeshed on a new name.  I think Semantic Conventions fits, and it's ok to have (non-normative) guidance on instrumentation in this repository.

Where I think we need more discussion is on where different non-normative "guidelines" belong overall.  For example, Metrics provides a lot of guidelines around how best to name metrics, units and dealing with fun scenarios like time-gaps, delta to cumulative conversions, etc.   Which of these belongs in the specification and which belong in semantic conventions?

My rule of thumb proposal is "Guidance on how best to implement instrumentation that abides by semantic conventions" can be in non-normative locations in the semantic convention repo.   All other guidelines and aides would remain in the specification.  WDYT?

text/0227-separate-semantic-conventions.md
Oberon00: > an entire InstrumentationScope (Meter, Tracer, Logger) MUST be the same schema version family

This is problematic anyways, at least for tracing, because e.g. an exception event might be created by a different instrumentation using a different semconv version than the main span.

Oberon00:
> I'm against normative instructions for instrumentation in the semantic convention repository. That's outside the scope.

Does this mean, they should be non-normative, or they need a different place, or they are entirely out of scope for OTel?

text/0227-separate-semantic-conventions.md
Aneurysm9: Changing the schema URL isn't an issue as there is a [constant](https://github.com/open-telemetry/opentelemetry-go/blob/2e54fbb3fede5b54f316b3a08eab236febd854e0/semconv/v1.18.0/schema.go#L20) for the URL in each semconv package.  The [package name](https://github.com/open-telemetry/opentelemetry-go/blob/v1.14.0/semconv/v1.18.0/doc.go#L20) we're using is `semconv` but the import paths include both `semconv` and the [corresponding spec versions](https://pkg.go.dev/go.opentelemetry.io/otel@v1.14.0/semconv).

What are other SIGs doing to keep versioned semconv constants available for use and how will this impact them?

text/0227-separate-semantic-conventions.md
lmolkova: How would we approach HTTP stability wrt versions? Sounds like we'll need to split schema versions almost right away. 

Perhaps we can solve it with two versions:
- the default repo-level version set to 0.X applied to all unstable conventions
- the 'stable' version (1.X) applied to all stable conventions

jsuereth: @lmolkova How would that work in relation to SchemaURL concerns I raised above?  I.e. is it technically feasible for us to have two versions split in one repo?

tigrannajaryan: >This is because there will be a version split from the specification.

Why this is necessary? Why can't we continue version numbers 1.20+ in the new repo?

text/0227-separate-semantic-conventions.md
tigrannajaryan: >Good callout, I can add this. I was planning to drop back to a v0.0.1 or v0.1 while we ramp-up to HTTP stability.

I don't understand why this decision. What is wrong with continuing the current version numbers? The Schema version numbers starting with 1.x don't imply any special stability that forces you to use 0.x to denote instability.

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,156 @@
...
+## Trade-offs and mitigations
+
+This proposal has a few drawbacks:
+
+- The semantic conventions will no longer be easily referencable form the specification.
+  - This is actually a benefit. We can ensure isolation of convention from
+    specification and require the Specification to use firm language for
+    attributes it requires, like `service.name`.
+  - We will provide x-links from existing location to the new location.
+- Semantic Convention version will no longer match the specification version.
```
tigrannajaryan, line 90: >- Semantic Convention version will no longer match the specification version.

I think this is fine, but IMO the next version of conventions and of SchemaURL should be v1.20.0 (or whatever is the version of the spec after the split happens).

text/0227-separate-semantic-conventions.md
tigrannajaryan: I disagree that we need to reset the version numbers and introduce a new package names. See my other comments.

tigrannajaryan:
To summarize my position:
- I agree it is a good idea to move semantic convention definitions to a new repo.
- I disagree we need to reset the version numbers to v0.0.0. I think we need to continue incrementing existing version numbers. This will prevent disruptions to many other things that rely on continuity of Otel's schema family and non-overlapping version numbers.

text/0227-separate-semantic-conventions.md
lmolkova: > How would that work in relation to SchemaURL concerns I raised above? I.e. is it technically feasible for us to have two versions split in one repo?

@jsuereth I can experiment to see if something goes wrong. On the high-level I don't see an issue if we think about version for the whole convention and not individual parts:
- Tracer/meter created for HTTP instrumentation sets schema to 1.X.  network conventions are shaded in and cannot move on their own.
- Tracer/meter for pure socket communication, gRPC, DB, etc set url to schema 0.X

tigrannajaryan: I have a feeling we are overcomplicating things. What exactly is the problem we are solving by giving different parts of semconv different version numbers?

lmolkova: We also want to ship stable instrumentation libraries for HTTP,  so we need to have different [semconv artifacts](https://mvnrepository.com/artifact/io.opentelemetry/opentelemetry-semconv) generated (stable and alpha).

So we have to have a way to distinguish stable and unstable conventions.

We can do it by separating semconv to `stable` and `experimental` folders and making tooling understand it.
Or by adding a `version` property in yaml that applies to the group.
E.g. 

```diff
  - id: network-core
    prefix: net
    type: attribute_group
    brief: These attributes may be used for any network-related operation.
+ version: 1.0.0

or

+ stable: true
    ...
````

In any case, this is needed regardless of where conventions live, @trask I hope we can discuss it with HTTP crew.

Comment (lmolkova) [reply text/0227-separate-semantic-conventions.md:79]: >  What exactly is the problem we are solving by giving different parts of semconv different version numbers?

@tigrannajaryan from my point of view, we can have the same version number, but the stable/experimental distinction is essential to ship instrumentation libraries with corresponding stability guarantees.

Comment (tigrannajaryan) [reply text/0227-separate-semantic-conventions.md:79]: Totally agree with the need to distinguish stable and unstable conventions. I am just saying I don't think we should use version numbers to be the designators of the stability.

File: text/0227-separate-semantic-conventions.md [hunk 15]
```diff
@@ -0,0 +1,156 @@
...
+
+Additionally, if the semantic conventions eventually move to domain-specific
+director structure (e.g. `docs/{domain}/README.md`, with trace, metrics, events
+in the same file), then this can be refactored in the new repository, preserving
+git history.
+
+There will also be the following exceptions in the specification:
+
+- Semantic conventions used to implement API/SDK details will be fully specified
+  and will not be allowed to change in the Semantic Convention directory.
```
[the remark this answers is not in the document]
jsuereth: Yes, I think that's likely what we do.

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,156 @@
...
+- A moratorium will be placed on Semantic Convention PRs to the specififcation
+  repository. (Caveat that PRs related to this proposal would be allowed).
+- All semantic conventions in the Specification will be copied to the new
+  repository.
+- Semantic conventions in the Specification will be marked as deprecated with
+  links to the new location.
+- The Specification will be updated to require "special" conventions, like
+  `service.name` and configuration interaction.
+- Instrumentation authors will update their `SchemaURL` to
+   `https://opentelemetry.io/semantic_conventions/schemas/{semconv_version}`
```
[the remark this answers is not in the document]
tigrannajaryan: Do we specify anywhere that convention version numbers follow semver rules? I can't find it in our spec. If we don't do that then why is it implied that 1.x is stable?

text/0227-separate-semantic-conventions.md
jsuereth: My main concern here is the tie of SchemaURL to semantic-versioning specification and the fact we're already at `1.x`.

Given discussion around what is considered breaking changes and the role of SchemaURL here, I don't think we have had any guarantee of stability for Semconv yet.

I'd like to get off on the right foot here with a `0.x` release denoting instability and a `1.x` denoting stability.

text/0227-separate-semantic-conventions.md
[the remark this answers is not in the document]
jsuereth: Here's the docs:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/README.md#schema-version-number

```
Version numbers use the [ordering rules](https://semver.org/#spec-item-11) defined by semver 2.0 specification. See how ordering is used in the [Order of Transformations](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/file_format_v1.1.0.md#order-of-transformations). Other than the ordering rules the schema version numbers do not carry any other semantic meaning.
```

So you're right there is no semantic meaning of breaking dependency.  However, effectively that has crippled SchemaURLs right now.  The only mechanism we have to denote "breaking" is changing the Schema family -> i.e. the URL.   OpenTelemetry is *required* to use the same URL in the current experimental specification, therefore OTEL cannot ever break.

I can update this proposal to include some fixes to how SchemaURL interacts.  I *am* suggested the current specification for SchemaURL is problematic.

text/0227-separate-semantic-conventions.md
jsuereth: @tigrannajaryan I agree we can avoid using version numbers to denote stability of conventions.

However, we *do* need something denote when stability has broken.  Without being able to leverage major version numbers, I'm concerned we're stuck.

text/0227-separate-semantic-conventions.md
jsuereth: I think there's a few "oddities" in the existing ecosystem we'll have to tease out and this migration is worth the effort.

E.g. the `enum` [discussion](https://github.com/open-telemetry/opentelemetry-specification/pull/3225/files#r1150119273) and its implications on the stability of existing conventions and changes required.

That said, if we just want to migrate things over with the existing version and evolve in place, I think this may wind up with *harder* to deal with breakages, but I can carve out a path for this.

text/0227-separate-semantic-conventions.md
jsuereth: @tigrannajaryan How would you ensure "safe" moving of semconv from current specification to the new repo?  I do *not* want a stop-the-world migration for all SiGs.

text/0227-separate-semantic-conventions.md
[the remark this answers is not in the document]
tigrannajaryan: If we need a way to denote breaking changes let's work on that. If we agree that the semver will be how we want to denote that then I will agree with you that we need to move back to 0.x.

However, I am not sure that that is the right solution to denote the breaking change. I think going to 0.x is unnecessarily disruptive. I think we may want to find a different solution that is less disruptive.

To understand what the solution should be I think we need to first explain what the problem is. I am not sure I understand the problem and I do not see it described anywhere. So to me the first question to answer is: why is it necessary to denote which semantic conventions are stable vs unstable and why does it need to be possible to understand by looking at the Schema URL? What is the expectation from the consumers of Schema URL? Why do the consumers care?

>I am suggested the current specification for SchemaURL is problematic.

That's entirely possible :-)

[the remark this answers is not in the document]
jsuereth: I understand you find this proposal disruptive.  I think moving semconv to a new location is *inherently* disruptive.  What this proposal does is try to embrace the distruptive in a way that's beneficial to community.  Specifically by:

1. Making it clear the disruption is happening.
2. Allowing Language SiGs to migrate-at-their-own pace.

If you think there's a way to do this switch with less disruption, I'd like to see the proposal.  I want to avoid the situation w/ Spec changes + Language SiGs are held "hostage" by a large forced migration to the new repository.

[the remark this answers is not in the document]
tigrannajaryan: >I think moving semconv to a new location is inherently disruptive.

Moving the content to a different repository does not need to be disruptive from schema publishing perspective. We can continue the current schema family with zero disruption.

>What this proposal does is try to embrace the distruptive in a way that's beneficial to community. Specifically by:
>Making it clear the disruption is happening.
>Allowing Language SiGs to migrate-at-their-own pace.

I think the disruption is avoidable, so we don't necessarily need to embrace it. :-)

I see 2 elements in this change:
1. Moving the source of truth for conventions to a different repository decouples them from the specification and allows conventions to be freed from overly strict processes we rightfully have in place for the spec repo and instead adopt their own processes. It also allows finer granularity in permission control. This will allow semantic convention contributors to move faster and in my opinion is a good idea and is a win for everyone.
2. Schemas that describe semantic conventions are published from this spec repo, which is currently the source of truth for conventions. We plan to move the source of truth to a different physical repository. This is fine and I do not see any reason for people who _consume_ schemas. They don't need to know which github repository contains the original schema files. Today the schemas are published at https://opentelemetry.io/schemas and they can continue being published at the same place after the original schema files are moved from spec repo to a semconv repo.


>If you think there's a way to do this switch with less disruption, I'd like to see the proposal. 

Here is I think what we can do:
1. Copy the existing semconv content and schema files from the spec repo to a new repo.
1. Create new teams and introduce any new processes wanted on the new repo.
1. Delete convention yaml files and schema yaml files from spec repo to prevent accidental usage in the future. 
1. Language SIGs can begin using the yaml files from the new repo next time they need update their generated semconvs. The [semconv generator](https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions) should work fine with the new repo. SIGs can do this at their pace.
1. Add a notice in convention md files in the spec repo saying that semconv is moved.
1. Change opentelemetry.io/schemas publishing process to consume schema files from the new repo.
1. Continue publishing schemas within the same schema family, without resetting version numbers.

Let me know if you want me to elaborate on the process.

>I want to avoid the situation w/ Spec changes + Language SiGs are held "hostage" by a large forced migration to the new repository.

I may be missing something but I don't see the large forced migration that you are referring to. Can you please clarify?

text/0227-separate-semantic-conventions.md
tigrannajaryan: >However, we do need something denote when stability has broken. Without being able to leverage major version numbers, I'm concerned we're stuck.

@jsuereth OK, sounds good. I am still not sure I understand _why_ we need denote the stability. For whom is this information? Is it for schema file consumers? Is it for semconv creators? Is it for instrumentation developers? Depending on the answer we may want a different solution.

text/0227-separate-semantic-conventions.md
tigrannajaryan: I posted a suggestion [here](https://github.com/open-telemetry/oteps/pull/227#discussion_r1168993110), let's continue in that thread.

tigrannajaryan:
@jsuereth thanks for updating the OTEP. I no longer see in the text whether you require that version numbers are reset and we start from 0.0.0 in the new repo. Is that the case or we can continue from the next available version number 1.21.0?

text/0227-separate-semantic-conventions.md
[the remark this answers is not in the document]
jsuereth: So before (1) we need to place a moratorium on semconv in the Specification so that no changes CAN occur during the transition.

Regarding (3) we cannot delete the yaml files until language sigs have moved their code generation, others we've possibly broken their builds or process.  Also, deleting this would break our existing checks on markdown + semconv.

Additionally, there should be a clear signal from language SiGs that the migration has completed so we know when we could remove semconv from the latest branches in the specification.

text/0227-separate-semantic-conventions.md
jsuereth: How would deal with the confusion where people would go to the specification version to look for semconv?  

SchemaURL does NOT denote where the definition lies, it only tracks changes.   Right now, the community is tuned into looking at the specification for the current URL.

My suggestion of a new URL is to help aide with training folks to look in a new place for the new version and NOT expect the specification to hold semconv going forward.

text/0227-separate-semantic-conventions.md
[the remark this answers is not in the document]
tigrannajaryan: >So before (1) we need to place a moratorium on semconv in the Specification so that no changes CAN occur during the transition.

Sounds good. We should aim to execute this plan as quickly as possible to minimize the time period when we cannot change/add conventions.

>Regarding (3) we cannot delete the yaml files until language sigs have moved their code generation, others we've possibly broken their builds or process. 

From what I understand code generation is used only when a new version of semconv is introduced (new version of spec today). Language SIGs will not need to use the code generator because we will not be making new releases until the transition to the new repo is complete.

We can double check this with language SIGs to make sure what I am saying is correct. If any of the language sigs uses the code generator for any other purpose (e.g. as part of regular builds somehow) then we can do what you suggest and delay the deletion of exiting yaml files until no-one depends on them anymore.

The code generator currently requires the path to the yaml files to be specified as a [command line parameter](https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions#code-generator). It is not hard-coded to the spec repo in any way. So there are no changes in the generator itself and for language SIGs the change is minimal, they will need to update their processes to reference the new repo (which the code generator expects to be somewhere in a local directory). Here is for example [how it is done in Go](https://github.com/open-telemetry/opentelemetry-go/blob/main/RELEASING.md#semantic-convention-generation).

I have not checked the other languages, we can work with all maintainers to confirm that they do something similar.

>Also, deleting this would break our existing checks on markdown + semconv.

We should also remove the checks when we delete the yamls. Markdown will be manually marked "deprecated" or "moved" and will be frozen in that state. Checks won't be necessary anymore.

text/0227-separate-semantic-conventions.md
tigrannajaryan: >How would deal with the confusion where people would go to the specification version to look for semconv?

We are planning to plaster the markdown in spec repo with "WARNING: DEPRECATED/MOVED TO FOO!!!" everywhere, right? We have to do it regardless since the community is tuned to go to the spec repo anyway. I am not sure how keeping the current Schema URL family makes it worse, especially since the Schema URL does not suggest in any way that the semconvs or schema files must be found in the spec repo.

>My suggestion of a new URL is to help aide with training folks to look in a new place for the new version and NOT expect the specification to hold semconv going forward.

I think this is solvable by warning messages in the markdown and links to the new repo. I think we can go a step further and as soon as the transition is complete delete the main content in the spec repo's markdown files and  only keep the warning and links to the new repo. There is no way to be confused when you land in the spec repo and see that.

text/0227-separate-semantic-conventions.md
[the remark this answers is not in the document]
jsuereth: > We can double check this with language SIGs to make sure what I am saying is correct. If any of the language sigs uses the code generator for any other purpose (e.g. as part of regular builds somehow) then we can do what you suggest and delay the deletion of exiting yaml files until no-one depends on them anymore.

We rely on the YAML files in the specification itself to ensure all Markdown tables are actually up-to-date with the YAML.  That's the main reason we can't delete them.

> The code generator currently requires the path to the yaml files to be specified as a [command line parameter](https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions#code-generator).

Yes, I'm aware of this.  The automation still hardcodes pulling down the specification repo, so that needs to change.  While it's not a large change, it's still something every langauge/collector will need to do.

> We should also remove the checks when we delete the yamls. Markdown will be manually marked "deprecated" or "moved" and will be frozen in that state. Checks won't be necessary anymore.

That's fair.  I still think we should delay doing this until we're comfortable with the new repository and have fully shifted to it.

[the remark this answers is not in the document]
tigrannajaryan: Sounds good. Let's just aim to make the transition period as short as possible. I can help with execution.

text/0227-separate-semantic-conventions.md
jsuereth: We chatted about this in person. I've updated the PR as follows:

1. we will do the process you suggest.
2. We will NOT change the schema url location
3. We WILL change the schema url version on our first release to denote moving away from schema version.

text/0227-separate-semantic-conventions.md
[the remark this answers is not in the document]
jsuereth: Think we're aligned, updated the specification appropriately.

text/0227-separate-semantic-conventions.md
jsuereth: We discussed this and, like it or not, we have "defacto stable" semconv today from remaining unchanged for so long.  As such, starting at 1.x seems reasonable.  We're going to have the first release from the new repo be 2.0 as there are some breaking changes, and we can figure out version denotations going forward from there.

text/0227-separate-semantic-conventions.md
jsuereth: Updated this to include first released version.  PTAL.

text/0227-separate-semantic-conventions.md
tigrannajaryan: >We WILL change the schema url version on our first release to denote moving away from schema version.

Didi you mean "to denote moving away from spec version"?

text/0227-separate-semantic-conventions.md
jsuereth: WE talked about this offline.  We want to denote stability in two fashions:

1. Programatically via SchemaURL
2. Intrinsically via Semantic Versioning of the SchemaURL version.

We expect (2) to catch the attention of OTEL Users who don't pay attention to details/nuances of our designs + discussions.

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,177 @@
...
+  `{specification}/semantic_conventions` today.
+- `docs/` A new directory that contains human readable documentation for how to
+  create instrumentation compliant with semantic conventions.
+  - `resource/` - Contents of `{specification}/resource/semantic_conventions`
+  - `trace/` - Contents of `{specification}/trace/semantic_conventions`
+  - `metrics/` - Contents of `{specification}/metrics/semantic_conventions`
+  - `logs/`- Contents of `{specification}/logs/semantic_conventions`
+  - `schemas/` - A new location for [Telemetry Schemas](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/README.md)
+    to live. This directory will be hosted at
+    `https://opentelemetry.io/semantic_conventions/schemas/`
```
tigrannajaryan, line 32: I think we agreed to contiunue the same Schema family, right?
```suggestion
    `https://opentelemetry.io/schemas/`
```

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,177 @@
...
+
+Existing semantic conventions in the specification would be marked as
+deprecated, with documentation denoting the move.
```
tigrannajaryan, line 35: Would it be more precise to say "moved" instead of "deprecated"?

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,177 @@
...
+
+Additionally, if the semantic conventions eventually move to domain-specific
+director structure (e.g. `docs/{domain}/README.md`, with trace, metrics, events
+in the same file), then this can be refactored in the new repository, preserving
+git history.
+
+There will also be the following exceptions in the specification:
+
+- Semantic conventions used to implement API/SDK details will be fully specified
```
tigrannajaryan, line 44:
```suggestion
- Semantic conventions used to implement API/SDK details will be fully specified in the `opentelemetry-specification` repo
```

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Internal details
+
+The following process would be used to ensure semantic conventions are
+seamlessly moved to their new location. This process lists steps in order:
+
+- A moratorium will be placed on Semantic Convention PRs to the specififcation
+  repository. (Caveat that PRs related to this proposal would be allowed).
+- Interactions between Semantic Conventions and the Specification will be
+  extracted such that the Specification can require things of Semantic
```
tigrannajaryan, line 69:
```suggestion
  extracted such that the Specification can place requirements on Semantic
```

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,177 @@
...
+      be adoptable/usable.**
+- Semantic conventions in the Specification will be marked as deprecated with
+  links to the new location.
+  - The semconv YAML files in the specification repository *will be deleted*.
+  - All semconv markdown files will be updated such that:
+    - They no longer generate from YAML files.
+    - They include a header denoting deprecation and move to the new repository.
+- Instrumentation authors will update their code generation to pull from the new
+  semconv repository instead of the specification repository..
+- The new repository will have its first released versioned be `2.0.0`, to avoid
```
tigrannajaryan, line 89: Just to confirm: as a consequence does this mean that the next SchemaURL will be `https://opentelemetry.io/schemas/2.0.0`?

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Trade-offs and mitigations
+
+This proposal has a few drawbacks:
+
+- The semantic conventions will no longer be easily referencable form the specification.
+  - This is actually a benefit. We can ensure isolation of convention from
+    specification and require the Specification to use firm language for
+    attributes it requires, like `service.name`.
+  - We will provide x-links from existing location to the new location.
```
tigrannajaryan, line 100: What's an x-link?

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,177 @@
...
+- Semantic Convention version will no longer match the specification version.
+  - Instrumentation authors will need to consume a separate semantic-convention
+    bundle from Specification bundle. What used to be ONE upgrade effort will
```
tigrannajaryan, line 103: We expect that the very small number of conventions in the spec repo will change very infrequently. Vast majority of upgrades are going to be induced by semantic-convetion repo releases.

Question: are you suggesting that we also keep the yaml files in spec repo, containing those few conventions that spec defines and that semconv code generator is able to consume the yaml files from both spec repo and semantic-conventions repo? Or does code generator only use semantic-conventions repo as the source?

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,177 @@
...
+    now be split into two (hopefully smaller) efforts.
+  - We expect changes from Semantic Conventions and Specification to be
+    orthogonal, so this should not add significant wall-clock time.
```
tigrannajaryan, line 106: I am not sure about complete orthogonality here. If `service.name` is in the spec and the rest of Service attributes are in the new repo and the identity of the Service is defined as a combination of all those attributes then there is still some (weak) coupling. I agree this should not add significant wall-clock, so in the end I don't think this is a problem, but want to make sure others don't mis-read this.

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,177 @@
...
+- Existing PRs against semantic conventions will need tobe regenerated.
```
tigrannajaryan, line 107:
```suggestion
- Existing PRs against semantic conventions will need to be regenerated.
```

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,177 @@
...
+
+Initially this repository would have the following ownership:
```
tigrannajaryan, line 109: I can be also added as a CODEOWNER for everything related to schemas to help review PRs.

text/0227-separate-semantic-conventions.md
jsuereth: Yes, this was just an oversite.

text/0227-separate-semantic-conventions.md
jsuereth: a cross-line, reworded to just link

text/0227-separate-semantic-conventions.md
jsuereth: Code generator only uses semantic convention repo as source.  The specification can require certain attributes have certain meaning, but we'd only YAML that in the semconv repo.

text/0227-separate-semantic-conventions.md
jsuereth: I can change it to be "mostly orthogonal". I agree there's nuance, but for the majority of semconv we don't see tight coupling between spec + semconv.

text/0227-separate-semantic-conventions.md
tigrannajaryan: Sounds good. 

Let's just keep in our mind that changing the major version number doesn't signify anything special, see  [here](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/schemas#schema-version-number):
>Other than the ordering rules the schema version numbers do not carry any other semantic meaning.

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+  - `resource/` - Contents of `{specification}/resource/semantic_conventions`
+  - `trace/` - Contents of `{specification}/trace/semantic_conventions`
+  - `metrics/` - Contents of `{specification}/metrics/semantic_conventions`
+  - `logs/`- Contents of `{specification}/logs/semantic_conventions`
+  - `schemas/` - A new location for [Telemetry Schemas](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/README.md)
+    to live. This directory will be hosted at
+    `https://opentelemetry.io/schemas/`
+
+Existing semantic conventions in the specification would be marked as
+moved, with documentation denoting the move, but preserving previous contents.
```
tigrannajaryan, line 35: Nit: we seem to use "moved" vs "deprecated" in some place. It would be good to settle on one term.

tigrannajaryan (approved):

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+  - Error/Exception handling will remain in the specification.
+  - SDK configuration interaction w/ semantic convention will remain in the
+    specification. Specifically `service.name`.
+- The specification may elevate some semantic conventions as necessary for
+  compatibility requirements, e.g. `service.instance.id` and
+  [Prometheus Compatibility](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md).
+
+These exceptions exist because:
+
+- Stable portions of the specification already rely on these conventions
```
jack-berg, line 55:
```suggestion
- Stable portions of the specification already rely on these conventions.
```

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+- A new repository `open-telemetry/semantic-conventions` will be constructed with
+  the proposed format and necessary `Makefile` / tooling.
+  - The new repository would be created by using `git filter-branch` to preserve
+    all existing semantic convention history. *This means all existing
+    semantic conventions will be in the new repository*.
+  - Github Actions, `Makefile` tooling and contributing / readmes would be
+    updated for the separate repository.
+  - **Note: At this point, the new location for semantic conventions should
+      be adoptable/usable.**
+- Semantic conventions in the Specification will be marked as deprecated with
```
jack-berg, line 81:
```suggestion
- Semantic conventions in the Specification will be marked as moved with
```

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+  links to the new location.
+  - The semconv YAML files in the specification repository *will be deleted*.
+  - All semconv markdown files will be updated such that:
+    - They no longer generate from YAML files.
+    - They include a header denoting deprecation and move to the new repository.
+- Instrumentation authors will update their code generation to pull from the new
+  semconv repository instead of the specification repository..
+- The new repository will have its first released versioned be `2.0.0`, to avoid
```
jack-berg, line 89: Would we plan on having a major version bump for the proposed changes to the http semantic conventions? If yes, should we bundle those changes with this? If no, ignore.

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+
+- The semantic conventions will no longer be easily referencable form the specification.
+  - This is actually a benefit. We can ensure isolation of convention from
+    specification and require the Specification to use firm language for
+    attributes it requires, like `service.name`.
+  - We will provide links from existing location to the new location.
+- Semantic Convention version will no longer match the specification version.
+  - Instrumentation authors will need to consume a separate semantic-convention
+    bundle from Specification bundle. What used to be ONE upgrade effort will
+    now be split into two (hopefully smaller) efforts.
```
jack-berg, line 104: In the java SIG, we discussed breaking out a new repository to hold the generated code for the semantic conventions. Something like `semantic-conventions-java`. This artifact could be versioned in sync with the `semantic-conventions` repo and the other java repo `opentelemetry-java` and `opentelemetry-java-instrumentation` would take a dependency on it.

jack-berg (approved):

trask (approved):

text/0227-separate-semantic-conventions.md
jsuereth: Jack fixed this one.

text/0227-separate-semantic-conventions.md
jsuereth: Agreed.  We still need to figure out how to denote this in Schemas

text/0227-separate-semantic-conventions.md
jsuereth: I would like to do that, yes.  That's a discussion that belongs outside this OTEP.  First is if we separate, next is details on HTTP semconv stability + breakage.

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+
+There will also be the following exceptions in the specification:
+
+- Semantic conventions used to implement API/SDK details will be fully specified in the `opentelemetry-specification` repo
+  and will not be allowed to change in the Semantic Convention directory.
+  - Error/Exception handling will remain in the specification.
+  - SDK configuration interaction w/ semantic convention will remain in the
+    specification. Specifically `service.name`.
+- The specification may elevate some semantic conventions as necessary for
+  compatibility requirements, e.g. `service.instance.id` and
```
yurishkuro, line 50: It's obvious the two repos will be referencing each other. Why can't there be a clean separation of _responsibilities_? Spec can say that `service.name` must be defined and how it impacts SDK etc, but at the same time refer to SemConv for usual details of an attribute.

If we have this explicit exclusion for _some_ attributes, we would also need to bifurcate the schemas, which would be bad.

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+      be adoptable/usable.**
+- Semantic conventions in the Specification will be marked as moved with
+  links to the new location.
+  - The semconv YAML files in the specification repository *will be deleted*.
+  - All semconv markdown files will be updated such that:
+    - They no longer generate from YAML files.
+    - They include a header denoting deprecation and move to the new repository.
+- Instrumentation authors will update their code generation to pull from the new
+  semconv repository instead of the specification repository..
+- The new repository will have its first released versioned be `2.0.0`, to avoid
```
yurishkuro, line 89: > to avoid confusion with specification version

Not very clear on this motivation. What confusion? The intention is that these will diverge quickly anyway, and if 2.x SC vs. 1.x Spec looks like a difference, at some point Spec will also may be 2.x

Basically, I don't think we should bump to 2.x, just let the versions roll.

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+This proposal has a few drawbacks:
+
+- The semantic conventions will no longer be easily referencable form the specification.
+  - This is actually a benefit. We can ensure isolation of convention from
+    specification and require the Specification to use firm language for
+    attributes it requires, like `service.name`.
+  - We will provide links from existing location to the new location.
+- Semantic Convention version will no longer match the specification version.
+  - Instrumentation authors will need to consume a separate semantic-convention
+    bundle from Specification bundle. What used to be ONE upgrade effort will
```
yurishkuro, line 103: what would be in the "Specification bundle"?

text/0227-separate-semantic-conventions.md
Aneurysm9: My biggest concern here is that we don't go backward.  I don't know that I have a strong preference for starting from `1.22` (or whatever the next minor version would be at the time) or `2.0`, but starting at `1.0` or `0.1` would be problematic for versioned semconv packages and schema URLs.

dyladan (approved):

jamesmoessis (approved):

text/0227-separate-semantic-conventions.md
dyladan: We have this in JS as well. In practice it is tough to version exactly with the semconv and we decided to keep our own version numbers. Now that it's a bit more stable and we're not refactoring it and making changes like that it might be easier.

text/0227-separate-semantic-conventions.md
tigrannajaryan: We definitely don't want to go backwards.

In addition @jsuereth suggests that we have a way to indicate in the version number that breaking changes occurred. I think it is not a bad idea, but we don't have to agree on it in this OTEP. It may be worth removing this particular sentence from the OTEP and make the OTEP strictly about the "separation from the spec".

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+A new github repository called `semantic-conventions` would be created in the
+OpenTelemetry organization.
+
+This would *initially* have the following structure:
+
+- Boilerplate files, e.g. `README.md`, `LICENSE`, `CODEOWNERS`, `CONTRIBUTING.md`
+- `Makefile` that allows automatic generation of documentation from model.
+- `semantic_conventions/` The set of YAML files that exist in
+  `{specification}/semantic_conventions` today.
+- `docs/` A new directory that contains human readable documentation for how to
```
carlosalberto, line 24: Is this the so-called "supplementary guidelines"?

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+As such, the Specification should define the absolute minimum of reserved or
+required attribute names and their interaction with the SDK.
+
+## Internal details
+
+The following process would be used to ensure semantic conventions are
+seamlessly moved to their new location. This process lists steps in order:
+
+- A moratorium will be placed on Semantic Convention PRs to the specififcation
+  repository. (Caveat that PRs related to this proposal would be allowed).
```
carlosalberto, line 67: What about existing experimental conventions in the SDK, such as `service.version`? I'm taking for granted they will remain in the SDK _and_ as experimental for now, but would like to clarify this.

text/0227-separate-semantic-conventions.md
carlosalberto: @jsuereth I'm also expecting you won't have monthly releases, now that this will be split off of the Spec?

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+There will also be the following exceptions in the specification:
+
+- Semantic conventions used to implement API/SDK details will be fully specified in the `opentelemetry-specification` repo
+  and will not be allowed to change in the Semantic Convention directory.
+  - Error/Exception handling will remain in the specification.
+  - SDK configuration interaction w/ semantic convention will remain in the
+    specification. Specifically `service.name`.
+- The specification may elevate some semantic conventions as necessary for
+  compatibility requirements, e.g. `service.instance.id` and
+  [Prometheus Compatibility](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md).
```
carlosalberto, line 51: On a related note, there's also the (experimental) Compatibility semantic conventions section, which only includes OpenTracing values at the moment: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/compatibility.md 

I'd vow to keep them in the Spec as well but I'm open to discuss it.

pyohannes (approved):

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+
+As such, the Specification should define the absolute minimum of reserved or
+required attribute names and their interaction with the SDK.
+
+## Internal details
+
+The following process would be used to ensure semantic conventions are
+seamlessly moved to their new location. This process lists steps in order:
+
+- A moratorium will be placed on Semantic Convention PRs to the specififcation
```
trask, line 66: if HTTP semconv PRs are going to be blocked until semconv 2.0 is released from the new repo, I'm interested to see language that this move is going to start (at least) soon

```suggestion
- As soon as this OTEP is merged,
- a moratorium will be placed on Semantic Convention PRs to the specififcation
```

dyladan: OTEPs go into effect when they are merged (unless stated otherwise)

```suggestion
- A moratorium is placed on Semantic Convention PRs to the specififcation
```

lmolkova (approved):

reyang (approved):

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+- `Elastic Common Schema` - the schema is its own project / document.
+- `Prometheus` - Prometheus does not define rigid guidelines for telemetry, like
+  semantic conventions, instead relying on naming conventions and
+  standardization through mass adoption.
+
+## Open questions
+
+This OTEP doesn't address the full needs of tooling and codegen that will be
+needed for the community to shift to a separate semantic convention directory.
+This will require each SiG that uses autogenarated semantic conventions to
```
reyang, line 151:
```suggestion
This will require each SIG that uses autogenerated semantic conventions to
```

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+  - [Liudmila Molkova](https://github.com/lmolkova), Microsoft
+  - [Sean Marcinak](https://github.com/MovieStoreGuy), Atlassian
+  - [Ted Young](https://github.com/tedsuo), Lightstep
+- Approvers (HTTP Only)
+  - [Trask Stalnaker](github.com/trask)
+- Approvers (SchemaUrl Files)
+  - [Tigran Najaryan](github.com/tigrannajaryan)
+- Maintainers
+  - [Josh Suereth](github.com/jsuereth)
+  - [Armin Reuch](github.com/arminru)
```
arminru, line 125:
```suggestion
  - [Armin Ruech](github.com/arminru)
```
🙂

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+## Motivation
+
+We need to allow semantic conventions to evolve mostly independent of the
+overall OpenTelemetry specification. Today, any breaking change in a semantic
+convention would require bumping the version number of the entirety of the
+OpenTelemetry specification.
+
+## Explanation
+
+A new github repository called `semantic-conventions` would be created in the
```
reyang, line 15: nit:

```suggestion
A new GitHub repository called `semantic-conventions` would be created in the
```

text/0227-separate-semantic-conventions.md
reyang: specififcation -> specification

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+- Interactions between Semantic Conventions and the Specification will be
+  extracted such that the Specification can place requirements on Semantic
+  Conventions and *normative* specification language will remain in the
+  core specification directories.
+- A new repository `open-telemetry/semantic-conventions` will be constructed with
+  the proposed format and necessary `Makefile` / tooling.
+  - The new repository would be created by using `git filter-branch` to preserve
+    all existing semantic convention history. *This means all existing
+    semantic conventions will be in the new repository*.
+  - Github Actions, `Makefile` tooling and contributing / readmes would be
```
reyang, line 77:
```suggestion
  - GitHub Actions, `Makefile` tooling and contributing / readmes would be
```

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+  - [Trask Stalnaker](github.com/trask)
+- Approvers (SchemaUrl Files)
+  - [Tigran Najaryan](github.com/tigrannajaryan)
+- Maintainers
+  - [Josh Suereth](github.com/jsuereth)
+  - [Armin Reuch](github.com/arminru)
+  - [Reiley Yang](github.com/reyang)
+
+That is, Maintenance would initially continue to fall on (a subset of) the
+Technical committee. Approvers would start with existing semconv approvers in
```
reyang, line 129:
```suggestion
Technical Committee. Approvers would start with existing semconv approvers in
```

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+- Instrumentation authors will update their code generation to pull from the new
+  semconv repository instead of the specification repository..
+- The new repository will have its first released versioned be `2.0.0`, to avoid
+    confusion with specification version.
+
+## Trade-offs and mitigations
+
+This proposal has a few drawbacks:
+
+- The semantic conventions will no longer be easily referencable form the specification.
```
reyang, line 96:
```suggestion
- The semantic conventions will no longer be easily referenceable form the specification.
```

text/0227-separate-semantic-conventions.md
```diff
@@ -0,0 +1,179 @@
...
+  new signals or hard-to-resolve new domains without breaking the Specification
+  version number.
+- Semantic Conventions can have dedicated maintainers and approvers.
+- Semantic Conventions can restructure to better enable subject matter experts
+  (SMEs) to have approver/CODEOWNER status on relevant directories.
+- Semantic Conventions can adopt semantic versioning for itself, helping clearly
+  denote breaking changes to users.
+
+There is a desire to move semantic conventions to domain-specific directories
+instead of signal-specific. This can occur after a separation of the repository
```
reyang, line 168:
```suggestion
instead of signal-specific. This can occur after the separation of the repository
```

text/0227-separate-semantic-conventions.md
jsuereth: This isn't how OTEPs work.  OTEPs are design docs, plans.  The actually changes need to hit the Specification and go through approval there.

SO yes, this OTEP (the plan) goes into effect when merged.  However, a moratorium on the Specificiation still needs to be a PR *in* the specification repository with clear messaging to specification contributors.  While most of us may follow OTEPS + specs, not everyone does.  This is why the execution of the plan hinges on that PR.

text/0227-separate-semantic-conventions.md
jsuereth: If you want them in the Specification, I'd suggest you put them here:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/opentracing.md


We can still move the YAML files to the new repository and do a similar "specification owns these attributes and requires them in semconv" behavior we're using for things like `service.name`

@carlosalberto would you be able to put that PR to the specification together?  For now I'll assume the existing YAML/MD is being pulled to the semconv repository based on my proposal above.

text/0227-separate-semantic-conventions.md
jsuereth: This moratorium will need to apply to ANY change right now.

The moratorium is only to ensure we get a clean copy of semconv history in the repository and it will be lifted ASAP (I'm hoping ~2 hours for the process to run).

After the moratorium is over, you could have a PR to the specification on `service.version` as this is the kind of attribute that *needs* to be mandated by the specification.

text/0227-separate-semantic-conventions.md
carlosalberto: Works, will send a PR to move them to the opentracind.md doc.

text/0227-separate-semantic-conventions.md
jsuereth: Maybe may language is not clear.  The idea here is what you're suggesting. 

Specification can enforce that certain attributes MUST exist in semantic conventions and defines usage/interaction w/ SDK or other OTel things.

The Semconv MUST include those attributes in its YAML and define semantics for them.

The requirements, though, should only flow one-way.  Specification can force semconv to include attributes, but semconv should NOT force any specification normative language.

text/0227-separate-semantic-conventions.md
jsuereth: As discussed in Specification SiG -

Here's my reasoning for 2.0.

- We want to clearly denote that the semconv no longer follows the specification version.  SO the first release will definitely be different than the specification.
- We *cannot* go backwards in number, as SchemaURL requires semconv *ordering* of versions.
- I think it's important we use semantic versioning as an *implicit* signal of breaking changes to semconv, where the Schema file should give *explicit* signal of breaking changes.  We know HTTP semconv has broken for this next release.

text/0227-separate-semantic-conventions.md
jsuereth: I would recommend at least using a bifurcated version strategy here:

1. The version of semconv
2. The version of the *codegen* for semconv.


This way if codegen changes in a breaking manner you can bump versions.  Otherwise you'd be forced to limit codegen fixes, and you may find you *need* to make breaking changes there.  The semconv version won't be able to reflect that.

Note: this is the same debate as OTLP versioning and generated protobuf repositories.

arminru (approved):

text/0227-separate-semantic-conventions.md
jsuereth: To resolve this, we moved this to "open questions".   We can continue this discussion of what version to use for the first release of semconv in the semconv repository.

carlosalberto:
Merging this as we have enough approvals and the feedback has been addressed (either by updating the content or by adding the discussion points to the open questions section, e.g. whether we should go 2.0 for the semantic conventions when this split happens).

Let's continue discussing the details once we start working on the actual changes.

PR #234, opened 2023-06-19
Trace Payload Collection

nirga:
Continuing the work started by @ronyis at #219. 

We're proposing semantic attributes for reporting payloads (of HTTP requests, DB queries, etc.) as part of OpenTelemetry. This should be optional for users who wish to use it and can provide consumers of OpenTelemetry with a standard option for getting payloads out of trace instrumentation.

Previous instrumentation libraries (pre-otel) have successfully implemented it so I think this is something that will be welcomed by the community. 

Would love to get your feedback and work on getting this accepted.

### DIFF
--- a/text/trace/0234-payload-collection.md
+++ b/text/trace/0234-payload-collection.md
@@ -1,0 +1,361 @@
+# Payload Collection APIs
+
+Define APIs in OpenTelemetry for capturing payload data.
+
+## Motivation
+
+This OTEP proposes adding APIs and semantic conventions for handling payload data,
+and capturing it in traces and logs, based on existing attributes.
+
+Instrumenting payload data seems to be a feature requested by many users
+[[1]](https://github.com/open-telemetry/opentelemetry-js/discussions/3446)
+[[2]](https://github.com/open-telemetry/opentelemetry-specification/issues/1062)
+[[3]](https://stackoverflow.com/questions/75260865/capture-request-body-using-opentelemetry-in-node-js)
+[[4]](https://github.com/open-telemetry/opentelemetry-specification/issues/376#issuecomment-1227501082)
+
+Tracing companies like Epsagon (now part of Cisco), and Aspecto (now part of SmartBear), have implemented payload collection in their custom instrumentation libraries.
+This capability was leveraged by many customers, including big companies that enabled it in production (and testing) environments.
+It was implemented in multiple runtimes (NodeJS, Python, Go, and others) and instrumentations (e.g. HTTP, AWS SDK, and many DB SDKs).
+
+
+What do we mean by payload data? While it’s hard to precisely define, the general guidelines are:
+- It includes the content of a message, rather than metadata or headers
+- It includes information from the “data plane”, rather than the “control plane”
+Some example of payload data includes HTTP bodies, database queries (including items read or written), and messages produced or consumed from a queue.
+
+Currently, the support in OpenTelemetry for collecting such data is quite limited. We propose in this OTEP to add an API that will define a standard and extendable way for collecting payloads in traces.
+The implmentation of this API is based on existing attributes, with proper semantic conventions.
+
+The possible value from collecting payloads is substantial. Using payload data, OpenTelemetry users can troubleshoot applications much more effectively in many cases - they can use that to understand exact data flows in their systems, reproduce problematic requests, or search for traces by specific payload information.
+Paylod data could also be used to create ad-hoc analytics which helps monitoring the system.
+
+As many could argue, there are many challenges for effieciently collecting payloads from production applications in a generic way: handling sensitive data, performance implications, telemetry vendors support and others. Nevertheless, we think that this should not block OpenTelemetry users from manually collecting payload data already, whenever it fits their requirements.
+
+Having a standard semantic conventions for payload data can allow monitoring tools and others to use them OOTB (if they exist), without relying on custom attribute names.
+
+Ultimatly, it could be beneficial if OpenTelemetry instrumentation SDKs will have OOTB optional payload collection as well, though this OTEP does NOT propose that, or takes into account that it will be added in the future. 
+
+## Explanation
+
+While payload data can just be a binary buffer, most of the times it has a defined structure. 
+For modern applications and APIs, the payload data is usually encoded as one of:
+* JSON
+* YAML
+* Protobuf
+* Avro
+* Plain strings
+* Blob
+
+Whenever the payload data has some meaningful structure (everything except a blob), that structure will be reflected in the semantic attribute of the relevant span. That way, processors downstream can easily access and manipulate the payload data.
+
+## Internal details
+
+### Semantic conventions
+
+We propose specifying a consistent naming for payload content attributes, and its
+related metadata. These conventions could help processors, backends and users to
+handle this kind of data.
+
+In this OTEP we only give ideas and recommondations for it, though
+a further discussion will be required for choosing the ideal naming when
+writing the actual specifications.
+
+The conventions we propose are added as a postfix to the base attribute name
+(for example, `http.request.body`):
+- **Data attribute**: `<attribute>.content`. Holds the decoded content of the payload data. Alternatives: `.data`, `.payload`.
+- **Length attribute**: `<attribute>.length`. Holds the bytes length of the encoded payload data. Alternatives: `.size`, `.bytes`.
+- **Encoding attribute**: `<attribute>.encoding`. Holds the original attribute encoding type.
+Predefined values should be declared (though users may decide using custom values as well).
+For example - `json`, `protobuf`, `avro`, `utf-8`.
+
+Adding each attribute should not be dependent on the others in any way.
+
+### API
+
+We propose specifying APIs the will abstract the handling and capturing of
+payload data, to allow customizing this functionallity and evolving it over time.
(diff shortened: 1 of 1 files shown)


text/trace/0234-payload-collection.md
```diff
@@ -0,0 +1,363 @@
...
+
+In this OTEP we only give ideas and recommendations for it, though
+a further discussion will be required for choosing the ideal naming when
+writing the actual specifications.
+
+The conventions we propose are added as a postfix to the base attribute name
+(for example, `http.request.body`):
+
+- **Data attribute**: `<attribute>.content`. Holds the decoded content of the payload data. Alternatives: `.data`, `.payload`.
+- **Length attribute**: `<attribute>.length`. Holds the bytes length of the encoded payload data. Alternatives: `.size`, `.bytes`.
```
kenfinnigan, line 68: How does the existing trace attribute `http.request.body.size` factor into the OTEP?

nirga: Hmm you're right @kenfinnigan. I think in that case I'd change the name to be `<attribute>.size` so it matched the current convention (and this can be the name for other payloads like DB query etc.)

text/trace/0234-payload-collection.md
```diff
@@ -0,0 +1,363 @@
...
+writing the actual specifications.
+
+The conventions we propose are added as a postfix to the base attribute name
+(for example, `http.request.body`):
+
+- **Data attribute**: `<attribute>.content`. Holds the decoded content of the payload data. Alternatives: `.data`, `.payload`.
+- **Size attribute**: `<attribute>.size`. Holds the number of bytes of the encoded payload data. Alternatives: `.length`, `.bytes`.
+- **Encoding attribute**: `<attribute>.encoding`. Holds the original attribute encoding type.
+  Predefined values should be declared (though users may decide using custom values as well).
+  For example - `json`, `protobuf`, `avro`, `utf-8`.
```
jsuereth, line 71: Are you going to allow "versioning" of these protocols?

What's the path for adding new (custom) encodings?

text/trace/0234-payload-collection.md
```diff
@@ -0,0 +1,363 @@
...
+  supported encodings, in an extendable fashion.
+- Aggregating multiple data chunks (to assist with handling asynchronous buffers)
+
+The APIs given here are only a draft to describe general characteristics.
+Like in the semantics, a further discussion will be required before creating actual specifications.
+
+#### `Payload` class
+
+We propose adding a class that will provide the related functionality for handling payload data.
+Both traces and logs could use instances of this class, which will assist sharing related functionality.
```
jsuereth, line 95: Since this was added in real tracing products, can you link to the APIs/SDKs they offered?

text/trace/0234-payload-collection.md
```diff
@@ -0,0 +1,363 @@
...
+  self.set_attributes(
+    {
+      f"{key}.content": payload.content,
+      f"{key}.encoding": payload.encoding,
+      f"{key}.size": payload.size
+    }
+  )
+```
+
+### Payload attributes in Logs
```
jsuereth, line 150: +1 to resolving this discussion as a priority.

text/trace/0234-payload-collection.md
```diff
@@ -0,0 +1,363 @@
...
+
+The payload content could be technically collected both as the the log record body, or as an attribute.
+We think that it's needed to specify the preferred way following a discussion in the logs SIG.
+[This](https://github.com/open-telemetry/opentelemetry-specification/pull/2926) PR references this issue as well.
+
+### Null value
+
+Representing Null values in payload contents is required, as it part of JSON and many other formats
+like Protobuf and Avro. Therefore we propose supporting it by spans and logs attributes APIs (while
+in OTel proto schema it is already supported).
```
jsuereth, line 160: Where is this supported in OTel proto schema?  Do you mean the "AnyValue" being allowed to be "emtpy"?

I'd like more details about where/how you need `Null`  support here.  Are you trying to record payload attributes even if the payload is null?  Could you just write an empty byte array for the Proto/Avro case?

text/trace/0234-payload-collection.md
```diff
@@ -0,0 +1,363 @@
...
+## Trade-offs and mitigations
+
+### Handling payloads with large size
+
+As payloads are saved as regular attributes, they must follow the defined [limits](https://github.com/alanwest/opentelemetry-specification/blob/eb551bc9cf50d93463f20585686194d62887e044/specification/common/README.md#attribute-limits).
+
+Capturing payloads using appropriate APIs could assist in specifying different limits in the future (either
+more are less strict than of 'regular' attributes), and mechanisms for shortening large payloads as well.
+
+### Handling sensitive data
```
jsuereth, line 171: I think it's worth calling out that payload collection likely requires some kind of flag to signal that sensitive data collection is ok.


Specifically - I think this proposal needs some kind of notion on a per-trace basis that denotes whether sensitive collection is "allowed" for that trace.

E.g. you can imagine a system where I issue an RPC with a special flag denoting that collection of payloads (and other sensitive data) is ok, where this would be turned on.

You can also imagine production systems that require user-consent before this data could be collected in a o11y system.  We should absolutely have a mechanism to check (dynamically, per-request / per-trace) whether this data *can* be collected.

text/trace/0234-payload-collection.md
```diff
@@ -0,0 +1,363 @@
...
+
+In this OTEP we only give ideas and recommendations for it, though
+a further discussion will be required for choosing the ideal naming when
+writing the actual specifications.
+
+The conventions we propose are added as a postfix to the base attribute name
+(for example, `http.request.body`):
+
+- **Data attribute**: `<attribute>.content`. Holds the decoded content of the payload data. Alternatives: `.data`, `.payload`.
+- **Size attribute**: `<attribute>.size`. Holds the number of bytes of the encoded payload data. Alternatives: `.length`, `.bytes`.
```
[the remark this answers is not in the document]
nirga: I think it's important to keep the size as a separate attribute, since we may have an actual different content size (for example if we truncated part of it).

text/trace/0234-payload-collection.md
nirga: It's a bit different, since they just offered this instrumentation OOTB, so the API for the payload was hidden from the developer. See for example [Epsagon's instrumentation for FastAPI](https://github.com/epsagon/epsagon-python/blob/master/epsagon/wrappers/fastapi.py) (The exact structure of the data is [seen here](https://github.com/epsagon/epsagon-python/blob/master/epsagon/runners/fastapi.py#L78))

text/trace/0234-payload-collection.md
nirga: But should it affect the attribute names for spans as proposed here?

text/trace/0234-payload-collection.md
nirga: I wonder if this might be out of scope for this as it should be resolved on the API level of the different SDKs. Here I suggest something much simpler - if we ever to auto-instrument collection of payloads, the default should always be `off` where users who decide to turn it on should know it will be turned on for everything.

text/trace/0234-payload-collection.md
ronyis: The point here is that we need to support JSON payloads like `'{"x": null}'` and also `null`, and similar examples in other formats.
If the payload is encoded with the "AnyValue" type, it can be achieved by using empty objects. Though it's still needed to be supported by the APIs.
In case we add a dedicated proto object, like the `Value` suggested in the alternative, we should also make sure it supports Null values.

text/trace/0234-payload-collection.md
ronyis: I'm also in favor of having this kind of flag regarding sensitive data collection. Though I don't think it should be a requirement for the changes proposed in this OTEP.
At this point, users already collect sensitive data using the existing APIs and attributes.

mishushakov:
Would love to see that happen at [Step CI](https://stepci.com) 😉

axlrate:
Should the payload be a span event instead of a span attribute? Most backends have higher size limits for span events.

mmanciop:
> Should the payload be a span event instead of a span attribute? Most backends have higher size limits for span events.

Lumigo does this with span attributes, and the experience was that a truncation to 1024, 2048 or 4096 or power thereof was often good enough. (The complex bit was performing truncation that would keep validity of the data format, e.g., JSON.) Users have means via SDK configuration to change the attribute truncation threshold. I think setting this data as span events would be very unintuitive in terms of data model. I don't think we should use an unintuitive data model to work around limitations of backends. Moreover, for an important use-case like this, I can imagine backends makings exceptions for the specific keys we'd use.

codefromthecrypt:
might be worthwhile to ping folks who did some things in this space, even if moved on since.

* @mchandramouli who created [blobs at expedia](https://github.com/ExpediaDotCom/blobs) for request/response logging. There are some [ancient notes here](https://github.com/openzipkin/openzipkin.github.io/wiki/2019-01-15-Expedia-Haystack-and-Adaptive-Alerting#blobs), but I suspect concepts are still relevant.

* @pavolloffay who worked on the [hypertrace java agent](https://github.com/hypertrace/javaagent) which adds request/response collection.

carlosalberto:

PR #235, opened 2023-08-09
Threshold propagation in trace state

kentquirk:
This is intended to capture the proposal developed by [the Sampling SIG](https://docs.google.com/document/d/1gASMhmxNt9qCa8czEMheGlUW2xpORiYoD7dBD7aNtbQ/edit#heading=h.mngoctm0c84w).


text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,85 @@
...
+The sampling decision is propagated with the following algorithm:
+* If the `th` key is not specified, then no previous sampling decision has been made.
+* If the value of the `th` key is `!`, always sample.
+* Else parse the `th` key as a hex value as described below.
+* Compare the 56 bits of `T` with the 56 bits of `R`. If `T <= R`, then do not sample.
+* This implies that if `T` is 0, then never sample.
+
+The `R` value MUST be derived as follows:
+* If the Random Trace ID Flag is `true` in the traceparent header, then `R` is the lowest-order 56 bits of the trace-id.
+* Else if the key `rv` is present in the Tracestate header, then `R = rv`.
```
PeterF778, line 31: The presence of `rv` should take precedence, even if the Random Trace ID is set.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,85 @@
...
+* Else if the key `r` is present in the Tracestate header, then `R = 2**(56-r)`.
```
oertl, line 32: I am against making the new proposal compatible with the old proposal. It makes it more complicated and leads to problems. For example, if `r` is present but not `rv` and a non-power-of-two threshold is chosen, the actual sampling probability does not fit the threshold.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,85 @@
...
+* Else `R` should be generated as a random value in the range `(0, (2**56)-1)` and added to the Tracestate header with key `rv`.
+
+The preferred way to propagate this value is as the lowest 56 bits of the trace-id. If these bits are in fact random, the `random` trace-flag SHOULD be set as specified in [the W3C trace context specification](https://w3c.github.io/trace-context/#trace-id).
+
+The value of the `rv` and `th` keys MUST be expressed as up to 14 hexadecimal characters from the set `[0-9a-f]`. Trailing zeros (but not leading zeros) may be omitted. _Question: should we permit [A-F] as well, or instead?_
+
+Examples:
+`th=!` -- "Always Sample".
```
oertl, line 40: I would prefer indicating "Always Sample" by a missing `th` field.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,85 @@
...
+* If the value of the `th` key is `!`, always sample.
+* Else parse the `th` key as a hex value as described below.
+* Compare the 56 bits of `T` with the 56 bits of `R`. If `T <= R`, then do not sample.
+* This implies that if `T` is 0, then never sample.
+
+The `R` value MUST be derived as follows:
+* If the Random Trace ID Flag is `true` in the traceparent header, then `R` is the lowest-order 56 bits of the trace-id.
+* Else if the key `rv` is present in the Tracestate header, then `R = rv`.
+* Else if the key `r` is present in the Tracestate header, then `R = 2**(56-r)`.
+* Else `R` should be generated as a random value in the range `(0, (2**56)-1)` and added to the Tracestate header with key `rv`.
```
oertl, line 33: I am wondering if it would be better to require the generation of `R` in this case. Then consistent sampling would always work if the Random Trace ID Flag is not set.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,87 @@
...
+However, that value is limited to powers of two, while this proposal is not; to distinguish the cases, this proposal uses the key `rv`.
+
+In order to make consistent sampling decisions across the entire path of the trace, two values SHOULD be propagated with the trace:
+
+1. A random (or pseudo-random) value of at least 56 bits, called `R` below.
+2. A 56-bit trace threshold as expressed in the TraceState, called `T` below.
+
+The sampling decision is propagated with the following algorithm:
+
+* If the `th` key is not specified, Always Sample.
```
yurishkuro, line 23: `th` was never introduced

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,87 @@
...
+This system can only achieve non-power-of-two sampling using interpolation between powers of two, which is unnecessarily restrictive.
+In existing sampling systems, sampling probabilities like 1%, 10%, and 75% are common, and it should be possible to express these without interpolation.
+There is also a need for consistent sampling in the collection path (outside of the head-sampling paths) and using inherent randomness in the traceID is a less-expensive solution than referencing a custom `r-value` from the tracestate in every span.
+This proposal allows for the continued expression of randomness using `r-value` as specified there using the key `r`.
+However, that value is limited to powers of two, while this proposal is not; to distinguish the cases, this proposal uses the key `rv`.
+
+In order to make consistent sampling decisions across the entire path of the trace, two values SHOULD be propagated with the trace:
+
+1. A random (or pseudo-random) value of at least 56 bits, called `R` below.
+2. A 56-bit trace threshold as expressed in the TraceState, called `T` below.
```
yurishkuro, line 19:
```suggestion
1. A _random_ (or pseudo-random) value of at least 56 bits, called `R` below.
2. A 56-bit _threshold_ as expressed in the TraceState, called `T` below.
```

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,87 @@
...
+
+The `T` value SHALL be derived as follows:
+
+* If the `th` key is not present in the Tracestate header, then `T` is effectively 2^56 (which doesn't fit in 56 bits).
+* Else the value corresponding to the `th` key should be interpreted as above.
+
+Sampling Decisions SHOULD be propagated by setting the value of the `th` key in the Tracestate header according to the above.
+
+In the case of a downstream sampler that is attempting to reduce the volume of traffic, the sampler MAY modify the `th` header.
+Note that the `T` value may be reduced by a downstream sampler, but should never be increased.
```
yurishkuro, line 54: Could you please elaborate why it cannot be increased? It seems like that would prevent a common use case of inflationary sampling.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,87 @@
+# Sampling Threshold Propagation in TraceState
+
+## Motivation
+
+Sampling can theoretically take place at nearly any point in a distributed tracing system. If sampling is to be performed at multiple points in the process, the only way to reason about it effectively is to make sure that the sampling decisions are **consistent**. In this context consistency is the property that allows different points in the sampling chain to make the same sampling decisions, based on information included with the trace.
```
yurishkuro, line 5: > In this context consistency is the property that allows different points in the sampling chain to make the same sampling decisions, based on information included with the trace.

This definition does not match the proposal, because in the proposal nodes are allowed to reduce the threshold-value, thus the sampling decision is not "the same".

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,87 @@
+# Sampling Threshold Propagation in TraceState
+
+## Motivation
```
yurishkuro, line 3: please add a discussion of calculating the weight for each sample

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,87 @@
...
+This system can only achieve non-power-of-two sampling using interpolation between powers of two, which is unnecessarily restrictive.
+In existing sampling systems, sampling probabilities like 1%, 10%, and 75% are common, and it should be possible to express these without interpolation.
+There is also a need for consistent sampling in the collection path (outside of the head-sampling paths) and using inherent randomness in the traceID is a less-expensive solution than referencing a custom `r-value` from the tracestate in every span.
+This proposal allows for the continued expression of randomness using `r-value` as specified there using the key `r`.
+However, that value is limited to powers of two, while this proposal is not; to distinguish the cases, this proposal uses the key `rv`.
+
+In order to make consistent sampling decisions across the entire path of the trace, two values SHOULD be propagated with the trace:
+
+1. A random (or pseudo-random) value of at least 56 bits, called `R` below.
```
PeterF778, line 18: I think it should be "at most", not "at least". There's no way to provide more than 56 bits for the R value, and the threshold values are encoded in 56 bits anyway

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,87 @@
...
+2. A 56-bit trace threshold as expressed in the TraceState, called `T` below.
+
+The sampling decision is propagated with the following algorithm:
+
+* If the `th` key is not specified, Always Sample.
+* Else derive `T` by parsing the `th` key as a hex value as described below.
+* If `T` is 0, Always Sample. This implies that non-probabalistic sampling is taking place.
```
jpkrohling, line 25:
```suggestion
* If `T` is 0, Always Sample. This implies that non-probabilistic sampling is taking place.
```

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Open questions
+
+This specification leaves room for different implementation options. For example, comparing hex strings or converting them to numeric format are both viable alternatives for handling the threshold.
+
+We also know that some implementations prefer to use a sampling probability (in the range from 0-1.0) or a sampling rate (1/probability); this design permits conversion to and from these formats without loss up to at least 6 decimal digits of precision.
+
+## Future possibilities
+
+This permits sampling systems to propagate consistent sampling information downstream where it can be compensated for. For example, this will enable the tail-sampling processor in the OTel Collector to propagate its sampling decisions to backends in a standard way.
```
jpkrohling, line 87: Can you expand a bit on that? Most of the policies in the tail-sampling processor are based on the characteristics of the trace itself. We do have one probabilistic policy, which is probably what you had in mind, but perhaps I'm misunderstanding parts of this proposal. For completeness, you might want to mention the probabilistic sampling processor as well.

oertl:
Currently, everything is very weakly worded, because the text contains many SHALLs and SHOULDs. Shouldn't we be stricter and require that any compliant consistent sampler implementation MUST follow all the rules defined here?

text/trace/0235-sampling-threshold-in-trace-state.md
oertl: I agree with @yurishkuro.

Maybe:

In this context, consistency means that a positive sampling decision made at a particular point with probability p1 implies a positive sampling decision made at another point that samples a different piece of information from the same trace with probability p2 >= p1.

text/trace/0235-sampling-threshold-in-trace-state.md
oertl: It should be exactly 56 bits.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,87 @@
...
+* If the key `rv` is present in the Tracestate header, then `R = rv`.
+* Else if the Random Trace ID Flag is `true` in the traceparent header, then `R` is the lowest-order 56 bits of the trace-id.
+* Else `R` SHALL be generated as a random value in the range `(0, (2**56)-1)` and added to the Tracestate header with key `rv`.
+
+The preferred way to propagate the `R` value is as the lowest 56 bits of the trace-id.
+If these bits are in fact random, the `random` trace-flag SHOULD be set as specified in [the W3C trace context specification](https://w3c.github.io/trace-context/#trace-id).
+There are circumstances where trace-id randomness is inadequate (for example, sampling a group of traces together); in these cases, an `rv` value is required.
+
+The value of the `rv` and `th` keys SHALL be expressed as up to 14 hexadecimal characters from the set `[0-9a-f]`. Trailing zeros (but not leading zeros) may be omitted. _Question: should we permit [A-F] as well, or instead?_
+
```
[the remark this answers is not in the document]
jmacd: I agree.

text/trace/0235-sampling-threshold-in-trace-state.md
jmacd: Some rewording is needed here -- I think by "downstream" @kentquirk means on the collection path, i.e., in a tail-sampling scenario. In tail sampling, to reduce the threshold means to increase sampling probability after the fact, which is a lie. I think this language can be as strong as `MUST NOT`, for tail sampling.

But note that it's OK for head samplers. A non-root sampler can raise or lower sampling threshold with the repercussions discussed e.g., in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md#recommenendation-use-non-descending-power-of-two-probabilities. If a non-root head sampler lowers probability, you may expect incomplete traces (missing their children, difficult to detect!), and if a non-root head sampler raises probability, you. may expect complete subtraces (missing their parents, easy to detect).

text/trace/0235-sampling-threshold-in-trace-state.md
jmacd: I intend to revive https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/24811 ASAP, which is I think an example of what Kent is predicting. Of course, our eyes are on more-configurable samplers, but that's future work.

kentquirk:
@oertl According to [this RFC](https://www.ietf.org/rfc/rfc2119.txt)https://www.[ietf.org/rfc/rfc2119.txt](https://www.ietf.org/rfc/rfc2119.txt), MUST and SHALL are equivalent. I thought I had read a document that said to prefer SHALL, but now I can't find it, and I'm finding lots of other ones that prefer MUST, so yeah, I've changed it.

I've tried to address most of the comments.

text/trace/0235-sampling-threshold-in-trace-state.md
kentquirk: I feel like I'd rather include some relatively generic code showing conversion between the different modes, rather than referring to a full implementation. In other words, I don't think we need executable code or full error checking, just a clear example of the math. Do you agree?

text/trace/0235-sampling-threshold-in-trace-state.md
kentquirk: I took a shot at it. Please tell me what you think.

text/trace/0235-sampling-threshold-in-trace-state.md
kentquirk: Thank you for the suggested wording!

text/trace/0235-sampling-threshold-in-trace-state.md
kentquirk: I added one line, but if you want more, we can go farther.

The general idea is that we want to be able to pass on the effective sample rate from dynamic sampling decisions, which today's tail sampling processor can't do.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,99 @@
...
+The trace state header SHOULD contain a field with the key `rv`, and a value that corresponds to a 56-bit sampling threshold.
+This value will be compared to the 56-bit random value associated with the trace.
+
+From a technical perspective, how do you propose accomplishing the proposal? In particular, please explain:
+
+* How the change would impact and interact with existing functionality
+* Likely error modes (and how to handle them)
+* Corner cases (and how to handle them)
+
+While you do not need to prescribe a particular implementation - indeed, OTEPs should be about **behaviour**, not implementation! - it may be useful to provide at least one suggestion as to how the proposal *could* be implemented. This helps reassure reviewers that implementation is at least possible, and often helps them inspire them to think more deeply about trade-offs, alternatives, etc.
```
pyohannes, line 79: That seems like boilerplate that needs to be removed.

text/trace/0235-sampling-threshold-in-trace-state.md
yurishkuro: yes, I just want to see math, not the code

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,99 @@
...
+* Else the value corresponding to the `th` key should be interpreted as above.
+
+Sampling Decisions SHOULD be propagated by setting the value of the `th` key in the Tracestate header according to the above.
+
+## Changing T and R values
+
+The T value MAY be modified.
+
+In the case of a downstream sampler -- a tail sampler on the collection path that is attempting to reduce the volume of traffic -- the sampler MAY modify the `th` header by reducing its value.
+It MAY NOT increase it, as it is not possible to retroactively adjust the sampling probability upward.
```
yurishkuro, line 61: why does anything need to be adjusted upward?

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,99 @@
+# Sampling Threshold Propagation in TraceState
+
+## Motivation
+
+Sampling can theoretically take place at nearly any point in a distributed tracing system. If sampling is to be performed at multiple points in the process, the only way to reason about it effectively is to make sure that the sampling decisions are **consistent**.
```
kalyanaj, line 5: In this statement "If sampling is to be performed at multiple points in the process", I feel there's a bit of ambiguity on if it is talking about sampling of the same span or sampling of different spans in a trace. It may be good to take the "Consistent probability sampling" motivation/definitions from the earlier experimental OTEP.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,99 @@
...
+To distinguish the cases, this proposal uses the key `rv`.
+
+In order to make consistent sampling decisions across the entire path of the trace, two values SHOULD be propagated with the trace:
+
+1. A _random_ (or pseudo-random) 56-bit value, called `R` below.
+2. A 56-bit trace _threshold_ as expressed in the TraceState, called `T` below.
+
+The sampling decision is propagated with the following algorithm:
+
+* If the `th` key is not specified, Always Sample.
```
kalyanaj, line 26: I didn't quite get the "Always Sample" decision here: if the 'th' key is not specified, it means that the parent/caller is not using consistent probability sampling, correct?  To still achieve some consistent sampling, couldn't the callee/child look at the sampled flag (i.e., parent-based sampling)?

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,99 @@
...
+* Else derive `T` by parsing the `th` key as a hex value as described below.
+* If `T` is 0 and the _sampled_ flag is set, Always Sample. This implies that non-probabilistic sampling is taking place.
+* Compare the 56 bits of `T` with the 56 bits of `R`. If `T <= R`, then do not sample.
```
kalyanaj, line 29: Isn't 'T' encoding the sampling probability of the parent?  Since we are talking about consistent probability sampling here, the callee could have a different sampling probability, correct? If so, shouldn't it be comparing **_its_** sampling probability (expressed in a similar 56 bits) with the 56 bits of 'R' value to make the consistent sampling decision?  This is based on my understanding that 'th' is the replacement for p-value (and hence encodes the adjusted count of the parent, but doesn't reflect the callee's sampling probability).

oertl: Yes, this only describes how the parent's threshold is determined. What is missing is that each sampler is completely free to choose its own sampling probability, but can make its choice based on the parent's threshold. In this way, it is possible, for example, to implement parent-based sampling by setting the threshold equal to that of the parent.

text/trace/0235-sampling-threshold-in-trace-state.md
kentquirk: I would very much appreciate your suggestion on exactly what parts to use. I have read it a couple of times and I'm not finding any parts that seemed like they would fit the description. 

Will you please do one of these?
* Give me a list of specific points you think should be made, and I'll try to write them out
* Copy or write new text and I'll incorporate it

text/trace/0235-sampling-threshold-in-trace-state.md
kentquirk: I believe the intent is that the Always Sample flag with a missing `th` value specifically indicates "always sample". @PeterF778 @oertl do I have this correct?

text/trace/0235-sampling-threshold-in-trace-state.md
kentquirk: Please see the section on Algorithms below and let me know what you think.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,144 @@
+# Sampling Threshold Propagation in TraceState
+
+## Motivation
+
+Sampling can theoretically take place at nearly any point in a distributed tracing system. If sampling is to be performed at multiple points in the process, the only way to reason about it effectively is to make sure that the sampling decisions are **consistent**.
+In this context, consistency means that a positive sampling decision made at a particular point with probability p1 implies a positive sampling decision made at another point that samples a different piece of information from the same trace with probability p2 >= p1.
+
```
PeterF778, line 7: I suggest:
In this context, consistency means that a positive sampling decision made for a particular span with probability p1 implies a positive sampling decision for any span belonging to the same trace, if it is made with probability p2 >= p1.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,144 @@
...
+The T value MAY be modified.
+
+In the case of a downstream sampler -- a tail sampler on the collection path that is attempting to reduce the volume of traffic -- the sampler MAY modify the `th` header by reducing its value.
+It MUST NOT increase it, as it is not possible to retroactively adjust the sampling probability upward.
+
+A non-root head sampler MAY raise or lower the T value.
+This may be necessary in situations where a non-root sampler wishes to ensure that the remainder of the trace is sampled due to some exceptional occurrence.
+Note that changing the probability of a trace in flight introduces inconsistency and may cause the trace to be incomplete.
+
```
PeterF778, line 66: I suggest to reword the above as:
A consistent head sampler MUST set the T value corresponding to the sampling probability it actually uses. If it samples a non-root span, it MAY use the sampling probability of the parent span and use its T value.
Using different sampling probabilities for spans belonging to the same trace will lead to incomplete traces.

jmacd (approved):
Approving, assuming the feedback @oertl and @PeterF778 will be incorporated. This is great!

text/trace/0235-sampling-threshold-in-trace-state.md
jmacd: There is a parallel discussion https://github.com/open-telemetry/opentelemetry-proto/issues/507 and https://github.com/open-telemetry/opentelemetry-proto/pull/484. 

Summarizing, I believe the Span's trace-state incorporates changes made by the span, considering the parent's input. We are discussing in those issues whether to add a bit to say whether the parent was sampled or not, since the span's flags refer to itself. However, I would not propose to record the parent's tracestate with the span, in addition to the span's tracestate.

In these abstract formulas, it doesn't matter which is which. There will be samplers that use the parent's probability to set the child's probability, among other approaches. The specification will say how the span should modify the parent's trace state to set it's own, as was done in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md.

jmacd (approved):

jmacd:
An illustration may help, here is one I've been using.
OTel tracestate sampling v2 (6)

kentquirk:
I've added Josh's image to the document and fixed lint errors. Are there further changes needed?

text/trace/0235-sampling-threshold-in-trace-state.md
jmacd: Answer is, I think: 
It is illogical to adjust probability upward. It is never intentional, but a series of samplers can be misconfigured in a way that if this property were not checked, would allow misstating of sampling probability.

Under development in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/24811, I have two sampler modes implemented as simple tail samplers. They are "equalizing" and "proportional". An equalizing sampler could be configured w/ an absolute threshold (e.g., 50% of all spans) while individual SDKs could be configured w/ 10% head sampling. When the 50%-equalizing tail sampler sees a 10%-head-sampled span, it will always pass the span through. It **must not** adjust the probability to 50% for the 10% head-sampled span, that's what this line is supposed to say.

With a proportional sampler, the resulting probabilities will be 50% and 5%.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,151 @@
...
+The system has the following invariant:
+
+`(T=0) OR ((R < T) = sampled flag)`
+
+The sampling decision is propagated with the following algorithm:
+
+* If the `th` key is not specified, Always Sample.
+* Else derive `T` by parsing the `th` key as a hex value as described below.
+* If `T` is 0 and the _sampled_ flag is set, Always Sample. This implies that non-probabilistic sampling is taking place.
```
jpkrohling, line 32: Wouldn't it be better to have this as the first step?

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Visual
+
+![Sampling decision flow](../img/0235-sampling-threshold-calculation.png)
+
+## Algorithms
+
+The `th` and `rv` values may be represented and manipulated in a variety of forms depending on the capabilities of the processor and needs of the implementation. As 56-bit values, they are compatible with byte arrays and 64-bit integers, and can also be manipulated with 64-bit floating point with a truly negligible loss of precision.
+
+The following examples are in Python3. They are intended as examples only for clarity, and not as a suggested implementation.
```
jpkrohling, line 85: Do you have a suggested implementation in any other language that can serve as a reference implementation? I think a table at the bottom with some tests would also be useful for implementors.

jpkrohling (approved):
LGTM.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,151 @@
...
+
+* If the key `rv` is present in the Tracestate header, then `R = rv`.
+* Else if the Random Trace ID Flag is `true` in the traceparent header, then `R` is the lowest-order 56 bits of the trace-id.
+* Else `R` MUST be generated as a random value in the range `[0, (2**56)-1]` and added to the Tracestate header with key `rv`.
+
+The preferred way to propagate the `R` value is as the lowest 56 bits of the trace-id.
+If these bits are in fact random, the `random` trace-flag SHOULD be set as specified in [the W3C trace context specification](https://w3c.github.io/trace-context/#trace-id).
+There are circumstances where trace-id randomness is inadequate (for example, sampling a group of traces together); in these cases, an `rv` value is required.
+
+The value of the `rv` and `th` keys MUST be expressed as up to 14 hexadecimal characters from the set `[0-9a-f]`. Trailing zeros (but not leading zeros) may be omitted.
```
oertl, line 45: Trailing zeros of the `th` value can be omitted. 

I would not allow the same for the `rv` value. The reason is that parsing is easier if you can expect exactly 14 digits.  Since `rv` is a random value, only 1 out of 16 random values will have a trailing zero, so nothing is gained by this.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Explanation
+
+The existing, experimental [specification for probability sampling using TraceState](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md) is limited to powers-of-two probabilities, and is designed to work without making assumptions about TraceID randomness.
+This system can only achieve non-power-of-two sampling using interpolation between powers of two, which is unnecessarily restrictive.
+In existing sampling systems, sampling probabilities like 1%, 10%, and 75% are common, and it should be possible to express these without interpolation.
+There is also a need for consistent sampling in the collection path (outside of the head-sampling paths) and using inherent randomness in the traceID is a less-expensive solution than referencing a custom `r-value` from the tracestate in every span.
+This proposal introduces a new value with the key `th` as a replacement for the `p` value in the previous specification.
+The `p` value is limited to powers of two, while this proposal is not.
```
kalyanaj, line 15:
```suggestion
The `p` value is limited to powers of two, while the `th` value in this proposal supports a large range of values.
```

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,151 @@
...
+This allows for the continued expression of randomness using `r-value` as specified there using the key `r`.
```
kalyanaj, line 16:
```suggestion
This proposal allows for the continued expression of randomness using `r-value` as specified there using the key `r`.
```

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,151 @@
...
+If these bits are in fact random, the `random` trace-flag SHOULD be set as specified in [the W3C trace context specification](https://w3c.github.io/trace-context/#trace-id).
+There are circumstances where trace-id randomness is inadequate (for example, sampling a group of traces together); in these cases, an `rv` value is required.
+
+The value of the `rv` and `th` keys MUST be expressed as up to 14 hexadecimal characters from the set `[0-9a-f]`. Trailing zeros (but not leading zeros) may be omitted.
+
+Examples:
+`th` value is missing: Always Sample (probability = 100%). The AlwaysSample sampler in the OTel SDK should do this.
+`th=8` -- equivalent to `th=80000000000000`, which is 50% probability.
+`th=08` -- equivalent to `th=08000000000000`, which is 3.125% probability.
+`th=0` -- equivalent to `th=00000000000000`, which means Always Sample; this is outside of probabalistic sampling.
```
kalyanaj, line 51: Should we look at the sampled flag here when th is set to zero?

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,151 @@
...
+There is also a need for consistent sampling in the collection path (outside of the head-sampling paths) and using inherent randomness in the traceID is a less-expensive solution than referencing a custom `r-value` from the tracestate in every span.
+This proposal introduces a new value with the key `th` as a replacement for the `p` value in the previous specification.
+The `p` value is limited to powers of two, while this proposal is not.
+This allows for the continued expression of randomness using `r-value` as specified there using the key `r`.
+To distinguish the cases, this proposal uses the key `rv`.
+
+In order to make consistent sampling decisions across the entire path of the trace, two values SHOULD be propagated with the trace:
+
+1. A _random_ (or pseudo-random) 56-bit value, called `R` below.
+2. A 56-bit trace _threshold_ as expressed in the TraceState, called `T` below. `T` represents the minimum threshold that was applied in all previous consistent sampling stages. If the current sampling stage applies a lower threshold than any stage before, it has to update (decrease) the threshold correspondingly.
```
kalyanaj, line 22: I want to understand the th value propagation when multiple participants are involved. Let me take one small example involving four participants A, B, C, and D:

A -> B -> C -> D

Let's say A uses consistent probability sampling with a sampling rate of 0.25.
Let's say B uses consistent probability sampling with a sampling rate of 0.5.
Let's say C uses parent-based sampler.
Let's say D uses consistent probability sampling with a sampling rate of 0.1.

When A samples a span, its outgoing traceparent will have the 'sampled' flag SET and the 'th' in its outgoing tracestate will be set to 0x40 0000 0000 0000.
When A does not sample a span, its outgoing traceparent will have the 'sampled' flag UNSET but the 'th' in its outgoing tracestate will still be set to 0x40 0000 0000 0000.
When B samples a span, its outgoing traceparent will have the 'sampled' flag SET and the 'th' in its outgoing tracestate will be set to 0x80 0000 0000 0000.
C (being a parent based sampler) samples a span purely based on its parent (B in this case), it will use the sampled flag to make the decision. Its outgoing 'th' value will continue to reflect what it got from B (0x80 0000 0000 0000), and this is useful to understand its adjusted count.
etc.

Is my above understanding correct?  If so, in this case, depending on each stage's sampling probability, it sets its own outgoing 'th' value (to encode its sampling probability), correct? This has an example of 'th' value increasing in this flow.

text/trace/0235-sampling-threshold-in-trace-state.md
oertl: I don't think it's necessary to look at the sampled flag if the span is never captured when the sampled flag is unset.

text/trace/0235-sampling-threshold-in-trace-state.md
oertl: My understanding is that this is correct.

tedsuo (approved):
LGTM

spencerwilson (commented):
From the perspective of someone familiar with the OTel spec but not engaged in the week-to-week Sampling SIG discourse, I think parts of this proposal could be clarified if they were described in terms of nouns and verbs used in the OTel spec. I was often unsure which components of the system statements were referring to or constraining the behavior of.

I left plenty of suggestions, but please feel free to pick and choose whatever subset you all think makes sense! Ideally any code maintainers read this (or the spec that will follow) and don't have any trouble understanding what changes they must make in order to comply.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,151 @@
...
+This system can only achieve non-power-of-two sampling using interpolation between powers of two, which is unnecessarily restrictive.
+In existing sampling systems, sampling probabilities like 1%, 10%, and 75% are common, and it should be possible to express these without interpolation.
+There is also a need for consistent sampling in the collection path (outside of the head-sampling paths) and using inherent randomness in the traceID is a less-expensive solution than referencing a custom `r-value` from the tracestate in every span.
+This proposal introduces a new value with the key `th` as a replacement for the `p` value in the previous specification.
+The `p` value is limited to powers of two, while this proposal is not.
+This allows for the continued expression of randomness using `r-value` as specified there using the key `r`.
+To distinguish the cases, this proposal uses the key `rv`.
```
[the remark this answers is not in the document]
spencerwilson: I should add that I'm somewhat skeptical of this point:
> - **Efficiency.** Samplers need not generate new random bytes when they can obtain it directly from the low 56 bits of the input trace ID.

It sounds like the claim is that masking the low 56 bits out of the trace ID costs significantly fewer CPU cycles than computing the next value from a pseudo-RNG. I'm skeptical that's true. Is there another angle to efficiency I'm missing, or should we just strike this and say that fundamentally the benefit is a simpler implementation?

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+### Converting threshold to an adjusted count (sampling rate)
+
+The adjusted count is an integer value, indicating the quantity of items from the population that this sample represents. It is 1/probability converted to an integer.
+
+```py
+maxth = 0x100_0000_0000_0000  # 2^56
+adjcount = int((maxth / float(threshold)) + 0.5)
+```
```
spencerwilson, line 131: I believe https://docs.python.org/3/library/math.html#math.floor and [`int(float)`](https://docs.python.org/3/library/stdtypes.html#typesnumeric:~:text=Conversion%20from%20float%20to%20int%20truncates%2C%20discarding%20the%20fractional%20part) are equivalent when their inputs are nonnegative. Given that the latter is somewhat obscure Python behavior, maybe this example would be more understandable to more programmers if written using `floor`?

Also it might be good to emphasize that this is a lossy mapping, an approximation.

```suggestion
The adjusted count is an integer value, indicating the approximate quantity of items from the population that this sample represents. It is 1/probability, [rounded half-up](https://en.wikipedia.org/wiki/Rounding#Rounding_half_up) to the nearest integer. It is not defined for spans that were obtained via non-probabilistic sampling (a sampled span with integer threshold = 0).

```py
maxth = 0x100_0000_0000_0000  # 2^56
adjcount = math.floor((maxth / float(threshold)) + 0.5)
```
```

File: text/trace/0235-sampling-threshold-in-trace-state.md (hunk 28 above)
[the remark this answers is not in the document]
Comment (oertl) [reply text/trace/0235-sampling-threshold-in-trace-state.md:17]: One could argue that it is more efficient if no additional field is needed to propagate the random number, as required for consistent sampling decisions.

File: text/trace/0235-sampling-threshold-in-trace-state.md (hunk 21 above)
Comment (kentquirk) [reply text/trace/0235-sampling-threshold-in-trace-state.md:32]: You first have to determine T before you can use it. Or am I misunderstanding you?

File: text/trace/0235-sampling-threshold-in-trace-state.md (hunk 22 above)
Comment (kentquirk) [reply text/trace/0235-sampling-threshold-in-trace-state.md:85]: We had several sample implementations in several languages during our internal discussions, but I felt that kind of thing was out of scope for a document like this. There were several examples cited for each of our proposals in https://github.com/open-telemetry/opentelemetry-specification/issues/3602. 

I don't object to having such things here, but I'm hesitant to spend another long stint of debating about implementation details before we can commit this and get to real implementations.

File: text/trace/0235-sampling-threshold-in-trace-state.md (hunk 27 above)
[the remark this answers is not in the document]
Comment (PeterF778) [reply text/trace/0235-sampling-threshold-in-trace-state.md:22]: The phrase "minimum threshold that was applied in all previous consistent sampling stages" can be confusing. The reader might get an impression that the set of the past threshold values is somehow available.

File: text/trace/0235-sampling-threshold-in-trace-state.md [hunk 30]
```diff
@@ -0,0 +1,151 @@
...
+To distinguish the cases, this proposal uses the key `rv`.
+
+In order to make consistent sampling decisions across the entire path of the trace, two values SHOULD be propagated with the trace:
+
+1. A _random_ (or pseudo-random) 56-bit value, called `R` below.
+2. A 56-bit trace _threshold_ as expressed in the TraceState, called `T` below. `T` represents the minimum threshold that was applied in all previous consistent sampling stages. If the current sampling stage applies a lower threshold than any stage before, it has to update (decrease) the threshold correspondingly.
+
+The system has the following invariant:
+
+`(T=0) OR ((R < T) = sampled flag)`
```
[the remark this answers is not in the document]
PeterF778: I believe we need to state explicitly that `sampled` flag and the `T` value in the trace state document the last decision wrt the span.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,151 @@
...
+
+The sampling decision is propagated with the following algorithm:
+
+* If the `th` key is not specified, Always Sample.
+* Else derive `T` by parsing the `th` key as a hex value as described below.
+* If `T` is 0 and the _sampled_ flag is set, Always Sample. This implies that non-probabilistic sampling is taking place.
+* Compare the 56 bits of `T` with the 56 bits of `R`. If `T <= R`, then do not sample.
```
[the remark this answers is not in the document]
PeterF778: "Always Sample" sounds prescriptive, but actually what it means in this context is "the last sampling decision for this span (or for the parent span if head-sampling) was made according to the AlwaysSample policy".

text/trace/0235-sampling-threshold-in-trace-state.md
kentquirk: I have attempted to integrate these suggestions. Please check.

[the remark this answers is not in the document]
kentquirk: I don't see the problem, personally. Can you suggest an alternative wording that expresses it better?

text/trace/0235-sampling-threshold-in-trace-state.md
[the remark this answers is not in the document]
kentquirk: I am not sure what to change based on this feedback.

text/trace/0235-sampling-threshold-in-trace-state.md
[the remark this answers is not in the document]
kentquirk: Does something need to change here?

text/trace/0235-sampling-threshold-in-trace-state.md
kentquirk: The Sampling SIG seems to agree that we don't need a suggested implementation in an OTEP.

text/trace/0235-sampling-threshold-in-trace-state.md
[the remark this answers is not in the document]
kentquirk: Reword: Samplers would not have to propagate different 56-bit random values when they can obtain it directly from the trace ID.

text/trace/0235-sampling-threshold-in-trace-state.md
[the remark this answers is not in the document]
kentquirk: There is a distinction between:
* Th missing and sample flag is set: means that the upstream sampling decision was 'yes' [assume 100% probability] - adjusted count is 1.
* th zero and sample flag is set: nonprobabalistic decision; adjusted count is 0
* th missing or zero, sample flag not set: upstream sampling decision was 'no'

If sample flag is not set, th might document the sample decision, but otherwise should be ignored.

(add note re th missing means 100% since it can't be represented)


NOTE -- continue conversation in Slack

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,151 @@
...
+
+The T value MAY be modified.
+
+In the case of a downstream sampler -- a tail sampler on the collection path that is attempting to reduce the volume of traffic -- the sampler MAY modify the `th` header by reducing its value.
+It MUST NOT increase it, as it is not possible to retroactively adjust the sampling probability upward.
+
+A consistent head sampler MUST set the T value corresponding to the sampling probability it actually uses. If it samples a non-root span, it MAY use the sampling probability of the parent span and use its T value.
+Using different sampling probabilities for spans belonging to the same trace will lead to incomplete traces.
+
+A sampler MUST introduce an R value to a trace that does not include one and does not have the `Random` trace-id flag set. It MUST use the `rv` key for this purpose. A sampler MUST NOT modify an existing R value or trace-id.
```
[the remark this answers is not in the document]
kentquirk: Better distinction: span origination vs making a decision on an existing span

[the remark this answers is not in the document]
spencerwilson: 👍 Updated the suggestions to use "downstream sampler" since "tail sampler" tends to refer to a special class of downstream sampler.

text/trace/0235-sampling-threshold-in-trace-state.md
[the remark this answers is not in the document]
spencerwilson: @kentquirk Not sure if you're addressing me or Peter. In case me: IMO it's unclear what "the system" refers to. My suggested edit attempts to be more precise by using API spec terminology. My current understanding is that L45 can be viewed as a predicate that's always true for any `Span` in any process's memory following the creation via a Tracer.

carlosalberto (approved):

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,169 @@
...
+In existing sampling systems, sampling probabilities like 1%, 10%, and 75% are common, and it should be possible to express these without interpolation.
+There is also a need for consistent sampling in the collection path (outside of the head-sampling paths) and using inherent randomness in the traceID is a less-expensive solution than referencing a custom `r-value` from the tracestate in every span.
+This proposal introduces a new value with the key `th` as a replacement for the `p` value in the previous specification.
+The `p` value is limited to powers of two, while the `th` value in this proposal supports a large range of values.
+This proposal allows for the continued expression of randomness using `r-value` as specified there using the key `r`.
+To distinguish the cases, this proposal uses the key `rv`.
+
+In order to make consistent sampling decisions across the entire path of the trace, two values SHOULD be propagated with the trace:
+
+1. A _random_ (or pseudo-random) 56-bit value, called `R` below.
```
PeterF778, line 21: We have picked 56 bits because that's how many random bits we expect in trace-id. However, I think that using 52-bit _randomness_ and _threshold_ is worth considering, with the main reason being that this would better match the precision of the floating point double arithmetic.
When calculating the _adjusted count_ it is hard to avoid floating point arithmetics. Yet, with 56-bit threshold we will get situations where _different_ threshold values T1 and T2 will result in the same adjusted count (without rounding to the nearest integer). This means that the last 4 digits of the 56-bit threshold is almost like noise.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,169 @@
...
+
+```py
+# embedded _ in numbers for clarity (permitted by Python3)
+maxth = 0x100_0000_0000_0000  # 2^56
+prob = float(threshold) / maxth
+```
+
+### Converting threshold to an adjusted count (sampling rate)
+
+The adjusted count is an integer value, indicating the approximate quantity of items from the population that this sample represents. It is 1/probability, [rounded half-up](https://en.wikipedia.org/wiki/Rounding#Rounding_half_up) to the nearest integer. It is not defined for spans that were obtained via non-probabilistic sampling (a sampled span with integer threshold = 0).
```
PeterF778, line 145: I do not thing we should mandate forcing the adjusted count to be an integer. Among motivations listed for this proposal is hypothetical user desire to sample with 75%. Using integer adjusted counts for this case would lead to high metrics errors.

jmacd:
For the record, this is not a stale OTEP. The Sampling SIG has met and discussed the merits of several alternatives and we expect this process to continue. We have some amount of uncertainty about how to reflect unknown sampling probability, zero sampling probability, and 100% sampling probability.

jmacd:
@kentquirk I'm ready to go with the rejection threshold approach.

Rejection threshold pros:
- A single tracestate field `tv:0` conveys 100% sampling with nothing more; max tracestate value is 14 hex bytes
Rejection threshold cons:
- Probabilities in the 0x1p-053 to 0x1p-56 create minor annoyance (but easily avoided for users)

Acceptance threshold pros:
- T-value reads as a hexadecimal fraction equal to sampling probability
Acceptance threshold cons:
- Extra tracestate needed for 100% sampling.

I still find myself confused how to reason about a span with no t-value. This can result from a legacy always-on sampler in the SDK. This can also result from a `sampling.priority` attribute causing sampling when the consistent sampler would not, in which case again no t-value. Is there some attribute we can / should use to indicate the difference between a span that was always-on from an old SDK and a span that was rejected by a consistent sampler and nevertheless exported. In last week's SIG, we discussed using attributes, not tracestate, to convey such information. For example, when a span is selected by a periodic, non-probabilistic process (e.g., one-per-minute) and rejected by a consistent sampler, how should we avoid counting it again as one span from a legacy SDK w/ no t-value, later in the pipeline? I'm willing to say this doesn't really matter. Each application can choose how it wants to treat spans with no t-value.

kentquirk:
Result of 7Dec23 Sampling SIG meeting: we have decided to convert this to a rejection threshold and I will make those edits this week.

kentquirk:
I have attempted to update the document based on a rejection threshold model. This is really breaking my brain. It solves one problem but I'm finding it extremely hard to think about.

I am not at all convinced that the example beginning at line 24 is correct (and maybe we should even delete it, as I don't find it helpful), and I would very much appreciate if people would carefully think about whether I correctly fixed all the places where we had <, >, increase, decrease, etc.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+### Converting threshold to a sampling probability
+
+The sampling probability is a value from 0.0 to 1.0, which can be calculated using floating point by dividing by 2^56:
+
+```py
+# embedded _ in numbers for clarity (permitted by Python3)
+maxth = 0x100_0000_0000_0000  # 2^56
+prob = float(threshold) / maxth
```
PeterF778, line 161: With rejection threshold, the formula should be `float(maxth - threshold) / maxth`

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+### Converting threshold to an adjusted count (sampling rate)
+
+The adjusted count indicates the approximate quantity of items from the population that this sample represents. It is equal to `1/probability`. It is not defined for spans that were obtained via non-probabilistic sampling (a sampled span with integer threshold = 0).
```
PeterF778, line 166: It is not defined for spans that were obtained via non-probabilistic sampling (a sampled span with no `th` value).

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,186 @@
...
+
+The system has the following invariant:
+
+`(R >= T) = sampled flag)`
+
+The sampling decision is propagated with the following algorithm:
+
+* If the `th` key is not specified, Always Sample.
+* Else derive `T` by parsing the `th` key as a hex value as described below.
+* If `T` is 0, Always Sample. This implies that non-probabilistic sampling is taking place.
```
PeterF778, line 51: I believe our intention was that missing `th` key implies non-probabilistic sampling. `T` equal to zero means 100% (probabilistic) sampling.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,186 @@
...
+* Compare the 56 bits of `T` with the 56 bits of `R`. If `T >= R`, then do not sample.
```
PeterF778, line 52: If `T > R` then do not sample.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,186 @@
...
+* Else `R` MUST be generated as a random value in the range `[0, (2**56)-1]` and added to the Tracestate header with key `rv`.
+
+The preferred way to propagate the `R` value is as the lowest 56 bits of the trace-id.
+If these bits are in fact random, the `random` trace-flag SHOULD be set as specified in [the W3C trace context specification](https://w3c.github.io/trace-context/#trace-id).
+There are circumstances where trace-id randomness is inadequate (for example, sampling a group of traces together); in these cases, an `rv` value is required.
+
+The value of the `rv` and `th` keys MUST be expressed as up to 14 hexadecimal digits from the set `[0-9a-f]`. For `th` keys only, trailing zeros (but not leading zeros) may be omitted. `rv` keys MUST always be exactly 14 hex digits.
+
+Examples:
+`th` value is missing: Always Sample (probability = 100%). The AlwaysOn sampler in the OTel SDK should do this.
```
PeterF778, line 67: Missing `th` value indicates non-probabilistic sampling decision.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,186 @@
...
+`th=4` -- equivalent to `th=40000000000000`, which is a 25% rejection threshold, corresponding to a 75% sampling probability.
+`th=c` -- equivalent to `th=c0000000000000`, which is a rejection threshold of 75%, corresponding to a sampling probability of 25%.
+`th=08` -- equivalent to `th=08000000000000`, which is a rejection threshold of 3.125%, corresponding to a sampling probability of 96.875%.
+`th=0` -- equivalent to `th=00000000000000`, which is a 0% rejection threshold, which means Always Sample.
+
+The `T` value MUST be derived as follows:
+
+* If the `th` key is not present in the Tracestate header, then `T` is effectively 0.
```
PeterF778, line 75: No, missing `th` implies that the adjusted count is undefined, so we cannot treat is the same way as `T` equal to zero. However, some tools MAY assume that missing `th` is like what you described, but such interpretation should not be mndatory.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Algorithms
+
+The `th` and `rv` values may be represented and manipulated in a variety of forms depending on the capabilities of the processor and needs of the implementation. As 56-bit values, they are compatible with byte arrays and 64-bit integers, and can also be manipulated with 64-bit floating point with a truly negligible loss of precision.
+
+The following examples are in Python3. They are intended as examples only for clarity, and not as a suggested implementation.
+
+### Converting t-value to a 56-bit integer threshold
+
+To convert a t-value string to a 56-bit integer threshold, pad it on the right with 0s so that it is 14 digits in length, and then parse it as a hexadecimal value.
```
PeterF778, line 129: For consistency, it might be better to use `T` and `R` in the following examples, as these symbols were already introduced before. So `T` instead of `threshold`, etc.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,186 @@
...
+
+```py
+h = hex(tvalue).rstrip('0')
+# remove leading 0x
+tv = 'tv='+h[2:]
+```
+
+### Testing rv vs threshold
+
+Given rv and threshold as 64-bit integers, a sample should be taken if rv is strictly less than the threshold.
```
PeterF778, line 148: Replace "strictly less" with "greater or equal".

text/trace/0235-sampling-threshold-in-trace-state.md
kentquirk: I see your point but I think it's clearer with the longer names

kentquirk:
I believe that this is now (finally!) ready to merge.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,189 @@
...
+When B samples a span, its outgoing traceparent will have the 'sampled' flag SET and the 'th' in its outgoing tracestate will be set to 0x80_0000_0000_0000.
+C (being a parent based sampler) samples a span purely based on its parent (B in this case), it will use the sampled flag to make the decision. Its outgoing 'th' value will continue to reflect what it got from B (0x80_0000_0000_0000), and this is useful to understand its adjusted count.
+
+This design requires that as a given span progresses along its collection path, `th` is non-decreasing (and, in particular, must be increased at stages that apply lower sampling probabilities).
+It does not, however, restrict a span's initial `th` in any way (e.g., relating it to that of its parent, if it has one).
+It is acceptable for B to have a lesser initial `th` than A has. It would not be ok if some later-stage sampler decreased A's `th`.
+
+The system has the following invariant:
+
+`(R >= T) = sampled flag)`
```
oertl, line 47:
```suggestion
`(R >= T) = sampled flag`
```

oertl (approved):

text/img/0235-sampling-threshold-calculation.png
jmacd, line 1: I updated the original drawing with minor edits. I noticed that the "also represented as" section was confusing and propose we remove it entirely. I also fixed one minor typo (decision still has a 0x8 in it, not 0xc). I can't write to your branch, or I'd just push it, but if you download from the drawing it's fixed there.

jmacd (approved):

jmacd:
@yurishkuro will you please review and consider approving? We have reached agreement in the Sampling SIG and have begun to press for approvals. Thank you.

text/img/0235-sampling-threshold-calculation.png
kentquirk: Thank you! I've updated the image from the source.

jmacd:
@bogdandrutu will you please review and consider approving? We have reached agreement in the Sampling SIG and have begun to press for approvals. Thank you.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,189 @@
...
+The `p` value is limited to powers of two, while the `th` value in this proposal supports a large range of values.
+This proposal allows for the continued expression of randomness using `r-value` as specified there using the key `r`.
+To distinguish the cases, this proposal uses the key `rv`.
+
+In the general case, in order to make consistent sampling decisions across the entire path of the trace, two values MUST be present in the `SpanContext`:
+
+1. A _random_ (or pseudo-random) 56-bit value, called `R` below.
+2. A 56-bit _rejection threshold_ (or just "threshold") as expressed in the TraceState, called `T` below. `T` represents the maximum threshold that was applied in all previous consistent sampling stages. If the current sampling stage applies a greater-valued threshold than any stage before, it MUST update (increase) the threshold correspondingly.
+
+One way to think about _rejection threshold_ is that is the number of spans that would be discarded out of 2^56 considered spans.
```
arminru, line 24: The concept/algorithm is only established with the invariant defined on line 47. At this point I think it's not clear where this statement is coming from. Maybe the invariant below could also be included in the definition of the rejection threshold above (e.g. by adding "spans with `R => T` will be sampled").

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,189 @@
...
+where -> indicates a parent -> child relationship.
+
+`A` uses consistent probability sampling with a sampling probability of 0.25 (this corresponds to a rejection probability of .75).
+`B` uses consistent probability sampling with a sampling probability of 0.5.
+`C` uses a parent-based sampler.
+
+When `A` samples a span, its outgoing traceparent will have the 'sampled' flag SET and the 'th' in its outgoing tracestate will be set to `0xc0_0000_0000_0000`.
+When `A` does not sample a span, its outgoing traceparent will have the 'sampled' flag UNSET but the 'th' in its outgoing tracestate will still be set to 0xc0_0000_0000_0000.
+When B samples a span, its outgoing traceparent will have the 'sampled' flag SET and the 'th' in its outgoing tracestate will be set to 0x80_0000_0000_0000.
+C (being a parent based sampler) samples a span purely based on its parent (B in this case), it will use the sampled flag to make the decision. Its outgoing 'th' value will continue to reflect what it got from B (0x80_0000_0000_0000), and this is useful to understand its adjusted count.
```
arminru, line 39:
```suggestion
When `A` does not sample a span, its outgoing traceparent will have the 'sampled' flag UNSET but the 'th' in its outgoing tracestate will still be set to `0xc0_0000_0000_0000`.
When B samples a span, its outgoing traceparent will have the 'sampled' flag SET and the 'th' in its outgoing tracestate will be set to `0x80_0000_0000_0000`.
C (being a parent based sampler) samples a span purely based on its parent (B in this case), it will use the sampled flag to make the decision. Its outgoing 'th' value will continue to reflect what it got from B (`0x80_0000_0000_0000`), and this is useful to understand its adjusted count.
```
nits - for consistency with the line above

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,189 @@
...
+There are two categories of sampler:
+
+- **Head samplers:** Implementations of [`Sampler`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.25.0/specification/trace/sdk.md#sampler), called by a `Tracer` during span creation.
+- **Downstream samplers:** Any component that, given an ended Span, decides whether to drop or forward ("sample") it on to the next component in the system. Also known as "collection-path samplers" or "sampling processors". _Tail samplers_ are a special class of downstream samplers that buffer the spans in a trace and select a sampling probability for the trace as a whole using data from any span in the buffered trace.
+
+This section defines behavior for each kind of sampler.
+
+### Head samplers
+
+A head sampler is responsible for computing a new span's initial [`TraceState`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.25.0/specification/trace/api.md#tracestate) (among other data). Notable inputs to that computation include the parent span's trace state (if a parent span exists) and the new span's trace ID.
```
arminru, line 94:
```suggestion
A head sampler is responsible for computing a new span's initial [`TraceState`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.29.0/specification/trace/api.md#tracestate) (among other data). Notable inputs to that computation include the parent span's trace state (if a parent span exists) and the new span's trace ID.
```

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,189 @@
...
+* If the `th` key is not present in the Tracestate header, then non-probabalistic sampling may be in use.
+* Else the value corresponding to the `th` key should be interpreted as above.
+
+Sampling Decisions MUST be propagated by setting the value of the `th` key in the Tracestate header according to the above.
+
+## Initializing and updating T and R values
+
+There are two categories of sampler:
+
+- **Head samplers:** Implementations of [`Sampler`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.25.0/specification/trace/sdk.md#sampler), called by a `Tracer` during span creation.
```
arminru, line 87:
```suggestion
- **Head samplers:** Implementations of [`Sampler`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.29.0/specification/trace/sdk.md#sampler), called by a `Tracer` during span creation.
```
No changes since then.

text/trace/0235-sampling-threshold-in-trace-state.md
arminru, line 94: Is the sampler responsible for the entire TraceState, or just `rv` and `th`?

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,189 @@
...
+
+This section defines behavior for each kind of sampler.
+
+### Head samplers
+
+A head sampler is responsible for computing a new span's initial [`TraceState`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.25.0/specification/trace/api.md#tracestate) (among other data). Notable inputs to that computation include the parent span's trace state (if a parent span exists) and the new span's trace ID.
+
+First, a consistent `Sampler` decides which sampling probability to use. The sampler MAY select any value of T. If a valid `SpanContext` is provided in the call to `ShouldSample` (indicating that the span being created will be a child span),
+
+- Choosing a T greater than the parent span's is expected to result in partial traces (the parent may be sampled but the child dropped).
```
arminru, line 98:
```suggestion
- Choosing a T greater than the parent span's is expected to result in partial traces (the parent may be sampled but its child, the current span, dropped).
```
This could make it clearer.

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,189 @@
...
+- Choosing a T less than or equal to the parent span is expected to result in complete traces (this is definition of consistent probability sampling).
+
+For the output TraceState,
+
+- The `th` key MUST be defined, with value describing the sampling probability the sampler actually used.
```
arminru, line 103:
```suggestion
- The `th` key MUST be defined with a value corresponding to the sampling probability the sampler actually used.
```

text/trace/0235-sampling-threshold-in-trace-state.md
```diff
@@ -0,0 +1,189 @@
...
+A downstream sampler, in contrast, may output a given ended Span with a _modified_ trace state, complying with following rules:
+
+- If the chosen sampling probability is 1, the sampler MUST NOT modify any existing `th`, nor set any `th`.
+- Otherwise, the chosen sampling probability is in `(0, 1)`. In this case the sampler MUST output the span with a `th` equal to `max(input th, chosen th)`. In other words, `th` MUST NOT be decreased (as it is not possible to retroactively adjust an earlier stage's sampling probability), and it MUST be increased if a lower sampling probability was used. This case represents the common case where a downstream sampler is reducing span throughput in the system.
+
+## Internal details
+
+The trace state header SHOULD contain a field with the key `rv`, and a value that corresponds to a 56-bit sampling threshold.
+This value will be compared to the 56-bit random value associated with the trace.
+
```
arminru, line 119: This section might be obsolete by now as this is covered above.

```suggestion
```

arminru (approved):

jmacd:
Thank you @kentquirk!

PR #238, opened 2023-10-31
Semantic conventions for telemetry pipeline monitoring

jmacd:
This is a continuation of https://github.com/open-telemetry/semantic-conventions/pull/184, which developed into a longer discussion that originally planned. 

This text includes results from auditing the OpenTelemetry Collector and makes recommendations consistent with existing practice that extend our ability to configure basic, normal, and detailed-level metrics about OpenTelemetry data pipelines.


text/metrics/0000-pipeline-monitoring.md
```diff
@@ -0,0 +1,347 @@
...
+2. No other outcomes are possible.
+
+There is a potential for pipeline metrics to be redundant, as
+described in these assertions.  In a pipeline with no fan-in or
+fan-out, each stage processes as many items as the stage before it
+did, minus the number of items dropped.
+
+#### Pipeline stage-name uniqueness
+
+The Pipeline Stage Name Uniqueness requirement developed here avoids
```
dashpole, line 184: This requirement doesn't seem necessary. A single metric name often counts the same thing from the perspective of multiple entities (e.g. faas.invocations will overcount if you have faas instances that talk to other faas instances).  Isn't resource enough to distinguish multiple chained sdk + collector(s)?

text/metrics/0000-pipeline-monitoring.md
```diff
@@ -0,0 +1,347 @@
...
+The relationship between items received, dropped, and exported is
+shown in the following diagram.
+
+![pipeline monitoring metrics](../images/otel-pipeline-monitoring.png)
+
+### Proposed metrics semantic conventions
+
+The proposed metric names are: 
+
+`otel.{station}.received`: Inclusive count of items entering the pipeline at a station.
```
dashpole, line 294: Apologies if this has been discussed before. Having wildcards in the name seems like a bad precedent to set, and makes constructing dashboards, etc. more difficult. It seems OK to have two versions of this, one for SDK and one for collector, but labels seem like a much better solution if this can actually be any string value.

text/metrics/0000-pipeline-monitoring.md
```diff
@@ -0,0 +1,347 @@
...
+
+There is one additional dimension that users may wish to opt-in to, in
+order to gain information about failures at a particular pipeline
+stage.  When detail-level metrics are requested, all three metric
+instruments specified for pipeline monitoring gain an additional
+`reason` attribute, with a short string explaining the failure.  
+
+For example, with detailed-level metrics in use, the
+`otel.{station}.received` and `otel.{station}.exported` counters will
+include additional `reason` information (e.g., `timeout`,
```
dashpole, line 332: I've been thinking about generally how domain-specific export metrics should relate to protocol-specific metrics.  otelhttp and otelgrpc should, in theory, be the canonical way to get a reason for receive/export failures, but that means looking at multiple metrics, and knowing which protocol the exporter uses. Would it be better for otelhttp/otelgrpc telemetry to be the "detailed" mechanism for finding the reason for failures?

text/metrics/0000-pipeline-monitoring.md
```diff
@@ -0,0 +1,347 @@
...
+
+Components are uniquely identified using a descriptive `name`
+attribute which encompasses at least a short name describing the type
+of component being used (e.g., `batch` for the SDK BatchSpanProcessor
+or the Collector batch proessor).
+
+When there is more than one component of a given type active in a
+pipeline having the same `domain` and `signal` attributes, the `name`
+should include additional information to disambiguate the multiple
+instances using the syntax `<type>/<instance>`.  For example, if there
```
djaglowski, line 278: In the collector, this is not necessarily enough to differentiate all components in a pipeline because we may have different components with the same type. e.g. `otlp` receiver and `otlp` exporter in the same pipeline.

Should this be addressed here, or if the problem is unique to the collector, should the collector adopt a strategy to incorporate the class into the type?

text/metrics/0000-pipeline-monitoring.md
jmacd: @dashpole thanks for an early review -- it was still a pretty rough draft that you saw but I've just updated it, and the detail you flagged here doesn't change. 

I am specifically concerned about wanting to automatically infer the health of the pipeline from all the pipeline-monitoring metrics. I'm worried about the 3-level layout that could be standard, with sdk sending to agent sending to gateway collectors.  If the agent and the gateway count items using the same metric name and I enter a metrics system with a query like `rate(otel.collector.received)` and compare it against `rate(otel.sdk.received)`, the number of collector items will be exactly double the number of SDK items. The resources would distinguish which was which, but not in a way that I can automatically apply to avoid this "incorrect" interpretation. I'm following the rule that it should be difficult to form meaningless aggregates. 

With that said, I don't want a dashboard problem and I don't expect more than 3 sets of 3 metric names, i.e., {sdk,agent,gateway} x {received,dropped,exported}. If a user were to configure a custom station name, they would just have to create a custom dashboard.

text/metrics/0000-pipeline-monitoring.md
jmacd: For the semantic-conventions repo, I would list three standard station names and give guidance on how to create more station names that follow the pattern. I don't think this will happen often, but I think it should be clear that automatic pipeline monitoring will break if ever   the same metric name counts items at more than one location in a pipeline.

(This uniqueness requirement is also the reason why I am not proposing to count the number of items that enter and exit each processor -- because then a chain of processors would require per-processor metrics for automatic monitoring to work. For this reason, it makes sense (as proposed here IMO) to count only the items dropped by processors.)

text/metrics/0000-pipeline-monitoring.md
jmacd: This is a top-of-mind question for me. I would definitely sacrifice detailed-level metrics from the pipeline-monitoring logic in favor of transport-specific metrics, especially when a user is not mixing transports. I am not sure this argument works quite as well for collectors.

I worry that the details are quite a bit different too. When I use gRPC-specific instrumentation, I can't be sure that the pipeline-monitoring equations stated in this document will work (e.g., how are retries counted? do those specifications exist yet?)--they could help diagnose different sorts of problem, but the metrics specific to a transport will not easily integrate with the pipeline-monitoring logic outlined here.

text/metrics/0000-pipeline-monitoring.md
jmacd: Good question. Since I've heavily updated this document and plan to present it in tomorrow's Collector SIG, I'm going to leave it open. As designed, it should be clear that `received` metrics are from receivers and `exported` metrics are from exporters. I want to say that everything else is a processor, but I realize this is a thorny question.

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,426 @@
...
+| `otel.sdk.dropped`  | Normal           |
+| `otel.sdk.exported` | Detailed         |
+
+#### Collector default configuration
+
+| Metric name               | Enablement level |
+|---------------------------|------------------|
+| `otel.{station}.received` | Basic            |
+| `otel.{station}.dropped`  | Basic            |
+| `otel.{station}.exported` | Basic            |
```
dmitryax, line 324: How would the `{station}` be set? A collector instance doesn't have that information. Will it be a mandatory configuration option?

What if we keep `otel.collector.(received|dropped|exported)` metric names and have another optional resource attribute to identify a station?

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,426 @@
...
+As stated above, SDKs SHOULD support configuring an alternate Meter
+Provider for pipeline-monitoring metrics.  When the global Meter
+Provider is used, the Metrics SDK's pipeline will receive its own
+pipeline-monitoring metrics.  When a custom Meter Provider is used, a
+secondary pipeline will receive the pipeline monitoring metrics, in
+which case the secondary pipeline may also self-report for itself.
+
+## Trade-offs and mitigations
+
+The use of three-levels of metric detail may seem like more freedom
```
dashpole, line 410: I buy the user experience you want to provide, but I wonder if creating a solution just for the sdk/collector is the correct scope.  For example, someone from gRPC recently asked how to provide a small set of metrics by default, but allow enabling more detailed metrics.  Should this functionality be part of the metrics API/SDK?  Otherwise, I suspect many instrumentation libraries will build their own, slightly different solutions to this problem.  I'm also not a huge fan of having two places (views and detail) to configure which attributes to keep/drop.

mtwo:
@jack-berg @carlosalberto can we mark this as triaged with priority p1?

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,426 @@
...
+observable to the service provider, as long as all original producers
+report pipeline-monitoring metrics.  For all telemetry producers with
+basic-level pipeline monitoring enabled, the telemetry system will be
+able to compare the actual number of OpenTelemetry spans, metric data
+points, and log records received against the numbers that entered the
+system's own pipelines.  In aggregate, the number of items entering
+the pipeline should match the number of items successfully received,
+otherwise the system is capable of reporting the combined losses to
+the user.
+
```
carlosalberto, line 355: This is a vital paragraph to understand why the default, different levels of detail for each station. Putting a end-to-end example, as you did with https://github.com/jmacd/oteps/blob/jmacd/drops/text/metrics/0238-pipeline-monitoring.md#pipeline-monitoring-diagram would work great here (albeit with numbers/metrics instead of charts)

djaglowski (commented):
To summarize my other comments and provide a suggestion:
- Per component metrics are valuable and if anything should be enhanced.
- The notion of diffing values between stations does not seem to allow sufficient accuracy in describing the way data may actually flow through a collector because it assumes linearity of data flow.

What seems like a better solution would be to add metrics which directly describe the aggregate behavior of the processors in each of the collector's pipelines.
1. The total number of items flowing into the first processor. 
2. The total number of items added by the processors.
3. The total number of items dropped by the processors.
4. (Implied by 1-3) The total number of items flowing out from the last processor.

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,426 @@
...
+4)](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/library-guidelines.md#requirements)
+describes a separation of protocol-dependent ("receivers",
+"exporters") and protocol-independent ("processors") parts.  We refer
+to the combination of parts as a station.
+
+The station concept is called out because within a station, we expect
+that the station (software) acts responsibly by design, for the
+integrity of the pipeline.  Stations allow data to enter a pipeline
+only through receiver components.  Stations are never responsible for
+dropping data, because only processor components drop data.  Stations
```
jack-berg, line 188: SDK processors do not drop data, and processors aren't responsible for passing data to the next process in a pipeline. Instead, higher order SDK code ensures that each registered processor is called. As the [spec puts it](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-processor), "Each processor registered on TracerProvider is a start of pipeline that consist of span processor and optional exporter".

The architectural differences between SDK and collector processor may impact your design. For example, further on you state:

> 1. Data that enters a station is eventually exported or dropped.
2. No other outcomes are possible.

For SDKs, its perfect valid for a processor to do nothing besides extract baggage and add it to the span. No filtering, no exporting.

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,426 @@
...
+of component being used (e.g., `batch` for the SDK BatchSpanProcessor
+or the Collector batch proessor).
+
+When there is more than one component of a given type active in a
+pipeline having the same `domain` and `signal` attributes, the `name`
+should include additional information to disambiguate the multiple
+instances using the syntax `<type>/<instance>`.  For example, if there
+were two `batch` processors in a collection pipeline (e.g., one for
+error spans and one for non-error spans) they might use the names
+`batch/error` and `batch/noerror`.
```
jack-berg, line 272: In the SDK, if two batch processors are present, they're only differentiable by the associated exporter. You can't configure the SDK to send some spans to one batch processor, and others to another. So maybe the name would have to be something like `batch/otlp`, `batch/zipkin`, etc.

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,426 @@
...
+These attributes can be applied to any of the pipeline monitoring
+metrics specified here.
+
+| Attributes     | Meaning                                     | Level of detail (Optional) | Examples                                                   |
+|----------------|---------------------------------------------|----------------------------|------------------------------------------------------------|
+| `otel.signal`  | Name of the telemetry signal                | Basic (Opt-out)            | `traces`, `logs`, `metrics`                                |
+| `otel.name`    | Type, name, or "type/name" of the component | Normal (Opt-out)           | `probabilitysampler`, `batch`, `otlp/grpc`                |
+| `otel.success` | Boolean: item considered success?           | Normal (Opt-out)           | `true`, `false`                                            |
+| `otel.reason`  | Explaination of success/failures.           | Detailed (Opt-in)          | `ok`, `timeout`, `permission_denied`, `resource_exhausted` |
+| `otel.scope`   | Name of instrumentation.                    | Detailed (Opt-in)          | `opentelemetry.io/library`                                 |
```
jack-berg, line 304: Is this intended to be different than the scope name? It seems like its the same. If so, then we can't opt-in or out of it. Maybe the prometheus exporter has options for that, but the scope isn't an optional part of the data model.

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,426 @@
...
+is different, because SDKs do not receive items from a pipeline and
+because they outnumber the other components.
+
+These attributes can be applied to any of the pipeline monitoring
+metrics specified here.
+
+| Attributes     | Meaning                                     | Level of detail (Optional) | Examples                                                   |
+|----------------|---------------------------------------------|----------------------------|------------------------------------------------------------|
+| `otel.signal`  | Name of the telemetry signal                | Basic (Opt-out)            | `traces`, `logs`, `metrics`                                |
+| `otel.name`    | Type, name, or "type/name" of the component | Normal (Opt-out)           | `probabilitysampler`, `batch`, `otlp/grpc`                |
```
jack-berg, line 301: `otel.component.name`?

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,426 @@
...
+`success=true`, but when a queue processor drops an item it may report
+`success=false`.
+
+#### SDK default configuration
+
+| Metric name         | Enablement level |
+|---------------------|------------------|
+| `otel.sdk.received` | Basic            |
+| `otel.sdk.dropped`  | Normal           |
+| `otel.sdk.exported` | Detailed         |
```
jack-berg, line 316: I'm struggling to see what metrics I would expect to get out of an SDK with different requirement levels. Would you be able to include an examples with a typical  SDK configuration? I'm imaging an SDK with logs, metrics, and traces enabled, each configured to export data via an OTLP exporter. What metrics and series do I see at Basic, Normal, and Detailed?

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,484 @@
...
+is supplied and is considered a closed set.  If these outcomes do not
+accurately explain the reason for a success or failure outcome, they
+SHOULD be extended by OpenTelemetry.
+
+For success:
+
+- `consumed`: Indicates a normal, synchronous request success case.
+  The item was consumed by the next stage of the pipeline, which
+  returned success.
+- `unsampled`: Indicates a successful drop case, due to sampling.
```
astencel-sumo, line 253: When the [Filter processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/filterprocessor/README.md) filters out data, should it count it as `unsampled`? The name doesn't sound right in this case.

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,484 @@
...
+has to query multiple metrics has to enter each metric name
+explicitly without help from the user interface.
+
+The proposed metric instrument would be named distinctly depending on
+whether it is a collector or an SDK, to prevent accidental aggregation
+of these timeseries.  The specified counter names would be:
+
+- `otelsdk.producer.items`: count of successful and failed items of
+  telemetry produced, by signal type, by an OpenTelemetry SDK.
+- `otelcol.receiver.items`: count of successful and failed items of
```
astencel-sumo, line 219: If I have a connector configured, I assume I get both the `otelcol.exporter.items` and `otelcol.receiver.items` metrics emitted for the connector, right?

Let's say I configured the [Count connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.92.0/connector/countconnector/README.md) on a traces pipeline, as described in the example in the component's README. The count connector then accepts traces on the `traces/in` pipeline and creates metrics on the `metrics/out` pipeline.

I imagine the `otelcol.exporter.items` metric for the count connector would count the incoming spans on the trace pipeline. What would be the `otel.outcome` for those correctly consumed spans? Would it be `consumed` or rather `unsampled`? These logs aren't shipped anywhere by the component, they are "swallowed" by the connector if I understand correctly.

I imagine the `otelcol.receiver.items` metric for the count connector would count the metrics created on the metrics pipeline, with the `otel.outcome` set to `consumed`.

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,484 @@
...
+
+### Error suppression behavior
+
+OpenTelemetry collector exporter components have existing error
+suppression behavior, optionally obtained through the `exporterhelper`
+library, which causes the `Consume()` function to return success for
+what would ordinarily count as failure.  This behavior makes automatic
+component health status reporting more difficult than necessary.
+
+One goal if this proposal is that Collector component health could be
```
astencel-sumo, line 313:
```suggestion
One goal of this proposal is that Collector component health could be
```

jmacd:
This is a work-in-progress. I will re-open it when it is ready for re-review.

text/metrics/0238-pipeline-monitoring.md
jmacd: The station concept was firmly rejected. This proposal is out of date, thanks.

text/metrics/0238-pipeline-monitoring.md
jmacd: The concept of additional detail seems to be unwelcome. I will remove before re-opening.

jmacd:
(I am incorporating all the feedback received so far. Thank you, reviewers!)

text/metrics/0238-pipeline-monitoring.md
jmacd: Is the consume operation synchronous? I think the traces/in pipeline will wait until the metrics/out pipeline finishes the consume operation, so the outcome for traces/in will depend on the outcome for metrics/out. If metrics/out fails w/ a retryable status, maybe the producer will retry.

Since the count connector can produce more or fewer metric data points than arriving spans, I do not expect the item counts to match between the exporter and receiver, but I think the outcomes could match for synchronous operations. If the operation is asynchronous, the rules discussed in this proposal would apply -- the traces/in might see `consumed` while the metrics/out sees some sort of failure.

I don't see any problems, per se, just that the monitoring equations for connectors don't apply. I can't assume that the items_in == items_dropped + items_out.

jmacd:
@kristinapathak will resume this effort, thank you!

PR #237, opened 2023-09-07
Introduces Profiling Data Model

petethepig:
OTel Profiling SIG has been working on this for a while and we're ready to present our Data Model.  So far we've done a number of things to validate it:
* we've written a [new profiles proto](https://github.com/open-telemetry/opentelemetry-proto-profile/blob/main/opentelemetry/proto/profiles/v1/profiles.proto)
* we've documented decisions made along the way in a [decision log](https://github.com/open-telemetry/opentelemetry-proto-profile/blob/main/opentelemetry/proto/profiles/v1/decision-log.md)
* we've done benchmarking to refine the data representation (see Benchmarking section in a [collector PR](https://github.com/petethepig/opentelemetry-collector/pull/1))
* we have a working [prototype collector](https://github.com/petethepig/opentelemetry-collector/pull/1) that supports profiling signal.
* we have working prototype [client](https://github.com/grafana/pyroscope-golang/pull/36) and [backend](https://github.com/grafana/pyroscope/pull/2177) implementations

We're seeking feedback and hoping to get this approved.

---

For (a lot) more details, see:
* [OTel Profiling SIG Meeting Notes](https://docs.google.com/document/d/19UqPPPlGE83N37MhS93uRlxsP1_wGxQ33Qv6CDHaEp0/edit)

PR #239, opened 2023-11-02
Introduces Profiling Data Model v2

petethepig:
This is second version of the Profiling Data Model OTEP. After [we've gotten feedback from the greater OTel community](https://github.com/open-telemetry/oteps/pull/237) we went back to the drawing board and came up with a new version of the data model. The main difference between the two versions is that the new version is more similar to the original pprof format, which makes it easier to understand and implement. It also has better performance characteristics. We've also incorporated a lot of the feedback we've gotten on the first PR into this OTEP.

Some minor details about the data model are still being discussed and will be flushed out in the future OTEPs. We intend to finalize these details after doing experiments with early versions of working client + collector + backend implementations and getting feedback from the community. The goal of this OTEP is to provide a solid foundation for these experiments.

So far we've done a number of things to validate it:
* we've written a new profiles proto described in this OTEP
* we've documented decisions made along the way in a [decision log](https://github.com/open-telemetry/opentelemetry-proto-profile/blob/main/opentelemetry/proto/profiles/v1/decision-log.md)
* we've done benchmarking to refine the data representation (see Benchmarking section in a [collector PR](https://github.com/petethepig/opentelemetry-collector/pull/1))

* diff between original pprof and the new proto: [link](https://github.com/open-telemetry/opentelemetry-proto-profile/compare/2cf711b3cfcc1edd4e46f9b82d19d016d6d0aa2a...petethepig:opentelemetry-proto:pprof-experiments#diff-9cb689ea05ecfd2edffc39869eca3282a3f2f45a8e1aa21624b452fa5362d1d2)

We're seeking feedback and hoping to get this approved. 

---

For (a lot) more details, see:
* [OTel Profiling SIG Meeting Notes](https://docs.google.com/document/d/19UqPPPlGE83N37MhS93uRlxsP1_wGxQ33Qv6CDHaEp0/edit)


mtwo:
@carlosalberto @jack-berg can we mark this as triaged with priority p0?

tigrannajaryan:
> The main difference between the two versions is that the new version is more similar to the original pprof format

@petethepig More similar, but not fully backwards compatible with pprof?

tigrannajaryan (commented):
How much difference does AnyValue/KeyValue interning make? I would prefer not to have a different type just for profiles. If it is a significant difference we may want to consider adding interning support to existing AnyValue/KeyValue.

aalexand:
> How much difference does AnyValue/KeyValue interning make? I would prefer not to have a different type just for profiles. If it is a significant difference we may want to consider adding interning support to existing AnyValue/KeyValue.

One thing is interning, another thing that pprof labels have is unit support. E.g. int64 bytes can be distinguished from int64 microseconds or count of something. This is used by pprof for using appropriate unit suffixes and scaling the label values appropriately.

petethepig:
@tigrannajaryan 

> How much difference does AnyValue/KeyValue interning make? I would prefer not to have a different type just for profiles. If it is a significant difference we may want to consider adding interning support to existing AnyValue/KeyValue.

We did some benchmarking for that. Here are the results (it's also in [this spreadsheet, "Attribute Represenations Summary" sheet](https://docs.google.com/spreadsheets/d/1Q-6MlegV8xLYdz5WD5iPxQU2tsfodX1-CDV1WeGzyQ0/edit?usp=sharing)):
Screenshot 2023-11-29 at 6 16 09 PM

And here's the difference between the 3 implementations:

```proto

// Extended
"message Sample {
  repeated opentelemetry.proto.common.v1.KeyValue attributes = 9;
}"


// ExtendedInterned
"message Sample {
  repeated opentelemetry.proto.common.v1.KeyValueInterned attributes = 6;
}"


// ExtendedLookup
"message Profile {
  // lookup table
  repeated opentelemetry.proto.common.v1.KeyValue attributes = 9;
}

message Sample {
  repeated uint64 attributes = 10;
}
```

To be clear, in the OTEP we're using `ExtendedLookup` represenation. It does not do the same kind of interning as it's done in original pprof (that would be `ExtendedInterned`).

`ExtendedInterned` uses a string_table for string values and `ExtendedLookup` uses an attributes lookup table. `Extended` uses a schema that's more common for other OTEL signals where the attributes are embedded inside other messages (`Sample` in our case).

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1458 @@
...
+  // discouraged case is having a string label and a numeric label of the same
+  // name on a sample.  Again, possible to express, but should not be used.
+  // [deprecated, superseded by attributes]
+  repeated Label label = 3;
+
+  // locations_start_index along with locations_length refers to to a slice of locations in Profile.location. Supersedes location_index.
+  uint64 locations_start_index = 4;
+
+  // locations_length along with locations_start_index refers to a slice of locations in Profile.location. Supersedes location_index.
+  uint64 locations_length = 5;
```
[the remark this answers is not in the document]
petethepig: Yes, the indexes are moved to `Profile.location` for performance (e.g less allocations) + smaller size.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1458 @@
...
+
+  // Timestamps associated with Sample represented in ms. These timestamps are expected to fall within the Profile's time range. [optional]
+  repeated uint64 timestamps = 6;
+
+  // References to attributes in Profile.attribute_table. [optional]
+  repeated uint64 attributes = 7;
+}
+
+message Label {
+  int64 key = 1;   // Index into string table
```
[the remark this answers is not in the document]
petethepig: I agree it would be nice if things were more consistent like this, but this might be too much of a change for pprof. @aalexand what do you think about this?

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1458 @@
...
+
+Here's a [link to a diff between original pprof and modified pprof](https://github.com/open-telemetry/opentelemetry-proto-profile/compare/2cf711b3cfcc1edd4e46f9b82d19d016d6d0aa2a...petethepig:opentelemetry-proto:pprof-experiments#diff-9cb689ea05ecfd2edffc39869eca3282a3f2f45a8e1aa21624b452fa5362d1d2) and here's a list of main differences between pprof and OTLP profiles:
+
+* Sharing of the call stacks between samples.
+* Sharing of labels (now called attributes) between samples.
+* Reuse of OpenTelemetry conventions and message types.
+* Semantic conventions for linking to other signals via `trace_id`s and `span_id`s.
+* First-class timestamp support.
+
+Below you will find the proto for the new Profiles signal. It is split into two parts: the first part is the OpenTelemetry specific part, and the second part is the modified pprof proto. Intention here is to make it easier to compare modified pprof proto to the original pprof proto.
```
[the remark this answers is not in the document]
petethepig: Added a section [here](https://github.com/open-telemetry/oteps/pull/239/files#diff-068ff81e247a4e046bbb0ee86c4c01b958c6c6ced699c46c3a41fa12d01ece12R107)

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1471 @@
...
+  // NOTE: While possible, having multiple values for the same label key is
+  // strongly discouraged and should never be used. Most tools (e.g. pprof) do
+  // not have good (or any) support for multi-value labels. And an even more
+  // discouraged case is having a string label and a numeric label of the same
+  // name on a sample.  Again, possible to express, but should not be used.
+  // [deprecated, superseded by attributes]
+  repeated Label label = 3;
+
+  // locations_start_index along with locations_length refers to to a slice of locations in Profile.location. Supersedes location_index.
+  uint64 locations_start_index = 4;
```
aalexand, line 503: I'm still not sure this single table of locations and pointing to a range in it from samples is worth it. I wonder if a separate Callstacks table that would contain specific stacks and then referring to it by index would be simpler to understand and not much less efficient.

The downsides of the current approach as I see them are:
* We are slicing a single array so it's effectively a little of custom memory management which means things like array-out-of-bounds can go unnoticed.
* Writing code for profile.Merge() is going to be tricky as it's not obvious what it should optimize for.
* Collector code usually needs to be as simple as possible, so putting there any heuristics for trying to minimize the locations table is probably unlikely.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1471 @@
...
+
+  // locations_length along with locations_start_index refers to a slice of locations in Profile.location. Supersedes location_index.
+  uint64 locations_length = 5;
+
+  // Timestamps associated with Sample represented in ms. These timestamps are expected to fall within the Profile's time range. [optional]
+  repeated uint64 timestamps = 6;
+
+  // References to attributes in Profile.attribute_table. [optional]
+  repeated uint64 attributes = 7;
```
aalexand, line 512: One downside that worries me a little here is that for numeric labels with high cardinality the key name is going to be duplicated a lot.

text/profiles/0239-profiles-data-model.md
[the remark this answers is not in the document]
aalexand: Given that this is a nit I'd prefer to keep the old name.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1471 @@
...
+  // The following fields indicate the resolution of symbolic info.
+  bool has_functions = 7;
+  bool has_filenames = 8;
+  bool has_line_numbers = 9;
+  bool has_inline_frames = 10;
+
+  // A string that uniquely identifies a particular binary with high probability.
+  // It is meant to work around deficiencies in the GNU Build ID that may compromise
+  // uniqueness / correlatability. Index into string table.
+  uint32 file_id_index = 11;
```
aalexand, line 562: Could we simply use the build_id field for this? The two fields are very close in semantics so I'm not sure it's great to carry both of them forward forever...

text/profiles/0239-profiles-data-model.md
aalexand: Also, if we do that, then we'd have a much more natural place to put `stacktrace_id_index` into. I think there it could even be just a string field, wouldn't have to be an index since it would be unique per unique stack trace anyway.

text/profiles/0239-profiles-data-model.md
aalexand: If the matter is being able to distinguish the semantics (GNU build ID vs alternative hash), maybe we could instead add an enum that tells the exact meaning of the build_id field with zero value enum member for the current semantics? I understand that build_id maybe non-ideal name, but perhaps it's good enough?

text/profiles/0239-profiles-data-model.md
aalexand: I mean something like [this](https://github.com/google/pprof/compare/main...aalexand:pprof:otel-proto?expand=1).

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1471 @@
...
+  // [deprecated, superseded by attributes]
+  repeated Label label = 3;
+
+  // locations_start_index along with locations_length refers to to a slice of locations in Profile.location. Supersedes location_index.
+  uint64 locations_start_index = 4;
+
+  // locations_length along with locations_start_index refers to a slice of locations in Profile.location. Supersedes location_index.
+  uint64 locations_length = 5;
+
+  // Timestamps associated with Sample represented in ms. These timestamps are expected to fall within the Profile's time range. [optional]
```
aalexand, line 508: Should probably be in microseconds or nanoseconds. Milliseconds might be not enough in precision. There are also duration_nanos and time_nanos in profile.proto already.

Timestamps should also be fixed protobuf types since they are typically large enough magnitude that varint-ness is not helpful. So, fixed64 perhaps.

I also wonder whether the addition of timestamps field could generally be postponed. It doesn't seem to be central to the format at least in the current form, it's just an additional field not referencing anything and not referenced by anything.

petethepig: I like the idea of postponing it. AFAIK We don't really have any profiler implementations that could use this feature and so a lot of conversations about timestamps have been quite speculative imo. 

I'd be fine with something like this:
* merge data model without timestamps
* have a solid base with a few working SDK implementations
* add timestamps then

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1471 @@
...
+  7. A request is received, the system measures 1 request.
+  8. The 1 second collection cycle ends. A metric is exported for the
+      number of requests received over the interval of time t_0 to
+      t_0+2 with a value of 5.
+  9. The system experiences a fault and loses state.
+  10. The system recovers and resumes receiving at time=t_1.
+  11. A request is received, the system measures 1 request.
+  12. The 1 second collection cycle ends. A metric is exported for the
+      number of requests received over the interval of time t_1 to
+      t_0+1 with a value of 1.
```
yanglong1010, line 461: Should it be `t_1+1` here?

text/profiles/0239-profiles-data-model.md
petethepig: We tried that in the past (For reference, we discussed it on October 5, I explained it around 23:00 mark) and unfortunately, that increases the number of objects allocated quite a bit, and is less efficient compared to the version where locations are in a shared array.

Here are the pseudocode diff between 2 versions:
* Oct 4 (pseudocode):
```proto
struct Profile {
  repeated Stacktrace stacktraces = 1;
  repeated Sample samples = 2;
}

struct Sample {
  uint64 stacktrace_id = 1;
}

struct Stacktrace {
  uint64 location_id = 1;
}
```


* After Stacktrace removal (Oct 6 2023) (pseudocode)
```proto
struct Profile {
  repeated uint64 location_ids = 1;
  repeated Sample samples = 2;
}

struct Sample {
  uint64 location_start_index = 1;
  uint64 location_length = 2;
}
```

And here's a screenshot from benchmarks:
Screenshot 2024-01-03 at 8 27 43 PM

[Link to the benchmarks](https://docs.google.com/spreadsheets/d/1Q-6MlegV8xLYdz5WD5iPxQU2tsfodX1-CDV1WeGzyQ0/edit?usp=sharing) ("Oct 4 vs Oct 6 After Stacktrace removal" sheet)

text/profiles/0239-profiles-data-model.md
petethepig: I know we've discussed this on Dec 14, but not sure what we've decided on here. @aalexand and @thomasdullien,  what do you guys think?

text/profiles/0239-profiles-data-model.md
petethepig: It's probably true for that specific use case where you have a lot of unique attributes. But for most other profiles this won't be an issue. One way I could see us solving it is if it ever does become an issue maybe we bring back the whole idea of interning attribute key/value pairs but make it a separate proposal that would work for all signals.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1458 @@
...
+According to the [data from Profilerpedia](https://docs.google.com/spreadsheets/d/1UM-WFQhNf4GcyXmluSUGnMbOenvN-TqP2HQC9-Y50Lc/edit?usp=sharing), it's one of the most widely used formats. Compared to other formats it has the highest number of profilers, UIs, formats it can be converted to and from.
+
+The original pprof data model underwent enhancements to more effectively manage profiling data within the scope of OpenTelemetry, and certain upgrades were implemented to overcome a few of the original format's constraints.
+
+Here's a [link to a diff between original pprof and modified pprof](https://github.com/open-telemetry/opentelemetry-proto-profile/compare/2cf711b3cfcc1edd4e46f9b82d19d016d6d0aa2a...petethepig:opentelemetry-proto:pprof-experiments#diff-9cb689ea05ecfd2edffc39869eca3282a3f2f45a8e1aa21624b452fa5362d1d2) and here's a list of main differences between pprof and OTLP profiles:
+
+* Sharing of the call stacks between samples.
+* Sharing of labels (now called attributes) between samples.
+* Reuse of OpenTelemetry conventions and message types.
+* Semantic conventions for linking to other signals via `trace_id`s and `span_id`s.
```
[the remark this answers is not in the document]
petethepig: @aalexand I don't remember exactly why we decided to go with semantic conventions. Could you share your thoughts on this?

text/profiles/0239-profiles-data-model.md
aalexand: I think what we've decided and what I'd prefer is that we keep the pprof's build_id field as is but also add a build_id_kind enum field with BuildIdKind enum that has members like `BUILD_ID_LINKER = 0` corresponding to the current semantics and `BUILD_ID_BINARY_HASH = 1` for the proposed new schema. In the future it can be extended to other hashing approaches if and as necessary. The only downside to this I see is imperfect name (one could argue that file_id is cleaner) but I think backward compatibility trumps the perfection in this case.

aalexand:
> > How much difference does AnyValue/KeyValue interning make? I would prefer not to have a different type just for profiles. If it is a significant difference we may want to consider adding interning support to existing AnyValue/KeyValue.
> 
> One thing is interning, another thing that pprof labels have is unit support. E.g. int64 bytes can be distinguished from int64 microseconds or count of something. This is used by pprof for using appropriate unit suffixes and scaling the label values appropriately.

We still need units for labels I think.

tigrannajaryan:
If label/attribute units are critical I suggest that we start this discussion early.

I looked at other signal types and there is a few attribute conventions that would benefit from units (but not many):

Span attributes (all bytes):
message.compressed_size
message.uncompressed_size
messaging.message.body.size
messaging.message.envelope.size
faas.max_memory

No many metric attributes that can have units, I was only able to find one (WAh):
hw.battery.capacity

Similarly just one resource attribute with unit (bytes):
host.cpu.cache.l2.size

There may be others that I missed.

You can file an issue in https://github.com/open-telemetry/opentelemetry-proto to begin the discussion.

petethepig:
I made a [few changes to this OTEP after our discussion today](https://github.com/open-telemetry/oteps/compare/a9038791636b6283e38c9b2ea704a0fe4624692b...82a083cf0e4ef954a0b558912d934ef426d02f8b):
* added a new `attribute_units` field to add support for units (Alexey & Tigran, curious what you think about this solution, it ended up being less ugly than I thought)
* introduced new `BuildIdKind` enum to handle different types of `build_id`s (need feedback from Elastic folks)
( brought back `Link` structure for `trace_id` / `span_id` linking (cc Alexey & Tigran)
* added a section describing “Open Questions” — things we intend to address in the experimentation phase

cc @aalexand @tigrannajaryan @thomasdullien

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1575 @@
...
+OpenTelemetry specific part:
+<!-- proto1 -->
+
+```proto
+syntax = "proto3";
+
+package opentelemetry.proto.profiles.v1;
+
+import "opentelemetry/proto/common/v1/common.proto";
+import "opentelemetry/proto/resource/v1/resource.proto";
```
Kielek, line 147:
```suggestion
import "opentelemetry/proto/common/v1/common.proto";
import "opentelemetry/proto/resource/v1/resource.proto";
import "opentelemetry/proto/profiles/v1/alternatives/pprofextended.proto";
```

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1575 @@
...
+//   mappings. For every nonzero Location.mapping_id there must be a
+//   unique Mapping with that id.
+
+syntax = "proto3";
+
+package opentelemetry.proto.profiles.v1.alternatives.pprofextended;
+
+import "opentelemetry/proto/common/v1/common.proto";
+
+option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1/alternatives/pprofextended";
```
Kielek, line 350:
```suggestion
option csharp_namespace = "OpenTelemetry.Proto.Profiles.V1.Alternatives.PprofExtended";
option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1/alternatives/pprofextended";
```

Probably also worth to add java related options as for all other OTLP proto files.

Kielek (commented):
I am playing a bit with exporter/receiver for C#/.NET in context of https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3074

Two minor issues:
1. profiles.proto needs explicit reference/import to pprofextended
2. pprofextended should sets option for generating classes for C#, and possibly Java, like for all others OTLP proto files.

aalexand:
> I made a [few changes to this OTEP after our discussion today](https://github.com/open-telemetry/oteps/compare/a9038791636b6283e38c9b2ea704a0fe4624692b...82a083cf0e4ef954a0b558912d934ef426d02f8b):
> 
> * added a new `attribute_units` field to add support for units (Alexey & Tigran, curious what you think about this solution, it ended up being less ugly than I thought)
> * introduced new `BuildIdKind` enum to handle different types of `build_id`s (need feedback from Elastic folks)
>   ( brought back `Link` structure for `trace_id` / `span_id` linking (cc Alexey & Tigran)
> * added a section describing “Open Questions” — things we intend to address in the experimentation phase
> 
> cc @aalexand @tigrannajaryan @thomasdullien

Some assorted comments :-) :

RE Link message - is this name standard in OTel and we borrow / reuse it? I found it overly generic. I'd name this SpanLink or something like that.

Can we have single link per sample instead of a repeated field? I think dealing with the semantics of linking to multiple spans is going to be trickier (tooling-wise, for example) so I want to make sure it's not a YAGNI. Documenting the motivation would be great.

For attribute units, I was envisioning something more like a single global table that would map attribute name to a unit and thus require that for all samples the unit is the same. With the current approach conflicts are possible. Plus I think the proposed approach can be expensive - we add a slice for each sample which means one more memory allocation. As a parallel, we've been discussing why we reference locations via subslicing rather than via a separate CallStack / Callchain entity and the answer there was to reduce memory allocations, but here we add a slice per sample which is even worse I think than the CallStack table since there is generally should be less call stacks than samples but here it's clearly 1:1 with samples.

RE Tigran - pprof definitely needs units for labels, the CLI uses them today. One thing is that if we introduce them as a field to OTel label without figuring out the interning then the space usage for the case of high-cardinality label would get even worse since we won't just duplicate the label name but also the unit string...

Maybe it's worth leaving out the units change completely for now and discuss it in follow-up PRs or a dedicated issue. Maybe same about Link? Independent focused discussions are kind of nice.

I would suggest organizing fields logically, not by proto ID simply. E.g. I'd put build_id_kind after build_id field.

For build_id, a couple of comment change suggestions:

```
// Indicates the semantics of the build_id field.
enum BuildIdKind {
  // Linker-generated build ID, stored in the ELF binary notes.
  BUILD_ID_LINKER = 0;
  // Build ID based on the content hash of the binary. Currently no particular
  // hashing approach is standardized, so a given producer needs to define it
  // themselves and thus unlike BUILD_ID_LINKER this kind of hash is producer-specific.
  // We may choose to provide a standardized stable hash recommendation later.
  BUILD_ID_BINARY_HASH = 1;
}
```

"See BuildIdKind enum for more detaiks" has a typo.

Nit nit: We should make headings consistent in the OTEP text on whether we follow "Sentence case" or "Title Case" in them. There is a mix right now. I'd take a look at what OTel does as a whole.

Nit nit: Does OTel have any guidelines of comment width? I know Go recommends keeping comments in 80 columns even though it recommends not wrapping code. I'm ambivalent on this, just want to make sure we follow local conventions.

petethepig:
> RE Link message - is this name standard in OTel and we borrow / reuse it? I found it overly generic. I'd name this SpanLink or something like that.

`Link` comes from traces spec: https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L238-L242

I agree SpanLink sounds a bit better, but feel strongly about this. @tigrannajaryan what do you think?

> Can we have single link per sample instead of a repeated field? I think dealing with the semantics of linking to multiple spans is going to be trickier (tooling-wise, for example) so I want to make sure it's not a YAGNI. Documenting the motivation would be great.

Agreed, addressed in the latest version of the PR.

> For attribute units, I was envisioning something more like a single global table that would map attribute name to a unit and thus require that for all samples the unit is the same.

Agreed, I like your way more. I addressed in the latest version of the PR, please take a look.

> I would suggest organizing fields logically, not by proto ID simply. E.g. I'd put build_id_kind after build_id field.

Agreed, changed the order in other messages too.

> For build_id, a couple of comment change suggestions:

Thanks for writing these up, addressed in the latest version of the PR.

> "See BuildIdKind enum for more detaiks" has a typo.

fixed

> Nit nit: We should make headings consistent in the OTEP text on whether we follow "Sentence case" or "Title Case" in them. There is a mix right now. I'd take a look at what OTel does as a whole.

fixed

> Nit nit: Does OTel have any guidelines of comment width? I know Go recommends keeping comments in 80 columns even though it recommends not wrapping code. I'm ambivalent on this, just want to make sure we follow local conventions.

I could not find anything specific.  cc @tigrannajaryan

text/profiles/0239-profiles-data-model.md
petethepig: Addressed this in the last commit.

text/profiles/0239-profiles-data-model.md
petethepig: Addressed this in the last commit. Thanks for trying it out!

Rperry2174 (approved):
I think this is a great milestone! Looking forward to the steps that follow!

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1601 @@
...
+  // Use arbitrary string (for example, "requests") as a custom count unit.
+  // If no unit is specified, consumer may apply heuristic to deduce the unit.
+  // Consumers may also  interpret units like "bytes" and "kilobytes" as memory
+  // units and units like "seconds" and "nanoseconds" as time units,
+  // and apply appropriate unit conversions to these.
+  int64 num_unit = 4;  // Index into string table
+}
+
+// Indicates the semantics of the build_id field.
+enum BuildIdKind {
```
beaubelgrave, line 577: For Windows, the equivalent of GNU build note are the various codeview sections in a PE file:
https://www.debuginfo.com/articles/debuginfomatch.html

I would suggest adding BUILD_ID_PDB_20 and BUILD_ID_PDB_70 as those are very commonly tracked values for Windows based profiling for symbol correlation.

The .NET team also added a newer convention for matching a Linux perfmap file in their PE files for ready-to-run compilation that should be covered here:
https://github.com/dotnet/runtime/blob/main/docs/design/coreclr/botr/r2r-perfmap-format.md
https://github.com/dotnet/runtime/pull/53792

I would suggest a BUILD_ID_R2R_PERF_MAP for this.

aalexand: I would suggest we submit this PR as is and then have subsequent change to add these other build ID types as enum values so that the discussion and evolution are clearer. You could send that PR, doesn't have to be Dmitry. We'll need to add proper comments to hint at what the semantics of each value are.

beaubelgrave: Sounds good to me.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1601 @@
...
+
+[Link](#message-link) is a message that is used to represent connections between profile [Samples](#message-sample) and trace spans. It uses `trace_id` and `span_id` as identifiers.
+
+For other signals, such as logs or metrics, because other signals use the same way of linking between such signals and traces (`trace_id` and `span_id`), it is possible to correlate profiles with other signals using this same information.
+
+#### From Other Signals to Profiles
+
+Other signals can use `profile_id` to reference a profile. For example, a log record can reference a profile that was collected at the time when the log record was generated by using `profile_id` as one of the attributes. This allows to correlate logs with profiles.
+
+Additionally, `trace_id`, `span_id` can be used to reference specific [Samples](#message-sample) in a Profile, since [Samples](#message-sample) are linked to traces with these same identifiers using [Links](#message-link).
```
aalexand, line 109: I think this can be slightly misleading since this might create an impression that this allows to link to a given sample, but this is not so because it's really more about linking to a given span and thus also, as a by-product, linking to all samples linked to that span.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1601 @@
...
+
+The exact details of such linking are out of scope for this OTEP. It is expected that the exact details will be defined in Profiles part of [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification).
+
+### Compatibility With Original pprof
+
+The proposed data model is backwards compatible with original pprof in a sense that a pprof file generated by existing software can be parsed using the new proto.
+
+It is not forwards compatible, meaning that a pprof file generated by the new proto cannot be parsed by existing software. This is mainly due to the sharing of the call stacks between samples + new format for labels (more on these differences below).
```
aalexand, line 117: nit: backwards compatible -> backward compatible, forwards compatible -> forward compatible.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1601 @@
...
+
+According to the [data from Profilerpedia](https://docs.google.com/spreadsheets/d/1UM-WFQhNf4GcyXmluSUGnMbOenvN-TqP2HQC9-Y50Lc/edit?usp=sharing), it's one of the most widely used formats. Compared to other formats it has the highest number of profilers, UIs, formats it can be converted to and from.
+
+The original pprof data model underwent enhancements to more effectively manage profiling data within the scope of OpenTelemetry, and certain upgrades were implemented to overcome a few of the original format's constraints.
+
+Here's a [link to a diff between original pprof and modified pprof](https://github.com/open-telemetry/opentelemetry-proto-profile/compare/2cf711b3cfcc1edd4e46f9b82d19d016d6d0aa2a...petethepig:opentelemetry-proto:pprof-experiments#diff-9cb689ea05ecfd2edffc39869eca3282a3f2f45a8e1aa21624b452fa5362d1d2) and here's a list of main differences between pprof and OTLP profiles:
```
aalexand, line 129: Is the link up-to-date? I noticed it links to specific commit hashes and personal fork, so asking.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1601 @@
...
+
+* Sharing of the call stacks between samples.
+* Sharing of labels (now called attributes) between samples.
+* Reuse of OpenTelemetry conventions and message types.
+* Semantic conventions for linking to other signals via `trace_id`s and `span_id`s.
+* First-class timestamp support.
```
aalexand, line 135: Remove? We postponed this.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1601 @@
...
+// └──────────────────┘
+//   │
+//   │ 1-1
+//   ▽
+// ┌──────────────────┐
+// │     Function     │
+// └──────────────────┘
+//
+
+// ProfilesData represents the profiles data that can be stored in a persistent storage,
```
aalexand, line 213: FWIW I think plural in noun phrases is uncommon, so ProfileData would be more common. But maybe this follows some common OTel convention.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1601 @@
...
+  // The OpenTelemetry API specification further restricts the allowed value types:
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute
+  // Attribute keys MUST be unique (it is not allowed to have more than one
+  // attribute with the same key).
+  repeated opentelemetry.proto.common.v1.KeyValue attributes = 4;
+
+  // dropped_attributes_count is the number of attributes that were discarded. Attributes
+  // can be discarded because their keys are too long or because there are too many
+  // attributes. If this value is 0, then no attributes were dropped.
+  uint32 dropped_attributes_count = 5;
```
aalexand, line 299: Similar to my other comment: Intentionally plural? `dropped_attribute_count` would be more common spelling I think.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1601 @@
...
+
+package opentelemetry.proto.profiles.v1.alternatives.pprofextended;
+
+import "opentelemetry/proto/common/v1/common.proto";
+
+option csharp_namespace = "OpenTelemetry.Proto.Profiles.V1.Alternatives.PprofExtended";
+option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1/alternatives/pprofextended";
+
+// Represents a mapping between Attribute Keys and Units.
+message AttributeUnit {
```
aalexand, line 357: I think the rest of the proto follows top-down declaration order (and protobuf allows that, unlike C or C++) so I'd move this down and keep the main proto (Profile) at the top.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1601 @@
...
+  // 16-byte array.
+  bytes trace_id = 1;
+
+  // A unique identifier for the linked span. The ID is an 8-byte array.
+  bytes span_id = 2;
+}
+
+// Specifies the method of aggregating metric values, either DELTA (change since last report)
+// or CUMULATIVE (total since a fixed start time).
+enum AggregationTemporality {
```
aalexand, line 439: FWIW I never reviewed this enum and all the extensive comments in it in details. I would probably prefer if we leave it out of this initial PR and add later, reviewing the concept and the comments holistically.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1601 @@
...
+  repeated Label label = 3;
+  // References to attributes in Profile.attribute_table. [optional]
+  repeated uint64 attributes = 10;
+
+  // Reference to link in Profile.link_table. [optional]
+  uint64 link = 12;
+
+  // Timestamps associated with Sample represented in ms. These timestamps are expected
+  // to fall within the Profile's time range. [optional]
+  repeated uint64 timestamps = 13;
```
aalexand, line 554: I thought we decided to remove timestamps from this initial PR.

text/profiles/0239-profiles-data-model.md
aalexand: No, this format does not specifically deal with JIT.

aalexand (approved):

etep (approved):

jhalliday (approved):
Great starting point for experimental implementations! Clear explanations and practical benchmark data. This should be merged as is and further tweaks deferred until we have feedback from API design and early implementation work.

text/profiles/0239-profiles-data-model.md
petethepig: This is copied from [other signals in OTEL](https://github.com/search?q=org%3Aopen-telemetry+dropped_attributes_count&type=code), would be impractical to change imo.

text/profiles/0239-profiles-data-model.md
petethepig: This is OTEL specific, going to go with convention:
* https://github.com/open-telemetry/opentelemetry-proto/blob/449a1b0219a699a60d9cbc3899c68cb1ec8c1139/opentelemetry/proto/metrics/v1/metrics.proto#L38
* https://github.com/open-telemetry/opentelemetry-proto/blob/449a1b0219a699a60d9cbc3899c68cb1ec8c1139/opentelemetry/proto/logs/v1/logs.proto#L38
* https://github.com/open-telemetry/opentelemetry-proto/blob/c26513af3305a845d1b38abea6244b074d21759d/opentelemetry/proto/trace/v1/trace.proto#L38

text/profiles/0239-profiles-data-model.md
petethepig: Yes, it is up-to-date. Right now it is convenient for me to keep this URL so that this link is always valid and I don't have to go through PR process in that repo every time I make a change.

My plan is to replace it with a better URL (no personal fork, shas instead of branch names) right before merging this OTEP.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1601 @@
...
+
+Other signals can use `profile_id` to reference a profile. For example, a log record can reference a profile that was collected at the time when the log record was generated by using `profile_id` as one of the attributes. This allows to correlate logs with profiles.
+
+Additionally, `trace_id`, `span_id` can be used to reference specific [Samples](#message-sample) in a Profile, since [Samples](#message-sample) are linked to traces with these same identifiers using [Links](#message-link).
+
+The exact details of such linking are out of scope for this OTEP. It is expected that the exact details will be defined in Profiles part of [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification).
+
+### Compatibility With Original pprof
+
+The proposed data model is backwards compatible with original pprof in a sense that a pprof file generated by existing software can be parsed using the new proto.
```
tigrannajaryan, line 115: Can we further clarify this? Does parsing existing pprof file result in all of its data fully present in the fields of the new proto? The reason I am asking is that it in Protobufs it is valid to parse data that contains unknown fields and it is not an error, however essentially doing so loses data.

Does new proto define all fields of original pprof such that parsing it is a looseless operation?

tigrannajaryan (approved):
Thank you @petethepig and all members of the Profiling SIG. I think the SIG did a great job in researching and making this proposal. I think it is now in a shape that allows to begin experimental implementations and collect feedback and data and resolve the remaining open questions.

text/profiles/0239-profiles-data-model.md
petethepig: Yes, we preserved all fields so that old pprof files can still be parsed using the new proto, and no data is lost. I'll clarify this in the OTEP too.

text/profiles/0239-profiles-data-model.md
petethepig: I know the comments are quite long, but the concept is realtively simple — there's two types of aggregation temporality, `delta` and `cumulative`. 
* `delta` is the default type that you'd see for example in CPU profiles where you reset all values for each new profile you send
* `cumulative` is a profile where the values always go up, e.g Go `alloc_objects` profiles.

This concept is used in [metrics spec as well](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L249-L318), and I just copied the type along with the comments.

Regarding removing it vs keeping it I would vote for keeping it here, because this way we can start experimenting with it. And if it turns out we decide to keep it, that's one less thing we'll need to change. And if we decide to remove it or change it we'll still have that option. But I'm happy to discuss this further live at the Profiling SIG meeting.

text/profiles/0239-profiles-data-model.md
petethepig: I addressed this in [Open Questions section](https://github.com/petethepig/oteps/blob/1e71ed0aa8d16c241ff292da48a44bf36988773f/text/profiles/0239-profiles-data-model.md#timestamps) — my idea was to have some field that we could start using in experiments but make it clear that it will most likely change by the time profiles reach `Stable` state.

I'm following similar logic to that in the [Aggregation Temporality discussion](https://github.com/open-telemetry/oteps/pull/239#discussion_r1462653940) — this gives us a good base to experiment with, while preserving optionality.

I'm open to change this though, let's further discuss Thursday.

text/profiles/0239-profiles-data-model.md
petethepig: Answered [above](https://github.com/open-telemetry/oteps/pull/239#discussion_r1462656265)

text/profiles/0239-profiles-data-model.md
petethepig: Addressed it in [this commit](https://github.com/open-telemetry/oteps/pull/239/commits/e61721cda19dab8a7a7db0520540747438e11248)

This is a tricky one, let me know if you think it's still misleading.

text/profiles/0239-profiles-data-model.md
petethepig: fixed

text/profiles/0239-profiles-data-model.md
petethepig: That's a good point. For reference, The way it's handled in metrics is by using two separate types of messages: [gauges and sums](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L193-L217), where gauges don't have aggregation temporality, but sums do, so you end up with 3 choices:
* gauge
* sum delta
* sum cumulative

I imagine this is not going to work in our case — let me think about this some more.

aalexand:
I'd probably prefer a single enum...

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1607 @@
...
+|denormalized|87,887,253|6,890,103|2,287,303|27|190,243,856|2,325,604|
+|normalized|2,528,337|953,211|333,565|27|46,449,824|1,274,000|
+|arrays|2,251,355|999,310|213,018|27|60,971,752|1,904,756|
+|pprofextended|2,398,961|872,059|274,140|27|38,874,712|353,083|
+
+#### Conclusions
+
+After running many benchmarks and analyzing the results, we came to the following conclusions:
+
+* `denormalized` representation is good but lacks deeper integration with OpenTelemetry standards and could be improved in terms of performance.
```
jpkrohling, line 1522:
```suggestion
* `pprof` representation is good but lacks deeper integration with OpenTelemetry standards and could be improved in terms of performance.
```

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1607 @@
...
+  // If the original payload is in pprof format, it SHOULD not be included in this field.
+  // The field is optional, however if it is present `profile` MUST be present and contain the same profiling information.
+  bytes original_payload = 7;
+
+  // This is a reference to a pprof profile. Required, even when original_payload is present.
+  opentelemetry.proto.profiles.v1.alternatives.pprofextended.Profile profile = 8;
+}
+```
+
+<!-- proto1 -->
```
jpkrohling, line 316: What's this comment about?

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1607 @@
...
+      - profiles:
+          - profile_id: 0102030405060708090a0b0c0d0e0f10
+            start_time_unix_nano: 1687841520000000
+            end_time_unix_nano: 1687841530000000
+            profile:
+              sample_type:
+                type: 4
+                unit: 5
+                aggregation_temporality: 1
+              attribute_table:
```
jpkrohling, line 1220: I was a bit confused about this at first, and I think we could have better clarification around attributes.

- resource attributes is what we have for other signals already, no need to further clarify
- profile container attributes -- among others, you used `"/http/server_latency": 300` as an example, but I'm not sure what this relates to in the context of a profile.
- profile attributes table (lookup table for things within the profile)
- sample attributes -- can you give examples of sample attributes?
- mapping attributes -- ditto, what do you envision this having?
- location attributes -- same here

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1607 @@
...
+  // Mapping from address ranges to the image/binary/library mapped
+  // into that address range.  mapping[0] will be the main binary.
+  repeated Mapping mapping = 3;
+  // Locations referenced by samples via location_indices.
+  repeated Location location = 4;
+  // Array of locations referenced by samples.
+  repeated int64 location_indices = 15;
+  // Functions referenced by locations.
+  repeated Function function = 5;
+  // A common table for strings referenced by various messages.
```
jpkrohling, line 379: Can you review this description? It seems to belong to the string_table.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1607 @@
...
+Array of locations referenced by samples.
+
+##### Field `function`
+
+Functions referenced by locations.
+
+##### Field `attribute_table`
+
+A common table for strings referenced by various messages.
+string_table[0] must always be "".
```
jpkrohling, line 859: This might be related to a previous comment, but can you double-check this as well? If you do mean that the first entry on the table should be empty, the example should also be updated.

jpkrohling (approved):
LGTM, and the few nits I caught can certainly be addressed in a follow-up PR.

tigrannajaryan:

text/profiles/0239-profiles-data-model.md
petethepig: Good catch, fixed it. Thank you!

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1610 @@
...
+
+[Benchmarking results](https://docs.google.com/spreadsheets/d/1Q-6MlegV8xLYdz5WD5iPxQU2tsfodX1-CDV1WeGzyQ0/edit#gid=0) spreadsheet shows the most recent benchmarking results as well as history of previous benchmarking runs. [Here](https://github.com/petethepig/opentelemetry-collector/pull/1#:~:text=in%20text%20form-,To%20run%20benchmarks%3A,-Clone%20this%20repo) are instructions on how to run the benchmarks. Here's a rough history of benchmarking our group has done:
+
+* "23 July 2023", "24 Aug 2023" — show differences between "pprof", "denormalized" and "normalized" representations
+* "04 Oct 2023" — introduces "pprofextended" representation, compares it to "pprof" and "arrays". Shoes that "pprofextended" version is overall better than "pprof", but not as a good as "arrays"
+* "06 Oct 2023" vs "After Stacktrace removal (Oct 6 2023)" — shows difference between representing stacktraces as separate struct vs a separate array of integers. Shows massive reduction in retained_objects. Demonstrates that after Stacktrace struct removal "pprofexntended" representation is better than "arrays" representation
```
christos68k, line 1468:
```suggestion
* "06 Oct 2023" vs "After Stacktrace removal (Oct 6 2023)" — shows difference between representing stacktraces as separate struct vs a separate array of integers. Shows massive reduction in retained_objects. Demonstrates that after Stacktrace struct removal "pprofextended" representation is better than "arrays" representation
```

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1610 @@
...
+<!-- proto2 -->
+
+### Message Descriptions
+
+These are detailed descriptions of protobuf messages that are used to represent profiling data.
+
+<!-- messages -->
+#### Message `ProfilesData`
+
+ProfilesData represents the profiles data that can be stored in a persistent storage,
```
christos68k, line 690:
```suggestion
ProfilesData represents the profiles data that can be stored in persistent storage,
```

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1610 @@
...
+  that is most useful to humans.  There should be enough
+  information present to determine the original sampled values.
+- On-disk, the serialized proto must be gzip-compressed.
+- The profile is represented as a set of samples, where each sample
+  references a sequence of locations, and where each location belongs
+  to a mapping.
+- There is a N->1 relationship from sample.location_id entries to
+  locations. For every sample.location_id entry there must be a
+  unique Location with that id.
+- There is an optional N->1 relationship from locations to
```
christos68k, line 814: This is also what we show in the ASCII diagram

```suggestion
- There is an optional 1->1 relationship from locations to
```

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1610 @@
...
+  set with their default value.
+- When possible, measurements should be stored in "unsampled" form
+  that is most useful to humans.  There should be enough
+  information present to determine the original sampled values.
+- On-disk, the serialized proto must be gzip-compressed.
+- The profile is represented as a set of samples, where each sample
+  references a sequence of locations, and where each location belongs
+  to a mapping.
+- There is a N->1 relationship from sample.location_id entries to
+  locations. For every sample.location_id entry there must be a
```
christos68k, line 812: I find this rewording to be less confusing:
```suggestion
- There is a 1->N relationship from samples to locations.
  For every sample.location_id entry there must be a
```

christos68k (approved):
Left a few (minor) suggestions and comments. We can iterate on this OTEP going forward after initial merge.

text/profiles/0239-profiles-data-model.md
petethepig: fixed as it's the same kind of issue as [the previous one](https://github.com/open-telemetry/oteps/pull/239#discussion_r1463712873)

text/profiles/0239-profiles-data-model.md
petethepig: This is for a script that I use to automatically update the protos from [opentelemetry-proto-profile repo](https://github.com/petethepig/opentelemetry-proto/tree/pprof-experiments) in the right place.

florianl (approved):

brancz:
So ultimately we don’t care we’ll always implement support, and implement a much deeper integrated custom format between our server and agent. However, I frankly don’t think this format is thought through enough nor has the space had enough time to innovate so I think it will stifle innovation in the space as enterprises are going to want exactly this format, it’s unsaid but to me it seems this is the primary motivation for this: to have a standard. And in my opinion this is not good for the greater community and ecosystem long term. Besides that I think we’re ignoring concerns from long term industry experts (eg. Russ Cox). It feels like we’re just trying to push a standard through to push a standard through.

I realize this sounds harsh, but I don’t think it’s the right thing to do.

tigrannajaryan:
> I think we’re ignoring concerns from long term industry experts

@brancz what concerns are you talking about? Please elaborate or provide some links.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1611 @@
...
+  // Locations referenced by samples via location_indices.
+  repeated Location location = 4;
+  // Array of locations referenced by samples.
+  repeated int64 location_indices = 15;
+  // Functions referenced by locations.
+  repeated Function function = 5;
+  // Lookup table for attributes.
+  repeated opentelemetry.proto.common.v1.KeyValue attribute_table = 16;
+  // Represents a mapping between Attribute Keys and Units.
+  repeated AttributeUnit attribute_units = 17;
```
jsuereth, line 383: There's a piece of me that wonders if we should put unit in `AttributeKey-Value`.

jsuereth (approved):
This LGTM as a way forward.  Great work profiling team!

text/profiles/0239-profiles-data-model.md
tigrannajaryan: Yes, we discussed that briefly as an option. I looked a bit into that and we don't have many semconvs at the moment that could use it (outside profiling). I only found a couple that are `Byte` measurements I think. Everything else seems unitless.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1607 @@
...
+
+### Compatibility With Original pprof
+
+The proposed data model is backward compatible with original pprof in a sense that a pprof file generated by existing software can be parsed using the new proto. All fields in the original pprof are preserved, so that original pprof files can still be parsed using the new proto, and no data is lost.
+
+It is not forward compatible, meaning that a pprof file generated by the new proto cannot be parsed by existing software. This is mainly due to the sharing of the call stacks between samples + new format for labels (more on these differences below).
+
+### Proto Definition
+
+Proto definition is based on [pprof format](https://github.com/google/pprof/blob/main/proto/profile.proto).
```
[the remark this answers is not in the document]
petethepig: At first I thought maybe I'll add a link, but then I realized the OTEP itself contains the whole proto definition, so I think it'd be extra to add another link to it, so I'm not adding it. Let me know if you have a better idea.

text/profiles/0239-profiles-data-model.md
petethepig: I think the diagram is incorrect, I'll fix.

mtwo:
Following up from today's meeting:

- @petethepig mentioned that he's going to take one last look at the documentation / rules for attribute encoding.
- We discussed two (minor) areas that are still under consideration (timestamps and aggregation temporality), and whether these should be included or not. The decision was to mark them as experimental so that we leave the door open to changes, them becoming 1.0, or their removal in the future. Implementations will not depend upon these.

We'll still need at least two more TC members to review and sign off before merging this. I'll let them know that this is ready for broad review.

mtwo (approved):

felixge (commented):
I still have a few more comments to submit along, and will hopefully heave them ready by tomorrow. Just submitting these WIP comments now since there is not reason to hold them.

thomasdullien (approved):
LGTM :-) broad strokes looking very good, thanks for the work! Let's implement this and put some strain on it :-)

felixge (approved):
Approving on behalf of Datadog.

Our approval is made under the assumption that the following things can be hashed out during the experimental phase:

* **Passthrough mode**: There should be a (configurable) passthrough mode for `pprofextended` as well as non-pprof data sources such as JFR and Go Execution traces. Without this, clients and the collector will be forced to add unnecessary encoding/decoding overhead that will significantly interfere with the stated design goal of providing low overhead. Additionally a good passthrough mode will allow ongoing innovation.
* **No pprof fork**: At this point we're still lacking an official confirmation from Google that `pprof` will adopt the modifications from `pprofextended` and support them in the pprof tool as well as the Go runtime. Forking pprof would be undesirable from our PoV, especially since the latest iteration of `pprofextended` has been designed with the goal of avoiding a fork, limiting itself to only small extensions.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1611 @@
...
+
+Additionally, `trace_id`, `span_id` can be used to reference groups of [Samples](#message-sample) (but not individual [Samples](#message-sample)) in a Profile, since [Samples](#message-sample) are linked to traces with these same identifiers using [Links](#message-link).
+
+The exact details of such linking are out of scope for this OTEP. It is expected that the exact details will be defined in Profiles part of [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification).
+
+### Compatibility With Original pprof
+
+The proposed data model is backward compatible with original pprof in a sense that a pprof file generated by existing software can be parsed using the new proto. All fields in the original pprof are preserved, so that original pprof files can still be parsed using the new proto, and no data is lost.
+
+It is not forward compatible, meaning that a pprof file generated by the new proto cannot be parsed by existing software. This is mainly due to the sharing of the call stacks between samples + new format for labels (more on these differences below).
```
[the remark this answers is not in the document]
petethepig: Agreed, I think this is very important and my plan with this is to get very specific about this particular issue and more broadly about "how do you handle old pprof" in the OTEL specification itself.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1611 @@
...
+                  value:
+                    Value:
+                      string_value: us
+              function:
+                - name: 1
+                - name: 2
+                - name: 3
+              location:
+                - line:
+                    - {}
```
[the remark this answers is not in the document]
petethepig: yeah, this is better - the reason this happened by the way is that the formatter I'm using drops empty fields (it thinks 0 is empty) by default

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1611 @@
...
+//
+// - The profile is represented as a set of samples, where each sample
+//   references a sequence of locations, and where each location belongs
+//   to a mapping.
+// - There is a N->1 relationship from sample.location_id entries to
+//   locations. For every sample.location_id entry there must be a
+//   unique Location with that id.
+// - There is an optional N->1 relationship from locations to
+//   mappings. For every nonzero Location.mapping_id there must be a
+//   unique Mapping with that id.
```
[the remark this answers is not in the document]
petethepig: fixed

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1611 @@
...
+  repeated ProfileContainer profiles = 2;
+
+  // This schema_url applies to all profiles and profile events in the "profiles" field.
+  string schema_url = 3;
+}
+
+// A ProfileContainer represents a single profile. It wraps pprof profile with OpenTelemetry specific metadata.
+message ProfileContainer {
+  // A unique identifier for a profile. The ID is a 16-byte array. An ID with
+  // all zeroes is considered invalid.
```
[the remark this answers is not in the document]
petethepig: I was thinking globally unique, similar to how trace id is globally unique. will clarify in the comments

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1611 @@
...
+  repeated ValueType sample_type = 1;
+  // The set of samples recorded in this profile.
+  repeated Sample sample = 2;
+  // Mapping from address ranges to the image/binary/library mapped
+  // into that address range.  mapping[0] will be the main binary.
+  repeated Mapping mapping = 3;
+  // Locations referenced by samples via location_indices.
+  repeated Location location = 4;
+  // Array of locations referenced by samples.
+  repeated int64 location_indices = 15;
```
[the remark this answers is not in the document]
petethepig: `location_indices` stores location ids in one array, and then later samples refer to slices from that array, e.g:

```yaml
location_indices:
  - 0
  - 1
  - 2
sample:
  - locations_start_index: 0
    locations_length: 3
    value:
      - 100
  - locations_start_index: 0
    locations_length: 2
    value:
      - 200
```

if location 0 is `foo`, 1 is `bar` and 2 is `baz`, first sample refers to `foo;bar;baz` and the second one refers to `foo;bar`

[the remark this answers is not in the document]
petethepig: It's also covered in **Relationship Between Samples And Locations** section: https://github.com/open-telemetry/oteps/blob/c9051417675e31f83771c4881339b3b39091061e/text/profiles/0239-profiles-data-model.md#relationship-between-samples-and-locations

text/profiles/0239-profiles-data-model.md
petethepig: * ProfileContainer is meant to be a container for a profile collected from some application or host. And so these attributes could be used to clarify how those profiles were collected, for example, we could include things that are relevant to the whole profile, e.g:
```
original_format: jfr
```
It is different from resource attributes because there can be multiple profiles per resource, so I figured it makes sense to have per profile attributes as well. 

* Sample attributes are things that could refer to a particular execution scope or request scope, e.g:
```yaml
controller: "payments"
path: "/v1/payments"
```

* Mapping attributes would be things that describe particular parts of executable, e.g. if it's JIT compiled or not. To be honest, I'm not an expert in this part, I think @florianl could explain this one better:
```yaml
jit_compiled: false
```

* Location attributes can be used to add additional information about code segments, e.g if you have a profiler that can distinguish between native stacks and interpreter stacks we could do something like this:
```yaml
language: python
native: false
```

---

By the way, none of these names are final, these are just examples. We're planning to use semantic conventions to standardize how these attributes are used after we merge this OTEP.

text/profiles/0239-profiles-data-model.md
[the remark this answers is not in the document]
felixge: Ah, sorry, the reason I got confused is because of this comment:

```
  // locations_start_index along with locations_length refers to to a slice of locations in Profile.location.
  // Supersedes location_index.
  uint64 locations_start_index = 7;
```

Based on this I had assumed that `locations_start_index` directly points to an entry in the `Profile.location` list. Maybe this comment could be adjusted to mention the indirection via `location_indices`?

carlosalberto (approved):

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1611 @@
...
+}
+
+// Details a specific line in a source code, linked to a function.
+message Line {
+  // The index of the corresponding profile.Function for this line.
+  uint64 function_index = 1;
+  // Line number in source code.
+  int64 line = 2;
+  // Column number in source code.
+  int64 column = 3;
```
P403n1x87, line 662: Python 3.11 added column-level information with [this PEP](https://peps.python.org/pep-0657/). The motivation is primarily for the tracebacks displayed when the application ends with an uncaught exception, but it is applicable to the broader spectrum of debugging tools (including profilers). The enhanced location information comes in the form of line and column ranges. That is, Python code objects have a lookup table that can map bytecode offsets to line start/end and column start/end pairs.

As some of the [critique](https://lists.archive.carbon60.com/python/dev/1531870) has pointed out, this new location information poses a problem for data "correctness". Because the PEP introduced ranges, simply extracting the line start to fill in the line field of the pprof spec could lead to potentially incorrect location information. For example, a profiler that is extracting/receiving only the line start information will not account for the wall time (or whatever other metric is being measured) for the full line range, thus introducing a form of bias that could be misleading when the data is (visually) represented (e.g. source code heat maps highlighting just the first line of the range).

Based on the points raised above, I would like to propose extending this structure to allow carrying the information of line end and column end, when this information is available from the runtime.

aalexand: This was discussed [here](https://github.com/google/pprof/pull/818#issuecomment-1896267978) a bit. I definitely wonder what others think but in my opinion expressing line or column ranges is fairly niche and should not be a part of the initial revision. I'd rather discuss this as a separate change request later, collecting use cases, pros and cons etc.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1611 @@
...
+
+## Design Notes
+
+### Design Goals
+
+These goals are based on the vision set out in [Profiling Vision OTEP](./0212-profiling-vision.md):
+
+* Make profiling compatible with other signals
+* Standardize profiling data model for industry-wide sharing and reuse
+* Profilers must be implementable with low overhead and conforming to OpenTelemetry-wide runtime overhead / intrusiveness and wire data size requirements.
```
reyang, line 76: minor:

```suggestion
* Make profiling compatible with other signals.
* Standardize profiling data model for industry-wide sharing and reuse.
* Profilers must be implementable with low overhead and conforming to OpenTelemetry-wide runtime overhead/intrusiveness and wire data size requirements.
```

image

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1611 @@
...
+
+The last point is particularly important in the context of profiling. Profilers generate large amounts of data, and users of profiling technology are very sensitive to the overhead that profiling introduces. In the past high overhead has been a blocker for wider adoption of continuous profiling and was one of the reasons why profiling was not used in production environments. Therefore it is important to make sure that the overhead of handling the profiling data on the client side as well as in intermediaries (e.g collector) is minimal.
```
reyang, line 78:
```suggestion
The last point is particularly important in the context of profiling. Profilers generate large amounts of data, and users of profiling technology are very sensitive to the overhead that profiling introduces. In the past high overhead has been a blocker for wider adoption of continuous profiling and was one of the reasons why profiling was not used in production environments. Therefore, it is important to make sure that the overhead of handling the profiling data on the client side as well as in intermediaries (e.g collector) is minimal.
```

jmacd (approved):
I ❤️ the compromises reached here, with an extended pprof representation.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1611 @@
...
+
+## Open Questions
+
+Some minor details about the data model are still being discussed and will be flushed out in the future OTEPs. We intend to finalize these details after doing experiments with early versions of working client + collector + backend implementations and getting feedback from the community. The goal of this OTEP is to provide a solid foundation for these experiments and [more](#future-possibilities).
+
+Here's a list of open questions:
+
+### Units in Attributes
+
+Original pprof format allows to specify units for attributes. The current data model supports a similar concept via the use of string table (see `attribute_units` in `Sample` message). It might be a good idea to have the units be specified directly in `KeyValue` message. However, such change would require changes in virtually all signals and it is not clear if it is worth it. We intend to research this question in the future and if it is worth it, we will submit a separate OTEP to make this change.
```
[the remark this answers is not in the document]
dashpole: I still think unit as metadata makes sense.  If we go that direction, I would want unit as metadata in addition to it being in the metric name.  It isn't a convention in Prometheus to have units as a suffix on attributes, though.  Do you see that as a problem, or are you just making the more general suggestion that units be part of names?

reyang (approved):

text/profiles/0239-profiles-data-model.md
felixge: I agree that I don't think this has to hold up this OTEP. But let's try to take another look at it as part of the upcoming spec work.

text/profiles/0239-profiles-data-model.md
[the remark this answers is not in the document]
petethepig: Great points! My immediate thought on this is that profile ID would act more like span id in a sense that it'd be unique per producer. The sampling part seems more relevant for us. 

But either way this is a great suggestion — We'll discuss this in a Profiling SIG meeting and we'll make sure to cover this in the upcoming profiling spec.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1611 @@
...
+
+We plan to leverage OTEL Semantic Conventions for various attributes and enums such as profile types or units. Here's a non-exhaustive list of semantic conventions that are used in data model. It is expected to be polished and extended in the future.
+
+#### Attributes
+
+| Attribute  | Type | Description  | Examples  | Requirement Level |
+|---|---|---|---|---|
+| `trace_id` | bytes | Reference to a trace | `0x01020304010203040102030401020304` | Conditionally Required: If `span_id` present. |
+| `span_id` | bytes | Reference to a span | `0x9999999999999999` | Conditionally Required: If `trace_id` present. |
+
```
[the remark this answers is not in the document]
petethepig: Oh, this is actually leftover section from when we were going to put trace_id and span_id in attributes. I'll remove the section. But we'll discuss adding flags to the data model in a SIG meeting.

text/profiles/0239-profiles-data-model.md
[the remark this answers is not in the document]
aalexand: I think there are some downsides of having the unit in the metric name as a "/" separated suffix:

1) The unit isn't that useful to see in the UI visualizing a profile since the unit is used to format values and will appear in the value in a scaled form. E.g. seeing "foo/bytes" for values like "1 KiB" would be redundant at best. So tools would have to strip this anyway.
2) It's a fairly implicit convention.

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1604 @@
...
+By storing locations in a separate table and referring to start_index+length in that table, we can reduce the size of the resulting protobuf payload and the number of objects that need to be allocated to parse such payload. With this approach we can also take advantage of the fact that many samples share the same locations. Below is a representation of the resulting protobuf payload (in YAML format for legibility):
+
+```yaml
+sample:
+  - locations_start_index: 0
+    locations_length: 3
+    value:
+      - 100
+  - locations_start_index: 2
```
florianl, line 1352: For `abc;def` the `locations_start_index` should be `4` 
```suggestion
  - locations_start_index: 4
```

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1604 @@
...
+  // name on a sample.  Again, possible to express, but should not be used.
+  // [deprecated, superseded by attributes]
+  repeated Label label = 3;
+  // References to attributes in Profile.attribute_table. [optional]
+  repeated uint64 attributes = 10;
+
+  // Reference to link in Profile.link_table. [optional]
+  uint64 link = 12;
+
+  // Timestamps associated with Sample represented in ms. These timestamps are expected
```
florianl, line 553: In `ProfileContainer` there are `start_time_unix_nano` and `end_time_unix_nano`. Should we have the same precission with `timestamps` in `Sample` and also use ns instead of ms?

text/profiles/0239-profiles-data-model.md
```diff
@@ -0,0 +1,1604 @@
...
+
+Benchmarking shows that this approach is significantly more efficient in terms of CPU utilization, memory consumption and size of the resulting protobuf payload. See [Prior art and alternatives](#prior-art-and-alternatives) for more details.
+
+#### Relationship Between Samples and Locations
+
+Relationship between Samples and Locations is using the referencing technique described above. However, there's an additional optimization technique used to reduce the size of the resulting protobuf payload and the number of objects that need to be allocated to parse such payload. The technique is based on the fact that many samples share the same locations.
+
+Considering the following example presented in a folded format:
+
+```
```
florianl, line 1332: Adding two more stacks to this example, so in the folded format it looks like this:
```
foo;bar;baz 100
abc;def 200
foo;bar 300
abc;ghi 400
foo;bar;qux 500
```
Like in most stack traces, the base frames are similar, but there is a variation in the leaf frames. To reflect this, the two added traces use different leaf frames, `ghi` and `qux`.

Should the resulting `sample` look like the following?
```yaml
sample:
  - locations_start_index: 0
    locations_length: 3
    value:
      - 100
  - locations_start_index: 3
    locations_length: 2
    value:
      - 200
  - locations_start_index: 0
    locations_length: 2
    value:
      - 300
  - locations_start_index: 5
    locations_length: 2
    value:
      - 400
  - locations_start_index: 7
    locations_length: 3
    value:
      - 500
location_indices:
  - 0 # foo
  - 1 # bar
  - 2 # baz
  - 3 # abc
  - 4 # def
  - 3 # abc 
  - 5 # ghi
  - 0 # foo
  - 1 # bar
  - 6 # qux
location:
  - line:
      - function_index: 0 # foo
  - line:
      - function_index: 1 # bar
  - line:
      - function_index: 2 # baz
  - line:
      - function_index: 3 # abc
  - line:
      - function_index: 4 # def
  - line:
      - function_index: 5 # ghi
   - line:
      - function_index: 6 # qux
function:
  - name: 1 # foo
  - name: 2 # bar
  - name: 3 # baz
  - name: 4 # abc
  - name: 5 # def
  - name: 6 # ghi
  - name: 7 # qux
```

In particular for deep stack traces with a high number of similar frames and where only leaf frames are different, the use of `locations_start_index`, `locations_length` with `location_indices` will get more complex than the (deprecated) `location_index` which just holds a list of IDs into the location table.

aalexand: The `location_indices` approach is primarily about reusing identical stacks between samples (e.g. when only labels are different) in a way that does not incur too many additional allocations. It is not an attempt to optimize encoding shared prefixes of a call tree. Neither previous nor current approach do that efficiently.

florianl: The original [pprof message Sample](https://github.com/google/pprof/blob/fb44976bdcd59b48559fb1763d69313508ad3a75/proto/profile.proto#L104-L128) does also not use the `_start_index` / `_length` approach. From my understanding all messages of type `Sample` within the same `Profile` groups stack traces from the same origin/with the same attributes.
For a different set of attributes, I think, a dedicated `Profile` should be preferred with its own attributes.

An alternative, to allow sharing `Mapping`, `Location` and `Function` information between stack traces with different attributes would be to move these three tables one layer up into `ProfileContainer`, so that they can be referenced from each `Profile`.

While the variety of leaf frames is usually high and attributes are often more static, can we remove the `deprecated` label from `location_index` in `message Sample` and let the user either set `location_index` or `location_start_index` with `locations_length`?

text/profiles/0239-profiles-data-model.md
jpkrohling: Would it be possible to address this on a follow-up PR?

florianl: As this is just a typo, I think it does not change the purpose or intention of this PR and should be good to be fixed before it get merged.

text/profiles/0239-profiles-data-model.md
florianl: With `Profile.time_nanos` there is another timestamp in the message that uses nanosecond precision.

text/profiles/0239-profiles-data-model.md
jsuereth: Let's address this in a follow up PR.

text/profiles/0239-profiles-data-model.md
jsuereth: Discussed with profiling SiG.  We believe this discussion can continue as the model is implemented in the proto directory, and does not need to be addressed in the proposal, as it doesn't change overall direction.

text/profiles/0239-profiles-data-model.md
jsuereth: Discussed with profiling SiG.  We believe this discussion can continue as the model is implemented in the proto directory, and does not need to be addressed in the proposal, as it doesn't change overall direction.

PR #225, opened 2023-02-23
Configuration proposal

codeboten:
This OTEP is the result of the Configuration working group. See https://github.com/open-telemetry/opentelemetry-specification/issues/2920 for additional details.

cc @MrAlias @jack-berg

On-going prototypes:

* [Python](https://github.com/MrAlias/otel-schema/pull/44)
* [Javascript](https://github.com/MrAlias/otel-schema/pull/49)
* [Java](https://github.com/MrAlias/otel-schema/pull/46)

PR #213, opened 2022-09-08
Add Sampling SIG research notes

spencerwilson:
- Create text/experimental/notes/ directory, containing a partial OTEP ("Motivation" section only) and two research documents on the sampling landscape

The scope of the OTEP (that appears in this PR in partial form) is a prerequisite to https://github.com/open-telemetry/oteps/issues/191.

This PR needn't be merged. Its purpose is to provide a vehicle for review and iteration by the Sampling SIG. I previously distributed these documents as GitHub gists, but the SIG decided this would be superior to that. I expect that at some point this PR will have served its purpose and at that point may be closed.

PR #240, opened 2023-11-08
A Sampling Configuration proposal

PeterF778:
Presents a configuration model for describing built-in and third-party Samplers. This is written to support a class of future samplers, as described in the OTEP.
 
(description written by @jmacd)


text/0240-Sampling_Configuration.md
```diff
@@ -0,0 +1,163 @@
...
+   samplerType: ParentBased
+   parameters:
+    - samplerType: TraceIdRatioBased  # for root spans
+      parameters:
+       - 0.75
+    - samplerType: AlwaysOn    # remote parent sampled
+    - samplerType: AlwaysOff   # remote parent not sampled
+    - samplerType: AlwaysOn    # local parent sampled
+    - samplerType: AlwaysOff   # local parent not sampled
+```
```
yurishkuro, line 77: this example is confusing - ParentBased takes only one base sampler as argument, does it not?

text/0240-Sampling_Configuration.md
```diff
@@ -0,0 +1,163 @@
...
+### Logical-Or Sampling
+```yaml
+   samplerType: AnyOf
+   parameters:
+    - - SAMPLER1
+      - SAMPLER2
+      ...
+      - SAMPLERn
+```
+The AnyOf sampler takes a list of Samplers as the argument. When making a sampling decision, it goes through the list to find a sampler that decides to sample. If found, this sampling decision is final. If none of the samplers from the list wants to sample the span, the AnyOf sampler drops the span.
```
yurishkuro, line 94: I don't believe we have such sampler type currently in the spec. The challenge with it is not introducing a configuration, but defining statistically meaningful behavior and sample weight calculation.

text/0240-Sampling_Configuration.md
PeterF778: No, it takes five arguments. Perhaps you got it confused with `ParentPasedSamplerBuilder`, which takes one.

text/0240-Sampling_Configuration.md
PeterF778: No, we don't. This proposal does not call for adding such sampler to the specs, it can remain an internal gadget, but adding it to the specs is also a possibility.
I admit I do not understand your second sentence.

text/0240-Sampling_Configuration.md
yurishkuro: if those are 5 distinct base samplers, why not name them accordingly in yaml instead of 
`-samplerType`?

text/0240-Sampling_Configuration.md
yurishkuro: > I admit I do not understand your second sentence.

there was a lot of effort put into the spec to develop sampling strategies that allow calculating summary statistics by capturing `weight` of each sample (simplest example: probabilistic 1-in-10 sampler gives each sample/span a `weight=10`). When you have a sequential composite sampler like you have here, how is the weight of a sample will be calculated?

PeterF778: Consistent Probability Sampling users have their own composite samplers, which properly handle logical-or and logical-and operations on consistent probability samplers, so they do not need AnyOf or AllOf samplers proposed here, and they should not use them.
Generally, composing consistent probability samplers with non-probabilistic samplers leads to undefined (or incorrect) adjusted counts anyway.

text/0240-Sampling_Configuration.md
PeterF778: Because that would require the parser to know that ParentBasedSampler takes 5 samplers as arguments (and not Strings, for example).

text/0240-Sampling_Configuration.md
```diff
@@ -130,6 +130,56 @@ where each RULE is
...
+- for non-root spans
+    - follow the parent sampling decision
+    - however, capture all calls to service `/foo` (even if the trace will be incomplete)
+- in any case, do not exceed 1000 spans/second
+
+Such sampling requirements can be expressed as:
+```yaml
+  samplerType: AllOf
+  parameters:
+  - - samplerType: AnyOf
```
trask, line 154: is the extra dash intentional here (and below)?

PeterF778: Yes, we have nested lists here, the list of samplers is an element of the parameter list.

text/0240-Sampling_Configuration.md
yurishkuro: parser already needs to know that. If this was a different sampler like TraceIdRatioBased it would be one parameter with numeric value - how is that knowledge different from knowing about 5 args?

text/0240-Sampling_Configuration.md
```diff
@@ -0,0 +1,232 @@
...
+
+Even when the code for these samplers is available, their use is not very simple. In case of Java, they require writing a custom agent [extension](https://opentelemetry.io/docs/instrumentation/java/automatic/extensions/). This can become a hurdle, especially for the Open Telemetry users with no hands-on coding experience.
+
+## The Goal
+
+The goal of this proposal is to create an open-ended configuration schema which supports not only the current set of SDK standard samplers, but also non-standard ones, and even samplers that will be added in the future. Furthermore the samplers should be composable together, as it is often required by the users.
+
+In contrast, the existing configuration schemas, such as [Jaeger sampling](https://www.jaegertracing.io/docs/1.50/sampling/#file-based-sampling-configuration), or Agent [OTEP 225 - Configuration proposal](https://github.com/open-telemetry/oteps/pull/225) address sampling configuration with a limited known set of samplers only.
```
jack-berg, line 32: > Furthermore the samplers should be composable together,

File configuration supports this.

File based configuration will support custom extension samplers. One of the stated goals is:

> Custom span processors, exporters, samplers, or other user defined extension components can be configured using this format.

There's an outstanding question of how exactly these extension components are called with configuration such that an instance can be configured. My working mental model is that we should have a generic configuration properties class which has accessors for reading configuration properties. Then, when file configuration comes across a reference to an extension component, it tries to load and obtain (language specific how this will work) an instance of it by passing the the generic configuration properties to it. The component will then be responsible for reading any of its fields from configuration properties and providing a configured instance.

With that said, I think we should extend the specification to include additional built-in samplers to accommodate common use cases. Having these would imply having dedicated types defined in the [opentelemetry-configuration](https://github.com/open-telemetry/opentelemetry-configuration) schema.

PeterF778: Yes, I had a look at the file based configuration and I saw the sampling configuration there, and it looks like it takes only samplers from a predefined set. 
If I understand you correctly, extending it to cover any (or almost any) sampler would require a 2-step process. Custom samplers would have to get their types and parameters declared there first. Do you expect the Open Telemetry end users being able to do that on their own?

text/0240-Sampling_Configuration.md
PeterF778: YAML parser natively provides Numbers, Strings, an possibly nested Lists and Maps (with Strings as keys). Detection of known higher-order types such as Samplers, Predicates and Rules should be easy. Then the parser will build an argument list for a parser to instantiate. Instantiation failure will indicate an error in the configuration.
So it is necessary to distinguish a case where a sampler takes 5 Samplers as arguments from the case where it takes a list of Samplers. This difference is reflected in the proposed YAML syntax.

yurishkuro: This does not explain why parameters cannot be named according to their semantics in the specific sampler.

mtwo:
@carlosalberto @jack-berg can we mark this as triaged with priority p1?

text/0240-Sampling_Configuration.md
```diff
@@ -0,0 +1,232 @@
+# Sampling Configuration
```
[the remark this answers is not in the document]
PeterF778: Yes, absolutely. I'll be happy to add this reference.

text/0240-Sampling_Configuration.md
```diff
@@ -0,0 +1,232 @@
...
+
+```yaml
+   samplerType: RuleBased
+   parameters:
+    - spanKind: SERVER | CONSUMER | ...   # optional
+    - - RULE1
+      - RULE2
+      ...
+      - RULEn
+    - FALLBACK_SAMPLER
```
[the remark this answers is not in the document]
PeterF778: @spencerwilson  I agree that `predicate`s should be powerful enough to access anything which is officially made available to the Sampler interface. However, I believe that handling span kind in a special way makes sense and actually improves readability. When users build their `predicate`s, they will often focus on just one kind, such as SERVER and the predicates will be constructed with that specific span category in mind. Repeating `spanKind == SERVER` as part of every rule will be less user friendly. Omitting it entirely might lead to undesired effects, such as CONSUMER or INTERNAL spans matching accidentally.

Eliminating the fallback sampler is, I think,  a good idea.

text/0240-Sampling_Configuration.md
PeterF778: A list is needed because the order of the arguments is essential. A map of parameter names to values would be unordered and the parser would have to know the order of the parameter names required for sampler instantiation.

text/0240-Sampling_Configuration.md
```diff
@@ -0,0 +1,232 @@
...
+
+If all of the child samplers agreed to sample, and some of them modified the trace state, the modifications are cumulative as performed in the given order. If the final decision is not to sample, the trace state remains unchanged.
+
+### Rule based sampling
+
+For rule-based sampling (e.g. decision depends on Span attributes), we need a RuleBasedSampler, which will take a list of Rules as an argument, an optional Span Kind and a fallback Sampler. Each rule will consist of a Predicate and a Sampler. For a sampling decision, if the Span kind matches the optionally specified kind, the list will be worked through in the declared order. If a Predicate holds, the corresponding Sampler will be called to make the final sampling decision. If the Span Kind does not match the final decision is not to sample. If the Span Kind matches, but none of the Predicates evaluates to True, the fallback sampler makes the final decision.
```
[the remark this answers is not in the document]
PeterF778: @spencerwilson Do you have any references for it? I'll be happy to include them as prior art.

text/0240-Sampling_Configuration.md
jack-berg: I'm not referring to jaeger file based configuration. I'm referring to the prototype OpenTelemetry file configuration specification, which was proposed in this [otep](https://github.com/open-telemetry/oteps/pull/225), and which is specified in this [document](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md), and includes a schema defined in [opentelemetry-configuration](https://github.com/open-telemetry/opentelemetry-configuration).

[Here's](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/kitchen-sink.yaml#L343-L349) some example config where we specify that the sampler is the parent based sampler, and go on to specify samplers as inputs to the parent based sampler's configuration. In this situation, the parent based sampler delegates to the trace id ratio sampler when a certain condition is true (i.e. the span has no parent), but the file configuration schema is designed to be able to accommodate arbitrarily complex configurations and support custom extension components like samplers. This means that if you can dream up how to specify it in YAML, you should be able to use file configuration to accomplish it. 

Here's a "[steel thread](https://github.com/open-telemetry/opentelemetry-java-examples/pull/227)" I've been working through to prove out the concepts, which uses file configuration to configure a custom sampler (one that is published form `opentelemetry-java-contrib`) to filter out spans with matching attribute conditions.

However, things are easier when you can reference components which are part of the specification and built into SDK implementations. And I think it would be good to spec out a rule based sampler, with a declarative syntax for specifying conditions and samplers that should be used when those conditions are met.

text/0240-Sampling_Configuration.md
```diff
@@ -0,0 +1,232 @@
...
+The focus is on head-based sampling, but a similar approach might be used for later sampling stages as well. Most of the technical details presented here assume Java as the platform, but should be general enough to have corresponding concepts and solutions available for other platformstoo.
+
+## Motivation
+
+The need for sampling configuration has been explicitly or implicitly indicated in several discussions, some of them going back a number of years, see for example
+
+- issue [173](https://github.com/open-telemetry/opentelemetry-specification/issues/173): Way to ignore healthcheck traces when using automatic tracer across all languages?
+- issue [679](https://github.com/open-telemetry/opentelemetry-specification/issues/679): Configuring a sampler from an environment variable
+- issue [1060](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1060): Exclude URLs from Tracing
+- issue [1844](https://github.com/open-telemetry/opentelemetry-specification/issues/1844): Composite Sampler
```
jack-berg, line 16: This issue is very relevant to my [comment here](https://github.com/open-telemetry/oteps/pull/240/files#r1399863342). If we had a spec'd component for a composable rule bases sampler, we could define its configuration across languages in `opentelemetry-configuration`, and allow users to specify complex rule-based sampling configuration in a YAML syntax with file configuration.

text/0240-Sampling_Configuration.md
yurishkuro: Sorry, but I don't see convincing arguments here why config needs to be so cryptic, with unnamed positional arguments, which is very user-unfriendly. @jack-berg just [pointed to an alternative](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/kitchen-sink.yaml#L343-L349) that does not sacrifice readability for some implementation details

text/0240-Sampling_Configuration.md
spencerwilson: Got it. Apologies for my misunderstanding! As someone relatively out-of-the-loop, the idea of embedding this configuration within a broader SDK schema seems very sensible to me.

text/0240-Sampling_Configuration.md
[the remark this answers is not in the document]
spencerwilson: No, I never pushed any of the actual config speccing to the internet (prior to these comments, anyway).

For X-Ray and Refinery, their documentation are the primary sources. I probably linked to the relevant parts in my notes in https://github.com/open-telemetry/oteps/pull/213.

PeterF778:
Thank you all for your questions, comments and sharing your ideas.
I'll work on a modification of this proposal and introduce a possibility to _declare_ a sampler. I hope this will improve:

- readability, as the sampler arguments no longer will have to be purely positional
- applicability for non-Java platforms
- error checking and messaging

and will bring the proposal one step further to become a true meta-sampler.

text/0240-Sampling_Configuration.md
PeterF778: Modified the schema to match exactly the syntax proposed by [OTEP 225](https://github.com/open-telemetry/oteps/pull/225) while retaining additional capabilities.

text/0240-Sampling_Configuration.md
PeterF778: Not applicable anymore, the new syntax mostly avoids these nested lists.

PeterF778:
> Can we clarify that this proposal is for sampling traces only? Similarly, should the proposed configuration structs be nested under a `traces` section?

Added the clarification. The proposal complements, but does not replace the file based configuration as defined by [OTEP 225](https://github.com/open-telemetry/oteps/blob/main/text/0225-configuration.md#configuration-file)., so there's no need to specify `traces`.

jmacd (approved):
Adding support to this proposal, which deserves extra attention from folks familiar with Jaeger remote sampling and from the OTel Configuration working group. 

I appreciate that this proposal includes support for custom samplers, as I think this is an important area for experimentation. Sampler configuration is similar to Exemplar configuration in the OTel Trace SDK spec, where we are making similar allowances for custom samplers to be installed.

text/0240-Sampling_Configuration.md
jmacd: @jack-berg it looks like your work in https://github.com/open-telemetry/opentelemetry-specification/pull/3802 supports this OTEP nicely. Thank you!

text/0240-Sampling_Configuration.md
jmacd: I like the way this OTEP enables us to experiment with sampler configuration generally, because (as I think other commenters, including @spencerwilson have shown) this is complicated. After we have integrated the new configuration model, with the new tracestate information in #235 and demonstrated prototypes compatible with head sampling, tail sampling, and Jaeger (for example), then I think we can work on specifying a standard rule-based sampler configuration for OTel.

jmacd:
:tada: @PeterF778 this is looking great!

text/0240-Sampling_Configuration.md
```diff
@@ -0,0 +1,232 @@
...
+```yaml
+   samplerType: AnyOf
+   parameters:
+    - - SAMPLER1
+      - SAMPLER2
+      ...
+      - SAMPLERn
+```
+
+The AnyOf sampler takes a list of Samplers as the argument. When making a sampling decision, it goes through the list to find a sampler that decides to sample. If found, this sampling decision is final. If none of the samplers from the list wants to sample the span, the AnyOf sampler drops the span.
```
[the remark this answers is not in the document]
PeterF778: There are challenges with stateful (or with other side effects) samplers, as well as with samplers that modify the tracestate. We've made some effort to provide a clear framework for these samplers to operate. We want to see how far we can go with useful enhancements to the samplers library without having to change the specs for the head samplers API, but that means we cannot provide support for your "used" and "pivotal" concepts.

text/0240-Sampling_Configuration.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Rule based sampling
+
+For rule-based sampling (e.g. decision depends on Span attributes), we need a `rule_based` sampler, which will take a list of Rules as an argument, and an optional Span Kind. Each rule will consist of a Predicate and a Sampler. For a sampling decision, if the Span kind matches the optionally specified kind, the list will be worked through in the declared order. If a Predicate holds, the corresponding Sampler will be called to make the final sampling decision. If the Span Kind does not match, or none of the Predicates evaluates to True, the final decision is not to sample.
```
brettmc, line 123: Could we break this idea of a rule-based sampler out into its own proposal for a new Sampler type? I'm working on something very similar at the moment, with a working name of AttributeSampler. It seems like it could be a first-class sampler alongside parent-based, traceidratio & friends.

PeterF778: Yes, after receiving some feedback I see that this OTEP is too large. I'm going to close it and re-open new ones with tailored scopes - easier to read, and hopefully easier to approve. I intend to work on them soon, but right now I'm really tied up in other things.

text/0240-Sampling_Configuration.md
```diff
@@ -0,0 +1,277 @@
...
+                rule_based:
+                  span_kind: SERVER
+                  rules:
+                   - predicate: http.target == /healthcheck
+                     sampler: always_off
+                   - predicate: http.target == /checkout
+                     sampler: always_on
+                   - predicate: true     # works as a fallback sampler
+                     sampler: trace_id_ratio_based
+                       ratio: 0.25
```
zeitlinger, line 187: this is not valid yaml

PeterF778: Yes, thanks for pointing it out.

jmacd:
@PeterF778 Would you say we can close this in favor of #250, or at least until there is progress with #250?

PeterF778:
> @PeterF778 Would you say we can close this in favor of #250, or at least until there is progress with #250?

Yes, we can close it, and revisit the ideas later.

PR #243, opened 2023-11-28
Introducing Application Telemetry Schema in OpenTelemetry - Vision and Roadmap

lquerel:
[Easy to read version with images: [https://github.com/lquerel/oteps/blob/app-telemetry-schema-vision-roadmap/text/0243-app-telemetry-schema-vision-roadmap.md](https://github.com/lquerel/oteps/blob/app-telemetry-schema-vision-roadmap/text/0243-app-telemetry-schema-vision-roadmap.md)] 

Unlike the traditional data ecosystem (OLTP and OLAP), the world of telemetry generally does not rely on the concept of a schema. Instrumentation is deeply embedded in the code of applications and libraries, making it difficult to discover all the possible telemetry signals an application can emit. This gap prevents or limits the development of CI/CD tools for checking, reporting, documenting, and generating artifacts from telemetry signals specific to an application. This document presents a long-term vision aimed at enabling the OpenTelemetry project to address this issue and extend its impact to a broader ecosystem. It proposes extending the initiatives of Telemetry Schema and Semantic Conventions to include concepts of Application Telemetry Schema and Resolved Telemetry Schema. A series of OTEPs and Tools will be proposed in this overarching document to detail each aspect of this vision.

Similar (but proprietary) initiative from Facebook: [Positional Paper: Schema-First Application Telemetry](https://research.facebook.com/publications/positional-paper-schema-first-application-telemetry/)

**EDIT 1: The OTel Weaver project (a PoC implementation of this OTEP and some of the others mentioned in the roadmap) is now available [here](https://github.com/f5/otel-weaver).**

**EDIT 2: The Slack channel [#otel-weaver](https://cloud-native.slack.com/archives/C0697EXNTL3) is dedicated to this OTEP and the associated OTel Weaver project.**


yurishkuro:
Easier to read version with images: https://github.com/lquerel/oteps/blob/app-telemetry-schema-vision-roadmap/text/0240-app-telemetry-schema-vision-roadmap.md

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,271 @@
...
+
+The internal catalog is used to define all the attributes and metrics in this
+artifact. This design allows for the reuse of the same attributes or metrics
+multiple times in different signals. It is expected to be a very common pattern
+to reuse the same subset of attributes or metrics across several signals.
+
+A dedicated OTEP will precisely define the structure and format of the Resolved
+Telemetry Schema.
+
+## Roadmap
```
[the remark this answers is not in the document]
lquerel: In fact, I started with the proof of concept but the repository is not yet public (project code name OTel Weaver). I am working on it.
 
I have already done a demo to some members of the community (mainly the semantic conventions SIG, @tigrannajaryan , @jsuereth , @jmacd , @pyohannes , @lmolkova) but I can organize another broader demo if you want. Maybe within a SIG? I'm just not sure which one would be the most appropriate.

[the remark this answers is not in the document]
yurishkuro: perhaps a demo to the full TC?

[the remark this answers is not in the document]
lquerel: Good idea. I'll see how to organize this with the full technical committee.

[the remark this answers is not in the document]
tigrannajaryan: @lquerel Dec 6, 8am PT is a TC meeting if it works for you.

tigrannajaryan:
@lquerel thank you for working on this, I think schemas are very important for Otel's evolution.

I already had a chance to discuss the roadmap and I support it. The details of each steps of course are going to be important and will need their own OTEPs or spec PRs.

text/0240-app-telemetry-schema-vision-roadmap.md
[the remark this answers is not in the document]
lquerel: @tigrannajaryan Yes that works. Thanks

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,271 @@
...
+application produces. Although there is no direct lineage between these systems,
+a similar approach was designed and deployed by Facebook to address the same
+type of problem but in a proprietary context (refer to this
+[positional paper](https://research.facebook.com/publications/positional-paper-schema-first-application-telemetry/)
+for more information).
+
+The following diagram illustrates the relationships between these concepts and
+the main changes from Telemetry Schema v1.1 to v1.2.
+
+![Telemetry Schema](./img/0240-telemetry-schema-v1.1-to-v1.2.png)
```
pyohannes, line 174: A question regarding this diagram: it shows resources and signals are defined as "siblings" of versions. So if I understand correctly, the schema defines resources and signals for the latest version in the schema, and then additional versioned transformations to transform attributes of older versions.

I wonder if there's value in defining those resources and signals for every version, as also those definitions can change across versions? Or asked in a different way: do we want to be able to create SDKs for each version contained in the schema file?

pyohannes (commented):
This is great, thanks @lquerel for this bold and ambitious proposal.

text/0240-app-telemetry-schema-vision-roadmap.md
lquerel: For simplicity's sake, I was thinking of following the principles adopted by protobuf (forward-/backward compatibility rules) in terms of signals and resources, with the slight difference that the elements defined in the semantic convention registries (mainly attribute and metric definitions) follow the rules of transformation of the telemetry schema v1.1. Therefore, by default, a client SDK will be generated for the most recent version mentioned in the schema. However, it will be entirely possible for a backend component to consume data corresponding to an earlier version of the attributes and metrics, although in terms of signals and resources, apart from these elements, it will be necessary to be able to support the rules of classical forward-backward compatibility. For the moment, this seems to me a good balance between user experience and adaptability. This topic is clearly a debate to be had and experimented with.
It would be interesting to have @yurishkuro's feedback on this point, as a similar approach has been implemented at Meta. I am due to discuss this with him soon.

jmacd (approved):
I support this vision and roadmap and look forward to reviewing future OTEPs.

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,293 @@
...
+> also identify attributes as PII (Personally Identifiable Information), and
+> privacy policy enforcement can be implemented at various levels (e.g., in the
+> generated client SDK or in a proxy).
+
+Several OTEPs will be dedicated to the precise definition of the structure and
+the format of Telemetry Schema v1.2. The rules for resolving overrides
+(inheritance), external references, and conflicts will also be described in
+these OTEPs. See the Roadmap section for a comprehensive list of these OTEPs.
+
+## Resolved Telemetry Schema
```
yurishkuro, line 203: It is unclear to me what the scope of the resolved schema is. Is it a combination of all possible schemas used in an enterprise? Or a strict subset used in a specific application? If the latter, how does that work with oss instrumentation libraries that could refer to arbitrary schemas and are not uniformly discoverable?

yurishkuro: The difference is we're not trying to obtain an exhaustive schema for a given application. An application produces one or more datasets, eg an RPC instrumentation can produce RPC metrics dataset. We want each dataset to have a well defined schema, which is an easier problem. Then the question of all datasets for an app is a discovery problem, which can be solved via search, especially if the app/workload identity is semantically annotated in the dataset schemas. Eg find all metrics/logs/etc with attribute service=myservice.

mtwo:
@jmacd can you also add the 'triaged' label?

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,305 @@
...
+* Search Engine: This tool will provide search capability within semantic
+convention registries and telemetry schemas.
+  * V1 focusing on the semantic convention registry.
+  * V2 focusing on telemetry schema v1.2.
+* Telemetry Schema Resolver: This tool will resolve the telemetry schema and its
+dependencies. The main outcome of this tool is a Resolved Telemetry Schema.
+* Schema-driven Client SDK Generator: This tool will generate a client SDK or a
+client API from a telemetry schema v1.2. It will be designed to be extensible,
+allowing each language SIG and the general community to create templates for
+generating client SDKs and APIs in specific programming languages.
```
jack-berg, line 293: This is conceptually similar to the java [instrumentation-api](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation-api). The docs on instrumentation-api are sparse, but the general idea is to provide a higher order API built on top of the metric, trace, log API (only metric and trace in practice) which makes it easy to build instrumentation which conforms to the semantic conventions. As an instrumentation author, you implement a hand full of interfaces to extract attributes relevant to the semantic convention being implemented, and the instrumentation API takes care of recording spans / metrics correctly.

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,305 @@
...
+![Application Telemetry Schema - Levels of responsibilities & Key design properties](./img/0240-app-telemetry-schema-design-properties.png)
+
+This design enables the definition of semantic conventions in a distributed
+manner. OpenTelemetry, vendors, and enterprises can define their own semantic
+conventions in different registries simplifying the existing process.
+
+The next two sections will describe the concepts of Telemetry Schema v1.2 and
+the Resolved Telemetry Schema.
+
+## Telemetry Schema v1.2 (aka Application Telemetry Schema)
```
jack-berg, line 153: Currently, the telemetry schemas [v1.1.0 file format](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/file_format_v1.1.0.md) only defines the changes in the schema from version to version. Presumably, you aim to expand this to be able to declare the schema itself, and not just the changes. If so, big 👍 from me.

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,305 @@
...
+  * V2 focusing on telemetry schema v1.2.
+* Telemetry Schema Resolver: This tool will resolve the telemetry schema and its
+dependencies. The main outcome of this tool is a Resolved Telemetry Schema.
+* Schema-driven Client SDK Generator: This tool will generate a client SDK or a
+client API from a telemetry schema v1.2. It will be designed to be extensible,
+allowing each language SIG and the general community to create templates for
+generating client SDKs and APIs in specific programming languages.
+* Telemetry Schema Compatibility Checker: This tool will verify that two
+successive versions of the same Telemetry Schema adhere to the specified
+compatibility rules.
```
jack-berg, line 296: Would be good to have some sort of component that could validate that incoming telemetry conforms to the telemetry schema it uses to describe itself. For a while, I've imagined a collector processor which inspects data and says whether or not the data conforms to the semantic conventions of its schemaUrl. This effort could be a good chance to build such a tool.

Generally, I do think its important that we have at least some sort of tool that consumes telemetry schemas. Building this tool confirms that the concepts in telemetry schema are suitable to build on top of, and acts as a reference implementation for how other consumers of telemetry schemas might work.

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,305 @@
...
+schema-driven end-to-end telemetry system in the future.
+
+![Application Telemetry Schema Overview](./img/0240-app-telemetry-schema-overview.png)
+
+The concept of the Application Telemetry Schema, as described in the previous
+diagram, will play a central role and unlock numerous use cases (this is not an
+exhaustive list):
+
+* Telemetry Client SDKs can be automatically generated from an application
+telemetry schema, enhancing both user experience and performance.
```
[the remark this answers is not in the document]
lquerel: I agree that this diagram is not clear enough. The approach is as follows: the developer of an application or a service defines an application telemetry schema describing the telemetry signals emitted by that component (also true for third party libraries). From the application telemetry schema, a "compiler" (a la protoc) produces a resolved telemetry schema that will be used to generate an Client SDK/API optimized for that specific schema. This same resolved telemetry schema can be used for other functions in the CI/CD pipeline (e.g., privacy policy enforcement, compatibility checker) or in a telemetry backend (e.g., dashboard generation, table schema generation, schema-driven data processing, ...). I will update the diagram accordingly.

text/0240-app-telemetry-schema-vision-roadmap.md
lquerel: I may not be familiar enough with the Java Instrumentation API, but my understanding so far was that the Instrumentation API contains all the attributes and metrics defined in the official semantic convention registry. So, in a way, it's not code generated for a specific application context (or library) but rather code generated for a registry or a catalog of conventions. If my interpretation is correct, then I think that this OTEP follows a somewhat different approach, although complementary.

text/0240-app-telemetry-schema-vision-roadmap.md
lquerel: Yes this is exactly the intent.

text/0240-app-telemetry-schema-vision-roadmap.md
lquerel: Indeed, one of the objectives of this project is to be able to validate incoming telemetry among other things (e.g., privacy policy enforcement, dashboard generation, database schema generation, schema-driven data transformation, ...).

I have just made public this [repo](https://github.com/f5/otel-weaver) containing a tool (PoC) that explores and implements some aspects of this OTEP.

text/0240-app-telemetry-schema-vision-roadmap.md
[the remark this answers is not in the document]
lquerel: Udpated first diagram. Let me know if it's not clear enough.

text/0240-app-telemetry-schema-vision-roadmap.md
lquerel: Added a new section discussing in more details these problematics.

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,370 @@
...
+### OTEPs
+
+* Telemetry Schema versions (incremental steps):
+  * V1.1.1: Add a section to import one or several Semantic Convention
+  Registries. Once implemented, the official Semantic Convention Registry will
+  be referenced from the official OpenTelemetry Root Telemetry Schema.
+  * V1.2: Add references to parent schemas and define a schema section to
+  describe application/library signals. In this version, only metrics, logs, and
+  spans signals are described. This is the first version of the telemetry schema
+  that can be used by the Client SDK generator. In this version only one schema
```
tigrannajaryan, line 324: >This is the first version of the telemetry schema
>  that can be used by the Client SDK generator.

Why can't V1.1.1 be used by the generator? I think it will contain enough information to generate code for Otel-wide conventions.

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,370 @@
...
+
+## Roadmap
+
+To facilitate the review process and progressively deliver value to the
+OpenTelemetry community, a series of OTEPs and tools are proposed.
+
+### OTEPs
+
+* Telemetry Schema versions (incremental steps):
+  * V1.1.1: Add a section to import one or several Semantic Convention
```
tigrannajaryan, line 318: I would like to discuss in a separate OTEP the exact structure of the schema files that "import" vs schema files that are called "resolved schemas" in this OTEP. I think it is important that we always publish the resolved schemas that don't require further link resolution to be useful for processing purposes. We don't have to discuss the details here, just wanted to call out that the ability "import" does not necessarily need to be a feature of the published schema files.

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,370 @@
...
+  upon the approval of events in OpenTelemetry.
+  * V1.4: Add support for metric_groups (multivariate metrics) in the schema
+  section. This will only be relevant if a support for multivariate metrics is
+  planned in the Client SDK Generator.
+  * V1.5: Add support for multi-inheritance and for schema hierarchies of more
+  than 2 levels. This is useful for supporting complex environments and will
+  allow the distribution of semantic convention registries across vendors and
+  enterprises. This OTEP depends on the “Schema Child/Parent relationship
+  resolution” OTEP.
+* Resolved Schema Format v1.0. This OTEP will define both the structure and
```
tigrannajaryan, line 336: I would like to consider flipping this the other way round. 

Instead of introducing a new concept of "resolved schema" I want to say that what we have today _is_ the resolved schema. That's what we publish today and what we will continue publishing in the future (with more features in newer versions). The "unresolved" schema is the new concept that we will introduce which possibly may even exist only internally at Otel and not intended for public consumption (unless we see good reasons for such consumption). The unresolved schema + semantic conventions will be the input to an internal tool that produces the Schema Files as we know them (and which are always resolved).

text/0240-app-telemetry-schema-vision-roadmap.md
lquerel: See https://github.com/open-telemetry/oteps/pull/243/commits/b1954142ff280067f16185d3f7c89e7985902bc3

text/0240-app-telemetry-schema-vision-roadmap.md
lquerel: I see. This is also related to your subsequent comment about considering the current telemetry schema version 1.1.0 as actually being a precursor to what I refer to here as the resolved telemetry schema. I need to go through the entire document again to make the necessary adjustments.

text/0240-app-telemetry-schema-vision-roadmap.md
lquerel: Yes, I understand the rationale and it doesn't seem to be a significant modification. It will require me to make some minor adjustments here and there in the OTEP. I will take care of that as soon as possible.

pyohannes: I fully agree with @tigrannajaryan, this also takes care of https://github.com/open-telemetry/oteps/pull/243/files#r1413837539.

Publishing resolved schemas would allow us to go without requiring a public specification of the attribute registry, which would be required if users were required to resolve schemas. The schema being the single point where users interface with semantic conventions makes life easier for users, as well as for spec authors.

text/0240-app-telemetry-schema-vision-roadmap.md
jsuereth: > I think it is important that we always publish the resolved schemas

+1

lquerel:
@jsuereth @yurishkuro @tigrannajaryan @pyohannes @jack-berg @jmacd 

To celebrate the end of the year, I have prepared an update to the OTEP which I hope takes into account all the feedback I received over the last few weeks. I suggest you read through the entire document as there are many differences compared to the previous version. Happy reading. I look forward to your feedback.

lquerel:
@jsuereth @yurishkuro @tigrannajaryan @pyohannes @jack-berg @jmacd 
Happy New Year to everyone! To start off 2024 on the right foot, I would like to finalize our first OTEP focusing on the concept of application telemetry schema. I believe all feedback has been incorporated (if not, please let me know), and it seems crucial to obtain approval before we proceed with the plan outlined at the end of the OTEP.

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,429 @@
...
+  same way, it is possible to use a dictionary encoding for values whose domain is
+  specified in the form of an enumeration in the schema definition.
+
+Databases and RPC systems (e.g., Protobuf & gRPC) have already addressed some of
+these issues with a schema-first approach. There is nothing to prevent adopting
+a similar approach in the context of telemetry. **This document discusses how to
+apply a schema-first approach in the OpenTelemetry project and its implications
+for the existing Telemetry Schema and Semantic Conventions.**
+
+The following diagram provides a conceptual overview of the relationships
```
yurishkuro, line 73: this still goes under Current State header, a bit confusing, maybe add "Possible/Desired Future State"

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,429 @@
...
+> recent [paper](https://arxiv.org/pdf/2311.07509.pdf#:~:text=The%20results%20of%20the%20benchmark%20provide%20evidence%20that%20supports%20our,LLM%20without%20a%20Knowledge%20Graph)
+> from [data.world](https://data.world/home/), along with
+> the [MetricFlow framework](https://docs.getdbt.com/docs/build/about-metricflow)
+> which underpins the [dbt Semantic Layer](https://www.getdbt.com/product/semantic-layer),
+> highlights the significance of adopting a schema-first approach in data
+> modeling, especially for Generative AI-based question answering systems. Tools
+> like Observability Query Assistants (
+> e.g. [Elastic AI Assistant](https://www.elastic.co/fr/blog/introducing-elastic-ai-assistant)
+>
+and [Honeycomb Query Assistant](https://www.honeycomb.io/blog/introducing-query-assistant?utm_source=newswire&utm_medium=link&utm_campaign=query_assistant))
```
yurishkuro, line 234:
```suggestion
> and [Honeycomb Query Assistant](https://www.honeycomb.io/blog/introducing-query-assistant?utm_source=newswire&utm_medium=link&utm_campaign=query_assistant))
```

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,429 @@
...
+  telemetry schema will include a registry of all standardized attributes,
+  metrics, and signals as defined by the semantic conventions.
+  * **Phase 2**: Add sections for resource, instrumentation library,
+  and schema sections to represent telemetry signals emitted by an application
+  (and its dependencies) or a library. Initially, only metrics, logs, and spans
+  definitions will be supported.
+  * **Phase 3**: Enable application or library authors to
+  override definitions inherited from a parent resolved schema.
+  * **Phase 4**: Implement support for events in the schema
+  section, pending the approval of events in OpenTelemetry.
```
yurishkuro, line 388: Is this included because you consider it a new signal type? It feels somewhat irrelevant to this project, a specific sub-domain that will simply follow the same rules as the other sub-domains.

yurishkuro (approved):
LGTM

The only addition that I think would be useful is to link to potential examples of component and resolved schemas (perhaps from the prototype), just to make it more concrete for people, vs. just the diagrams that frankly look pretty similar.

lquerel:
> The only addition that I think would be useful is to link to potential examples of component and resolved schemas (perhaps from the prototype), just to make it more concrete for people, vs. just the diagrams that frankly look pretty similar.

@yurishkuro, I will add an appendix with example schemas to illustrate the different concepts.

text/0240-app-telemetry-schema-vision-roadmap.md
lquerel: I have included events in a separate phase for the following reasons:
- It's a new type of signal that I believe has not yet been approved.
- If it is approved, and although it is protocol-compatible with log-type signals, it will be necessary to reify this concept at the telemetry schema level to make the most of it.

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,430 @@
...
+![Use cases](./img/0240-otel-weaver-use-cases.svg)
+
+The main components of a Resolved Telemetry Schema are illustrated in the
+diagram below. The 'OTel Weaver' tool is used to create these schemas. It can
+also extend an existing schema or import a Semantic Convention Registry.
+Resolved Telemetry Schema serves as a key mechanism for interoperability,
+feeding various external tools, including SDK generators, documentation
+generators, policy enforcers, and more.
+
+![Resolved Telemetry Schema](./img/0240-otel-weaver-resolved-schema.svg)
```
pyohannes, line 279: Nit: according to the diagram, a Resolved Telemetry Schema refers to _exactly one_ instrumentation library. Will this also true for the Resolved Telemetry Schema created directly from OpenTelemetry semantic conventions and provided by OpenTelemetry?

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,430 @@
...
+  straightforward to avoid any misinterpretation.
+* **Platform- and Language-agnostic**: This artifact must be independent of any
+  platform architectures and programming languages.
+
+The following diagram describes two main use cases for the Resolved Telemetry
+Schema. The key points to remember are: 1) both use cases result in a Resolved
+Telemetry Schema, and 2) Resolved Telemetry Schemas serve as the mechanism for
+distributing Telemetry Schemas throughout the entire ecosystem.
+
+![Use cases](./img/0240-otel-weaver-use-cases.svg)
```
pyohannes, line 270: Is it required for end-users to use a Component Telemetry Schema? Or could they just use the Resolved Telemetry Schema provided by OTel, assuming that no customizations are needed?

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,430 @@
...
+In the first model, each build process can independently apply telemetry schema
+policies in parallel, knowing that the policies are shared and the entire
+environment is controlled by the same entity.
+
+In the second model, the application or service environment does not control the
+telemetry schema policies of external dependencies. There is a need for a method
+to retrieve telemetry schemas from these dependencies. The mechanism for
+distributing these schemas is still being discussed (refer to the Open Questions
+section). Ultimately, this will enable the CI/CD pipeline of the application or
+service to apply its own local policies to these schemas from the dependencies.
```
pyohannes, line 308: It's not clear to me how this should work. 

I guess this might work if those "local policies" only have an effect on the SDK, but I assume they can't change any API that is used by the dependencies?

Anyway, this is covered by an open question, and doesn't need to be answered in this OTEP.

pyohannes (approved):
Thanks @lquerel for all your efforts on this topic. This OTEP does a great job of describing the overall idea and of highlighting open issues to be discussed in further OTEPs.

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,430 @@
...
+* **Easy to parse**: A widespread and well-defined format should be preferred.
+  JSON is an example of such a format.
+* **Easy to interpret**: The internal structure of this artifact must be
+  straightforward to avoid any misinterpretation.
+* **Platform- and Language-agnostic**: This artifact must be independent of any
+  platform architectures and programming languages.
+
+The following diagram describes two main use cases for the Resolved Telemetry
+Schema. The key points to remember are: 1) both use cases result in a Resolved
+Telemetry Schema, and 2) Resolved Telemetry Schemas serve as the mechanism for
```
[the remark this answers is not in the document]
lquerel: Exact that's the idea.

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,430 @@
...
+> thanks to the adoption of a schema-first approach.
+
+Several OTEPs will be dedicated to the precise definition of the structure and
+the format of this/these schema(s). The rules for resolving overrides
+(inheritance), external references, and conflicts will also be described in
+these OTEPs. See the Roadmap section for a comprehensive list of these OTEPs.
+
+## Resolved Telemetry Schema
+
+A Resolved Telemetry Schema is the outcome of the schema resolution process.
```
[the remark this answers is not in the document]
lquerel: Indeed. I will make this change in the OTEP.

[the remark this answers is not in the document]
lquerel: Done. See https://github.com/open-telemetry/oteps/pull/243/commits/87840e2533ff1f89955909e9e40438406982a45d

text/0240-app-telemetry-schema-vision-roadmap.md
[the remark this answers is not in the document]
lquerel: Updated the OTEP to clarify this point. See https://github.com/open-telemetry/oteps/pull/243/commits/9369b8a23049bb752078291daa30fcd7aad4eca8

text/0240-app-telemetry-schema-vision-roadmap.md
lquerel: Indeed, this will be covered in a future OTEP.

text/0240-app-telemetry-schema-vision-roadmap.md
lquerel: Not exactly. The dependencies section will contains the instrumentation libraries of each dependency (see the appendix B I recently added to this OTEP). Regarding a resolved telemetry schema representing a semantic convention registry, the instrumentation library is not defined.

lquerel:
> LGTM
> 
> The only addition that I think would be useful is to link to potential examples of component and resolved schemas (perhaps from the prototype), just to make it more concrete for people, vs. just the diagrams that frankly look pretty similar.

@yurishkuro I added 2 appendices to describe the structure of the component and resolved telemetry schemas. This is only given as an example and represents my current view of things. A dedicated OTEP will be devoted in the future to the precise definition of their respective structure and format.

text/0240-app-telemetry-schema-vision-roadmap.md
lquerel: Yes it is conceivable to use a resolved telemetry schema, representing a compact and resolved version of a semantic convention registry, in a manner similar to that of current SDK/API clients, where nothing specific is specified/known.
However, the preferred mode (or main purpose) of use for the component telemetry schema in this proposal is to define the telemetry objects produced by an application or a library.

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,659 @@
...
+semantic_conventions:
+  - git_url: <git-url-of-the-semantic-conventions-repository>
+    path: <path-to-the-semantic-conventions-directory-inside-the-git-repo>
+  - url: <url-of-the-semantic-conventions-file>
+
+# The resource field is defined when the component schema is that of an
+# application (as opposed to that of a library). The resource field contains a
+# list of local references to attributes defined in the shared catalog within
+# this file.
+resource:
```
yurishkuro, line 471: This is interesting. When I think of a "component schema" I think of a compile unit, i.e. the schema of the telemetry is "fixed" after the code is compiled. But the resource attributes are not scoped to source code, they can be quite often added dynamically depending on where that compiled unit is deployed. Strictly speaking, you can even say that about other attributes, e.g. a processor in the collector could add certain attributes to telemetry (e.g. "processing timestamp"). So the resulting schema seen at rest may not match the component schema. 

How are the dynamic enrichments should be accounted for in the schema management lifecycle?

lquerel: For this specific case, I was thinking of offering the possibility to define the value of an attribute dynamically by referencing an environment variable (or any other value injection mechanism at the time of deployment). Thus, the attribute is defined in the schema with a clearly identified placeholder. The value of this placeholder is resolved at the time of deployment or by the SDK at runtime in a particular environment. We can also support cases where the identifiers of the attributes themself are not known before deployment, but in this case, these attributes will not be controlled/known at the time of the build.

yurishkuro: but I am not talking about values, rather the actual attributes. If your app is deployed on AWS, it may get AWS-specific resource attributes, which it wouldn't get on GCP.

lquerel: Yes, I had a doubt, which is why the second part of my response includes the case you are talking about, that is, the case where certain attributes (identifier and value) are only known at the time of deployment.

yurishkuro: seems like it's often going to be the case, so if the ingestion pipeline does any kind of schema validation it would need to know these deployment-enriched materialized schemas

text/0240-app-telemetry-schema-vision-roadmap.md
lquerel: I have added an 'Open Questions' section at the end of this OTEP. The purpose is to clearly expose the important questions that will need to be resolved in the following stages/OTEPs. The current OTEP focuses on the overall vision and the major concepts of this initiative. One of the questions is about having one or two types of schema and also includes a format aspect for the schema(s). The point discussed in your comment will be covered in the context of responding to this question in the next OTEP (see roadmap), which I hope to start soon.

jack-berg:
I just wanted to drop and leave a comment since I left some early comments on this OTEP: I'm pretty excited / interested in a lot of these ideas, but don't have the capacity to review deeply or be a big part of this effort, at least for the moment. I haven't seen anything in this proposal that concerns me - please don't wait for my approval to move forward.

jmacd (approved):
Re-read this, still approve.

lquerel:
@tigrannajaryan @yurishkuro @jsuereth @jmacd @pyohannes 
Hello everyone. After discussing with Tigran, I have significantly updated the OTEP to mainly simplify it, focus on the vision and roadmap, and eliminate any aspects that might limit our options in the future. No new concepts are introduced. As this represents a significant change, I would like to request a new approval from you, apologizing for the additional work this entails.

tigrannajaryan (approved):
Thanks for your patience @lquerel !

tigrannajaryan:
I had some old unpublished comments remaining from reviewing the previous iteration of the OTEP, which accidentally got published when I approved this OTEP. Ignore the comments if you receive them in emails, they are no longer relevant. I also deleted the comments from this PR to avoid confusion.

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,384 @@
...
+schemas addressable, easy to traverse, validate, and diff.
+
+The key design principles to be followed in the definition of the Resolved
+Telemetry Schema are:
+
+* **Self-contained**: No external references are allowed. This artifact contains
+  everything required to determine what an application or a library produces in
+  terms of telemetry.
+* **Easy to exchange**: This artifact must be easily accessible from a web
+  server via a URL. This artifact must be small and avoid the repetition of
```
yurishkuro, line 178:
```suggestion
* **Easy to exchange**: This artifact must be easily accessible to actors
  via a URI. This artifact must be small and avoid the repetition of
```

yurishkuro: "web server" is too specific, doesn't need to be used (e.g. maybe schemas can be loaded from a jar)

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,384 @@
...
+feeder, and more.
+
+![Example of Telemetry Schema Hierarchy](./img/0240-otel-weaver-hierarchy.svg)
+
+This hierarchy of telemetry schemas helps large organizations in
+collaborating on the Component Telemetry Schema. It enables different
+aspects of a Component Telemetry Schema to be managed by various teams.
+
+For all the elements that make up the Component Telemetry Schema, a
+general mechanism of annotation or tagging will be integrated in order to
```
yurishkuro, line 215: > annotation or tagging

I understand this is vague on purpose, but one thing I'd want to avoid is having two independent DSLs, one for schema definition and another for annotations. If you noticed in Meta's schema-first telemetry paper, the annotations and the schema both used the exact same Thrift IDL mechanism (which, in turn, means annotations can be arbitrarily nested).

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,384 @@
...
+process would then remove these references.
+
+Proponents of a two-schema approach believe that each format is intended for
+different users and use cases (app/lib developers vs telemetry tool developers),
+and therefore, having two distinct structures and formats would make it easier
+to optimize each for its specific use case (in multiple dimensions). The first
+group, app and lib developers, is much larger than the second and would not need
+to understand the details of the second format.
+
+### What should be the schema(s) be named?
```
yurishkuro, line 291: This is misleading, you can be asking about conceptual names (component vs. resolved schemas), or how they are implemented (same DSL or different)

yurishkuro (approved):

text/0240-app-telemetry-schema-vision-roadmap.md
lquerel: Yes we are on the same page. I will clarify this part of the OTEP.

lquerel: Added the following sentence: This annotation mechanism will be included as part of the Component Telemetry Schema definition.

text/0240-app-telemetry-schema-vision-roadmap.md
lquerel: I have updated this sentence as follows: 
> Furthermore, the group of developers using the Component Telemetry Schema will most likely be much larger than the group of developers who need to understand the details of the Resolved Telemetry Schema. 

I hope it's less misleading.

pyohannes (approved):

jmacd (approved):
This content is substantially similar to the copy I reviewed last time, with less content so that we can discuss/debate more in future OTEPs. Thanks @lquerel!

jsuereth (approved):

tigrannajaryan:

text/0240-app-telemetry-schema-vision-roadmap.md
```diff
@@ -0,0 +1,386 @@
...
+conventions establish a vendor-agnostic and standardized vocabulary to ensure
+more consistent instrumentation over time. This standardization facilitates the
+analysis and utilization of these metadata data across the entire telemetry
+pipeline.
+
+But this approach is not without challenges:
+
+* **Discoverability and Interoperability**: It is difficult to discover a priori
+  and automatically what an application as a whole specifically generates in terms
+  of telemetry and associated metadata.This makes it difficult to integrate with
```
reyang, line 42:
```suggestion
  of telemetry and associated metadata. This makes it difficult to integrate with
```

reyang (approved):

carlosalberto (approved):

tigrannajaryan:
@open-telemetry/specs-approvers this is a major OTEP that will impact many parts of Otel. We have enough approvals to merge it. Last call to review it.

tigrannajaryan:
Thank you @lquerel for the OTEP!

PR #250, opened 2024-02-29
Composite Samplers

PeterF778:
This is another approach for introducing Composite Samplers. The previous one ([A Sampling Configuration proposal](https://github.com/open-telemetry/oteps/pull/240)) proved too large, with some controversial elements.
This OTEP is a split-off from that one, focusing just on one area - new Composite Samplers.


yurishkuro (commented):
The bifurcation of AnyOf/Conjunction and RuleBased doesn't feel right. They are all in the realm of Boolean logic, with or without short circuiting, can we not come up with a more streamlined scheme?

For example, why does RuleBased sampler need to take a list of predicates, instead of one predicate and achieving multi-predicate behavior through composition?

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,118 @@
...
+
+Upon invocation of its `shouldSample` method, it MUST go through the whole list and invoke `shouldSample` method on each delegate sampler, passing the same arguments as received.
+
+`AnyOf` sampler MUST return a `SamplingResult` which is constructed as follows:
+
+- The sampling Decision is based on the delegate sampling Decisions. If all of the delegate Decisions are `DROP`, the composite sampler MUST return `DROP` Decision as well.
+If any of the delegate Decisions is `RECORD_AND_SAMPLE`, the composite sampler MUST return `RECORD_AND_SAMPLE` Decision.
+Otherwise, if any of the delegate Decisions is `RECORD_ONLY`, the composite sampler MUST return `RECORD_ONLY` Decision.
+- The set of span Attributes to be added to the `Span` is the sum of the sets of Attributes as provided by the delegate samplers within their `SamplingResults`s.
+- The `TraceState` to be used with the new `Span` is obtained by cumulatively applying all the potential modfications of the parent `TraceState` by the delegate samplers.
```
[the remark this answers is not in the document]
PeterF778: Good point. I'll add the relevant details soon.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,118 @@
...
+- The `TraceState` to be used with the new `Span` is obtained by cumulatively applying the potential modfications from the First and Second sampler.
+
+If the sampling Decision from the Second sampler is `RECORD_ONLY` or `DROP`, the Conjunction sampler MUST retrun a `SamplingResult` which is constructed as follows:
+
+- The sampling Decision is `DROP`.
+- The set of span Attributes to be added to the `Span` is empty.
+- The `TraceState` to be used with the new `Span` is the passed-in `TraceState`.
+
+
+## Rule based sampling
```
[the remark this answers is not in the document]
PeterF778: Yes, it requires some tidying, thanks.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,118 @@
...
+- `TraceId` of the Span to be created.
+- Name of the Span to be created.
+- Initial set of Attributes of the Span to be created.
+- Collection of links that will be associated with the Span to be created.
+
+## RuleBased
+
+`RuleBased` is a composite sampler which performs `Span` categorization and sampling.
+It takes the following arguments:
+- `SpanKind`
```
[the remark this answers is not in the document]
PeterF778: Better ergonomics and (slightly) better performance. Most non-trivial head-based sampling decisions will be made at ROOT spans. Embedding tests for the span kind in each Predicate would be a nuisance and could be easily omitted by mistake.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,118 @@
...
+
+- The sampling Decision is based on the delegate sampling Decisions. If all of the delegate Decisions are `DROP`, the composite sampler MUST return `DROP` Decision as well.
+If any of the delegate Decisions is `RECORD_AND_SAMPLE`, the composite sampler MUST return `RECORD_AND_SAMPLE` Decision.
+Otherwise, if any of the delegate Decisions is `RECORD_ONLY`, the composite sampler MUST return `RECORD_ONLY` Decision.
+- The set of span Attributes to be added to the `Span` is the sum of the sets of Attributes as provided by the delegate samplers within their `SamplingResults`s.
+- The `TraceState` to be used with the new `Span` is obtained by cumulatively applying all the potential modfications of the parent `TraceState` by the delegate samplers.
+
+Each delegate sampler MUST be given a chance to participate in the sampling decision as described above and MUST see the same _parent_ state. The order of the delegate samplers does not matter, as long as there's no overlap in the Attribute Keys or the trace state keys they use.
+
+## Conjunction
```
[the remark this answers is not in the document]
PeterF778: They are all useful for the example I provided, but I'll make it more explicit.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,118 @@
...
+
+`RuleBased` is a composite sampler which performs `Span` categorization and sampling.
+It takes the following arguments:
+- `SpanKind`
+- list of `Predicate`s
+- list of `Sampler`s
+
+The `RuleBased` sampler SHOULD NOT accept lists of different length, i.e. it SHOULD report an error. Implementations MAY allow for replacing both lists with a list of pairs (`Predicate`, `Sampler`), if this is supported by the platform.
+
+For making the sampling decision, if the `Span` kind matches the specified kind, the sampler goes through the lists in the declared order. If a Predicate matches the `Span` in question, the corresponding `Sampler` will be called to make the final sampling decision. If the `SpanKind` does not match, or none of the Predicates evaluates to `true`, the final decision is `DROP`.
```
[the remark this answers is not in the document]
yurishkuro: In the example at the top you have English description of different categories of spans being sampled. As a user, I would appreciate a mechanism to not only keep those description (maybe in a shorter form of "names"), but also to have observability where I can see which triggers fire how often (either by recording the trigger name as span attribute or emitting metrics labeled with those names). Your current mechanism does not allow for that afaict.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,118 @@
...
+If any of the delegate Decisions is `RECORD_AND_SAMPLE`, the composite sampler MUST return `RECORD_AND_SAMPLE` Decision.
+Otherwise, if any of the delegate Decisions is `RECORD_ONLY`, the composite sampler MUST return `RECORD_ONLY` Decision.
+- The set of span Attributes to be added to the `Span` is the sum of the sets of Attributes as provided by the delegate samplers within their `SamplingResults`s.
+- The `TraceState` to be used with the new `Span` is obtained by cumulatively applying all the potential modfications of the parent `TraceState` by the delegate samplers.
+
+Each delegate sampler MUST be given a chance to participate in the sampling decision as described above and MUST see the same _parent_ state. The order of the delegate samplers does not matter, as long as there's no overlap in the Attribute Keys or the trace state keys they use.
+
+## Conjunction
+
+`Conjunction` is a composite sampler which takes two Samplers (delegates) as the arguments. These delegate samplers will be hereby referenced as First and Second.
```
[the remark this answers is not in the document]
PeterF778: We can always look for better names, but I think EachOf would a particularly bad choice, as it would imply some deeper sort of similarity or symmetry between the two, which is not there.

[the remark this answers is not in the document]
PeterF778: Actually, yes, I realized that we need EachOf, in addition to Conjunction. An update is coming.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,118 @@
...
+
+The need for configuring head sampling has been explicitly or implicitly indicated in several discussions, both within the [Samplig SIG](https://docs.google.com/document/d/1gASMhmxNt9qCa8czEMheGlUW2xpORiYoD7dBD7aNtbQ) and in the wider community. Some of the discussions are going back a number of years, see for example
+
+- issue [173](https://github.com/open-telemetry/opentelemetry-specification/issues/173): Way to ignore healthcheck traces when using automatic tracer across all languages?
+- issue [1060](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1060): Exclude URLs from Tracing
+- issue [1844](https://github.com/open-telemetry/opentelemetry-specification/issues/1844): Composite Sampler
+
+## The Goal
+
+The goal of this proposal is to extend the set of standard SDK samplers by _Composite Samplers_ which make building advanced sampling configuration easier.
```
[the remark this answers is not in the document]
PeterF778: Good point, I'll re-word it.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,118 @@
...
+`AnyOf` is a composite sampler which takes a non-empty list of Samplers (delegates) as the argument.
+
+Upon invocation of its `shouldSample` method, it MUST go through the whole list and invoke `shouldSample` method on each delegate sampler, passing the same arguments as received.
+
+`AnyOf` sampler MUST return a `SamplingResult` which is constructed as follows:
+
+- The sampling Decision is based on the delegate sampling Decisions. If all of the delegate Decisions are `DROP`, the composite sampler MUST return `DROP` Decision as well.
+If any of the delegate Decisions is `RECORD_AND_SAMPLE`, the composite sampler MUST return `RECORD_AND_SAMPLE` Decision.
+Otherwise, if any of the delegate Decisions is `RECORD_ONLY`, the composite sampler MUST return `RECORD_ONLY` Decision.
+- The set of span Attributes to be added to the `Span` is the sum of the sets of Attributes as provided by the delegate samplers within their `SamplingResults`s.
```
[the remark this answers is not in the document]
PeterF778: Attribute collision is a possibility, but it is of course unlikely. None of the standard samplers or AFAIK any samplers from the contrib repository set any attributes today. The situation here is not much different from Span attributes that can be set by different instrumentation actors.

PeterF778:
> The bifurcation of AnyOf/Conjunction and RuleBased doesn't feel right. They are all in the realm of Boolean logic, with or without short circuiting, can we not come up with a more streamlined scheme?
> 
> For example, why does RuleBased sampler need to take a list of predicates, instead of one predicate and achieving multi-predicate behavior through composition?

`RuleBased` is based on easy to understand _if-then-else_ paradigm. Alternative schemas may look attractive, but I'm afraid they would provide less readable configurations.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,204 @@
...
+If the sampling Decision from the Second sampler is `RECORD_AND_SAMPLE`, the Conjunction sampler MUST return a `SamplingResult` which is constructed as follows:
+
+- The sampling Decision is `RECORD_AND_SAMPLE`.
+- The set of span Attributes to be added to the `Span` is the sum of the sets of Attributes as provided by the First and the Second sampler.
+- The `TraceState` to be used with the new `Span` is obtained by cumulatively applying the potential modfications from the First and Second sampler, with special handling of the `th` sub-key (the sampling rejection `THRESHOLD`) for the `ot` entry as described below.
+
+If both First and Second samplers provided `th` entry in the returned `TraceState`, the resulting `TraceState` MUST contain `th` entry with the `THRESHOLD` being maximum of the `THRESHOLD`s provided by the First and Second samplers. Otherwise, the `th` entry MUST be removed.
```
oertl, line 86: Doesn't the composite sampler break consistent sampling? The threshold value must always be chosen independently of the `R`-value. However, the use of the threshold value of the second sampler depends on the decision of the first sampler, whose decision on sampling depends on `T <= R` and thus also on `R`. Therefore, there is an indirect dependency on `R` which is not allowed.

PeterF778: Yes, you are right. It looks like the `Conjunction` sampler cannot be made compatible with consistent probability sampling. However, it is still needed, for example if someone want sto use a reservoir-based rate limiting sampler as the Second sampler.

oertl: When composing samplers, the composition must be based on the thresholds and must happen before the actual sampling decision, which compares the threshold with the random value. In other words, referring to the Java prototypes, there must be only a single `shouldSample` invovation for every span (https://github.com/open-telemetry/opentelemetry-java-contrib/blob/6a44e04a76bbe6b9844555f9b37e777c2a23103f/consistent-sampling/src/main/java/io/opentelemetry/contrib/sampler/consistent/ConsistentSampler.java#L232).

However, the sampler may delegate during the threshold computation to other samplers like it is done, for example, in the parent-based sampler (https://github.com/open-telemetry/opentelemetry-java-contrib/blob/6a44e04a76bbe6b9844555f9b37e777c2a23103f/consistent-sampling/src/main/java/io/opentelemetry/contrib/sampler/consistent56/ConsistentParentBasedSampler.java#L40).

To cover the example above, it should be possible to extend the rate-limiting consistent sampler and to provide a factory method such as (compare https://github.com/open-telemetry/opentelemetry-java-contrib/blob/6a44e04a76bbe6b9844555f9b37e777c2a23103f/consistent-sampling/src/main/java/io/opentelemetry/contrib/sampler/consistent/ConsistentSampler.java#L117)
```
ConsistentSampler rateLimited(double targetSpansPerSecondLimit, double adaptationTimeSeconds, ConsistentSampler delegateSampler);
```
that returns a rate-limiting sampler, that delegates the threshold computation to the `delegateSampler` and adapts the returned threshold in such a way to meet the rate limit. For that the rate-limiting sampler has to keep track of some statistics about the number of invovations and the thresholds returned by the delegate.

PeterF778: I had considered adding a delegate sampler to the `RateLimitingConsistentSampler`, but opted out, because it addresses just one sampler, and, furthermore, I believe that when dealing with consistent probability sampling, `EachOf` handles this case correctly.
`Conjunction` is needed for reservoir based samplers.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,118 @@
...
+
+## RuleBased
+
+`RuleBased` is a composite sampler which performs `Span` categorization and sampling.
+It takes the following arguments:
+- `SpanKind`
+- list of `Predicate`s
+- list of `Sampler`s
+
+The `RuleBased` sampler SHOULD NOT accept lists of different length, i.e. it SHOULD report an error. Implementations MAY allow for replacing both lists with a list of pairs (`Predicate`, `Sampler`), if this is supported by the platform.
```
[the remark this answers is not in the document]
PeterF778: Good point. I changed the description, hopefully it is less problematic now.

text/0250-Composite_Samplers.md
```diff
@@ -76,20 +76,20 @@ Each delegate sampler MUST be given a chance to participate in the sampling deci
...
 
 - The sampling Decision is `RECORD_AND_SAMPLE`.
 - The set of span Attributes to be added to the `Span` is the sum of the sets of Attributes as provided by the First and the Second sampler.
 - The `TraceState` to be used with the new `Span` is obtained by cumulatively applying the potential modfications from the First and Second sampler, with special handling of the `th` sub-key (the sampling rejection `THRESHOLD`) for the `ot` entry as described below.
 
-If both First and Second samplers provided `th` entry in the returned `TraceState`, the resulting `TraceState` MUST contain `th` entry with the `THRESHOLD` being maximum of the `THRESHOLD`s provided by the First and Second samplers. Otherwise, the `th` entry MUST be removed.
+If both First and Second samplers provided `th` entry in the returned `TraceState`, and the value of the `THRESHOLD` from the First sampler is `0`, then the resulting `TraceState` MUST contain `th` entry with the `THRESHOLD`as provided by the Second sampler. Otherwise, the `th` entry MUST be removed.
```
oertl, line 86: I do not understand how this definition allows to satisfy the last requirement in the given example without subsequent reservoir sampling. To my understanding, the sampling probability should be chosen first according to the rules
> - for root spans:
>   - drop all `/healthcheck` requests
>   - capture all `/checkout` requests
>   - capture 25% of all other requests
> - for non-root spans
>   - follow the parent sampling decision
>   - however, capture all calls to service `/foo` (even if the trace will be incomplete)

If the resulting sampling probabilities would lead to more than 1000 sampled spans/minute, the sampling probabilities should be adjusted (reduced) to satisfy the last rule
> * in any case, do not exceed 1000 spans/minute

In my interpretation this means that, if the rate of sampled spans would exceed 1000/min, the sampling probabilities as a result of the preceding rules need to be adjusted by the same reduction factor such that the limit is finally met. For that, the Second sampler must know the chosen threshold of the First sampler in order to be able to estimate the rate of sampled spans from its history (exponential smoothing) and also to apply the corresponding adjustment to the threshold. This is not possible, if the threshold is taken from the second sampler only if the threshold of the first sampler is zero.

PeterF778: If we want to support the given example exclusively using consistent probability samplers, then the right solution is to use `EachOf` (as provided in the document) rather than `Conjunction`. In particular, using the rate limiting CP sampler with exponential smoothing as the Second delegate for `Conjunction` will not work correctly. The issue is not with the `th` value, but with overshooting the limit if the First delegate does not always decide to sample.
However, `Conjunction` is expected to work correctly if a reservoir (non CP) sampler is used there.

Admittedly, `Conjunction` looks strange. I'm trying to kill a few birds with a single stone here. It should be used with caution, and only when `EachOf` is not sufficient.

oertl: > However, Conjunction is expected to work correctly if a reservoir (non CP) sampler is used there.

Unfortunately, reservoir sampling is not possible with the current sampler interface, as it requires an immediate sampling decision for every span. 

I am still wondering what is expected in the given example. For simplicity, assume that there are 2 categories `A` and `B` of spans. Assume there are 100 spans/min and 50% belong to `A` and 50% belong to `B`. Now we define to sample `A` with a probability of 60% and `B` with a probability of 40%. That means we would sample 30 spans/min from `A` and 20 spans/min from `B`.

If we do not want to sample more than 10 spans/min overall, what is our expectation?
1) Sample on average 5 spans/min from `A` and 5 spans/min from `B`?
2) Or, sample on average 6 spans/min from `A` and 4 spans/min from `B`?

With `EachOf`only the first can be achieved. For the second, we would need a rate-limiting consistent sampler, that adjusts the thresholds of the delegates by a factor that is determined from the history of thresholds using, for example, some kind of exponential smoothing.

oertl: No, this is not the same. Just consider the case where the ratio between the number of A and B spans changes. For example, from the assumed 50:50 to 100:0. Then I still want to sample 10 spans/min without reconfiguring the samplers.

PeterF778: > Unfortunately, reservoir sampling is not possible with the current sampler interface, as it requires an immediate sampling decision for every span.

An example of sampler that I had in mind here is Jaeger [RateLimitedSampler](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/RateLimitingSampler.java).

oertl: > An example of sampler that I had in mind here is Jaeger [RateLimitedSampler](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/RateLimitingSampler.java).

I fear that this sampling approach cannot be turned into a consistent sampler, as the sampling probability is not known.

PeterF778: > No, this is not the same. Just consider the case where the ratio between the number of A and B spans changes. For example, from the assumed 50:50 to 100:0. Then I still want to sample 10 spans/min without reconfiguring the samplers.

I've read your requirement as to provide the ratio of sampled spans 6:4. The solution works for a wide range of inputs, but obviously breaks if one of the span categories is critically underrepresented. A "fix" would require spilling some capacity from one category to another. How would your proposed adjustment of threshold handle this?

oertl: A solution could be the following:

Dependent on the category, the first layer of samplers would choose the (normalized, decimal) thresholds `T_A=0.4` and `T_B=0.6` corresponding to 60% and 40% sampling probability, respectively. The chosen thresholds are further passed to the rate limiting sampler. If the ratio is 50:50, this sampler observes using exponential smoothing that the (weighted) average sampling probability is 50% and also that the overall rate of spans is 100/min. Therefore the sampler concludes, if the thresholds are kept as they are, that the rate of sampled spans would be 50/min. Therefore the rate limiting sampler decides to adjust the thresholds such that the goal of 10/min is achieved. It reduces the sampling probability by a factor 5, which means that the thresholds get adjusted to `T_A=0.88` and `T_B=0.92`, respectively.

If now the ratio changes suddenly, for example, to 100:0, sooner or later (dependent on the smoothing parameter) the sampler will realize that the average sampling probability has changed to 60% and therefore it has to reduce all sampling probabilities by a factor of 6 and to adjust all thresholds correspondingly.

PeterF778: Got it. So the rate limiting sampler keeps two averages: incoming span rate and threshold value as reported by the delegate. And this gives _balancing_ behavior rather than _equalizing_.

jmacd (commented):
Haven't been able to finish this review. (Still need to understand Approach 2.) Sorry for the delay!

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,305 @@
...
+`AnyOf` is a composite sampler which takes a non-empty list of Samplers (delegates) as the argument. The intention is to make `RECORD_AND_SAMPLE` decision if __any of__ the delegates decides to `RECORD_AND_SAMPLE`.
+
+Upon invocation of its `shouldSample` method, it MUST go through the whole list and invoke `shouldSample` method on each delegate sampler, passing the same arguments as received.
+
+`AnyOf` sampler MUST return a `SamplingResult` which is constructed as follows:
+
+- The sampling Decision is based on the delegate sampling Decisions. If all of the delegate Decisions are `DROP`, the composite sampler MUST return `DROP` Decision as well.
+If any of the delegate Decisions is `RECORD_AND_SAMPLE`, the composite sampler MUST return `RECORD_AND_SAMPLE` Decision.
+Otherwise, if any of the delegate Decisions is `RECORD_ONLY`, the composite sampler MUST return `RECORD_ONLY` Decision.
+- The set of span Attributes to be added to the `Span` is the sum of the sets of Attributes as provided by the delegate samplers within their `SamplingResults`s.
```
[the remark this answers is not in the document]
PeterF778: The samplers are free, in general, to set span attributes even when their decision is DROP - although I do not see a use case for that. However, I'd like `AnyOf` to preserve the behavior of the component samplers as much as possible. In particular, for any sampler X, I'd like AnyOf([X]) to be equivalent of X.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,305 @@
...
+If any of the delegate Decisions is `DROP`, the composite sampler MUST return `DROP` Decision.
+Otherwise, if any of the delegate Decisions is `RECORD_ONLY`, the composite sampler MUST return `RECORD_ONLY` Decision.
+- The set of span Attributes to be added to the `Span` is the sum of the sets of Attributes as provided by the delegate samplers within their `SamplingResults`s.
+- The `TraceState` to be used with the new `Span` is obtained by cumulatively applying all the potential modfications of the parent `TraceState` by the delegate samplers, with special handling of the `th` sub-key (the sampling rejection `THRESHOLD`) for the `ot` entry as described below.
+
+If the final sampling Decision is `DROP` or `RECORD_ONLY`, the `th` entry MUST be removed.
+If the sampling Decision is `RECORD_AND_SAMPLE`, and there's no `th` entry in any of the `TraceState` provided by the delegates that decided to `RECORD_AND_SAMPLE`, the `th` entry MUST be also removed.
+Otherwise, the resulting `TraceState` MUST contain `th` entry with the `THRESHOLD` value being the maximum of all the `THRESHOLD` values as reported by those delegates that decided to `RECORD_AND_SAMPLE`.
```
oertl, line 77: I think the `th` field must always be removed, except in the case where all delegate samplers define a `th` value.

oertl: Consider the case with one consistent and one inconsistent sampler, both sampling with a probability of 50%. Therefore, a span is recorded with a probability of 25%. The `th` value would be taken from the consistent sampler representing a sampling probability of 50%, which is a mismatch with the actual sampling probability.

PeterF778: Yes, and this is what I tried to say, I guess I was not clear enough.

oertl: Yes, you are right. I'm sorry, I overlooked it.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,305 @@
...
+
+`EachOf` sampler MUST return a `SamplingResult` which is constructed as follows:
+
+- The sampling Decision is based on the delegate sampling Decisions. If all of the delegate Decisions are `RECORD_AND_SAMPLE`, the composite sampler MUST return `RECORD_AND_SAMPLE` Decision as well.
+If any of the delegate Decisions is `DROP`, the composite sampler MUST return `DROP` Decision.
+Otherwise, if any of the delegate Decisions is `RECORD_ONLY`, the composite sampler MUST return `RECORD_ONLY` Decision.
+- The set of span Attributes to be added to the `Span` is the sum of the sets of Attributes as provided by the delegate samplers within their `SamplingResults`s.
+- The `TraceState` to be used with the new `Span` is obtained by cumulatively applying all the potential modfications of the parent `TraceState` by the delegate samplers, with special handling of the `th` sub-key (the sampling rejection `THRESHOLD`) for the `ot` entry as described below.
+
+If the final sampling Decision is `DROP` or `RECORD_ONLY`, the `th` entry MUST be removed.
```
oertl, line 75: If all delegated samplers are consistent, it still makes sense to set the `th` value even in the case of negative sampling decisions. 

One could perhaps say that, in general, for both `EachOf` and `AnyOf`, the `th` value must be set if and only if all delegated samplers are consistent (regardless of the sampling decision).

PeterF778: I like this idea, but didn't we expect that `th` is removed after a negative sampling decision? Similarly to how the p-values were handled?

oertl: I see. If we never put `th` to the trace state for negative sampling decisions, which I think makes sense, then again what is the purpose of an empty `th` field representing 0% probability? @jmacd

text/0250-Composite_Samplers.md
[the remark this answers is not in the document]
PeterF778: I take it back. We have to distinguish between attributes added for DROP decisions from all other attributes, so yes, you are right.

text/0250-Composite_Samplers.md
PeterF778: Added Approach Two, describing this idea.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,305 @@
...
+If the sampling Decision is `RECORD_AND_SAMPLE`, and there's no `th` entry in any of the `TraceState` provided by the delegates that decided to `RECORD_AND_SAMPLE`, the `th` entry MUST be also removed.
+Otherwise, the resulting `TraceState` MUST contain `th` entry with the `THRESHOLD` value being the minimum of all the `THRESHOLD` values as reported by those delegates that decided to `RECORD_AND_SAMPLE`.
+
+Each delegate sampler MUST be given a chance to participate in the sampling decision as described above and MUST see the same _parent_ state. The order of the delegate samplers does not matter, as long as there's no overlap in the Attribute Keys or the trace state keys (other than `th`) that they use.
```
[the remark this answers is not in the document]
PeterF778: I updated the description for that.

text/0250-Composite_Samplers.md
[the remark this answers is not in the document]
PeterF778: Added clarification on how this should be handled.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,308 @@
...
+`Conjunction` is a composite sampler which takes two Samplers (delegates) as the arguments. These delegate samplers will be hereby referenced as First and Second. This kind of composition forms conditional chaining of both samplers. 
+
+Upon invocation of its `shouldSample` method, the Conjunction sampler MUST invoke `shouldSample` method on the First sampler, passing the same arguments as received, and examine the received sampling Decision. Upon receiving `DROP` or `RECORD_ONLY` decision it MUST return the SamplingResult from the First sampler, and it MUST NOT proceed with querying the Second sampler. If the sampling decision from the First sampler is `RECORD_AND_SAMPLE`, the Conjunction sampler MUST invoke `shouldSample` method on the Second sampler, effectively passing the `TraceState` received from the First sampler as the parent trace state.
+
+If the sampling Decision from the Second sampler is `RECORD_AND_SAMPLE`, the Conjunction sampler MUST return a `SamplingResult` which is constructed as follows:
+
+- The sampling Decision is `RECORD_AND_SAMPLE`.
+- The set of span Attributes to be added to the `Span` is the sum of the sets of Attributes as provided by both samplers.
```
oertl, line 90: Conflict resolution of attributes is missing here. In contrast to `EachOf` and `AnyOf`, conjunction does not suggest symmetry. So conflict resolution does not need to be symmetric in this case.
```suggestion
- The set of span Attributes to be added to the `Span` is the union of the sets of Attributes as provided by both samplers.
```

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,308 @@
...
+#### GetThreshold
+
+This is a routine/function/method for all Consistent Probability Samplers. Its purpose is to query the sampler to provide the rejection threshold value they would use had they been asked to make a sampling decision for a given span, however, without constructing the actual sampling Decision.
+
+#### Required Arguments for GetThreshold:
+
+The arguments are the same as for [`ShouldSample`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#shouldsample).
+
+- `Context` with parent Span.
+- `TraceId` of the Span to be created.
```
oertl, line 225: I think `TraceId` should not be an argument as the threshold must not depend on the randomness. By providing the trace ID, there is some risk that the lower 56-bits will be (mis-)used for the threshold computation.
```suggestion
```

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,308 @@
...
+
+To make this approach possible, all Consistent Probability Samplers need to implement the following API, in addition to the standard Sampler API. This extension will be used by all composite samplers in Approach Two, as listed in the next sections.
+
+#### GetThreshold
+
+This is a routine/function/method for all Consistent Probability Samplers. Its purpose is to query the sampler to provide the rejection threshold value they would use had they been asked to make a sampling decision for a given span, however, without constructing the actual sampling Decision.
+
+#### Required Arguments for GetThreshold:
+
+The arguments are the same as for [`ShouldSample`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#shouldsample).
```
oertl, line 222: (see my comment below)

```suggestion
The arguments are the same as for [`ShouldSample`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#shouldsample) except for the `TraceId`.
```

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,308 @@
...
+         (http.url == /foo) => ConsistentAlwaysOn
+       }
+     ),
+     1000 * 60
+   )
+```
+
+### Limitations of composite samplers in Approach Two
+
+While making sampling decisions with samplers from Approach Two is more efficient and avoids dealing with non-mainstream cases, it puts some limits on the capabilities of the Consistent Probability Samplers. In particular, a custom CP sampler that wishes to add a span `Attribute` or modify TraceState will be out of luck if it is used as a delegate.
```
oertl, line 299: As it is explicitly mentioned as a limitation. I wonder why a consistent sampler would add any span attributes or change the trace state at all (other than `th` and `rv`)? What would be a use case for this?

text/0250-Composite_Samplers.md
PeterF778: Great suggestion, thanks!

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,308 @@
...
+
+Upon invocation of its `shouldSample` method, it MUST go through the whole list and invoke `shouldSample` method on each delegate sampler, passing the same arguments as received, and collecting the delegates' sampling Decisions.
+
+`EachOf` sampler MUST return a `SamplingResult` with the following elements.
+
+- If all of the delegate Decisions are `RECORD_AND_SAMPLE`, the composite sampler MUST return `RECORD_AND_SAMPLE` Decision as well.
+If any of the delegate Decisions is `DROP`, the composite sampler MUST return `DROP` Decision.
+Otherwise, if any of the delegate Decisions is `RECORD_ONLY`, the composite sampler MUST return `RECORD_ONLY` Decision.
+- If the resulting sampling Decision is `DROP`, the set of span Attributes to be added to the `Span` is empty. Otherwise, it is the sum of the sets of Attributes as provided by the delegate samplers within their `SamplingResults`s. In case of conflicting attribute keys, the attribute definition from the last delegate that uses that key takes effect.
```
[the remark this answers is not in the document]
PeterF778: Removing conflicting attributes is an interesting concept. However, if we were to set attributes using `Span.SetAttribute`, the specification says
_Setting an attribute with the same key as an existing attribute SHOULD overwrite the existing attribute's value_, so there would be an inconsistency. Adding entries with conflicting keys to `TraceState` also overwrites the previous entries - another inconsistency.
And we would be left with an interesting corner case: what to do if two samplers attempt to set Attributes with the same keys _and_ values?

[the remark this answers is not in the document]
oertl: > Upon invocation of its `shouldSample` method, it MUST go through the whole list and invoke `shouldSample` method on each delegate sampler, passing the same arguments as received, and collecting the delegates' sampling Decisions.

I understand this means the delegate samplers must be called with the same attributes. After invocating the delegates we have different sets of attributes that must be merged to a final result. Given an initial set of attributes, individual copies of this initial set are then passed to the delegates, who may set/modify attributes. As the final step we have to combine all the resulting sets of attributes including the initial one. To achieve symmetry, one could define that an attribute is only changed/set if the modifications by the delegates are not conflicting meaning that they don't try to set different values for the same key. If we take the initial set of attributes and apply only the non-conflicting modifications using `Span.SetAttribute` we should be fine.

text/0250-Composite_Samplers.md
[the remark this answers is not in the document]
PeterF778: Yes, I expect both approaches to be useful, targeting different users. I do not think we should abandon either one in favor of the other.
With respect to your point 1 above, the intention is to handle the `th` value "correctly". But we cannot just say that in all cases if any non-CP sampler is used then the `th` value needs to be erased. Both `RuleBased` and `AnyOf` samplers can have non-CP delegates, yet still deliver valid `th` values.

It is possible to use Approach Two delegates within Approach One samplers, and this is more or less what your example does. So it is a workaround for some limitations of Approach Two.

PeterF778: Good question. With trace state, I imagine a head sampler can _mark_ some traces for downstream (head) samplers and for further trace processing, effectively classifying traces according to arbitrary criteria. Obviously, this requires writing custom samplers, but these samplers can be still CP compliant.
Adding Attributes is similar, even though it works only on the Span level and only for sampled Spans.

oertl: Got it. I think this could also be achieved with Approach Two by extending the ConsistentSampler interface by methods other than `getThreshold`such as `getAttributes` and defining some logic how those attributes are finally added in case the overall sampling decision is positive. So I would not necessarily see that as limitation of Approach Two as we also have to define some logic for the attributes for Approach One.

PeterF778: Yes, and we could do a similar thing with `TraceState` by providing `adjustTraceState`. For that call, we could pass the final `SamplingDecision` (which can be different from what the delegate decided). One difficulty is that for `RuleBased`, the Predicates would get re-evaluated in order to route the call to the correct delegate.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,308 @@
...
+#### Required Arguments for ConsistentRateLimiting
+
+- ConsistentProbabilitySampler (delegate)
+- maximum sampling (throughput) target rate
+
+The sampler SHOULD measure and keep the average rate of incoming spans, and therefore also of the desired ratio between the incoming span rate to the target span rate.
+Upon invocation of its `GetThreshold` function, the composite sampler MUST get the threshold from the delegate sampler, passing the same arguments as received.
+If using the obtained threshold value as the final threshold would entail sampling more spans than the declared target rate, the sampler SHOULD increase the threshold to a value that would meet the target rate. Several algorithms can be used for threshold adjustment, no particular behavior is prescried by the specification though.
+
+When using `ConsistentRateLimiting` in our requirements example as a replacement for `EachOf` and `RateLimiting`, we are left with no use case for any direct equivalent to `EachOf` in Approach Two.
```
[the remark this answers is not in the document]
jmacd: I see --
would you agree that the concept of a `ConsistentEachOf` sampler can be defined easily, and that it reduces to the maximum-threshold of the delegates. It's well-defined, but not necessary for the example being developed, I think.

text/0250-Composite_Samplers.md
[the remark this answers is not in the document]
jmacd: Got it. I think we're in agreement on this topic. I think our next step is really to prototype what this looks like in a prototype implementation.

jmacd: I am thinking in the same direction. I have a sandbox where I've begun drafting my hopes&dreams for a V2 sampler API, right now I have something like:

```
type SamplingResult struct {
	Decision SamplingDecision
	Mutation func(trace.TraceState) ([]attribute.KeyValue, trace.TraceState)
}
```

for an Approach-2 sampler, the idea is that a composition of sampler delegates will first evaluate the decision, then it would modify the special TraceState fields, then for the positive decision makers it would allow them (sequentially) to apply modified attributes and modify the trace state in other ways. I think it will be fine to say that attributes are applied in the same order as the original delegate list, and that later samplers clobber earlier samplers in case of conflicts. I will flush this out more, but I think we can get past the trouble associated with modifying attributes and tracestate independent of the decision & essential tracestate updates.

text/0250-Composite_Samplers.md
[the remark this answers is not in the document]
PeterF778: Yes. I do not see any difficulties in defining the behavior of `ConsistentEachOf`, it is just its usefulness that is questionable. As soon as we find a nice use case for it, it can be added to this spec.

text/0250-Composite_Samplers.md
PeterF778: I modified the document to present this two-phase process.

jmacd: I've begun to prototype https://github.com/jmacd/opentelemetry-go/pull/877

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,337 @@
...
+
+If the sampling Decision from the Second sampler is `RECORD_AND_SAMPLE`, the Conjunction sampler MUST return a `SamplingResult` which is constructed as follows:
+
+- The sampling Decision is `RECORD_AND_SAMPLE`.
+- The set of span Attributes to be added to the `Span` is the union of the sets of Attributes as provided by both samplers.
+- The `Tracestate` to be used with the new `Span` is as provided by the Second sampler, with special handling of the `th` sub-key (the sampling rejection `THRESHOLD`) for the `ot` entry.
+If the First sampler did not provide `th` entry in the returned `Tracestate`, or if the value of the corresponding `THRESHOLD` is not `0`, then the `th` entry MUST be removed from the resulting `Tracestate`.
```
kalyanaj, line 96: I didn't quite get the part about "or if the value of the corresponding `THRESHOLD` is not `0`,". If the first sampler returned RECORD_AND_SAMPLE, it is possible that the threshold could be non-zero, correct?  Can you please clarify?

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,337 @@
...
+
+If the sampling Decision from the Second sampler is `DROP` or `RECORD_ONLY`, the Conjunction sampler MUST return a `SamplingResult` which is constructed as follows:
+
+- The sampling Decision is `DROP`.
```
kalyanaj, line 100: If the sampling decision from the second sampler is RECORD_ONLY, shouldn't this composite sampler be returning that - didn't quite get why the final decision is DROP.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,337 @@
...
+
+Each delegate sampler MUST be given a chance to participate in the sampling decision as described above and MUST see the same _parent_ state. The resulting sampling Decision does not depend on the order of the delegate samplers.
+
+### Conjunction
+
+`Conjunction` is a composite sampler which takes two Samplers (delegates) as the arguments. These delegate samplers will be hereby referenced as First and Second. This kind of composition forms conditional chaining of both samplers. 
+
+Upon invocation of its `shouldSample` method, the Conjunction sampler MUST invoke `shouldSample` method on the First sampler, passing the same arguments as received, and examine the received sampling Decision. Upon receiving `DROP` or `RECORD_ONLY` decision it MUST return the SamplingResult from the First sampler, and it MUST NOT proceed with querying the Second sampler. If the sampling decision from the First sampler is `RECORD_AND_SAMPLE`, the Conjunction sampler MUST invoke `shouldSample` method on the Second sampler, effectively passing the `Tracestate` received from the First sampler as the parent trace state.
```
kalyanaj, line 89: For the case where the first sampler returns DROP or RECORD_ONLY, can we also clarify the behavior of the span attributes and tracestate?  For the RECORD_ONLY case, should they be retained? Currently, the spec doesn't seem to be specifying that aspect.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,337 @@
+# Composite Samplers Proposal
+
+This proposal addresses head-based sampling as described by the [Open Telemetry SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#sampling).
+It introduces additional _composite samplers_. Composite samplers use other samplers (_delegates_ or _children_) to make sampling decisions. The composite samplers invoke the delegate samplers, but eventually make the final call.
+
+The new samplers proposed here are mostly compatible with Consistent Probability Samplers. For detailed description of this concept see [probability sampling](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md). However, the technical details in that document are outdated. For the current proposal see [OTEP 235](https://github.com/open-telemetry/oteps/blob/main/text/trace/0235-sampling-threshold-in-trace-state.md). Also see Draft PR 3910 [Probability Samplers based on W3C Trace Context Level 2](https://github.com/open-telemetry/opentelemetry-specification/pull/3910).
```
kalyanaj, line 6: Since OTEP 235 supersedes the earlier experimental spec, my suggestion is to link *only* to OTEP 235 in this spec (both here and later below in the section Example - sampling configuration 2). Otherwise, it can be confusing for readers. If they need more info on the earlier version, I believe in OTEP 235 we link to the earlier version of the spec.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,337 @@
...
+
+The need to encode and decode the `Tracestate` multiple times affects performance of the composite samplers. This drawback is eliminated in Approach Two.
+
+## Approach Two
+
+A principle of operation for Approach Two is that `ShouldSample` is invoked only once, on the root of the tree formed by composite samplers. All the logic provided by the composition of samplers is handled by calculating the threshold values, delegating the calculation downstream as necessary.
+
+### New API
+
+To make this approach possible, all Consistent Probability Samplers which participate in the samplers composition need to implement the following API, in addition to the standard Sampler API. We will use the term _Composable Sampler_ to dentote Consistent Probability Samplers which provide the new API and conform to the rules described here.
```
kalyanaj, line 220:
```suggestion
To make this approach possible, all Consistent Probability Samplers which participate in the samplers composition need to implement the following API, in addition to the standard Sampler API. We will use the term _Composable Sampler_ to denote Consistent Probability Samplers which provide the new API and conform to the rules described here.
```

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,337 @@
...
+
+### Limitations of composite samplers in Approach Two
+
+Making sampling decisions with samplers from Approach Two is more efficient than in Approach One, especially if, platform permitting, `null` values can be used for `GetAttributes` and `UpdateTraceState` functions to represent the prevailing trivial cases of _no-new-attributes_ and _no-special-trace-state-keys_. The only limitation of this approach that it operates exclusively within the domain of Composable Samplers (a subset of Consistent Probability Samplers).
+
+Developers of Composable Samplers should consider that the sampling Decision they advise might be different from the final sampling Decision.
+
+## Prior art
+
+A number of composite samplers are already available as independent contributions
```
kalyanaj, line 332: FYI, I have a few more here (in case you want to link to them) at:
- https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/trace/links-based-sampler
- https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/trace/stratified-sampling-example
- https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/trace/tail-based-sampling-span-level

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,337 @@
...
+
+#### Constructing `SamplingResult`
+
+The process of constructing the final `SamplingResult` in response to a call to `ShouldSample` on the root sampler of the composite samplers tree consists of the following steps.
+
+- The sampler gets its own `SamplingAdvice`, it is a recursive process as described below (unless the sampler is a leaf),
+- The sampler compares the received THRESHOLD value with the trace Randomness value to arrive at the final sampling `Decision`,
+- In case of a positive sampling decision the sampler calls the received `GetAttributes` function to determine the set of Attributes to be added to the Span, in most cases it will be a recursive step,
+- The sampler calls the received `UpdateTraceState` function passing the parent `Tracestate` and the final sampling `Decision` to get the new `Tracestate` to be associated with the Span - again, in most cases this is a recursive step,
+- The sampler modfies (or removes) the `th` value for the `ot` key in the `Tracestate` according to the final sampling `Decision` and the THRESHOLD used in the second step above.
```
kalyanaj, line 253:
```suggestion
- The sampler modifies (or removes) the `th` value for the `ot` key in the `Tracestate` according to the final sampling `Decision` and the THRESHOLD used in the second step above. 
```

text/0250-Composite_Samplers.md
PeterF778: Yes, this is a tricky part. If the First sampler decides to sample only _some_ spans (i.e. the `THRESHOLD` is not `0`), we cannot, in general, assure that the composite sampler's behavior is correct (even if we will use maximum of the `THRESHOLD` values from First and Second delegate). See @oertl's comments from March 7 (click on Show resolved).

PeterF778: Actually, giving it more thoughts, I'm inclined to relax this condition and allow any `THRESHOLD` values from the First delegate. It opens an opportunity for abuse by a malicious Second sampler, but such abuse (using the `r` value for determining sampling priority) can happen anywhere. I believe that as long as First and Second delegates are well-behaved, we can use the maximum `THRESHOLD` value for the final result. CC: @oertl

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,307 @@
...
+- The `TraceState` to be used with the new `Span` is obtained by cumulatively applying all the potential modfications of the parent `TraceState` by the delegate samplers. In case of conflicting entry keys, the entry definition provided by the last delegate that uses that key takes effect. However, the `th` sub-key (the sampling rejection `THRESHOLD`) for the `ot` entry gets special handling as described below.
+
+If the final sampling Decision is `DROP` or `RECORD_ONLY`, the `th` entry MUST be removed.
+If the sampling Decision is `RECORD_AND_SAMPLE`, and there's no `th` entry in any of the `TraceState` provided by the delegates, the `th` entry MUST be also removed.
+Otherwise, the resulting `TraceState` MUST contain `th` entry with the `THRESHOLD` value being the maximum of all the `THRESHOLD` values as reported the delegates.
+
+Each delegate sampler MUST be given a chance to participate in the sampling decision as described above and MUST see the same _parent_ state. The resulting sampling Decision does not depend on the order of the delegate samplers.
+
+### Conjunction
```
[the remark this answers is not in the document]
PeterF778: Both Chain and Sequence suggest, IMHO, a possibility to have more steps (delegates) than just two. And I do not want to deal with more than 2 delegates - this kind of composition is rarely needed for two, and I do not have a use case for more than two. It is a bit like IfThen construct (I'm not suggesting this name though, this could be too confusing).
I think Chain is acceptable, and I'm open to any new suggestions.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,307 @@
...
+- The `TraceState` to be used with the new `Span` is obtained by cumulatively applying all the potential modifications of the parent `TraceState` by the delegate samplers. In case of conflicting entry keys, the entry definition provided by the last delegate that uses that key takes effect. However, the `th` sub-key (the sampling rejection `THRESHOLD`) for the `ot` entry gets special handling as described below.
+
+If the final sampling Decision is `DROP` or `RECORD_ONLY`, the `th` entry MUST be removed.
+If the sampling Decision is `RECORD_AND_SAMPLE`, and there's no `th` entry in any of the `TraceState` provided by the delegates that decided to `RECORD_AND_SAMPLE`, the `th` entry MUST be also removed.
+Otherwise, the resulting `TraceState` MUST contain `th` entry with the `THRESHOLD` value being the minimum of all the `THRESHOLD` values as reported by those delegates that decided to `RECORD_AND_SAMPLE`.
+
+Each delegate sampler MUST be given a chance to participate in the sampling decision as described above and MUST see the same _parent_ state. The resulting sampling Decision does not depend on the order of the delegate samplers.
+
+### EachOf
```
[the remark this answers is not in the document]
PeterF778: Interestingly, I had considered AllOf originally, but decided against it because it looks too much like AnyOf ... especially in small font and for tired eyes ... but definitely an option.

text/0250-Composite_Samplers.md
PeterF778: Good point, I'll change that.

text/0250-Composite_Samplers.md
PeterF778: Thanks! Good stuff!

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,337 @@
...
+## Approach Two
+
+A principle of operation for Approach Two is that `ShouldSample` is invoked only once, on the root of the tree formed by composite samplers. All the logic provided by the composition of samplers is handled by calculating the threshold values, delegating the calculation downstream as necessary.
+
+### New API
+
+To make this approach possible, all Consistent Probability Samplers which participate in the samplers composition need to implement the following API, in addition to the standard Sampler API. We will use the term _Composable Sampler_ to dentote Consistent Probability Samplers which provide the new API and conform to the rules described here.
+The composite samplers in Approach Two are Composable Samplers as well.
+
+#### GetSamplingAdvice
```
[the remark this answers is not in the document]
PeterF778: Sure, how about GetSamplingIntent?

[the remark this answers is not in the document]
kalyanaj: sure, that sounds good to me.

text/0250-Composite_Samplers.md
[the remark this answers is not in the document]
PeterF778: How about just Junction?

text/0250-Composite_Samplers.md
PeterF778: Yes, you are right - it is more logical to return RECORD_ONLY in that case. I'll change the description.

text/0250-Composite_Samplers.md
PeterF778: Sure, I'll clarify that.

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,353 @@
...
+
+```
+S4 = Conjunction(S3, RateLimitingSampler(1000 * 60))
+```
+
+### Example - sampling configuration 2
+
+Many users are interested in [Consistent Probability Sampling](https://opentelemetry.io/docs/specs/otel/trace/tracestate-probability-sampling/#consistent-probability-sampler) (CP), as it gives them a chance to calculate span-based metrics even when sampling is active. The configuration presented above uses the traditional samplers, which do not offer this benefit.
+
+Here is how an equivalent configuration can be put together using [CP samplers](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md#samplers). In this example, the following implementations are used:
```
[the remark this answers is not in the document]
PeterF778: > EachOf(
>   RuleBased(ROOT, {IsTypeA => ConsistentFixedThreshold(0.4), IsTypeB => ConsistentFixedThreshold(0.2)}), 
>   ConsistentRateLimiting(750 * 60)
> )
> ```
> 
> This would give 250/min for A and 400/min for B, which gives a total of 650/min. Hence, the result differs from that of configuration 1.
> 
The total of 650/min is a concern, I'm less worried about the individual distribution between type A and B. After all, Consistent samplers have very different behavior than leaky-bucket (balancing vs. proportional).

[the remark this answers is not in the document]
PeterF778: > 
> To make configuration 2 equivalent, we must replace `EachOf` with `Conjunction`. 

Yes, `Conjunction` will work better here. Which raises another question: do we need `EachOf` at all (neither in Approach One nor Approach Two)?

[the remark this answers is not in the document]
PeterF778: > Furthermore, we would need a weighted variant of the `ConsistentRateLimiting` sampler, which analyzes the recent distribution of the `th` values given by the first sampler of the conjunction and which adapts the `th` values in such a way that the desired sampling rate is met. Ideally, the `ConsistentWeightedRateLimiting` would increase the thresholds for A and B from 0.6 to 0.675 and from 0.8 to 0.8375, respectively. (Assuming thresholds are values from [0,1].) This would again give 325/min for both, like configuration 1.
> 
Looking at the `th` values only makes sense if the given sampler is used as the Second delegate in `Conjunction`.
So how about generalizing this idea by defining `Weighted` or `Proportional` sampler which analyzes the population of the `rv` values in the population to sample. We cannot use the sampling probability value if it is influenced by the span's `rv`, but it does not mean that it cannot be influenced by the `rv` values from the _previous_ spans. Keeping track of the distribution of the `rv` values at a given sampling stage allows us not only to define a proportional variant of `RateLimitingSampler`, but also something like `ProportionalFixedRate` sampler. It would always reduce the volume of spans according to the defined factor. Not much use in head sampling perhaps, but I believe it would be useful in tail sampling.

More specifically, such a sampler would track the average `rv` of the incoming spans. For an unsampled population this average is expected to be 0.5. If the population has been pre-sampled, the spans with smaller `rv` would have been eliminated, and the observed average would be higher. This would allow the samplers with _proportional_ behavior to adjust their sampling probability accordingly (just as you proposed in your example).

text/0250-Composite_Samplers.md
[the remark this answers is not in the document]
jmacd: `Conditional`, maybe, or `Contingent`?

text/0250-Composite_Samplers.md
```diff
@@ -0,0 +1,293 @@
...
+- Name of the `Span` to be created.
+- `SpanKind` of the `Span` to be created.
+- Initial set of `Attributes` of the `Span` to be created.
+- Collection of links that will be associated with the `Span` to be created.
+
+#### Return value:
+
+The return value is a structure (`SamplingIntent`) with the following elements:
+
+- The THRESHOLD value represented as a 14-character hexadecimal string, with value of `null` representing non-probabilistic `DROP` decision,
```
jmacd, line 196: Suggest permitting an implementation to use an unsigned representation, too.
```suggestion
- The THRESHOLD value indicating the 56-bits defined in OTEP 235, with a special value (e.g., `null`) representing the non-probabilistic `DROP` decision,
```

jmacd (approved):
I will re-review in three weeks as I set out to prototype this. I expect to find / hope to find little resistance to Approach 2 samplers from the community. Thanks @PeterF778!

text/0250-Composite_Samplers.md
jmacd: (Now I see you used "lexicographical" below, which explains why you spelled out the 14-character hex-string part here. I can accept it this way, but in https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/93a78c6ad8f57c5e4ff29a5f41779af03fc85145/pkg/sampling/threshold.go#L37 I've used an unsigned representation ;)

text/0250-Composite_Samplers.md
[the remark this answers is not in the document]
samsp-msft: There seems to be a lot of text here to describe what is essentially boolean/tri-state conditions. We want to be able to AND and OR together a set of conditions.

yurishkuro:
Seems related to https://github.com/open-telemetry/opentelemetry-specification/issues/1844, I was looking for a link to this OTEP.

jmacd:
@yurishkuro Yes, this group is beginning to prototype the approach described in this OTEP in order to address https://github.com/open-telemetry/opentelemetry-specification/issues/1844.

PeterF778:
Here is a follow-up to @jmacd's suggestion posted in the slack channel (#otel-sampling) about backward compatibility of Consistent Probability Samplers.

A particular challenge while trying to inter-operate with legacy samplers is with `RateLimiting` sampler. If its delegate (like `ParentBasedSampler`) does not provide a valid threshold, but still wants to sample, the `RateLimiting` sampler could potentially do two things:
 - either (somehow) enforce the final sampling decision by looking at trace randomness, or
 - provide a threshold value in its `SamplingIntent` that would ensure the correct span rate.
 
The former breaks the new structural approach to Consistent Probability Sampling, and fails if the `RateLimiting` sampler is used not as the top-level sampler but is embedded somewhere in the sampler tree. The later results in potentially incorrect `th` value, because a valid threshold must be used.

The suggested solution then is that `SamplingIntent` contains not only the recommended threshold, but also a boolean flag that would indicate that the sampling recommendation is not solely based on Consistent Probability Sampling principles.
This single flag would mean that:
  - something interfered with purely probabilistic sampling decision making, for example, the `sampled` flag from the parent span had influence
  - the adjusted count remains undefined
  - the `th` value should be removed from the `tracestate`, even for positive sampling decisions

The composite samplers would propagate this flag upstream. And in some cases reset it when it can be determined that the adjusted count is correct after all, in given circumstances.

carlosalberto:

PeterF778:
This PR has been re-opened at the new repository, see https://github.com/open-telemetry/opentelemetry-specification/pull/4321

PR #254, opened 2024-03-06
profiles: fix index in example

florianl:
For `abc;def` the `locations_start_index` should be `4` as `2` points to `baz`.

Follow up of https://github.com/open-telemetry/oteps/pull/239#discussion_r1496062414

### DIFF
--- a/text/profiles/0239-profiles-data-model.md
+++ b/text/profiles/0239-profiles-data-model.md
@@ -1349,7 +1349,7 @@ sample:
     locations_length: 3
     value:
       - 100
-  - locations_start_index: 2
+  - locations_start_index: 4
     locations_length: 2
     value:
       - 200
--- a/text/profiles/0239-profiles-data-model.md
+++ b/text/profiles/0239-profiles-data-model.md
@@ -1361,8 +1361,8 @@ location_indices:
   - 0 # foo
   - 1 # bar
   - 2 # baz
-  - 4 # abc
-  - 5 # def
+  - 3 # abc
+  - 4 # def
 location:
   - line:
       - function_index: 0 # foo
--- a/text/profiles/0239-profiles-data-model.md
+++ b/text/profiles/0239-profiles-data-model.md
@@ -1349,7 +1349,7 @@ sample:
     locations_length: 3
     value:
       - 100
-  - locations_start_index: 4
+  - locations_start_index: 3
     locations_length: 2
     value:
       - 200


joaopgrassi (approved):

text/profiles/0239-profiles-data-model.md
```diff
@@ -1349,7 +1349,7 @@ sample:
     locations_length: 3
     value:
       - 100
-  - locations_start_index: 2
+  - locations_start_index: 4
```
petethepig, line 1352: Ah, my bad, thanks for fixing this, but I think it should be `3` actually:
```suggestion
  - locations_start_index: 3
```

And also, I think `location_indices` should be slightly different, there's a off by 1 error right now:
```
location_indices:
  - 0 # foo
  - 1 # bar
  - 2 # baz
  - 3 # abc
  - 4 # def
```

@florianl Is this right or am I missing something?

petethepig (changes requested):

mtwo:
Comment from the OTel maintainer meeting: could / should this be moved to a comment on the current [Profiling PR in the OTLP](https://github.com/open-telemetry/opentelemetry-proto/pull/534) repository?

text/profiles/0239-profiles-data-model.md
florianl: sorry for the delay - you are right :sweat_smile:

text/profiles/0239-profiles-data-model.md
```diff
@@ -1349,7 +1349,7 @@ sample:
     locations_length: 3
     value:
       - 100
-  - locations_start_index: 2
+  - locations_start_index: 3
```
trask, line 1352: also, should below

```
location_indices:
  - 0 # foo
  - 1 # bar
  - 2 # baz
  - 4 # abc
  - 5 # def
```

be

```
location_indices:
  - 0 # foo
  - 1 # bar
  - 2 # baz
  - 3 # abc
  - 4 # def
```

trask (approved):
I think it's worth fixing here since it's just editorial and can cause confusion for readers.

text/profiles/0239-profiles-data-model.md
florianl: Good spot! :+1: I have fixed also the indices of `location_indices`.

carlosalberto (approved):

PR #255, opened 2024-04-30
Sensitive Data Redaction

svrnm:
This OTEP proposes changes that will enable sensitive data redaction. This is a follow up to https://github.com/open-telemetry/semantic-conventions/pull/971 and https://github.com/open-telemetry/semantic-conventions/pull/961 and the SemConv spec meeting (2024-04-29)


text/0000-sensitive-data-redaction.md
```diff
@@ -0,0 +1,269 @@
...
+This proposal aims to provide the following features:
+
+- methods for consumers of the OpenTelemetry API to enrich collected Attributes with sensitivity information and hooks to apply different levels of redaction.
+- related to those methods a similar way to enrich attributes in the semantic conventions with sensitivity information and ways of redaction.
+- a consistent way to configure sensitivity requirements for end-users of the OpenTelemetry SDK and in instrumentation libraries, including predefined “configuration profiles” and ways for fine grain configuration.
+- A redactor implements the logic to apply redaction and that owns predefined helpers for redaction in the SDK (URLParams filtering, Zeroing IPs, etc.).
+
+The following limitations apply:
+
+- Only Attributes are treated, although it is possible that sensitive data is contained in other data generated by OpenTelemetry as well (e.g. the span name, or the instrumentation scope name could contain sensitive data)
```
lmolkova, line 30: The log bodies can be an interesting problem, but maybe we can treat payload fields in a similar way as attributes or assume that it's instrumentation-side responsibility to redact things there. 

Some context - https://github.com/open-telemetry/semantic-conventions/pull/980 attempts to introduce event bodies with sensitive opt-in fields.

text/0000-sensitive-data-redaction.md
```diff
@@ -0,0 +1,269 @@
...
+As a first building block the OpenTelemetry API needs to provide capability to enrich collected attributes with sensitivity details and with potential hooks to redact those. Those API changes then also can be used to describe sensitivity information in the semantic conventions.
+
+#### API
+
+Every API that sets an attribute consisting of a key and a value, needs to be enhanced by an additional functionality that allows to add details about the potential sensitivity of this data and a hooks how it may be redacted. This can be an additional set of parameters for an existing method or a method that can be called after the attribute has been set, if adding parameters to a method signature would lead to a breaking change.
+
+An additional method for setting the sensitivity information for a span attribute might look like the following:
+
+```
+span.setAttribute("url.query", url.toString(), <SENSITIVITY_DETAILS>);
```
lmolkova, line 43: It's probably more of a spec concern and we'll have more time to design and polish API, but I'd like to entertain the idea of

- Some `SensitivityConfig` instrumentation can request from `Tracer`
- `SensitivityConfig` can be configured by the SDK/distro/etc and can have different implementations for different regulations
- `SensitivityConfig` has convenience methods to redact common things like URLs. E.g. `String SensitivityConfig.redactQueryParams(Uri uri)`
- It may recognize common attributes which are known to be problematic (e.g. `client.address` on the server spans is potentially PII) and could be applied implicitly without instrumentation code

E.g. 
```java
  Span clientSpan= tracer.startSpan("GET");
  String redactedUri = tracer.getSensitivityConfig().redactQueryParams(uri);
  clientSpan.setAttribute("url.full", redactedUri);
}
```

or 

```java
  Span serverSpan= tracer.startSpan("GET /foo", SERVER);
  // the `SensitivityConfig` associated with the tracer knows that `client.address` 
  // may be sensitive and would apply whatever it's configured to do 
  // allow, drop, anonymize, etc)
  serverSpan.setAttribute("client.address", request.getclientIP()); 
}
```

text/0000-sensitive-data-redaction.md
```diff
@@ -0,0 +1,269 @@
...
+**Note 2**: Those requirements are subject of change and outside of the control of the OpenTelemetry community.
+
+## Trade-offs and mitigations
+
+### Performance Impact
+
+By adding an extra layer of processing every time an attribute value gets set, might have an impact on the performance. There might be ways to reduce that overhead, e.g. by only redacting values which are finalized and ready to exported such that updated values or sampled data does not need to be handled.
+
+## Prior art and alternatives
+
```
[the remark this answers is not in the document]
svrnm: Oh, how could I miss those! I didn't check OTEPs, will review them and add them to the prior art. Thanks

text/0000-sensitive-data-redaction.md
svrnm: log bodies is an interesting case indeed! It might be possible to cover it with this OTEP as well but maybe we need to decide that it's out of scope for the first iteration. I think it is too early to decide that.

text/0000-sensitive-data-redaction.md
lmolkova: I think it's both :)

I'd like to be able to add redaction/sanitization without **requiring** instrumentation to change their code or worry about it.

I.e. the alternative proposal is to build a reasonable story within existing API surface first.
But the proposal you made in the OTEL will still be necessary - we'll need to redact/sanitize custom attributes or to optimize and fine-tune instrumentation code for those who want it.

lmolkova: Throwing in one more suggestion - we can auto-generate semantic convention code along with the redaction.

E,g, today we generate constants like:

```java
class UrlAttributes {
   public static final String URL_FULL = "url.full";
}
```

we could instead generate a method

```java
class UrlAttributes {
   public static final String URL_FULL = "url.full";
   public static void recordUrlFull(Span span, Uri uri) {
       String redacted = getSensitivityConfig().redactQueryParams(uri);
       span.setAttribute("url.full", redacted);
   }
}
```

We still need to have a central config and a way to pass it around and make it accessible on the API level, but we can hide a lot of boilerplate and be able to partially automate it.

/cc @lquerel

svrnm: > I'd like to be able to add redaction/sanitization without requiring instrumentation to change their code or worry about it.

Yes, that's an important feature! With me suggestions for how to write that down in the semantic conventions this should be feasible. Let me add some wording to make this clear.

Looking at the solutions I would prefer the first one over the updated code generation for the semantic conventions, i.e. `setAttribute` should do the redaction internally, because having a method per attribute seems excessive to me (and people **will** forget to use it, vs not using the constants comes with less penalties) and people can configure their own redaction through configuration (let's say the use an application they do not have code access and this application emits "enduser.id" without redaction, they can update their sensitivity config and make sure that it is redacted)

dmathieu: One other problem with the suggestion. to specify redaction rules when the attribute is set is that every time someone defines an attribute, they have to consider redaction, which can fairly easily lead to forgetting about it and data leakage.

IMHO, I'd see a redaction system as something which runs in the background (@lmolkova's second suggestion), and can make redaction decisions based on various factors (the attribute name, or the value matching a regexp for example).

text/0000-sensitive-data-redaction.md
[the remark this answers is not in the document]
svrnm: I added both to the prior art section

text/0000-sensitive-data-redaction.md
```diff
@@ -0,0 +1,276 @@
+# Sensitive Data Redaction
+
+This is a proposal for adding treatment of sensitive data to the OpenTelemetry (OTel) Specification and semantic conventions.
+
+## Motivation
+
+When collecting data from an application, there is always the possibility, that the data contains information that shouldn’t be collected, because it is either leaking (parts of) credentials (passwords, tokens, usernames, credit card information) or can be used to uniquely identify a person (name, IP, email, credit card information) which may be protected through certain regulations. By adding OTel to a library or instrumentation an end-user of OTel is facing exactly this challenge: values of an [Attribute](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute) may carry such sensitive data.
```
reyang, line 7:
```suggestion
When collecting data from an application, there is always the possibility, that the data contains information that shouldn't be collected, because it is either leaking (parts of) credentials (passwords, tokens, usernames, credit card information) or can be used to uniquely identify a person (name, IP, email, credit card information) which may be protected through certain regulations. By adding OTel to a library or instrumentation an end-user of OTel is facing exactly this challenge: values of an [Attribute](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute) may carry such sensitive data.
```

reyang, line 7: In additional to "values of an Attribute", the name of the span, name of the event and any other strings that come from the user could share the same problem.

Consider something like "any string value that comes from the user could carry sensitive data".

In theory non-string types may also have the issue (e.g. if folks embed identifiable information in a GUID), probably fine to leave it as doing such thing is already very anti-pattern.

text/0000-sensitive-data-redaction.md
```diff
@@ -0,0 +1,276 @@
...
+
+- Only Attributes are treated, although it is possible that sensitive data is contained in other data generated by OpenTelemetry as well (e.g. the span name, or the instrumentation scope name could contain sensitive data)
+
+### Annotate attributes with sensitivity information
+
+As a first building block the OpenTelemetry API needs to provide capability to enrich collected attributes with sensitivity details and with potential hooks to redact those. Those API changes then also can be used to describe sensitivity information in the semantic conventions.
+
+#### API
+
+Every API that sets an attribute consisting of a key and a value, needs to be enhanced by an additional functionality that allows to add details about the potential sensitivity of this data and a hooks how it may be redacted. This can be an additional set of parameters for an existing method or a method that can be called after the attribute has been set, if adding parameters to a method signature would lead to a breaking change.
```
reyang, line 38: I feel metrics should take an exception here due to the nature of cardinality and performance. Do we know if there were credential/privacy leaks in metrics systems?

text/0000-sensitive-data-redaction.md
```diff
@@ -0,0 +1,276 @@
...
+The `DROP` keyword means that the value is replaced with `REDACTED` (or set to null, …).
+
+It is the responsibility of the OpenTelemetry SDK to implement those sensitivity details provided by the semantic conventions. This means that an instrumentation library does not need to add `<sensitivityDetails>` when calling an `setAttribute` method or does not need to call the additional `redactAttribute` method as outlined above. An instrumentation library may choose to apply additional redactions (leveraging the OpenTelemetry APIs or doing it before calling `setAttribute` in their own business logic).
+
+### SDK Configuration of sensitivity requirements
+
+The annotations and APIs as outlined above will allow SDK users to provide their sensitive requirements as configuration (here: environment variable, but can be encoded in future config files as well), e.g.
+
+```
+env OTEL_SENSITIVE_DATA_PROFILE=”STRICT” ./myInstrumentedApp
```
reyang, line 103: Would this apply to the SDK globally or can be applied on specific pipelines/processors/exporters? e.g. if the goal is to send audit logs (which have EUII such as email address and IP address) to destination A without performing any redaction, and send normal logs to destination B with redaction.

text/0000-sensitive-data-redaction.md
```diff
@@ -0,0 +1,276 @@
...
+- [Filter sensitive data in AppDynamics Java Agent](https://docs.appdynamics.com/appd/24.x/24.4/en/application-monitoring/install-app-server-agents/java-agent/administer-the-java-agent/filter-sensitive-data)
+- [GA4 data redaction](https://support.google.com/analytics/answer/13544947?sjid=3336918779004544977-EU)
+- [Configure Privacy Settings in Matamo](https://matomo.org/faq/general/configure-privacy-settings-in-matomo/)
+- [DataDog Sensitive Data Redaction](https://docs.datadoghq.com/observability_pipelines/sensitive_data_redaction/)
+
+## Open questions
+
+- **Question 1**: Should sensitivity details for an attribute in the semantic conventions be excluded from the stability guarantees? This means, updating them for a **stable** attribute is not a breaking change. The idea behind excluding them from the stability guarantees is that the requirements are subject of change due to changes in technology (see [#971](https://github.com/open-telemetry/semantic-conventions/pull/971), the list of query string values will evolve over time) or changes in regulatory requirements or both.
```
reyang, line 270: I think yes.

text/0000-sensitive-data-redaction.md
svrnm: Thanks, I can update it accordingly. I agree that non-string values _could_ have a similar issue but it is relatively unlikely

text/0000-sensitive-data-redaction.md
svrnm: I would need to research on that, but I think performance will be an issue for all signals, I added that to the trade-off section already. So I wonder if we need to do some experiments eventually.

text/0000-sensitive-data-redaction.md
svrnm: The suggestion here was globally, but I see the point you are making for having it split out by exporter.

text/0000-sensitive-data-redaction.md
reyang: Metrics API is designed to be much faster than other APIs (e.g. it's not uncommon to have metrics APIs that are 20x faster than tracing APIs), so the perf impact would be bigger if measured by the % drop of calls/second.

Cardinality could also be an issue, for example, if the intention is to get the "unique count of users via email address", the result will be `1` if email address got redacted to "redacted@email.address".

text/0000-sensitive-data-redaction.md
yurishkuro: I would suggest to start with the most generic solution that can address all leaks, even if it's less efficient. Anything that relies on instrumentation callsite changes is fundamentally not capable of full coverage - someone will always forget some sensitivity parameter or a call to sanitizer. Whereas having a config-driven sanitizer will be able to see all data and apply rules reliably.

text/0000-sensitive-data-redaction.md
svrnm: Thanks for clarification, that's a signification performance difference indeed. 

On cardinality I think there are 2 options to address that:
* if the redaction happens late (before exporting data) it is still possible to create that metric properly
* there are options to redact emails and still keep their uniqueness, e.g. by hashing the emails (or parts of it), note that hashing is a suboptimal solution since the hashed words are easy to guess & break (use a dictionary with common first and lastnames, etc.)

text/0000-sensitive-data-redaction.md
svrnm: If you want to accomplish to address all leaks you need both: a config-driven sanitizer and one that can be used on instrumentation call site. Let me make an example:

Assume that `url.query` is sanitized as outlined in [971](https://github.com/open-telemetry/semantic-conventions/pull/971), so there is a list of terms that will be droppes, such that a query like `?q=OpenTelemetry&sig=Sp8h9C9xJotR` will turn into `?q=OpenTelemetry&sig=REDACTED`. Now imagine a library that communicates with a 3rd party service that the use a key for the signature that's not in that list (let's say X-Example-Signature), the implementor of that library needs a way to address that. Without an API interface for that the only option is going through the semantic conventions and waiting for it to be included + the languages adding it to their implementation, whereas an API interface allows them to do the following:

```
span.setAttribute(LANGUAGE_CONSTANT_FOR_URL_QUERY, theirValue, { DEFAULT => REDACT_INSECURE_URL_PARAM_VALUES AND values for X-Example-Signature })
```

Of course this requires some semantics to append (or overwrite) existing sensitivity details.

Also, thinking about what you say, if the "configuration-driven sanitizer" is also available to an instrumentor (means via the API), they could express their redaction needs in a centralized way as well:

```
// somewhere at library entry/setup/...
OpenTelemetry.registerSensitivityDetails({
  LANGUAGE_CONSTANT_FOR_URL_QUERY => { DEFAULT => REDACT_INSECURE_URL_PARAM_VALUES AND values for X-Example-Signature },
  /* other configurations that are relevant for the use of their library */
})
```

text/0000-sensitive-data-redaction.md
```diff
@@ -0,0 +1,276 @@
...
+
+### SDK Configuration of sensitivity requirements
+
+The annotations and APIs as outlined above will allow SDK users to provide their sensitive requirements as configuration (here: environment variable, but can be encoded in future config files as well), e.g.
+
+```
+env OTEL_SENSITIVE_DATA_PROFILE=”STRICT” ./myInstrumentedApp
+```
+
+This call will make sure that redactions applied follow the `STRICT` profile. If not set the `DEFAULT` will be used. Additionally there are 2 levels that can not be used in sensitivity details:
```
dmathieu, line 106: Is the **not** on purpose?
It seems like at least `ALWAYS` could be useful, for example as a temporary mitigation when a leakage is discovered, and before it can be properly fixed.

```suggestion
This call will make sure that redactions applied follow the `STRICT` profile. If not set the `DEFAULT` will be used. Additionally there are 2 levels that can be used in sensitivity details:
```

text/0000-sensitive-data-redaction.md
dmathieu: It feels like there should still be some stability definition.
I wouldn't want to see something that was redacted suddenly not be.

text/0000-sensitive-data-redaction.md
svrnm: I think what I wanted to express here is that ALWAYS and NEVER can not be reconfigured through `<SENSITIVITY_DETAILS>` as outlined above in the document. Because if allowed it would break expectations

text/0000-sensitive-data-redaction.md
svrnm: That's a good point, maybe "adding redactions" may be outside, but "removing redactions" may require additional guard rails. It is probably unlikely that redactions will be removed, the only case I can think of is that something is deprecated for a very long time

text/0000-sensitive-data-redaction.md
```diff
@@ -0,0 +1,276 @@
...
+
+- Only Attributes are treated, although it is possible that sensitive data is contained in other data generated by OpenTelemetry as well (e.g. the span name, or the instrumentation scope name could contain sensitive data)
+
+### Annotate attributes with sensitivity information
+
+As a first building block the OpenTelemetry API needs to provide capability to enrich collected attributes with sensitivity details and with potential hooks to redact those. Those API changes then also can be used to describe sensitivity information in the semantic conventions.
+
+#### API
+
+Every API that sets an attribute consisting of a key and a value, needs to be enhanced by an additional functionality that allows to add details about the potential sensitivity of this data and a hooks how it may be redacted. This can be an additional set of parameters for an existing method or a method that can be called after the attribute has been set, if adding parameters to a method signature would lead to a breaking change.
```
danielgblanco, line 38: Are we considering Baggage as a signal here? And would we consider in-band data (context propagation) as well as out-of-band data?

I think https://github.com/open-telemetry/opentelemetry-specification/issues/1633 may benefit from any spec changes related to being able to specify sensitivity of Baggage attributes to be propagated across system boundaries. I don't think this OTEP would solve the  context propagation boundary issue (Propagators would still need to define what they consider a system boundary) but I wonder, providing that Baggage is considered here, if the `SensitivityConfig` should contain something related to defining if a Baggage attribute is safe to propagate or not. I'm not convinced either way myself just now, but thought I'd raise it as we're saying "Every API".

svrnm: That's a very interesting thought. Since baggage may carry sensitive data it's worth looking into it (we might later to move into "future considerations"). @lmolkova example with `client.address` in https://github.com/open-telemetry/oteps/pull/255#discussion_r1585199168 already called out that it might be relevant to add contextual information to the redaction, the baggage propagation is another example of that.

text/0000-sensitive-data-redaction.md
```diff
@@ -0,0 +1,276 @@
...
+
+While it’s ultimately the responsibility of the legal entity operating an application to protect sensitive data, end-users of OpenTelemetry (developers, operators working for that entity) are turning to the authors of OpenTelemetry – or to those of libraries that implement OpenTelemetry, like the Azure SDK – to have means in place to redact/filter sensitive data. Without that capability provided, they will raise security issues and/or will drop OpenTelemetry eventually due to it not meeting their security/legal requirements.
+
+In this OTEP you will find a proposal for adding treatment of sensitive data to OpenTelemetry.
+
+By adding the proposed features, OpenTelemetry will provide its end-users the tooling needed to make sure that sensitive data is treated according to their requirements.
+
+This will make sure that these end-users can use OpenTelemetry within their secure and legal requirements and that OpenTelemetry (and implementing libraries) are able to avoid vulnerabilities.
```
reyang, line 15: 💯 Nice summary @svrnm!

I suggest that we turn these into a list of principles and call it out clearly in this OTEP.
Here are some examples I could think of:

1. OpenTelemetry MUST allow the end-user to meet with their security/privacy/compliance requirements regarding the data being collected.
2. OpenTelemetry MUST not provide redaction offerings that lead to bigger security issues such as https://en.wikipedia.org/wiki/ReDoS (e.g. the redaction logic is poorly implemented, so a hacker could forge certain input to DDoS the redaction engine itself).

jpkrohling: About 2: while we cannot guarantee all SDKs will provide a bug-free implementation of the redaction logic, we could offer a logic to be followed by all SDKs.

text/0000-sensitive-data-redaction.md
jpkrohling: I think this is already implicit, but to me, it's important to let users disable the redaction (ie, having a sensitivity details config that is permissive). This would allow them to not pay a performance penalty at runtime, and apply the redaction logic at the collector if this model satisfies their requirements.

text/0000-sensitive-data-redaction.md
```diff
@@ -0,0 +1,276 @@
...
+
+Example:
+
+| Attribute | ... existing columns ... | sensitivity details |
+|-----------|--------------------------|---------------------|
+| `url.query`|                         | Rationale: Some verbatim wording why this is the way it is below<br>Type: `mixed`<br>`DEFAULT`: `REDACT_INSECURE_URL_PARAM_VALUES`<br>STRICT: `REDACT_ALL_URL_VALUES`<br>`STRICTER`: `DROP` |
+| `client.address`|                    | Rationale: some reasons why dropping octets may be required<br>Type: `maybe_pii`<br>`DEFAULT`: `NONE`<br>`STRICT`: `'s/([0-9]+\.[0-9]+\.[0-9]+\.)[0-9]+/\10/'`<br>`STRICTER`: `'s/([0-9]+\.[0-9]+\.)[0-9]+\.[0-9]+/\10.0/'` |
+| `enduser.creditCardNumber`**[1]** |     | Rationale: ...<br>Type: `always_pii`<br>DEFAULT: `EXTRACT_IIN`<br>`STRICT`: `DROP`|
+
+**[1]**: _This is a made-up example for demonstration purpose, it’s not part of the current semantic conventions. It gives a more nuanced example, e.g. that extracting the IIN might be an option over dropping the number completely. It also demonstrated the value of “type”, which can enable Data lineage use cases_
```
jpkrohling, line 92: s/IIN/Issuer Identification Number (IIN)/

text/0000-sensitive-data-redaction.md
```diff
@@ -0,0 +1,276 @@
...
+### Performance Impact
+
+By adding an extra layer of processing every time an attribute value gets set, might have an impact on the performance. There might be ways to reduce that overhead, e.g. by only redacting values which are finalized and ready to exported such that updated values or sampled data does not need to be handled.
+
+## Prior art and alternatives
+
+### OTEPS
+
+- [OTEP 100 - Sensitive Data Handling](https://github.com/open-telemetry/oteps/pull/100)
+- [OTEP 187 - Data Classification for resources and attributes](https://github.com/open-telemetry/oteps/pull/187)
```
jpkrohling, line 196: I think we'll eventually need to have something similar to this as well: semconv attributes would need a marker, about the common sensitivity of the attribute's value.

jpkrohling (approved):
I believe this goes in the right direction. You did list a "NEVER" option, effectively disabling redaction, but I think it would be more prominently mentioned that and end-user should be able to easily enable/disable this globally.

text/0000-sensitive-data-redaction.md
svrnm: Thanks @reyang, that's a good proposal, will incorporate that (+ the suggestion from @jpkrohling which I agree to, I'll try to find some wording to express that)

text/0000-sensitive-data-redaction.md
svrnm: I need to review those 2 PRs in detail and incorporate whatever we want to include here as well. I added a little bit of wording on classification (see my credit card example)

text/0000-sensitive-data-redaction.md
```diff
@@ -0,0 +1,276 @@
...
+```
+env OTEL_SENSITIVE_DATA_PROFILE=”STRICT” ./myInstrumentedApp
+```
+
+This call will make sure that redactions applied follow the `STRICT` profile. If not set the `DEFAULT` will be used. Additionally there are 2 levels that can not be used in sensitivity details:
+
+- `NEVER`: No redaction is applied, SDKs may choose to log a warning that this is a risky choice
+- `ALWAYS`: All values with sensitivity details will be dropped
+
+Additionally SDK end users can provide advanced configuration (through code, configuration file, probably not environment variable) to add specific needs:
```
pyohannes, line 111: I like the direction this proposal is going toward, however, I'd like to challenge the notion that redaction is treated as an instrumentation issue, as opposed to a configuration issue. I wonder if it would be possible to treat redaction as a configuration issue, like we do sampling.

Reading through this document and arriving at this part, I was thinking about a solution that relies solely on SDK components (without any API changes needed), by providing SDK redactors and configuring them via "redaction profiles" (which consists of mappings from attribute names to redaction rules, similar to the example below). SDKs could provide some default redaction profiles (for example for redacting strict, stricter, or never), which are based on definitions in semantic conventions.

The main challenges of this approach would be considering multiple versions of semantic conventions, furthermore it wouldn't be that straightforward for authors of instrumentation libraries to add redaction rules.

However, it would give the benefit to have redaction rules defined at a central place. Given that the examples above (different regular expressions for different redaction levels), it might be error prone to implement those consistently across all usages of an attribute in different instrumentation libraries in different languages.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,276 @@
...
+
+### Existing Solutions outside OpenTelemetry
+
+There are many solutions outside OpenTelemetry that help to filter or redact sensitive data based on security and privacy requirements:
+
+- [sanitize_field_name in Elastic Java](https://www.elastic.co/guide/en/apm/agent/java/1.x/config-core.html#config-sanitize-field-names)
+- [Filter sensitive data in AppDynamics Java Agent](https://docs.appdynamics.com/appd/24.x/24.4/en/application-monitoring/install-app-server-agents/java-agent/administer-the-java-agent/filter-sensitive-data)
+- [GA4 data redaction](https://support.google.com/analytics/answer/13544947?sjid=3336918779004544977-EU)
+- [Configure Privacy Settings in Matamo](https://matomo.org/faq/general/configure-privacy-settings-in-matomo/)
+- [DataDog Sensitive Data Redaction](https://docs.datadoghq.com/observability_pipelines/sensitive_data_redaction/)
```
joaopgrassi, line 266:
```suggestion
- [DataDog Sensitive Data Redaction](https://docs.datadoghq.com/observability_pipelines/sensitive_data_redaction/)
- [Dynatrace data privacy and security configuration](https://docs.dynatrace.com/docs/shortlink/global-privacy-settings)
```

Dynatrace also has offerings for this :)

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,276 @@
...
+This proposal aims to provide the following features:
+
+- methods for consumers of the OpenTelemetry API to enrich collected Attributes with sensitivity information and hooks to apply different levels of redaction.
+- related to those methods a similar way to enrich attributes in the semantic conventions with sensitivity information and ways of redaction.
+- a consistent way to configure sensitivity requirements for end-users of the OpenTelemetry SDK and in instrumentation libraries, including predefined “configuration profiles” and ways for fine grain configuration.
+- A redactor implements the logic to apply redaction and that owns predefined helpers for redaction in the SDK (URLParams filtering, Zeroing IPs, etc.).
+
+The following limitations apply:
+
+- Only Attributes are treated, although it is possible that sensitive data is contained in other data generated by OpenTelemetry as well (e.g. the span name, or the instrumentation scope name could contain sensitive data)
```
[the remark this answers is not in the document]
joaopgrassi: Ah disregard, I see it was already talked about above.

text/0000-sensitive-data-redaction.md
joaopgrassi: Yeah that statement was a bit confusing to me as well, maybe it can be rephrased to express what you said above?

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,276 @@
...
+- `DEFAULT`: What should be the default redaction applied to this value
+- `STRICT`: What should be a basic redaction that is applied to this value
+- `STRICTER`: What should be an advanced redaction that is applied to this value
+- …
+
+The value `<REDACTION_RULE>` can be one of the following:
+
+- A regular expression (or sed expression?) which when applied turns parts of a string into their redacted version,e.g. `s/([0-9]+\.[0-9]+\.[0-9]+\.)[0-9]+/\10/` applied on an IP address will replace the last octet with 0: `1.2.3.4` becomes `1.2.3.0`
+- A constant that represents pre-defined redaction methods (see below for Redaction helpers), e.g. `REDACT_INSECURE_URL_PARAM_VALUES` will apply what is required for [#971](https://github.com/open-telemetry/semantic-conventions/pull/971)
+- A callback that will call a function on that value and apply advanced redaction
```
lmolkova, line 72: what's the motivation? As a user I'd always prefer a callback to regex for usability, perf, level of control

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,276 @@
...
+
+It should be recommended to either use the expression or the constant and only in rare circumstances a callback function should be applied.
```
lmolkova, line 74: what's the motivation? As a user I'd always prefer a callback to regex for usability, perf, level of control

text/0000-sensitive-data-redaction.md
svrnm: > I wonder if it would be possible to treat redaction as a configuration issue, like we do sampling.

For 90% of the cases, I would say the answer to this is yes, but especially library authors need ways to express additional redaction requirements very locally, e.g. if they call a 3rd party service via a HTTP GET and they know that in this one specific case they need to apply additional redaction. Without a functionality for that they can either over-configure (e.g. apply it globally but potentially have other calls redacted without the need for that) or do their own logic on top of what we provide them. Both things are **OK** and we might decide against having the API for that (in a first version, or forever), but then we need to indiciate that somehow in the specification.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,276 @@
...
+- a consistent way to configure sensitivity requirements for end-users of the OpenTelemetry SDK and in instrumentation libraries, including predefined “configuration profiles” and ways for fine grain configuration.
+- A redactor implements the logic to apply redaction and that owns predefined helpers for redaction in the SDK (URLParams filtering, Zeroing IPs, etc.).
+
+The following limitations apply:
+
+- Only Attributes are treated, although it is possible that sensitive data is contained in other data generated by OpenTelemetry as well (e.g. the span name, or the instrumentation scope name could contain sensitive data)
+
+### Annotate attributes with sensitivity information
+
+As a first building block the OpenTelemetry API needs to provide capability to enrich collected attributes with sensitivity details and with potential hooks to redact those. Those API changes then also can be used to describe sensitivity information in the semantic conventions.
```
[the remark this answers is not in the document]
svrnm: My wording is poor here, what I tried to express here is that by having an API to express this (even if it is one working in the background), we can build the configuration on top of it. I will try to rephrase that accordingly.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,276 @@
...
+
+With the definitions above values in the semantic conventions can be annotated with `<SENSITIVITY_DETAILS>` as outlined above, with the exception that no callback functions can be supplied as redaction rules.
+
+Example:
+
+| Attribute | ... existing columns ... | sensitivity details |
+|-----------|--------------------------|---------------------|
+| `url.query`|                         | Rationale: Some verbatim wording why this is the way it is below<br>Type: `mixed`<br>`DEFAULT`: `REDACT_INSECURE_URL_PARAM_VALUES`<br>STRICT: `REDACT_ALL_URL_VALUES`<br>`STRICTER`: `DROP` |
+| `client.address`|                    | Rationale: some reasons why dropping octets may be required<br>Type: `maybe_pii`<br>`DEFAULT`: `NONE`<br>`STRICT`: `'s/([0-9]+\.[0-9]+\.[0-9]+\.)[0-9]+/\10/'`<br>`STRICTER`: `'s/([0-9]+\.[0-9]+\.)[0-9]+\.[0-9]+/\10.0/'` |
+| `enduser.creditCardNumber`**[1]** |     | Rationale: ...<br>Type: `always_pii`<br>DEFAULT: `EXTRACT_IIN`<br>`STRICT`: `DROP`|
```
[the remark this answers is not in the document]
svrnm: What algorithm would you have in mind for anonymization? Often hashing is used for that, but depending on the underlying data it's pseudo-anonymization (there is a paper on that from the EU, I need to dig up), e.g. email addresses can be guessed easily by tuning your dictionaries.

text/0255-sensitive-data-redaction.md
svrnm: Thanks for calling this out, I didn't add my motivation here. Here was my line of thinking:

This is for **end-users** doing redaction themselves, not for the Semantic Convention. So in their case, the best they can do is reuse some helpers we predefine in the SDK (expressed by constants, eg REDACT_INSECURE_URL_PARAM_VALUES or ANONYMIZE_IP or DROP, ...). I assumed the regex to be second because a regex is much easier to translate them into a helper once again if end-users bring them to the SIG, while a callback can be very convoluted and also have a huge performance impact if not done properly (although I agree the same applies to regular expressions)

Let me rephrase that.

svrnm:
I reworked the document to address (most) of the feedback, please take another look, thanks!

text/0000-sensitive-data-redaction.md
svrnm: I reworked that section to lead with the configuration part (used some wording used here and in @pyohannes' suggestion), please take another look. I kept a section on the API but tried to express it's optional.

text/0000-sensitive-data-redaction.md
svrnm: I kept it as generic as possible: `emitted telemetry may carry such sensitive data.`

text/0000-sensitive-data-redaction.md
svrnm: I still need to add this into the document.

text/0000-sensitive-data-redaction.md
svrnm: I still need to add this into the document.

text/0000-sensitive-data-redaction.md
svrnm: I will leave the question open, I can add some words on adding/removing redaction

text/0000-sensitive-data-redaction.md
svrnm: I still need to add this into the document.

text/0000-sensitive-data-redaction.md
svrnm: I still need to add this into the document.

text/0000-sensitive-data-redaction.md
svrnm: I added those at the top of the document. I used "MUST avoid" vs "MUST not provide" since as @jpkrohling said this is impossible to guarantee, but we can put mechanisms in place to avoid them

text/0000-sensitive-data-redaction.md
svrnm: applied

text/0000-sensitive-data-redaction.md
svrnm: See the updated version, I moved towards leading with the configuration centric approach and mentioned API capabilities towards the end

text/0255-sensitive-data-redaction.md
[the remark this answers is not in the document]
svrnm: This wording does no longer exist.

text/0255-sensitive-data-redaction.md
svrnm: I removed that wording for now, it adds no immediate value

text/0255-sensitive-data-redaction.md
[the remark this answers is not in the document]
lmolkova: good question, I don't have an algorithm in mind. My intention was to have some form of AMONYMIZE mode listed, but details can be out of scope of this otep.

If you believe there is no common and reliable algorithm, then I'm taking my comment back.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+One option is to add additional attributes to methods like `setAttribute`:
+
+```
+span.setAttribute("url.query", url.toString(), <SENSITIVITY_DETAILS>);
+```
+
+Or, if the method can not be extended, an additional method can be called:
+
+```
+span.setAttribute("url.query", OpenTelemetry.redact(url.toString(), <SENSITIVITY_DETAILS>);
```
lmolkova, line 179:
```suggestion
span.setAttribute("url.query", OpenTelemetry.redact(url.toString(), <SENSITIVITY_DETAILS>));
```

lmolkova: In this example the caller is the one doing redaction, but from the OTel API side it doesn't know it already happened and the SDK would do the redaction again.

In the best case it's redundant, but it's also possible that the different redaction layers are inconsistent.

I believe we should allow instrumentations to do the redaction because they may have more context on the data, but then we should avoid doing it again.

----

The suggestion here is to
1. keep `span.setAttribute(string name, any value)` for the cases when SDK does redaction. 
2. allow to express that data is already safe and sanitized with a different API. E.g. `span.setAttribute(string name, any value, <SAFE_DO_NOT_CHANGE>)

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+- [GA4 data redaction](https://support.google.com/analytics/answer/13544947?sjid=3336918779004544977-EU)
+- [Configure Privacy Settings in Matamo](https://matomo.org/faq/general/configure-privacy-settings-in-matomo/)
+- [DataDog Sensitive Data Redaction](https://docs.datadoghq.com/observability_pipelines/sensitive_data_redaction/)
+- [Dynatrace data privacy and security configuration](https://docs.dynatrace.com/docs/shortlink/global-privacy-settings)
+
+## Open questions
+
+- **Question 1**: Should sensitivity details for an attribute in the semantic
+  conventions be excluded from the stability guarantees? This means, updating
+  them for a **stable** attribute is not a breaking change. The idea behind
```
lmolkova, line 381: The stability guarantees cover the part of telemetry that goes over the wire, which, AFAIK is not the case for the sensitivity details.

The part sensitivity details affect is the attribute value, but we don't (at least yet) provide any guarantees there.

So I think we can remove this open question.

text/0000-sensitive-data-redaction.md
yurishkuro: > Without an API interface for that the only option is going through the semantic conventions and waiting for it to be included + the languages adding it to their implementation

The user deploying the library would already have the capability to redact any attribute produced in the telemetry through a central config. Why does it need to wait for anything in semantic conventions?

There is an aspect that the user may not know that there is an attribute that should be redacted, but that problem is solved by the schemas proposal (otel-weaver), where there is a standard way of introspecting all the shapes that instrumentation might produce.

svrnm: The user deploying that application has this capability through the configuration, yes, but what I was talking about is the library author only having the OpenTelemetry API available? How would they be able to express their redaction requirements? I see 3 options:

- They can also add configuration through the API, which is noop until the SDK is loaded. This has the advantage of them using the same logic as the user deploying the library.
- They can use a method to express their redactions. This gives them the most local control, i.e. when there is a specific place in their code where they want to have very specific redactions.
- They can not use OpenTelemetry for that and have their own logic for it. This is OK, if we want to keep that out of the conversation for now.

(related: https://github.com/open-telemetry/oteps/pull/255#discussion_r1593692898)

text/0255-sensitive-data-redaction.md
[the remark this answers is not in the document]
svrnm: https://ec.europa.eu/justice/article-29/documentation/opinion-recommendation/files/2014/wp216_en.pdf is the document I was referring to, they give a really good overview of different approaches to Anonymization (e.g. stripping octets from IPs would fall into "Generalization" or "Aggregation") and also discuss Pseudonymization (e.g. hashing).

I don't know if this is doable, my point is that if we provide ANONYMIZE as an algorithm it should do exactly that (choosing from the approaches listed in that document), and if we want to use hashing etc an appropriate name should be used (PSEUDONYMIZE?)

It's worth to explore this, but before doing that we need to have the tooling for that in place.

text/0255-sensitive-data-redaction.md
svrnm: I would prefer to add parameters to setAttribute, I just wanted to explore how to do it if we want to avoid that. 

In my previous proposal I had something like

```
span.setAttribute(name, value)
span.redactAttribute(name, <SENSITIVITY_DETAILS>)
```

I removed and replaced that because I was not sure anymore why I did it, your comment reminded me of that! My idea was that the end user can give redaction hints, but I also like your suggestion of signaling that the application/library took care of redaction themselves (which is related to https://github.com/open-telemetry/oteps/pull/255#discussion_r1592144764 and https://github.com/open-telemetry/oteps/pull/255#discussion_r1594511964)

text/0255-sensitive-data-redaction.md
svrnm: Agreed, only guarantee we should consider to give is that redactions are never "downgraded" (see https://github.com/open-telemetry/oteps/pull/255#discussion_r1587875097), such that endusers can trust the redaction to work as expected when they update opentelemetry.

text/0000-sensitive-data-redaction.md
yurishkuro: Library authors do not have enough context to make reduction decisions. The best they can do is tag attributes with some semantic annotation saying "this is sensitive" (e.g. in my company we have a taxonomy of "purpose policy" annotations for data elements). If I deploy this library as a user, I should be able to decide if I actually want "sensitive" to be redacted or not - I may be debugging locally and need to see all data, or maybe in production I have different pathways where the data flows, some require reduction and others require raw data.

And to add these annotations it's better to go with the schemas approach, without changing the API.

reyang: > And to add these annotations it's better to go with the schemas approach, without changing the API.

+1

svrnm: I think we are talking about the same thing, @yurishkuro: Indeed a library author can not make reduction decision, this is the responsibility of the application operator, but the library author needs capabilities to express that some data may be sensitive and provide redaction options similar to what we want to do in the semantic conventions. Here is an example: A library author writes a method that calls a payment provider HTTP endpoint, which has a unique query parameter to carry the token (let's say `acmeToken`). That library author now wants to say "when I call this endpoint, and a HTTP span is generated, make sure that `url.query` is redacted in a way that `acmeToken` is treated like other potentially sensitive attributes (see [#971](https://github.com/open-telemetry/semantic-conventions/pull/971)). Ideally they can do something similar to what is proposed for the semantic conventions, e.g. a way to write down what to do when `DEFAULT`, `STRICT` or `STRICTER` (or any other profile is selected). Then, the end user of that library and of opentelemetry can make the decision of redaction.

yurishkuro: @svrnm yes, this is a pretty complex use case, but I would still err on the side of a centralized solution over federating these responsibilities to the call sites. It's conceivable that the semantic convention / schema for `url.query` could be extended with conditional clauses like "if I am in this instrumentation scope or the URL contains this domain name, these are the additional rules / annotations for the query elements". And I think this approach is going to be necessary in many cases anyway, because the specific HTTP instrumentation may not have any clue about this business use case, e.g. it could be a generic HTTP client interceptor that generates client spans and records URL, and only user code would have the theoretical capability of understanding the sensitivity of the query params.

text/0255-sensitive-data-redaction.md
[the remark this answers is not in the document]
lmolkova: thanks for the context! I'm going to resolve this comment since  there is nothing in the OTEP that contradicts adding it in the future.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+There may be multiple redaction profiles (like `DEFAULT`, `STRICT`, etc.), and
+for each one a configuration (as outlined above) may be applied.
+
+any sensitive data.
+
+## Internal details
+
+### Redactor
+
+To accomplish redaction the SDK needs a component (similar to a sampler) that inspects attributes when they are set and applies the required redactions:
```
jmacd, line 202: For the trace SDK-- yes, this is functionality that is similar to samplers. Sampling SIG has been working on requirements for a V2 sampler API, and I would include redaction functionality there. See https://github.com/open-telemetry/opentelemetry-specification/issues/4012.

For the metrics SDK -- there will be questions about whether redaction happens before or after aggregation. The present SDK specification doesn't support a processor that happens before aggregation; the filtering that is possible through metrics Views doesn't exactly support redaction, since modifications to the attribute set need to be considered before aggregation (or else re-aggregation has to be performed following redaction). There has been a requested feature, which goes by the name `MeasurementProcessor` in past discussions, that would be the proper place to (a) redact / drop attributes, (b) extend attributes from context.

In the logs SDK, this is form of processing is the subject of an open discussion, see https://github.com/open-telemetry/opentelemetry-specification/issues/4010.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+#### Redaction requirement profiles
+
+An end user should be able to pick from a set of redaction requirement profiles
+that cover common configurations:
+
+- `NONE` or `NEVER`: No redaction is applied. This is useful if an end user
+  chooses to run a collector nearby the application that will take care of
+  redaction out of process. If this profile is selected the SDK must make sure
+  that (almost) no overhead remains from the redaction methods.
+- `DEFAULT`: A set of rules that especially make sure that no security-sensitive
```
jsuereth, line 106: I'm not sure I see a lot of difference between DEFAULT/STRICT/STRICTER.  At a minimum, I see too many "vague" notions where we could wind up bikeshedding what belong in one vs. another.

Can we be more principled in terms of defining:

1. What each profile means.
2. A set of annotations / meta-semantics for attributes?


E.g.

```
NONE - no redaction
NO_USER - redact any information that is input by a user of the system or could be used to identify a user of the system (email, name, IP, etc.)
NO_LOCATION - redact any location information (geo lat/lon, address, etc.)
```

Basically - I don't think the current set is going to be something we can specify.  This needs more rigor and strict lines.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+
+- Conditions when a redaction should be applied
+- Instructions for the redaction that will be applied when the conditions are met
+
+In an example configuration file a list could look like the following:
+
+```yaml
+  sensitiveData:
+      - attribute: url.query
+        redaction: REDACT_ALL_URL_VALUES
```
jsuereth, line 132: Do we need to only specify "redaction" that can be implemented in all languages?    Will this be a new specification of allowable redactions?

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+      - attribute: com.example.customer.name
+        redaction: DROP
+      - attribute: com.example.customer.email
+        redaction: 's/^[^@]*@/REDACTED@/'
```
jsuereth, line 136: This looks like a regex rule.  The reality here is there's a lot of nuanced-differences between regex across languages and libraries.  You'll want to pick a specific implementation to reference or avoid exposing it here.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+```
+
+In this example the condition is expressed via `attribute` and the instructions
+via `redaction`, where either a constant is selected that runs a predefined helper
+method, or a sed-like expression that uses a regular expression to apply redaction.
+
+Another possibility is to re-use a language like
+[OTTL](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/README.md).
```
jsuereth, line 144: OTTL needs better formal specificity to be language-neutral before this is viable.   

I'd actually like to see that formal specification happen, but I also don't think its current state is a good candidate for this, as it relegates all actual runtime behavior to functions (Today it even relgates the data model via implementation-specific meaning for `.`).   

TL;DR; This is akin to suggesting we have a set of hardcoded redaction functions.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+### Semantic Conventions
+
+To enable the redaction requirement profiles the semantic conventions need to be
+annotated with sensitivity details.
+
+Example:
+
+| Attribute | ... existing columns ... | sensitivity details |
+|-----------|--------------------------|---------------------|
+| `url.query`|                         | Rationale: Some verbatim wording why this is the way it is below<br>Type: `mixed`<br>`DEFAULT`: `REDACT_INSECURE_URL_PARAM_VALUES`<br>STRICT: `REDACT_ALL_URL_VALUES`<br>`STRICTER`: `DROP` |
```
jsuereth, line 155: Can you have this proposal be against semconv YAML model (or the model in general) vs. the documentation?

The model is what we'll be generating code against and what we'll be advertising via SchemaURL.  It'd be a lot easier to comment/design around it.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+
+### Annotate attributes with sensitivity information
+
+As an additional building block the OpenTelemetry API should provide capabilities
+to library (and application) authors to apply in-place redaction
+
+One option is to add additional attributes to methods like `setAttribute`:
+
+```
+span.setAttribute("url.query", url.toString(), <SENSITIVITY_DETAILS>);
```
jsuereth, line 173: Would it make more sense to include sensitivity details in the *attribute itself*?

E.g.
```java
// When defining a key, we define the sensitivity under which it would be redacted.
final AttributeKey<String> myKey = AttributeKey.stringKeyWithSensitivity("url.query", Sensitivity.NO_USER, redactionFunction);
```

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+```
+
+Or, if the method can not be extended, an additional method can be called:
+
+```
+span.setAttribute("url.query", OpenTelemetry.redact(url.toString(), <SENSITIVITY_DETAILS>);
+```
+
+Note that those API calls are no-op and will be implemented by the SDK (as we do it with other API methods as well), this way (almost?) no additional overhead will be created by introducing those annotations.
```
jsuereth, line 182: Do you mean in a no-impl API sense or do you mean after the SDK implements it?


Having this be low/no overhead is a general requirement of our API.  We shouldn't have to explicitly state it.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+```
+
+There may be multiple redaction profiles (like `DEFAULT`, `STRICT`, etc.), and
+for each one a configuration (as outlined above) may be applied.
+
+any sensitive data.
+
+## Internal details
+
+### Redactor
```
jsuereth, line 200: Can you call out how this works across signals?

Right now we have no direct ties between Span/Metric/Trace via the SDK.  They actually only communicate with each other via `Context` API.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+Treating sensitive data properly is a very complex multi-dimensional topic.
+Below you will find some additional context on that subject.
+
+### Types of sensitive data
+
+There are different kinds of “sensitivity” that may apply to data. The ones most
+relevant in this proposal are “security” and “privacy”. They may overlap but we
+distinguish them as follows:
+
+- security-relevant sensitive data: any information that when exposed
```
jsuereth, line 225: Seeing this no, would love to go back to your definitions above and have sensitivity/redaction be:

`ALL_KNOWN`, `KNOWN_SECURITY`, `KNOWN_PRIVACY`, `NONE`

I pick these words because we can offer NO GUARANTEES on redaction given our open API.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+  "client IP" in an client-to-server communication, where the client can be
+  linked to a human.
+- API tokens of a demo system are less sensitive than API tokens for a production
+  system
+- The license plate of an individual’s car is less sensitive than their social
+  security number
+- The full name of a user in a social network is less sensitive than the full
+  name of a user in a medical research database
+
+Depending on the sensitivity data an end-user of an observability system may
```
jsuereth, line 250: I'm not sure we need to be defining the level of sensitivity for users.  For example, a system able to be Privacy compliant could likely handle wipeout rules and protection for all of these.  OpenTelemetry should not enforce level of sensitivity, but *allow our users* to interact with their own definitions using our system.

I'd cut this section or call it out as something for our users to define, not us.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+- [more…](https://en.wikipedia.org/wiki/Information_privacy)
+
+Additionally the entities operating the applications who leverage OpenTelemetry
+may have their own requirements for treating certain sensitive data.
+
+Finally end-users may want to apply recommendations for
+[Data Minimization](https://en.wikipedia.org/wiki/Data_minimization), to avoid
+"unnecessary risks for the data subject".
+
+**Note 1**: it is not (and can not be) the responsibility of the OpenTelemetry
```
jsuereth, line 272: This should be the first thing stated in this section.   We provide tooling to allow users to be compliant, but they are ultimately responsible.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+**Note 2**: Those requirements are subject of change and outside of the control
+of the OpenTelemetry community.
+
+## Trade-offs and mitigations
+
+### Performance Impact
+
+By adding an extra layer of processing every time an attribute value gets set,
+might have an impact on the performance. There might be ways to reduce that
+overhead, e.g. by only redacting values which are finalized and ready to exported
```
jsuereth, line 286: This has security implications.  E.g. in Java we you wouldn't want sensitive passwords interned in Strings because they can be exfiltrated.

I like what you're suggesting, but I'd go the other way.  You could memoize the redaction function when generating the attribute value in some fashion, *or* use codegen/weaver to optimise these use cases (if we get the yaml model sorted).

jsuereth (commented):
Overall want to give a HUGE THANK YOU for making a dent in this problem and throwing a proposal out there.

I have a lot of comments.  I wish I had more time to give viable solutions to my comments, but I do not.  (I also lack the time to be brief, so this is long winded).

Overall -

This has the bones of what I think we want to build.

- A configurable set of redaction rules
- A mechanism to annotate attributes with meta-infromation about sensitivity/security
- A simple user knob of "ON/OFF/my-custom-thing" for important use cases.

The details need a bit more fleshing out in particular:

- The YAML model (and OTEL API) for annotating attributes with baseline sensitivity/redaction controls.  You may need MULTIPLE redaction methods for each level of redaction you support, e.g.
- Stricter lines / definitions (less possible bikeshed) on Sensitivity levels.
- More details on the interaction with the SDK, in particular is it the same interface for all three signals or different?
- More granular model (Span name redaction, e.g.)

Overall a great start.  Let me know if there's anything I can do to help push areas of this forward or flesh out any of your ideas.

text/0255-sensitive-data-redaction.md
svrnm: Based on the feedback so far major consensus is that the redaction _mainly_ should happen in the background, so even if you provide `url.query` as a string the SDK knows what to do with that (e.g. by leveraging code generated by the weaver).

The big discussion point is what to do if someone wants to add additional redaction requirements. We can not expect that what we provide via the semconv + code-generation is covering all the requirements that users have.

text/0255-sensitive-data-redaction.md
svrnm: Thanks, I think I wrote that "similar to a sampler" without giving it a lot of thought beyond "this needs to be a component that can be evolved independently (like samplers) and is also independent of the signal.

I'll review your feedback and try to incorporate it into the document

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+ and some custom requirements like the following:
+
+* They provide a profile name `STRICTER` via an environment variable, e.g.
+  OTEL_SENSITIVE_DATA_PROFILE="STRICTER"
+* They add custom requirements in a
+  [configuration file](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md), e.g.
+  
+  ```yaml
+  sensitiveData:
+      - attribute: url.query
```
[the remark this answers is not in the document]
svrnm: This is just an example, I think we need more than attributes, see below I even suggested to use OTTL (might be an overkill thinking about the performance impact)

text/0255-sensitive-data-redaction.md
svrnm: Wow, I like that, this is a great idea!

text/0255-sensitive-data-redaction.md
svrnm: maybe? I think at some point we need to do some prototyping to figure out what redaction is allowed to do and what not. Performance is the main reason but "this needs to be implemented in "any" language" is another complex requirement.

text/0255-sensitive-data-redaction.md
svrnm: Yes, regular expressions might not be the right tool to express the redaction requirements. Having differences in languages is the one thing, but also the ReDOS example comes to mind.

text/0255-sensitive-data-redaction.md
svrnm: Yes, I can do that!

text/0255-sensitive-data-redaction.md
svrnm: > Having this be low/no overhead is a general requirement of our API. 

This is what I meant yes, so I can remove it

text/0255-sensitive-data-redaction.md
svrnm: This is already a kind of "appendix" section, I can make it clearer from the document structure.

svrnm:
> Overall want to give a HUGE THANK YOU for making a dent in this problem and throwing a proposal out there.

Happy to get this conversation started, a big thank you from me to everyone who provided their feedback so far. 

> I have a lot of comments. I wish I had more time to give viable solutions to my comments, but I do not. (I also lack the time to be brief, so this is long winded).

I appreciate your feedback, a lot of things I can take and try to provide a solution for.


> This has the bones of what I think we want to build.
> 
>     * A configurable set of redaction rules
> 
>     * A mechanism to annotate attributes with meta-infromation about sensitivity/security
> 
>     * A simple user knob of "ON/OFF/my-custom-thing" for important use cases.

💯 

> The details need a bit more fleshing out in particular:
> 
>     * The YAML model (and OTEL API) for annotating attributes with baseline sensitivity/redaction controls.  You may need MULTIPLE redaction methods for each level of redaction you support, e.g.
> 
>     * Stricter lines / definitions (less possible bikeshed) on Sensitivity levels.
> 
>     * More details on the interaction with the SDK, in particular is it the same interface for all three signals or different?
> 
>     * More granular model (Span name redaction, e.g.)

I'll take another look into all feedback and try to evolve the document with it.

> Overall a great start. Let me know if there's anything I can do to help push areas of this forward or flesh out any of your ideas.

Thanks!

text/0000-sensitive-data-redaction.md
svrnm: > It's conceivable that the semantic convention / schema for url.query could be extended with conditional clauses like "if I am in this instrumentation scope or the URL contains this domain name, these are the additional rules / annotations for the query elements".

That's what is part of this proposal, I need to optimize the wording for that, but eventually the configuration for redaction needs to be that fine-grained.

> but I would still err on the side of a centralized solution over federating these responsibilities to the call sites.

Yes, centralized is the core of it, but library/application authors still need a way to provide a hint that a redaction needs to be applied. Rethinking my example, we actually have the following situation:

A library author is likely using a HTTP client library as dependency themselves to interact with a 3rd partyAPI, so in some pseudo code they might have something like the following:

```
function authenticateWithMyService($token) {
  // they want to let OpenTelemetry know that $token is sensitive, and when
  // it is injected into the `url.query` attribute certain redaction may be required.
  use3rdPartyHttpClient("https://myservice//?myAcmeToken=$token")
}
```

How would they implement that? They need some ways to annotate that. Also those annotations need to be composable somehow.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+
+- `Redactor::setup(profile)` will setup the redactor with the given profile,
+  maybe a constructor? depends on the language
+- `Redactor::redact(value, <sensitivityDetails>)` will return the redacted value
+  using the provided profile
+
+The redactor will also have all the methods to apply predefined redactions
+(`REDACT_ALL_URL_VALUES`, `REDACT_INSECURE_URL_PARAM_VALUES`, `DROP`, etc.).
+If a method is not implemented (either by the SDK or by the end-user choosing
+one that does not exist), it will default to apply `DROP` to avoid leakage of
```
lquerel, line 212:
```suggestion
one that does not exist), it will default to apply `DROP` to avoid leakage of data.
```

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+
+There are different kinds of “sensitivity” that may apply to data. The ones most
+relevant in this proposal are “security” and “privacy”. They may overlap but we
+distinguish them as follows:
+
+- security-relevant sensitive data: any information that when exposed
+  [weakens the overall security of the device/system](https://en.wikipedia.org/wiki/Vulnerability_(computing)).
+- privacy-relevant sensitive data: any information that when exposed
+  [can adversely affect the privacy or welfare of an individual](https://en.wikipedia.org/wiki/Information_sensitivity).
+
```
lquerel, line 229: The identification of sensitive data could also be applied at a semantic level. For example, as a company, I might want to apply a redaction rule to ALL URLs, another rule to ALL IP addresses, and so on. One possible approach would be to introduce a new type of group in the semconv that would allow the definition of semantic types, which could be attached to existing attributes in addition to the "physical" data type that we already have. The redaction rules would then apply to the semantic types.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+may have their own requirements for treating certain sensitive data.
+
+Finally end-users may want to apply recommendations for
+[Data Minimization](https://en.wikipedia.org/wiki/Data_minimization), to avoid
+"unnecessary risks for the data subject".
+
+**Note 1**: it is not (and can not be) the responsibility of the OpenTelemetry
+project to provide compliance with any of those regulations, this is a
+responsibility of the OTel end-user. OTel can only facilitate parts of those
+requirements.
```
lquerel, line 275: Having support for lineage associated with this type of metadata could help identify where a redaction operation originates from. We have started to support the concept of lineage in Weaver, and I think it could be used in this specific case.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+As listed above there are multiple processors available that can be used to
+redact or filter sensitive data with the OpenTelemetry collector. The challenge
+with that is that it is unknown to an application (owner) if data is processed
+in the collector as expected. Also, the data leaving the application might
+already be a risk (non-encrypted or compromised network) or may not be allowed
+(collector is hosted in a different country, which may conflict with a regulation)
+
+Ideally a combination is used.
+
+### Alternative 2: Backend
```
lquerel, line 360: Another complementary option could be to support this type of sensitive data management at the code generation level with a tool like Weaver.

svrnm:
Thank you all for the great feedback. I was not sure what a good practice is, but I reset the PR to `draft` because I am in the process of incorporating all the feedback and coming back with a significant change.

text/0255-sensitive-data-redaction.md
svrnm: Yes 🙌 , I mentioned lineage somewhere in the doc already because it seems to be highly related here.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+The `DROP` keyword means that the value is replaced with `REDACTED` (or set to null, …).
+
+It is the responsibility of the OpenTelemetry SDK to implement those sensitivity details provided by the semantic conventions.
+
+### Annotate attributes with sensitivity information
+
+As an additional building block the OpenTelemetry API should provide capabilities
+to library (and application) authors to apply in-place redaction
+
+One option is to add additional attributes to methods like `setAttribute`:
```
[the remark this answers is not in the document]
svrnm: Developers **need** to be able to provide an annotation(!) that data is "potentially sensitive". I try to express this better, that it's not their job to make decisions on that but that they need to be part of highlighting that.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+
+| Attribute | ... existing columns ... | sensitivity details |
+|-----------|--------------------------|---------------------|
+| `url.query`|                         | Rationale: Some verbatim wording why this is the way it is below<br>Type: `mixed`<br>`DEFAULT`: `REDACT_INSECURE_URL_PARAM_VALUES`<br>STRICT: `REDACT_ALL_URL_VALUES`<br>`STRICTER`: `DROP` |
+| `client.address`|                    | Rationale: some reasons why dropping octets may be required<br>Type: `maybe_pii`<br>`DEFAULT`: `NONE`<br>`STRICT`: `'s/([0-9]+\.[0-9]+\.[0-9]+\.)[0-9]+/\10/'`<br>`STRICTER`: `'s/([0-9]+\.[0-9]+\.)[0-9]+\.[0-9]+/\10.0/'` |
+| `enduser.creditCardNumber`**[1]** |     | Rationale: ...<br>Type: `always_pii`<br>DEFAULT: `EXTRACT_IIN`<br>`STRICT`: `DROP`|
+
+**[1]**: _This is a made-up example for demonstration purpose, it’s not part of the current semantic conventions. It gives a more nuanced example, e.g. that extracting the Issuer Identification Number (IIN) might be an option over dropping the number completely. It also demonstrated the value of “type”, which can enable Data lineage use cases_
+
+The `DROP` keyword means that the value is replaced with `REDACTED` (or set to null, …).
```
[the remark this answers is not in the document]
svrnm: You are right, I plan to change this. I consider to suggest that we keep some kind of "suggested default redaction" if that is feasible, but beyond that it's something that needs to be decided locally.

text/0255-sensitive-data-redaction.md
```diff
@@ -0,0 +1,397 @@
...
+  that (almost) no overhead remains from the redaction methods.
+- `DEFAULT`: A set of rules that especially make sure that no security-sensitive
+  data is exposed, only limited PII-related redactions are applied.
+- `STRICT`: A set that builds on top of `DEFAULT` and also applies commonly required
+  PII-related redactions (anonymize emails, names, IPs, etc)
+  `STRICTER`: A set that builds on top of `STRICT` and applies even more redactions,
+  e.g. drop all query attribute values, drop PII-related data, etc.
+- `DROP` or `ALWAYS`: All values with (potentially) sensitivity details will be dropped
+
+Those profiles will be defined through the semantic conventions (see below).
```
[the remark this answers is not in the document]
svrnm: I am reworking that

text/0255-sensitive-data-redaction.md
svrnm: @lquerel my knowledge about telemetry schema is very limited, can you provide me some examples and pointers to learn more about this? Overall I am a fan of making one thing a "subset of another", but I would need to understand the impact of that, technically but also organizational (e.g. what would be the extended timeline for that?)

text/0255-sensitive-data-redaction.md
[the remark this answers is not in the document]
svrnm: I am thinking about that as well right now, I try to come up with a suggestion in the updated proposal.

svrnm:
https://github.com/open-telemetry/semantic-conventions/pull/971#issuecomment-2437014611

> I tried to rework it but there are so many moving pieces that I couldn't get a good handle of it, while also running into some bandwidth issues on my end, since other things required my attention. I still want to follow up on this, but it will take much more time than anticipated.

pellared:
Related to https://github.com/open-telemetry/semantic-conventions/issues/128

carlosalberto:

PR #256, opened 2024-05-07
Introduce Entities Data Model, Part 1

tigrannajaryan:
This is a proposal of a data model to represent entities. The purpose of the data model is to have a common understanding of what an entity is, what data needs to be recorded, transferred, stored and interpreted by an entity observability system.

This data model sets the foundation for adding entities to OpenTelemetry. The data model is largely borrowed from
[the initial proposal](https://docs.google.com/document/d/1VUdBRInLEhO_0ABAoiLEssB1CQO_IcD5zDnaMEha42w/edit) that was accepted for entities SIG formation.

This OTEP is step 1 in introducing the entities data model. Follow up OTEPs will add further data model definitions, including the linking of Resource information to entities.


text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,765 @@
...
+unique. For example a process entity's LID is its PID number and process start time.
+The (PID,StartTime) pair is unique only in the context of a host where the process runs
+(and the host in this case is the identification context).
+
+The GID of an entity is globally unique, in the sense that for the entire set of entities
+in a particular telemetry store no 2 entities exist that have the same GID value.
+
+The GID of an entity E is defined as:
+
+`GID(E) = { LID(E), GID(IDCONTEXT(E)) }`
```
HunterSherms, line 407: Something I’m not super clear on:

My understanding of the formula here suggests that it is recursive, in the sense that in order to calculate the GID for an entity, you will need to know the LID of the parent, and the parent’s parent, until you hit a GID.

Is it expected that the sender will always have enough available information/access to parent entity information to send a GID? What if it doesn't?

jack-berg (approved):
Couple of comments that can be worked out in future work but this looks great to me. Thanks for putting this together!

text/entities/0256-entities-data-model.md
tigrannajaryan: >My understanding of the formula here suggests that it is recursive, in the sense that in order to calculate the GID for an entity, you will need to know the LID of the parent, and the parent’s parent, until you hit a GID.

That is exactly correct, it is recursive. I should probably also include the base case `GID(ROOT)={}` to make it clear.

>Is it expected that the sender will always have enough available information/access to parent entity information to send a GID? What if it doesn't?

No, this is not expected. It is explicitly called out on line 62 that this information may not be available in its entirety in one place. The sender may send an incomplete GID to the collection pipeline and elements of pipeline that have additional knowledge about entity's IDCONTEXT can add further elements to the ID. This is also known as "enrichment" and in the early version of the proposal we described how Collector does it [here](https://docs.google.com/document/d/1VUdBRInLEhO_0ABAoiLEssB1CQO_IcD5zDnaMEha42w/edit#heading=h.ij8yjheo645z).

Here is an example to visualize the process of computing a Process GID step by step (warning: no claims are made that Process will have exactly this ID attributes - the following diagram is just to illustrate the principle):
image

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,765 @@
...
+
+In this example we assume this to be a valid GID because Pod is the root
+IDCONTEXT, since Pod's LID includes the cluster name, which is expected
+to be globally unique. If this assumption about global uniqueness of
+cluster names is wrong then another containing IDCONTEXT within which
+cluster names are unique will need to be applied and so forth.
+
+Note also how we used a pair (`k8s.pod.uid`, `k8s.cluster.name`).
+Alternatively, we could say that Kubernetes Cluster is a separate entity
+we care about. This would mean the Pod's IDCONEXT is the cluster. The
```
mx-psi, line 544:
```suggestion
we care about. This would mean the Pod's IDCONTEXT is the cluster. The
```

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,765 @@
...
+for arrays and maps is allowed.
+<p>
+SHOULD follow OpenTelemetry <a
+href="https://github.com/open-telemetry/semantic-conventions">semantic
+conventions</a> for attributes.
+    </td>
+   </tr>
+</table>
+
+### Minimally Sufficient Id
```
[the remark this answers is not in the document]
tigrannajaryan: This is recorded as a TODO here https://github.com/orgs/open-telemetry/projects/85?pane=issue&itemId=57053415

I would like to address it after the OTEP, and it is likely going to be in the form of supplementary guidelines.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,765 @@
...
+```
+
+## Prototypes
+
+A set of prototypes that demonstrate this data model has been implemented:
+
+- [Go SDK Prototype](https://github.com/tigrannajaryan/opentelemetry-go/pull/244)
+- [Collector Prototype](https://github.com/tigrannajaryan/opentelemetry-collector/pull/4)
+- [Collector Contrib Prototype](https://github.com/tigrannajaryan/opentelemetry-collector-contrib/pull/1/files)
+- [OTLP Protocol Buffer changes](https://github.com/tigrannajaryan/opentelemetry-proto/pull/2/files)
```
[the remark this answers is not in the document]
tigrannajaryan: It is an option. That's in fact how they are implemented experimentally in the Collector. We will work on this in a later OTEP as called out in "Future Work".

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,765 @@
...
+    <td>A set of descriptive (non-identifying) attributes of the Entity.
+<p>
+MAY change over the lifetime of the entity. MAY be empty. These
+attributes are not part of Entity's identity.
+<p>
+“value” follows <a
+href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#type-any">any</a>
+value definition in the OpenTelemetry spec - it can be a scalar value,
+byte array, an array or map of values. Arbitrary deep nesting of values
+for arrays and maps is allowed.
```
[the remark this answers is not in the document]
tigrannajaryan: >I like that you've called out that the id key/value entries are common attributes. This ensures that all identifying attributes are representable in a resource.

That is exactly the reason why I chose it that way. In a future OTEP we will see why it is important, when we show the required changes in the Resource.

>This means that all resource attributes are valid entity attributes, but not all entity attributes are valid resource attributes.

Yes. I don't like this non-uniformness, but it seems to be necessary for modelling arbitrary entities. The representation of entities may and often will come outside of OpenTelemetry, so we don't control the data structure that describes such entities.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,765 @@
...
+    </td>
+   </tr>
+   <tr>
+    <td>Type
+    </td>
+    <td>string
+    </td>
+    <td>Defines the type of the Entity. MUST not change during the
+lifetime of the entity. For example: "service" or "host". This field is
+required and MUST not be empty for valid entities.
```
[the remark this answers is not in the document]
tigrannajaryan: Yes, we should look into that as part of that Open Question. My current thinking is that it is not a runtime data model question per se, but rather a matter of modelling the entity semantic conventions (aka our type definitions). I may change my opinion if we show use cases where having this information at runtime is useful because we have entities the type of which within the hierarchy may change dynamically and cannot be a static knowledge recorded in semantic conventions. So far I can't see a use case like that.

text/entities/0256-entities-data-model.md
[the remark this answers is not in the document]
mx-psi: Makes sense feel free to resolve this then. Maybe adding a link to that project board in the Open Questions section would help here?

text/entities/0256-entities-data-model.md
HunterSherms: Given that, since we know in advance that there will be cases where an entity is not uniquely identifiable from the data available to the sender, can this still be said to be true?

`Meaningful (especially human-readable) IDs are more valuable than random-generated IDs`

I would argue that a randomly generated UUID would be far more valuable than a meaningful id that does not uniquely distinguish the entity.

text/entities/0256-entities-data-model.md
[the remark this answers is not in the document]
tigrannajaryan: Added links to all TODOs.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+    </td>
+   </tr>
+   <tr>
+    <td>Id
+    </td>
+    <td>key/value pair list
+    </td>
+    <td>A set of attributes that identifies the Entity.
```
jaronoff97, line 120: Will it make sense to re-evaluate the opamp spec when this is merged? Right now an Agent sends its identifying and non-identifying attributes which is very similar to the entity definition stated here.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+    <td>Id
+    </td>
+    <td>
+    </td>
+    <td>See data model
+</td>
+   </tr>
+</table>
+
+Note that EntityDelete is optional and is not guaranteed to be sent when
```
jaronoff97, line 376: I think it would be incredibly valuable to include an optional reason field in the entity delete event. Knowing _why_ an entity is being deleting would help users easily identify unexpected changes in their system's steady state. For example in a kubernetes cluster, a pod's EntityDelete event may be due to a new version being rolled out (completely normal state) or it could be due to the pod crashlooping and being deleted by a controller (something a user would want to be informed about). 

Similarly, the reason would be very useful for users running instances in a cloud provider. A user may initiate a delete for an EC2 instance or the EC2 instance may be killed by the provider.

mx-psi (approved):
LGTM. I haven't been able to join the Entities meeting but this looks like a good start to me and it seems like this way of approaching identity is better than the alternatives

jsuereth (approved):

text/entities/0256-entities-data-model.md
HunterSherms: @jsuereth - yes, to be clear I don't disagree with any of this **except in a context where an entity cannot be uniquely identified give available data**, in which case wouldn't we prefer to guarantee that it is uniquely identifiable?

text/entities/0256-entities-data-model.md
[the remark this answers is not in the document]
tigrannajaryan: We have a TODO for this: https://github.com/orgs/open-telemetry/projects/85/views/1?pane=issue&itemId=57173884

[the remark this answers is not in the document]
jsuereth: I'd like to see these as events as well.

text/entities/0256-entities-data-model.md
tigrannajaryan: Yes, we should re-evaluate. I think what OpAMP does matches this proposal's ideas fairly well, so I don't anticipate major changes, but it may result in clarifications in OpAMP.

text/entities/0256-entities-data-model.md
tigrannajaryan: Added a TODO: https://github.com/orgs/open-telemetry/projects/85/views/1?pane=issue&itemId=62416924

text/entities/0256-entities-data-model.md
tigrannajaryan: @HunterSherms you are absolutely right that exceptions exist. In fact we had to recently make this painful exception for [service.instance.id](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/resource#service-experimental) field and recommend using a UUID for it.

The design principle of course is not absolute (none of the principles are) and is a merely a guide that helps make decisions. The principles are intentionally soft-worded and don't use the RFC-normative language (e.g. capitalized MUST or SHOULD) you see elsewhere.

HunterSherms: Excellent, thank you for the clarification!

HunterSherms (approved):

jaronoff97 (approved):

carlosalberto (approved):

aknuds1 (approved):

tigrannajaryan:

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+
+We propose a new concept of Entity.
+
+Entity represents an object of interest associated with produced telemetry:
+traces, metrics or logs.
+
+The Entity that produces the telemetry is called the Producing Entity. For example
+telemetry produced using OpenTelemetry SDK is normally associated with a Service.
+The Service is the Producing Entity. Similarly, OpenTelemetry defines system metrics
+for a host. The Host is the Producing Entity in this case.
```
djaglowski, line 81: I find the term "Producing" to be confusing here because although the host is described by the telemetry, it is not necessarily _producing_ it. For example, an external observer may be generating the telemetry based on observations of the host, in which case it seems more natural to describe the external observer as the producer.

To highlight this further, I believe line 440 uses the term "producer" in the "external observer" sense rather than the "described entity" sense: `The entity information producer is responsible for determining the identification context of each entity it is producing information about.`

If I understand the intention here, we're really just trying to be clear that the Entity is the object that is described by the telemetry, so would it be clearer to call it the "Described Entity". Moreover, is either adjective really adding value here? Why not just use the term Entity?

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+   </tr>
+</table>
+
+### Minimally Sufficient Id
+
+Often a number of attributes of an entity is readily available for the telemetry
+producer to compose an Id from. Of the available attributes the Entity Id should
+include the minimal set of attributes that is sufficient for uniquely identifying
+that entity. No superfluous attributes should be included in the Id set. For example
+a Process on a host can be uniquely identified by `process.id` attribute. Adding for
```
djaglowski, line 161: We use process as an example in the LID/GID/IDContext discussion as well but define it differently. (PID/StartTime)

I recognize that the example is intentionally simple to illustrate a point, but it seems we should either be consistent across the doc or find another example for this.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+    <td>nanoseconds
+    </td>
+    <td>The time since when the entity state is described by this event.
+The time is measured by the origin clock. The field is required.
+    </td>
+   </tr>
+   <tr>
+    <td>Interval
+    </td>
+    <td>milliseconds
```
djaglowski, line 278: Are we sure sub-millisecond intervals won't ever be useful? On the consumer side it might be easier anyways to just have nanoseconds here so that Timestamp+Interval does not require a conversion.

text/entities/0256-entities-data-model.md
jsuereth: This is a good call out.  In the previous "full" proposal, there's a link between "Resource" and "Producing Entity".  I think the language you see here is a bit of a preview of a Resource<->Entity mapping.   I think it'd be better to cut this from the proposal (just use entity) prior to needing the term "Producing Entity" and associations between Resource<->Entity, which should be the next OTEP we read/agree on.

tigrannajaryan: I agree. We arlready have a TODO to find a better name: https://github.com/orgs/open-telemetry/projects/85/views/1?pane=issue&itemId=57054361

I removed references to "Producing Entity" from this OTEP for now in commit https://github.com/open-telemetry/oteps/pull/256/commits/a1f126f339a9078235819371af24b2a31153a968.

We will find a better name in follow up OTEPs.

text/entities/0256-entities-data-model.md
tigrannajaryan: Good catch, fixed in https://github.com/open-telemetry/oteps/pull/256/commits/a1c99e1da8d2385eeb3ddb09151e236126ec1d85

text/entities/0256-entities-data-model.md
tigrannajaryan: Added a TODO to discuss separately: https://github.com/orgs/open-telemetry/projects/85/views/1?pane=issue&itemId=63753636

text/entities/0256-entities-data-model.md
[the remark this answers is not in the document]
tigrannajaryan: We have a TODO to discuss this further: https://github.com/orgs/open-telemetry/projects/85/views/1?pane=issue&itemId=62411493

Resolving this comment for now.

alanwest (approved):

pellared (commented):
(blank)

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,789 @@
...
+<!-- tocstop -->
+
+## Motivation
+
+This data model sets the foundation for adding entities to OpenTelemetry. The data model
+is largely borrowed from
+[the initial proposal](https://docs.google.com/document/d/1VUdBRInLEhO_0ABAoiLEssB1CQO_IcD5zDnaMEha42w/edit)
+that was accepted for entities SIG formation.
+
+This OTEP is step 1 in introducing the entities data model. Follow up OTEPs will add
```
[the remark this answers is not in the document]
tigrannajaryan: The OTEP does not introduce the _entire_ data model, which is important to note in some way. There will be "step 2 of data model". Otherwise it would be unclear why the OTEP is incomplete.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,789 @@
...
+unnecessary and violates the Minimally Sufficient Id rule.
+
+### Examples of Entities
+
+_This section is non-normative and is present only for the purposes of demonstrating
+the data model._
+
+Here are examples of entities, the typical identifying attributes they
+have and some examples of non-identifying attributes that may be
+associated with the Entity.
```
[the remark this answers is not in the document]
tigrannajaryan: It should be capitalized when introducing the term the first time and when used as part of multi-word identifiers such as EntityState event, lower case elsewhere.

Fixed now.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,789 @@
...
+
+- Meaningful (especially human-readable) IDs are more valuable than random-generated IDs.
+  Long-lived IDs that survive state changes (e.g. entity restarts) are more valuable than
+  short-lived, ephemeral IDs.
+  See [the need for navigation](https://docs.google.com/document/d/1Xd1JP7eNhRpdz1RIBLeA1_4UYPRJaouloAYqldCeNSc/edit#heading=h.fut2c2pec5wa).
+
+- We cannot make an assumption that the entirety of information that is necessary for
+  global identification of an entity is available at once, in one place. This knowledge
+  may be distributed across multiple participants and needs to be combined to form a
+  globally unique identifier.
```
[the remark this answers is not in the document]
tigrannajaryan: Rephrased to avoid saying "globally distinct identifier".

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,789 @@
...
+    </td>
+    <td><strong>Non-identifying Attributes</strong>
+    </td>
+   </tr>
+   <tr>
+    <td>Service
+    </td>
+    <td>"service"
+    </td>
+    <td>service.name (required)
```
pellared, line 189: Maybe instead of "required" we should say "not empty"?
Or even more explicitly `(required, not empty)`.

Same comment applies to other places.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,789 @@
...
+EntityState event. The recipient can use this value to compute when to expect the next
+EntityState event and if the event does not arrive in a timely manner (plus some slack)
+it can consider the entity to be gone even if the EntityDelete event was not observed.
+
+## Entity Identification
+
+_This section is a supplementary guideline and is not part of logical data model._
+
+The data model defines the structure of the entity ID field. This section explains
+how the ID field is computed.
```
pellared, line 391: nit: In other places it is `Id`.

Same comment applies to other occurrences of `ID and `Id`.

Personally I favor `ID`.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,789 @@
...
+    </td>
+    <td>Defines the type of the Entity. MUST not change during the
+lifetime of the entity. For example: "service" or "host". This field is
+required and MUST not be empty for valid entities.
+    </td>
+   </tr>
+   <tr>
+    <td>Id
+    </td>
+    <td>key/value pair list
```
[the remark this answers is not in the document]
tigrannajaryan: Reworded to make this more precise and match what we have in log data model.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,789 @@
...
+SHOULD follow OpenTelemetry <a
+href="https://github.com/open-telemetry/semantic-conventions">semantic
+conventions</a> for attributes.
+    </td>
+   </tr>
+</table>
+
+### Minimally Sufficient Id
+
+Often a number of attributes of an entity are readily available for the telemetry
```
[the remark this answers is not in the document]
tigrannajaryan: Fixed.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,789 @@
...
+producer to compose an Id from. Of the available attributes the Entity Id should
+include the minimal set of attributes that is sufficient for uniquely identifying
+that entity. No superfluous attributes should be included in the Id set. For example
```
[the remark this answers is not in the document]
tigrannajaryan: Doesn't this section explain it clearly enough? Can you suggest a different wording?

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,789 @@
...
+for "k8s.node" and "k8s.cluster" can prescribe that the IDCONTEXT of entity of type
+"k8s.node" is always an entity of type "k8s.cluster".
+
+Important: semantic conventions can NEVER prescribe the complete GID composition.
+Semantic conventions SHOULD prescribe LID and MAY prescribe IDCONTEXT, but GID
+composition, generally speaking, cannot be known statically.
+
+For example: a host's LID should be a `host.id` attribute. A host running on a cloud
+should have an IDCONTEXT of "cloud.account" and the LID of "cloud.account" entity
+is (`cloud.provider`, `cloud.account.id`). However semantic conventions cannot prescribe
```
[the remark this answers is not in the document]
tigrannajaryan: Good point. Reworded.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,789 @@
...
+If we assume that we have only one data center and host ids are globally
+unique then the above id is globally unique and is the GID of the
+process. If this assumption is not valid in our situation we would
+continue applying additional IDCONTEXT's until the GID is globally
+unique. See for example the
+[Host in Cloud Account](#host-in-cloud-account) example below.
+
+#### Process in Kubernetes
+
+A Kubernetes Collector that produces information about process entities
```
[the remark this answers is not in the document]
tigrannajaryan: Reworded.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,789 @@
...
+traces, metrics or logs.
+
+For example telemetry produced using OpenTelemetry SDK is normally associated with
+a Service. Similarly, OpenTelemetry defines system metrics for a host. The Host is the
+entity we want to associate metrics with in this case.
+
+Entities may be also associated with produced telemetry indirectly.
+For example a Service that produces
+telemetry is also related with a process in which the Service runs, so we say that
+the Service entity is related to the Process entity. The process normally also runs
```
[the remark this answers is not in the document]
tigrannajaryan: Fixed.

text/entities/0256-entities-data-model.md
tigrannajaryan: Fixed.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,789 @@
...
+that the GID of a host is (`host.id`, `cloud.provider`, `cloud.account.id`) because not all
+hosts run on cloud. A host that runs on prem in a single data center may have a GID
+of just (`host.id`) or if a customer has multiple on prem data centers they may use
+data.center.id as its identifier and use (`host.id`, `data.center.id`) as GID of the host.
+
+### Examples
+
+#### Process in a Host
+
+A locally running host agent (an Otel Collector) that produces
```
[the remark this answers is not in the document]
tigrannajaryan: Fixed.

text/entities/0256-entities-data-model.md
pellared: The definition in sem conv does not require the value to be non-empty.

jmacd (approved):

djaglowski (approved):

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+    <td>map&lt;string, any&gt;
+    </td>
+    <td>Descriptive (non-identifying) attributes of the entity.
+<p>
+MAY change over the lifetime of the entity. MAY be empty. These
+attributes are not part of entity's identity.
+<p>
+Follows <a
+href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#type-any">any</a>
+value definition in the OpenTelemetry spec - it can be a scalar value,
```
MrAlias, line 143:
```suggestion
value definition in the OpenTelemetry log spec - it can be a scalar value,
```

MrAlias: This means logs attributes will be mixed with the common attributes in the Id in the same type, right?

I see we capture this in the "open questions", but it isn't clear from that why we are doing this?  Is there a reason we need the log attribute types here?

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+    </td>
+    <td><strong>Non-identifying Attributes</strong>
+    </td>
+   </tr>
+   <tr>
+    <td>Service
+    </td>
+    <td>"service"
+    </td>
+    <td>service.name (required)
```
MrAlias, line 189: Does this mean that an entity type is not guaranteed to always have the same identifying attribute key set?

How will we distinguish between something with `service.name=foo` and `service.name=foo, service.instance.id=1`? Won't there be ambiguity when the entities are retrieved?

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+added, an existing descriptive attribute may be deleted or a value of an
+existing descriptive attribute may be changed. All these changes
+represent valid mutations of an entity over time. When these mutations
+happen the identity of the entity does not change.
+
+When the entity's state is changed it is expected that the source will
+emit a new EntityState event with a fresh timestamp and full list of
+values of all other fields.
+
+Entity event producers are recommended to periodically emit events even
```
MrAlias, line 326: To prevent confusion about intent for this to be normative:

```suggestion
Entity event producers are expected to periodically emit events even
```

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+    <td>Type
+    </td>
+    <td>
+    </td>
+    <td>See data model.
+
+</td>
+   </tr>
+   <tr>
+    <td>Id
```
MrAlias, line 298: Is the entity state used to create entities, and does the uniqueness of this Id cause the creation when the entity is not already found?

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+    </td>
+    <td>
+    </td>
+    <td>See data model
+</td>
+   </tr>
+</table>
+
+Note that EntityDelete is optional and is not guaranteed to be sent when
+the entity is gone. Recipients of entity signals should be prepared to
```
MrAlias, line 376:
```suggestion
the entity is gone. Recipients of entity signals need to be prepared to
```

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+    <td>Id
+    </td>
+    <td>
+    </td>
+    <td>See data model
+</td>
+   </tr>
+</table>
+
+Note that EntityDelete is optional and is not guaranteed to be sent when
```
MrAlias, line 375:
```suggestion
Note that transmitting EntityDelete is voluntary and is not guaranteed to be sent when
```

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+the entity is gone. Recipients of entity signals should be prepared to
+handle this situation by expiring entities that are no longer seeing
+EntityState events reported (i.e. treat the presence of EntityState
+events as a liveliness indicator).
+
+The expiration mechanism is based on the previously reported `Interval` field of
+EntityState event. The recipient can use this value to compute when to expect the next
+EntityState event and if the event does not arrive in a timely manner (plus some slack)
+it can consider the entity to be gone even if the EntityDelete event was not observed.
```
MrAlias, line 384: What should be done to EntityState values received and post-dated to an EntityDelete value?

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+unique. For example a process entity's LID is its PID number and process start time.
+The (PID,StartTime) pair is unique only in the context of a host where the process runs
+(and the host in this case is the identification context).
+
+The GID of an entity is globally unique, in the sense that for the entire set of entities
+in a particular telemetry store no 2 entities exist that have the same GID value.
+
+The GID of an entity E is defined as:
+
+`GID(E) = { LID(E), GID(IDCONTEXT(E)) }`
```
MrAlias, line 407: Can we use the word union instead of `{}`. I'm not aware of that being a standard syntax.

```suggestion
`GID(E) = UNION( LID(E), GID(IDCONTEXT(E)) )`
```

text/entities/0256-entities-data-model.md
tigrannajaryan: From [definition](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/#required):

>All instrumentations MUST populate the attribute.

I think this is good enough for the purposes of this OTEP. This is just an example, it is not a normative language and is not an OTEP topic.

We can dive deeper into this when we work on how we define entity semantic conventions. I added a TODO: https://github.com/orgs/open-telemetry/projects/85/views/1?pane=issue&itemId=63999527

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+
+In certain cases, where only one possible IDCONTEXT definition is meaningful, the
+IDCONTEXT MAY be defined in the semantic conventions. For example Kubernetes nodes
+always exist in the identifying context of a Kubernetes cluster. The semantic convention
+for "k8s.node" and "k8s.cluster" can prescribe that the IDCONTEXT of entity of type
+"k8s.node" is always an entity of type "k8s.cluster".
+
+Important: semantic conventions are not expected to (and normally won't) prescribe
+the complete GID composition.
+Semantic conventions SHOULD prescribe LID and MAY prescribe IDCONTEXT, but GID
```
MrAlias, line 450: These are normative statements in a supplementary guideline. Restructuring to move these into the non-supplementary guideline sections will help it from being missed and prevent confusion.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+container. If we begin with the same process LID:
+
+```json5
+{
+  "process.pid": 12345,
+  "process.start_time": 1714491491
+}
+```
+
+the Collector will then add the IDCONTEXT of container and of
```
MrAlias, line 514:
```suggestion
the Collector will then add the IDCONTEXT of container and
```

text/entities/0256-entities-data-model.md
tigrannajaryan: >Won't there be ambiguity when the entities are retrieved?

Yes, there will be so, the answer is "don't do that". The guarantees will come from the producers of the entity information. It's their responsibility to produce valid, non-ambiguous Ids. Generally, we heavily rely on producers doing the right thing.

Also, this is just an example, it is not normative for this OTEP. The exact conventions for services and what is a required vs non-required attribute TBD, we have a TODO for that: https://github.com/orgs/open-telemetry/projects/85/views/1?pane=issue&itemId=63999527

text/entities/0256-entities-data-model.md
tigrannajaryan: Yes and yes.

text/entities/0256-entities-data-model.md
tigrannajaryan: That indicates an entity that was gone and came back. It is a valid situation.

text/entities/0256-entities-data-model.md
tigrannajaryan: We want to limit Id key/value pairs to common attribute types. The reason for this is that Id key/values will be also duplicated as Resource attributes (more on this in the next OTEP). We control what the Id is and we can place this restriction.

As opposed to that we often don't control Entity's descriptive attributes. The information about the Entity may originate externally and have arbitrary complex structure. We need to be able to record this information in its original, structured form. This requirement is exactly similar to what we have for logs, where log attributes may also originate externally.

I am not particularly happy about this situation, that's why the open question. We will work on this and will try to find a better approach.

text/entities/0256-entities-data-model.md
tigrannajaryan: Good catch. I think this really needs to be normative. Let me change it.

text/entities/0256-entities-data-model.md
tigrannajaryan: I am not sure the word "voluntary" fits. The intent here is that if the producer is able it absolutely must sent the event, but we know that sometimes the producers won't be able to and that is a valid situation and recipients must be prepared for that. Perhaps I should simply drop the word "optional" and say that it is not guaranteed to be transmitted, e.g.

>Note that transmitting EntityDelete is not guaranteed when

text/entities/0256-entities-data-model.md
tigrannajaryan: Fixed.

text/entities/0256-entities-data-model.md
tigrannajaryan: Reworded.

text/entities/0256-entities-data-model.md
tigrannajaryan: Fixed.

text/entities/0256-entities-data-model.md
tigrannajaryan: Fixed.

text/entities/0256-entities-data-model.md
MrAlias: Do we expect producers will need to send "unset" values to try guarantee the uniformity of Id attributes? Does that mean we need a `nil` value for attributes?

text/entities/0256-entities-data-model.md
MrAlias: > Perhaps I should simply drop the word "optional" and say that it is not guaranteed to be transmitted, e.g.
> 
> > Note that transmitting EntityDelete is not guaranteed when

LGTM :+1:

bidetofevil (approved):
This new concept is great, and (I think) it provides a much more elegant solution to how we can model frequently-changing, mutable state(s) that have an outsized impact of how mobile applications run in production. The ability to model network connections, OS resource provisioning (e.g. CPU throttling, process priority, etc.), concurrent app execution, OS services and state, etc. as one or more entities in a consistent, semantic convention'd manner will go a long way into standardizing the execution context of mobile apps, something we sorely need.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+    <td><strong>Identifying Attributes</strong>
+    </td>
+    <td><strong>Non-identifying Attributes</strong>
+    </td>
+   </tr>
+   <tr>
+    <td>Service
+    </td>
+    <td>"service"
+    </td>
```
zeitlinger, line 188: A service name is unique in a service namespace, k8S cluster, k8S namespace, deployment environment, etc.

Is one producer required to produce all those attributes - or can a producer say that the service name is unique in something else, e.g. k8S cluster?

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+{
+  "process.pid": $pid,
+  "process.start_time": $starttime
+}
+```
+
+For custom entity types (not defined in OTel semantic conventions) the end-user is
+responsible for defining their custom semantic conventions in a similar way.
+
+The entity information producer is responsible for determining the identification
```
zeitlinger, line 439: How should the service know it's running in a k8S cluster for example?

text/entities/0256-entities-data-model.md
tigrannajaryan: You are right. This should be in a normative section.

text/entities/0256-entities-data-model.md
tigrannajaryan: The service likely won't know. It will be the job of the Collector to figure it out. See this section that explains how the Collector can enrich data emitted by Otel SDK: https://docs.google.com/document/d/1VUdBRInLEhO_0ABAoiLEssB1CQO_IcD5zDnaMEha42w/edit#heading=h.ij8yjheo645z

text/entities/0256-entities-data-model.md
tigrannajaryan: The [current convention](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/resource#service) is that (service name, namespace, instance id) is unique. There is nothing defined about uniqueness of service names in k8s clusters, k8s namespaces, etc.

Otel SDK is required to produce service attributes that it knows about (normally the name and instance id). Collector can enrich the produced data by service namespace.

The exact details of what attributes are used for services (and other entity types) is out of this OTEP's scope and will be a future work.

text/entities/0256-entities-data-model.md
tigrannajaryan: Fixed.

text/entities/0256-entities-data-model.md
tigrannajaryan: >Does that mean we need a nil value for attributes?

I had thought about that a while back and decided against it. It is a valid situation that different instances of entities of the same type have different sets of Id attributes. I don't think there is a need to include `nil` valued attributes just to ensure entities of the same type always have the same exact set of attributes.

If there is a use case where such uniformity is desired it can be achieved by read-time on-the-fly backfilling of missing attributes by `nil` values (and I am not sure there is a use case like that).

text/entities/0256-entities-data-model.md
zeitlinger: thanks

text/entities/0256-entities-data-model.md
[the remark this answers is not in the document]
joaopgrassi: I was thinking on re-reading that maybe we can just drop 

>No superfluous attributes should be included in the Id set.

Because above we already say:

>Of the available attributes the Entity Id should
include the minimal set of attributes that is sufficient for uniquely identifying
that entity

text/entities/0256-entities-data-model.md
MrAlias: I imagine the use-case is the error case where we want to degrade gracefully by reporting we know we should add an Id attribute but couldn't determine the value.

I ask about `nil` though, because the [common attributes](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/common#attribute) call out that setting a `null` value is not allowed. If we do want to allow this possibility in the future we would need to use different attributes now or we would have to change the attributes specification.

This doesn't seem like a blocker, but I want to call it out that if we go with this design there will not be a way for a producer to always guarantee the uniformity of Ids with the common attributes.

zeitlinger (approved):

text/entities/0256-entities-data-model.md
tigrannajaryan: >I imagine the use-case is the error case where we want to degrade gracefully by reporting we know we should add an Id attribute but couldn't determine the value.

Good point. I added a TODO to think about this: https://github.com/orgs/open-telemetry/projects/85/views/1?pane=issue&itemId=64166000

MrAlias (approved):

text/entities/0256-entities-data-model.md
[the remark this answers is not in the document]
tigrannajaryan: Good point. Done.

tigrannajaryan:
All reviewers, thank you for your comments and improvement suggestions. I think I addressed all questions. 

I will keep the PR open for 2 more business days and will merge after that if I see no objections.

text/entities/0256-entities-data-model.md
tigrannajaryan: Resolving. Feel free to reopen if there are further questions.

text/entities/0256-entities-data-model.md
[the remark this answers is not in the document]
tigrannajaryan: Resolving. Feel free to reopen if there are further questions.

text/entities/0256-entities-data-model.md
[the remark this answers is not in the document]
tigrannajaryan: Resolving. Feel free to reopen if there are further questions.

text/entities/0256-entities-data-model.md
```diff
@@ -0,0 +1,790 @@
...
+
+A locally running host agent (e.g. an OTel Collector) that produces
+information about "process" entities has the knowledge that the
+processes run in the particular host and thus the "host" is the
+identification context for the processes that the agent observes. The
+LID of a process can look like this:
+
+```json5
+{
+  "process.pid": 12345,
```
mjwolf, line 473: Hi @tigrannajaryan, I know this is only an example, but I'd like to point out this won't be sufficient for a unique process entity ID on Windows.  

On Windows, PIDS can be re-used in less than 1s, so to be unique, the start time could be more precise (nanoseconds), or more attributes are needed

tigrannajaryan: Yes, just an example and thanks for pointing out. We will work on this most likely within the semconv group when the Entities SIG makes a bit more progress on the basics.

PR #257, opened 2024-05-10
Invalid UTF-8 error handling policy

jmacd:
This is a draft following research done into 

https://github.com/open-telemetry/opentelemetry-specification/issues/3421

and

https://github.com/open-telemetry/opentelemetry-specification/issues/3950.


text/0000-utf8-handling.md
```diff
@@ -0,0 +1,205 @@
...
+### Responsibility to the user
+
+When a user uses a telemetry SDK to gain observability into their
+application, it can lead to quite the opposite result when invalid
+UTF-8 causes loss of telemetry.  When a user turns to a telemetry
+system to obseve their system and makes a UTF-8 mistake, the
+consequent total loss of telemetry is a definite bad outcome.  For the
+user to be well-served in this situation, OpenTelemetry SDKs and
+Collectors SHOULD correct invalid UTF-8 using the Unicode replacement
+character, `�` (U+FFFD) as a replacement for all invalid substrings.
```
XSAM, line 148: I just want to know the details of how the collector should deal with this requirement (it may be out of the scope of this proposal, but it is good to have a guideline for implementing this feature). Should the OpenTelemetry Collectors provide a recommended opt-in option (probably a processor) to correct invalid UTF-8?

XSAM (commented):
@jmacd Thanks for inviting me to share my thoughts.

Overall, I like this proposal. We might need to mention how to deal with the current [mapping rules](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-type-mapping.md#string-values) since if we convert invalid utf-8 string into �; we probably don't need to convert invalid utf-8 into bytes.

And, I prefer to convert invalid utf-8 into � instead of silently changing the type if we expose the byte-slice valued attributes on our API.

text/0000-utf8-handling.md
jmacd: I finished this and moved it out of draft. I propose that Collectors should be more graceful -- receivers should allow the data to be parsed by disabling UTF-8 validation in their protocol buffer libraries (as current behavior). Exporters should expect UTF-8 validation was done by a hidden processor. There may be processors that want to have valid UTF-8 and there may be processors that output modified-invalid UTF-8, which leaves open the opportunity to re-validate UTF-8 between processors, if desired.

I'd like to see this processor automatically generated from the `pdata` definitions.

jmacd:
@XSAM This was discussed in the Spec SIG today. There appears to be not much support for binary-attribute values. I think it's bad for the users, but it's not so bad if we automatically correct invalid UTF-8. Therefore, I will move forward with only half of this proposal.

text/0257-utf8-handling.md
```diff
@@ -0,0 +1,223 @@
...
+happen.  As a project, OpenTelemetry should give its component authors
+guidance on how to treat invalid UTF-8 so that users can confidently
+rely on OpenTelemetry software/systems.
+
+## Explanation
+
+OpenTelemetry has existing [error handling
+guidelines](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md)
+which dictate the project's general posture towards handling.  Users
+expect telemety libraries to be harmless and safe to use, so that they
```
Flarna, line 18:
```suggestion
expect telemetry libraries to be harmless and safe to use, so that they
```

text/0257-utf8-handling.md
```diff
@@ -0,0 +1,223 @@
...
+
+This document proposes to update OpenTelemetry error handling
+guidelines to require SDKs and Collectors to protect users from loss
+of telemetry caused by string-valued fields that are not well-formed.
+A number of options are available, all of which are better than doing
+nothing about this problem.
+
+This document also proposes to extend the OpenTelemetry API to support
+byte-slice valued attributes.  Otherwise, users have no good
+alterntives to handling this kind of data, which is often how invalid
```
joaopgrassi, line 53:
```suggestion
alternatives to handling this kind of data, which is often how invalid
```

joaopgrassi (approved):
I think it's great that this is handled. For ex, if a receiver uses the bindings we offer here https://github.com/open-telemetry/opentelemetry-proto-java they will drop the entire batch if anything contains invalid UTF-8.

text/0000-utf8-handling.md
jmacd: I revised this proposal to simplify requirements for the collector.
The current PR specifies that Collectors SHOULD validate UTF-8 after they receive data, not after processors or before exporters. I have added an Alternatives Considered documenting more-safe variations and rationale for not pursuing them.

I still believe we should auto-generate validation routines from pdata defs.

jmacd:
@open-telemetry/specs-logs-approvers @open-telemetry/specs-metrics-approvers @open-telemetry/specs-trace-approvers 

Please consider this updated OTEP. 

The changes I have applied:
- The OTel group has already decided not to support byte-valued attributes: document this. (Tough!)
- Specific wording for SDK requirements: SHOULD be opt-out, SHOULD replace invalid sequences w/ �, etc.
- Specific wording for Collector "behavior": SHOULD be opt-out, SHOULD follow each receiver for validation of external data, not recommended for processor manipulations.

text/0257-utf8-handling.md
```diff
@@ -0,0 +1,245 @@
...
+#### Automatic correction
+
+This document proposes that OpenTelemetry's default approach should be
+the one taken by the Rust `String::from_utf8_lossy` method, which
+converts invalid UTF-8 byte sequences to the Unicode replacement
+character, `�` (U+FFFD).  This is the most preferable outcome as it is
+simple and preserves what valid content can be recovered from the
+data.
+
+#### Dropping data
```
jsuereth, line 90: What's the cost of performing this check on valid strings in the collector?

I see this as a performance tradeoff for where to do the enforcement of utf-8, and my preference would be to push as much to generation side as possible.

I'll read your alternatives considered, as you probably call this out.

text/0257-utf8-handling.md
```diff
@@ -0,0 +1,245 @@
...
+telemetry as an unacceptable outcome.  OpenTelemetry SDKs and
+Collectors SHOULD prefer to drop individual items of telemetry as
+opposed to dropping whole batches of telemetry.
+
+When a Collector drops individual items of telemetry as a result of
+invalid UTF-8, it SHOULD return a `rejected_spans`,
+`rejected_data_points`, or `rejected_log_records` count along with a
+message indicating data loss.
+
+### Survey of existing systems
```
jsuereth, line 102: You should add a few more.

E.g. Java - only enforces UTF-8 when attempting to  read the bytes into Java's String format.  See: https://github.com/protocolbuffers/protobuf/blob/0bfe41b27e3dd8a30ae383210d7af10c28a642ea/java/core/src/main/java/com/google/protobuf/Internal.java#L56

text/0257-utf8-handling.md
reyang: "Generation side" - SDK/Exporter? Then the question becomes "should the collector trust the input" (I think the answer is "no").

text/0257-utf8-handling.md
```diff
@@ -0,0 +1,245 @@
...
+#### Permissive transport
+
+We observe that some protocol buffer implementations are permissive
+with respect to invalid UTF-8, while others are strict.  It can be
+risky to combine dissimilar protocol buffer implementations in a
+pipeline, since a receiver might accept data that an exporter cannot
+send, simply resulting from invalid UTF-8.
+
+Considering whether components could support permissive data
+transport, it appears to be risky and for little reward.  If
```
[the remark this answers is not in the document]
reyang: > I like having validaiton as opt-in/opt-out, I'm not sure which should be the default though.

+1, I don't just "like" it, I think we SHOULD do this.

Here are my reasons:

1. The collector is normally sending the data to some endpoints. Many backend services already perform such validation/correction, so folks might want to just do it once in the backend rather than duplicating the effort.
2. There are cases where data can be consumed directly on the collector side (e.g. a collector running in a local data center might decide to trigger a rollback due to certain metrics KPI drop during a deployment), I think this is a general pattern for things that run on Edge.
3. Depending on the ownership of different parts of the system, the parts could be designed to be trusting each other or not. Collector needs to provide flexibility.
4. Things could break between Collector and backend (e.g. bit flips caused by high energy particles from the universe, hardware failures), certain software needs to handle these as part of the design.

Regarding which one should be the default, based on what I've seen in Microsoft across Windows/Office/Azure/etc. I think it should be off-by-default and allow folks to opt-in.

jmacd:
@jsuereth and @reyang I appreciate the feedback. Both of you are, I think, suggesting to make UTF-8 validation an opt-in instead of an opt-out feature. I support that motion. The most critical thing for me is that if the SDK is configured with a permissive stance (opt-out), the SDK "MUST" configure its underlying technologies in support. 

Opting-out does not mean doing nothing, in other words, it means explicitly configuring a pipeline to permit invalid UTF-8 unless a user opts-in to UTF-8 validation.

When UTF-8 validation is selected (opt-in), it seems we have two options: (a) reject individual items, (b) correct invalid UTF-8. Do either of you think both of these options are worthwhile? I think (b) should be preferred, but I would accept (a) too.

reyang:
> When UTF-8 validation is selected (opt-in), it seems we have two options: (a) reject individual items, (b) correct invalid UTF-8. Do either of you think both of these options are worthwhile? I think (b) should be preferred, but I would accept (a) too.

I think if we have very limited bandwidth, we should do (b). (a) can be added later if we see a huge demand.
One technical detail - I think for attribute values with string type, we should do some correction, for attribute names that have invalid UTF-8, it could be a very bad idea. I'm a bit on the fence here...

text/0257-utf8-handling.md
```diff
@@ -0,0 +1,245 @@
...
+SDKs MAY permit users to opt-out of UTF-8 validation for performance reasons.
+
+Optional UTF-8 validation MUST be applied before the attribute
+transformation rule stated for invalid UTF-8 strings.  This means
+UTF-8 validation prevents downgrading to a `bytes_value` encoding
+for string attributes.
+
+#### No byte-slice valued attribute API
+
+As a caveat, the OpenTelemetry project has previously debated and
```
tsloughter, line 168: Was this ever formally rejected? 

I'd like this option because we technically already support it in Erlang. The fact the main string type in Erlang/Elixir is `binary` and the SDK safety mechanism mentioned elsewhere in this doc that stores invalid utf8 in `bytes_value` of the proto. 

So because attribute values are already type `binary` the user can pass any binary data they want as an attribute value and it gets used.

I recall it being rejected informally in spec sig meetings but maybe different luck with a formal proposal. Do you think there would be any chance of that?

carlosalberto:

PR #241, opened 2023-11-14
Inter-process context propagation requirements

deejgregor:
https://github.com/open-telemetry/opentelemetry-specification/issues/740

PR #258, opened 2024-05-25
[otep] Propose adding env variables as context carriers to specification

adrielp:
Based on conversations last week in the Specification and Semantic Conventions SIGs, I'm opening this duplicate pull request which was originally set as a [Draft](https://github.com/open-telemetry/oteps/pull/241/files) and hasn't had movement since last November. 

There are real use cases that are coming to fruiting, namely in the CI/CD working group, that will benefit from this being accepted. Once accepted we can work on getting the specification added for both general context propagation and baggage.

On the note of baggage; baggage is a form of context propagation and was not originally mentioned directly by name in this OTEP. It is however, absolutely essential. I've had the pleasure of prototyping out tracing within an OpenTofu controller system where context on available in parent/child at the very start of the trace was available. Baggage was the means of transferring this critical context to subsequent siblings that would've not had it otherwise. 

Thanks for all the hard work to the original author (@deejgregor) and opening the draft #241 

CC. TC sponsors @jsuereth @carlosalberto


deejgregor:
@adrielp thank you for picking up this work. I appreciate that while I’ve been going through a busy time.

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,359 @@
...
+
+This OTEP is based on discussion in [opentelemetry-specification #740](https://github.com/open-telemetry/opentelemetry-specification/issues/740#issue-665588273), builds on the [Context Propagators
+API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md)
+and follows [OTEP 0205: Context propagation requirements for messaging semantic conventions](https://github.com/open-telemetry/oteps/blob/main/text/trace/0205-messaging-semantic-conventions-context-propagation.md)
+as a guide.
+
+## Motivation
+
+There are existing requirements for context propagation over HTTP,
+messaging systems, and others, but there are no requirements for
```
yurishkuro, line 14: > there are no requirements

in accurate - there are both requirements and implementations, what's lacking is a standard/specification.

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,359 @@
...
+
+On Windows, in particular, because environment variable keys are case
+insensitive, there is a (low?) chance that automatically instrumented context
+propagation variables could conflict with existing application environment
+variables. It will be important that this instrumentation can be disabled
+should such cases arise.
+
+### Privacy and Security
+
+Environment variables are visible to any process on the system, so it is
```
yurishkuro, line 298: only root can see env vars of other processes, not any process.

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,359 @@
...
+          |
+          | Environment variable(s) used as the carrier
+          | for context propagation, e.g.: TRACEPARENT
+          v
+  +----------------+
+  | Child Process  |
+  +----------------+
+```
+
+## Proposed addition to Process and Process Runtime Resources Semantic Conventions
```
yurishkuro, line 50: I am not following how propagation format is in any way related to semantic conventions and resources.

text/0258-env-context-baggage-carriers.md
adrielp: Fair point on the wording here, I'll adjust.

text/0258-env-context-baggage-carriers.md
adrielp: I agree. This should be `specification` imo.

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,359 @@
+# Environment Variables as Context and Baggage Carriers
+
+Add requirements for inter-process context and baggage propagation using
+environment variables as the carrier.
```
yurishkuro, line 4: the purpose of the OTEP is not clear to me. What does "add requirements" mean? Requirements for whom? Is it proposing to extend specification? Why does it need to be extended? 

Context propagation in general is meant to enhance inter-process communication (IPC) with the ability to pass context/baggage. It puts very high-level requirements on the communication protocol, namely to be able to carry additional k-v metadata that propagators use to encode baggage. HTTP as protocol satisfies that requirement, so does parent-child IPC via env vars.

Really the only new thing needed to enable env var IPC is a "carrier" that can deal with case (in)sensitivity of the names, similar to HTTP carrier that does the same. It would be much clearer if the OTEP was explicit about that and limited to just that.

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,359 @@
...
+inter-process context propagation, although there is considerable
+ad-hoc propagation being done using environment variables today (see [Prior Art](#prior-art)).
+This OTEP is to formalize a standard based on the existing informal
+consensus.
+
+These are some examples of batch and system use cases that could
+benefit from this standardization:
+
+* CI/CD systems
+* Build and deployoment tooling
```
christophe-kamphaus-jemmic, line 24: Typo
```suggestion
* Build and deployment tooling
```

text/0258-env-context-baggage-carriers.md
adrielp: @yurishkuro  - thank you for reviewing! The conversation getting started was exactly what I was looking for when copying this forward into a new pull request. After reading your comments & since the original content has now had one review, I've re-written in based on additional context and your feedback. I've also changed the title to make more clear what is being requested. Pretty much everything has been re-written so please feel free to give another review and let me know if this is more clear & and in-line with what you're looking for in an OTEP.

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,236 @@
...
+Python automatically upper-cases environment variables. The issue was merged and
+this [documentation][cpython-doc] was added to clarify the behavior.
+
+[cpython]: https://github.com/python/cpython/issues/101754
+[cpython-doc]: https://docs.python.org/3/library/os.html#os.environ
+
+### Allowed characters
+
+The characters allowed in keys by the `TextMapPropagator` are also allowed as
+environment variable keys on UNIX and Windows.
```
yurishkuro, line 156: I am not following this. Are you saying that say `&` is allowed in the name of the environment variable?

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,236 @@
...
+> / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
+> / DIGIT / ALPHA
+
+To ensure compatibility, specification for Environment Variables SHOULD adhere
+to the current specification for `TextMapPropagator` where key/value pairs MUST
+only consist of US-ASCII characters that make up valid HTTP header fields as
+per RFC 7230.
+
+One key note is that windows disallows the use of the `=` character in
+environment variables. See [MS Env Vars][ms-env] for more information.
```
yurishkuro, line 170: > environment variables

names or values?

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,236 @@
...
+in systems where HTTP communication does not occur, a specification for
+Environmental Variables can be used.
+
+Consider the following diagram:
+
+![Environment Variable Context Propagation](./img/0258-env-context-parent-child-process.png)
+
+In this diagram, a parent process is forked to spawn a child process,
+inheriting the environment variables from the original parent. The environment
+variables defined here, `TRACEPARENT`, `TRACESTATE`, and `BAGGAGE` are used to
```
yurishkuro, line 56: > The environment variables defined here

What does this phrase mean? Where is "here"? These concrete three names are what W3C Trace Context spec defines, they can be one possible implementation, e.g. if the SDK is configured with W3CPropagator. But the actual "env vars as carriers" should not care about which names are used, it should worry about the encoding of those names (and values) in such a way that the resulting output is safe for env vars (ideally across many OSs, ie. the lowest common denominator).

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,236 @@
...
+propagate context to the child process such that it can be tied to the parent.
+Without `TRACEPARENT`, a tracing backend would not be able to connect the child
```
yurishkuro, line 58: one thing that is unclear to me, are we proposing that the parent process should alter the env vars before creating a child process? It's never stated, the text reads like the child just inherits the same vars as parent. I think this would be a problem because if you imaging a hierarchy of child processes and they all inherit TRACEPARENT from the top parent process, then the hierarchy will not be represented in the trace (all spans will be children of the same root span). A common way to deal with this in OpenTelemetry is by having two levels of spans in each process, one for the entry point (`server` or `consumer` kinds) and another for each outbound IPC (`client` or `producer` spans). It's the latter that are propagated to downstream processes and form a call tree.

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,236 @@
...
+
+To propagate context and baggage between parent, and child processes
+in systems where HTTP communication does not occur, a specification for
+Environmental Variables can be used.
+
+Consider the following diagram:
+
+![Environment Variable Context Propagation](./img/0258-env-context-parent-child-process.png)
+
+In this diagram, a parent process is forked to spawn a child process,
```
yurishkuro, line 54: Why should parent be forked to spawn a child process? It's one possible implementation, but if I don't need to clone the parent memory space and just need to start a child process (e.g. via `os.system()` in Python), then it's not "forking".

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,236 @@
...
+Defining the specification for Environment Variables as carriers will have a
+wide impact to the industry in enabling better observability to systems outside
+of the normal HTTP microservice architecture.
+
+[w3c-parent]: https://www.w3.org/TR/trace-context/#traceparent-header
+[w3c-state]: https://www.w3.org/TR/trace-context/#tracestate-header
+[w3c-bag]: https://www.w3.org/TR/baggage/#header-name
+[bag-api]: https://opentelemetry.io/docs/specs/otel/baggage/api/
+
+## Internal details
```
yurishkuro, line 108: I think this should be the core of the proposed spec changes, not "internal details". The spec should be for env-var carrier, which should describe how key-value pairs produced by an arbitrary TextMapPropagator can be encoded in a safe and lossless way in env variables.

text/0258-env-context-baggage-carriers.md
yurishkuro: So I think it would be useful if the spec outlined how instrumentation using the new env-var carrier should behave, i.e. if I create a "client" style span, how should its context be transmitted to the child process.

text/0258-env-context-baggage-carriers.md
adrielp: This diagram was simply showing an example and not language specific. It's simply a visual aid. If it needs to be adjusted with the different values to reflect possibilities, then I can work on that.

text/0258-env-context-baggage-carriers.md
adrielp: This entire paragraph is discussing the example diagram above. "Here" means in that diagram. 

Are you saying that we should standardize on what the environment variable names are for context propagation? OpenTelemetry already has standards for environment variable names in its conventions. I don't understand why we wouldn't do the same here. Am I misunderstanding your statement?

text/0258-env-context-baggage-carriers.md
yurishkuro: Ok, I mistook the scope of the statements. I think forking is still an interesting use case, this comment https://github.com/open-telemetry/oteps/pull/258#discussion_r1616264598 still applies.

text/0258-env-context-baggage-carriers.md
adrielp: That's a good point. I agree that there would be issues if they weren't able to alter the env vars and I agree that the spec should outline the behavior you mention. I can call out that specific need for the spec to define the behavior in this OTEP.

text/0258-env-context-baggage-carriers.md
yurishkuro: > OpenTelemetry already has standards for environment variable names in its conventions.

(a) they are used to configure SDK, not to affect per-request behavior, 

(b) we are looking to deprecate them in favor of configs, 

(c) the names of the metadata keys that constitute context/baggage are determined by the individual propagators. E.g. Jaeger propagator would use different HTTP headers than W3C Propagator. The same applies to env vars - there is no need to standardize on them, the names are already defined by propagators. But we need to standardize on encoding rules for them.

text/0258-env-context-baggage-carriers.md
carlosalberto: only for names, so we are fine, value-wise (but this should be clarified).

text/0258-env-context-baggage-carriers.md
carlosalberto: +1

text/0258-env-context-baggage-carriers.md
carlosalberto: > But the actual "env vars as carriers" should not care about which names are used

Agreed. We can recommend a few environment variables names for consistency purposes (e.g. BAGGAGE), but I think that should be all.

text/0258-env-context-baggage-carriers.md
kuisathaverat: Since the context propagation can use environment variables, a fork is not needed.

for example, you can run any command with the TRACEPARENT environment variable exported, then all processes in that shell will use the TRACEPARENT to propagate the context

```shell
export TRACEPARENT="00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01" 
my-new-process0
my-new-process1
my-new-process2
my-new-process3
```

text/0258-env-context-baggage-carriers.md
kuisathaverat: >(a) they are used to configure SDK, not to affect per-request behavior,

AFAIU, the environment variables will affect any process that is running in the context where they are defined, so if you send and OpenTelemetry span (for example) in that context those requests should be affected by the context in the environment variables that will be the parent context of everything in that shell context (I say shell by simplicity).

kuisathaverat: >(b) we are looking to deprecate them in favor of configs,

There is previous art in a few places, so IMO we should support the current environment variables in use. But it is not bad idea to allow use `OTEL_*` variants too.

kuisathaverat: >(c) the names of the metadata keys that constitute context/baggage are determined by the individual propagators. E.g. Jaeger propagator would use different HTTP headers than W3C Propagator. The same applies to env vars - there is no need to standardize on them, the names are already defined by propagators. But we need to standardize on encoding rules for them.

We are talking about passing the OpenTelemetry parent context between processes, so it should not affect the propagation of HTTP headers.

For example, Suppose I run the application in the command line and pass a TRACEPARENT environment variable with the parent context related to my build in progress. In that case, all the spans and distributing tracing generated spans by that application will have that TRACEPARENT as the parent span. All the rest should work as work now in distributing tracing.

text/0258-env-context-baggage-carriers.md
kuisathaverat: 🤔 when you pass TRACEPARENT you do it on purpose so your intention is that all the child spans have that parent context. If you do not want to pass that parnet context to all spans you should use OTEL_EXPORTER_OTLP_HEADERS or OTEL_EXPORTER_OTLP_TRACES_HEADERS, or OTEL_EXPORTER_OTLP_LOGS_HEADERS or OTEL_EXPORTER_OTLP_METRICS_HEADERS, or other implemented variables by the application used

text/0258-env-context-baggage-carriers.md
adrielp: That's the way it reads. This was one of the original statements. I've removed It. From what I'm reading, the only values allowed in the keys are letters, numbers and hyper/underscore. Updating this to reflect.

text/0258-env-context-baggage-carriers.md
adrielp: I updated the core of the spec to encompass this. Let me know if this is the direction you're thinking. Im going to be re-reading and making further edits to, but wanted to gauge direction.

text/0258-env-context-baggage-carriers.md
adrielp: I think this has been addressed through refinement of the OTEP and gearing towards TextMap Propagation where  injector/extractor is defined, etc. Feel free to give another review and see if this change makes sense.

text/0258-env-context-baggage-carriers.md
adrielp: I think this has been addressed through a couple iterations.

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,294 @@
...
+* [Open Questions](#open-questions)
+* [Future Possibilities](#future-possibilities)
+
+## Motivation
+
+The motivation for defining the specification for context and baggage
+propagation by using environment variables as carriers stems from a long open
+issue on the OpenTelemetry Specification repository, [issue #740][issue-740].
+This issue has been open for such a long time that multiple groups have gone
+forward in implementing their own solutions to the problem using `TRACE_PARENT`
```
pellared, line 31:
```suggestion
forward in implementing their own solutions to the problem using `TRACEPARENT`
```

pellared (commented):
👍

pellared (approved):
Awesome OTEP 👍

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,294 @@
...
+* `parent-id`
+* `trace-flags`
+
+This could be set in the environment as follows:
+
+```bash
+export TRACEPARENT=version=2HEXDIGLC,trace-id=32HEXDIGLC,parent-id=16HEXDIGLC,trace-flags=2HEXDIGLC
+```
+
+`tracestate` (lowercase), originates in the [W3C Specification][w2c-state] and
```
lmolkova, line 102: same not on "originates" - could we call out that the format is the same?

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,294 @@
...
+can include any opaque value in a key-value pair structure. Its goal is to
+provide additional vendor-specific trace information.
+
+`baggage` (lowercase), also is defined in the [W3C Specification][w3c-bag] and
```
lmolkova, line 106:
```suggestion
`baggage` (lowercase), also is defined in the [W3C Baggage][w3c-bag] and
```

lmolkova (approved):
Added some minor comments.

Approving under the assumption that traceparent and tracestate format is intended to be identical to the W3C Trace-Context spec (and that we document it explicitly)

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,294 @@
...
+to the root span, creating a single trace as shown in the image of a real
+OpenTofu trace below.
+
+![OpenTofu Trace](./img/0258-env-context-opentofu-trace.png)
+
+Additionally, the `init` span is able to pass baggage to the `plan` and `apply`
+spans. One example of this is module version and repository information. This
+information is only determined and known during the `init` process. Subsequent
+processes only know about the module by name. With `BAGGAGE` the rest of the
+proccesses are able to understand a key piece of information which allows
```
magnusbaeck, line 135:
```suggestion
processes are able to understand a key piece of information which allows
```

carlosalberto:
Overall LGTM, albeit I'd like to have a pair of things called out:

* On propagating other formats: we are taking for granted we will **only** want to propagate `traceparent` (tracecontext), whereas there is a small chance that for legacy reasons b3 or similar is supported in a given environment, hence:
* Suggest/standardize even the usage of `TRACEPARENT`, but allow the user to override the environment variable name (useful for overriding purposes + allowing the easy usage of any legacy format).
  - I believe this is mentioned as part of the "Changed" section, but a clearer note would help IMHO.

carlosalberto (approved):

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,294 @@
...
+
+The motivation for defining the specification for context and baggage
+propagation by using environment variables as carriers stems from a long open
+issue on the OpenTelemetry Specification repository, [issue #740][issue-740].
+This issue has been open for such a long time that multiple groups have gone
+forward in implementing their own solutions to the problem using `TRACEPARENT`
+and `TRACESTATE` environment variables.
+
+[Issue #740][issue-740] identifies several use cases of systems that do not
+communicate across bounds by leveraging HTTP communications such as:
```
svrnm, line 35:
```suggestion
[Issue #740][issue-740] identifies several use cases of systems that do not
communicate across bounds by leveraging network communications such as:
```

I guess HTTP is fine, but my understanding is that the use cases below are process-boundaries vs network-boundaries which have headers, meta data and other existing mechanisms to transfer context.

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,294 @@
...
+
+[cicd-wg]: https://github.com/open-telemetry/community/blob/main/projects/ci-cd.md
+[issue-740]: https://github.com/open-telemetry/opentelemetry-specification/issues/740#issue-665588273
+
+## Design
+
+To propagate context and baggage between parent, sibling, and child processes
+in systems where HTTP communication does not occur between processes, a
+specification using key-vaulue pairs injected into the environment can be read
+and produced by an arbitrary TextMapPropagator.
```
svrnm, line 62:
```suggestion
To propagate context and baggage between parent, sibling, and child processes
in systems where network communication does not occur between processes, a
specification using key-vaulue pairs injected into the environment can be read
and produced by an arbitrary TextMapPropagator.
```

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,294 @@
...
+is a set of key-value pairs to propagate context between signals. In
+OpenTelemetry, baggage is propagated through the [Baggage API][bag-api].
+
+[w3c-parent]: https://www.w3.org/TR/trace-context-2/#traceparent-header-field-values
+[w3c-state]: https://www.w3.org/TR/trace-context-2/#tracestate-header
+[w3c-bag]: https://www.w3.org/TR/baggage/#baggage-http-header-format
+
+#### Distributed Tracing in OpenTofu Prototype Example
+
+Consider this real world example OpenTofu Controller Deployment.
```
svrnm, line 116: Maybe add a few words to the end of this section, that there is an ongoing discussion in the OpenTofu project around introducing OpenTelemetry support (thanks @apparentlymart for linking this OTEP and the OpenTofu RFC [here](https://github.com/opentofu/opentofu/pull/2028#issuecomment-2397631674))

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,294 @@
...
+[python-env]: https://github.com/Div95/opentelemetry-python/tree/feature/env_propagator/propagator/opentelemetry-propagator-env
+[swift-env]: https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/OpenTelemetrySdk/Trace/Propagation/EnvironmentContextPropagator.swift
+
+### UNIX Limitations
+
+UNIX system utilities use upper-case for environment variables and lower-case
+are reserved for applications. Using upper-case will prevent conflicts with
+internal application variables.
+
+Environment variable names used by the utilities in the XCU specification
```
svrnm, line 191:
```suggestion
Environment variable names used by the utilities in the Shell And Utilities (XCU) specification
```

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,294 @@
...
+3. This would deviate from significant prior art that currently uses
+   environment variables.
+
+## Open questions
+
+The author has no open questions at this point.
+
+## Future possibilities
+
+1. Enabling distributed tracing in systems that do not communicate over HTTP.
```
svrnm, line 294:
```suggestion
1. Enabling distributed tracing in systems that do not communicate over network protocols, that allow trace context being propagated through headers, meta data or other means.
```

svrnm (commented):
overall lgtm, some comments, especially around the use of HTTP vs network in general.

svrnm (approved):
LGTM!

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,306 @@
...
+To propagate context and baggage between parent, sibling, and child processes
+in systems where network communication does not occur between processes, a
+specification using key-vaulue pairs injected into the environment can be read
+and produced by an arbitrary TextMapPropagator.
+
+### Example Context
+
+Consider the following diagram in the context of process forking:
+
+> Note: The diagram simply an example and simplification of process forking.
```
trask, line 68:
```suggestion
> Note: The diagram is simply an example and simplification of process forking.
```

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,306 @@
...
+
+#### Distributed Tracing in OpenTofu Prototype Example
+
+Consider this real world example OpenTofu Controller Deployment.
+
+![OpenTofu Run](./img/0258-env-context-opentofu-tracing.png)
+
+In this model, the OpenTofu Controller is the start of the trace, containing
+the actual trace_id and generating the root span. The OpenTofu controller
+deploys a runner which has it's own environment and processes to run OpenTofu
```
trask, line 126:
```suggestion
deploys a runner which has its own environment and processes to run OpenTofu
```

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,306 @@
...
+commands. If one was to trace these processes without a carrier mechanism, then
+they would all show up as unrelated root spans as separate traces. However, by
```
trask, line 128:
```suggestion
they would all show up as unrelated root spans in separate traces. However, by
```

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,306 @@
...
+  `BAGGAGE` that can be used to propagate context between processes.
+* A specification for allowed environment names and values due to operating
+  system limitations.
+* A specification for how implementers can inject and extract context from the
+  environment through a TextMapPropagator.
+* A specification for how processes should update environment variables before
+  spawning new processes.
+
+Defining the specification for Environment Variables as carriers for context
+will enable SDK's and other tools to implement getters and settings of context
```
trask, line 173:
```suggestion
will enable SDK's and other tools to implement getters and setters of context
```

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,306 @@
...
+to the specification.
+
+Two implementations already exist within OpenTelemetry for environment
+variables through the TextMap Propagator:
+
+* [Python SDK][python-env] - This implementation uses environment dictionary as
+  the carrier in Python for invoking process to invoked process context
+  propagation. This pull request does not appear to have been merged.
+* [Swift SDK][swift-env] - This implementation uses `TRACEPARENT` and
+  `TRACESTATE` environment variables alongside the w3cPropagator to inject and
```
trask, line 185:
```suggestion
  `TRACESTATE` environment variables alongside the W3C propagator to inject and
```

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,306 @@
...
+
+* `version`
+* `trace-id`
+* `parent-id`
+* `trace-flags`
+
+This could be set in the environment as follows:
+
+```bash
+export TRACEPARENT=version=2HEXDIGLC,trace-id=32HEXDIGLC,parent-id=16HEXDIGLC,trace-flags=2HEXDIGLC
```
trask, line 103: why not use the same format as the `traceparent` header? (sorry if I missed in previous discussions)

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,294 @@
...
+propagate context to the child process such that it can be tied to the parent.
+Without `TRACEPARENT`, a tracing backend would not be able to connect the child
+process spans to the parent span, forming an end-to-end trace.
+
+> Note: While the below exclusively references the W3C Specification, this
+> proposal is not exclusive to W3C and is instead focused on the mechanism of
+> Text Map Propagation with a potential set of well-known environment variable
+> names.
+
+`traceparent` (lowercase), originates in the [W3C Specification][w3c-parent]
```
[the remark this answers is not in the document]
lmolkova: I'm not sure if there was any clarification made, I still find "originates" problematic and would like to see something definitive saying that the format is exactly the same as W3C trace-context..

adrielp:
> Overall LGTM, albeit I'd like to have a pair of things called out:
> 
> * On propagating other formats: we are taking for granted we will **only** want to propagate `traceparent` (tracecontext), whereas there is a small chance that for legacy reasons b3 or similar is supported in a given environment, hence:
> * Suggest/standardize even the usage of `TRACEPARENT`, but allow the user to override the environment variable name (useful for overriding purposes + allowing the easy usage of any legacy format).
>   
>   * I believe this is mentioned as part of the "Changed" section, but a clearer note would help IMHO.

I've adjust the changed section slightly. Let me know if that tracks with what you were hoping for.

text/0258-env-context-baggage-carriers.md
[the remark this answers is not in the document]
adrielp: I've changed `originates` to `is defined in`. This whole section is the "example context" section and those fields/environment variable names are those that were defined in W3C for this particular example.

austinlparker (approved):

adrielp:
> Added some minor comments.
> 
> Approving under the assumption that traceparent and tracestate format is intended to be identical to the W3C Trace-Context spec (and that we document it explicitly)

@lmolkova - after the SIG, I pushed another update which hopefully makes this more clear within the document. Please feel free to give it another review and let me know if the changes do indeed add that clarity.

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,311 @@
...
+
+[cicd-wg]: https://github.com/open-telemetry/community/blob/main/projects/ci-cd.md
+[issue-740]: https://github.com/open-telemetry/opentelemetry-specification/issues/740#issue-665588273
+[tmp]: https://opentelemetry.io/docs/specs/otel/context/api-propagators/#textmap-propagator
+
+## Design
+
+To propagate context and baggage between parent, sibling, and child processes
+in systems where network communication does not occur between processes, a
+specification using key-vaulue pairs injected into the environment can be read
```
christophe-kamphaus-jemmic, line 61:
```suggestion
specification using key-value pairs injected into the environment can be read
```

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,311 @@
...
+process spans to the parent span, forming an end-to-end trace.
+
+> Note: While the below exclusively follows the W3C Specification translated
+> into environment variables, this proposal is not exclusive to W3C and is
+> instead focused on the mechanism of Text Map Propagation with a potential set
+> of well-known environment variable names. See the [Core Specification
+> Changes](#core-specification-changes) section for more information.
+
+Given the above example aligning with the W3C Specification, the following is
+an contextual mapping of environment variables to headers defined by W3C.
```
christophe-kamphaus-jemmic, line 88:
```suggestion
a contextual mapping of environment variables to headers defined by W3C.
```

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,311 @@
...
+[ms-env]: https://learn.microsoft.com/en-us/windows/win32/procthread/environment-variables
+
+## Trade-offs and Mitigations
+
+### Case-sensitivity
+
+On Windows, because environment variable keys are case insensitive, there is a
+chance that automatically instrumented context propagation variables could
+conflict with existing application environment variables. It will be important
+to denote this behavior and identify document how languages mitigate this
```
christophe-kamphaus-jemmic, line 257:
```suggestion
to denote this behavior, identify and document how languages mitigate this
```

trask (approved):

christophe-kamphaus-jemmic (approved):

text/0258-env-context-baggage-carriers.md
```diff
@@ -0,0 +1,310 @@
...
+to denote this behavior and document how languages mitigate this issue.
+
+### Security
+
+Do not put sensitive information in environment variables. Due to the nature of
+environment variables, an attacker with the right access could obtain
+information they should not be privy too. Additionally, the integrity of the
+environment variables could be compromised.
+
+## Prior Art and Alternatives
```
tylerbenson, line 266: This is very similar to something the Lambda SIG had to deal with around AWS Lambda's Active Tracing context propagation via an X-Ray Environment Variable.  One area we ran into when trying to move over to handling this in a propagator instead of directly in instrumentation was around Span Linking.  If the Environment Variable propagator takes priority over other listed propagators in the global propagator chain, then it breaks span linking.  One solution we considered to this problem is allowing the span link to consider all propagators' results, and not just the winning one. I still think this is a good solution that should be considered.

carlosalberto: @tylerbenson actually there was a plan to work on this even without env vars, so we should look into that (independently from this OTEP).

tylerbenson: There was a fairly lengthy discussion here: https://github.com/open-telemetry/semantic-conventions/pull/354#discussion_r1414457602

adrielp: Thanks! I'm going to resolve this comment.

PR #259, opened 2024-05-29
WIP: pipeline monitoring otep

kristinapathak:
A continuation of @jmacd's work on https://github.com/open-telemetry/oteps/pull/238 and https://github.com/open-telemetry/oteps/pull/249

The goal of this OTEP is to define a semantic convention for metrics that provide information on the flow of data through a pipeline, providing insights both between and within segments of telemetry pipelines.

WIP.

Main focuses currently:
1. Measuring loss in the pipeline: providing examples that show how failures show up in the proposed metric instruments.
2. Example scenarios and adding descriptions to diagrams.


text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,495 @@
...
+## Explanation
+
+This document proposes two metric instrument semantics and three overall metric 
+instruments for pipeline accounting, considering SDK and Collector pipeline 
+metrics as separate, non-intersecting signals for reasons discussed below. The 
+three instruments are:
+
+- `otelcol_outgoing_items`: Exported, dropped, and discarded items (Collector)
+- `otelcol_incoming_items`: Received and inserted data items (Collector)
+- `otelsdk_outgoing_items`: Exported, dropped, and discarded items (SDK)
```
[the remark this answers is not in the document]
kristinapathak: @TylerHelmuth, you left feedback on Josh's PR about this.  Please let me know your preference.

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,495 @@
...
+
+The specified counter names are:
+
+- `otelcol_incoming_items`: The number of items received or inserted into a pipeline.
+- `otelcol_outgoing_items`: The number of items discarded, dropped, or exported by a Collector pipeline segment.
+- `otelsdk_outgoing_items`: The number of items discarded, dropped, or exported by a SDK pipeline segment.
+
+### Recommended conventional attributes
+
+- `otel.error` (boolean): This is true or false depending on whether the
```
kristinapathak, line 214: @TylerHelmuth, you [recommended](https://github.com/open-telemetry/oteps/pull/249#discussion_r1552062062) adding `pipeline` to the `otel.` prefix here (ie `otel.pipeline.error`.  Do you mean for all of the attributes below? I'm not sure how these are all attributes of an otel pipeline.

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,495 @@
...
+  outcome is considered a failure or a success.  See the chart below.
+- `otel.outcome` (string): This describes the outcome in a more specific
+  way than `otel.error`.
```
0x006EA1E5, line 217: @kristinapathak As mentioned at the SIG, I am interested in adding a similar attribute to the `otelcol_exporter_send_failed_*` metrics as part of https://github.com/open-telemetry/opentelemetry-collector/pull/10158 

I would be happy to use `outcome` if that is thought best.  

My one concern is that I see that attribute is used on another metric we use in our org, specifically the Micrometer generated `http.server.requests` metric, see [here for the ENUM values](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/http/Outcome.java).  I see that this is **not** defined for the http [semconv](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/), but I just thought it worth noting here for reference.

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,495 @@
...
+- `deferred:<failure outcome>`: Deferred cases are where the preceder receives 
+  a success response and the true outcome is failure, but this is not known 
+  until later. The item is counted as `deferred:` combined with the failure 
+  outcome that would otherwise have been counted. The failure outcome could be 
+  transient or permanent. An example of when deferred would be used is in an 
+  asynchronous Collector pipeline.
+
+Additional examples of these outcomes can be found in the Appendix.
+
+### Collector Pipelines With Multiple Exporter Components
```
djaglowski, line 145: There are some other arrangements of components which I'm trying to fit to the proposed model. Some of them may be worth including in this document as well. Here's how I'm understanding them:
1. _A single collector pipeline with multiple exporters_. As already noted, this includes a fanout point. The document describes how to aggregate outcomes from multiple exporters, effectively ensuring that `Incoming(Segment) == Outgoing(Segment)`.
2. _A single collector pipeline with multiple receivers_. This is fairly straightforward and common. The incoming items are just summed together. Probably doesn't require a dedicated section but we could include it for symmetry.
3. _A single receiver shared by multiple collector pipelines_. This is the other type of fanout point in the collector. From here there are some similar considerations to the first case, but instead of fanning out to multiple exporters, we instead fanout to entire pipelines. What is the relationship between an item arriving at such a receiver and the outcomes of passing it to multiple pipelines? For example, one pipeline may successfully export the item while the other encounters an error which propagates back to the receiver. Does this count as "origin:received" for both pipelines and then each pipeline show a different outcome?
4. _A single exporter shared by multiple collector pipelines_. This is the other type of merge point in the collector. In this case, I think synchronous outcomes can resolve back to a specific pipeline, but async outcomes may not be related to any specific pipeline. For example, if 2 pipelines each send 10 items to an exporter, which then batches all 20 items together into a single export request, the outcome may be "20 deferred:rejected", but it would be incorrect for either pipeline to incorporate that count directly. Is there any way to handle this? Otherwise, maybe this is just a caveat for interpreting deferred outcomes.

djaglowski (commented):
Overall this proposal strikes me as more intuitive than previous iterations but I still have a few questions regarding collector pipelines.

text/metrics/0238-pipeline-monitoring.md
kristinapathak: I can see how `outcome` will be used by a variety of metrics. My hope is that with the `otel` prefix (ie `otel.outcome`) there is no conflict with the attribute name.

Do the outcome values defined here work to solve https://github.com/open-telemetry/opentelemetry-collector/issues/10157 or is more detail needed? If others are also in favor of this attribute, my hope is that you can update your PR to match this. 🙂

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,495 @@
...
+  telemetry.
+
+The terms "following"/"follower" and "preceding"/"preceder" refer to the 
+relative services after or before another service in a pipeline. The preceding 
+service ("preceder") transmits telemetry to the following service ("follower").
+
+An arrangement of discrete pieces of functionality ("components") within a 
+service acting as a single unit is called a segment. Segments can be found in 
+the OpenTelemetry Collector as Collector pipelines. A Collector can contain 
+multiple Collector pipelines which can contain multiple segments. Each segment
```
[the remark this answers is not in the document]
kristinapathak: The segments would be defined as:
- `r1, r2, p1, p2, e1`
- `r1, r2, p1, p2, e2`

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,495 @@
...
+exporters. If the pipeline is synchronous, the outcome for the incoming item is
+recorded based on the rules in the below order:
+
+1. If there is a permanent error, that is used as the outcome. If there are
+   multiple permanent errors, choose them in the following order:
+   `rejected`, `deferred:rejected`, `unknown`, `deferred:unknown`.
+2. If there is a transient error, that is used as the outcome. If there are
+   multiple transient errors, choose them in the following order:
+   `dropped`, `deferred:dropped`, `timeout`, `deferred:timeout`, `exhausted`, 
+   `deferred:exhausted`, `retryable`, `deferred:retryable`.
```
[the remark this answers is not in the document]
kristinapathak: Oops that's a good point! Deferred doesn't belong here

text/metrics/0238-pipeline-monitoring.md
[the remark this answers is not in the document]
djaglowski: If I understand correctly, each collector pipeline contains a segment per exporter, where each of these segments contains all the receivers and all the processors of the pipeline, in addition to the exporter.

Can we update the language in this section to state this more clearly? Currently, it reads as "a receiver, zero or more processors, and an exporter" which doesn't appear accurate.

Can we also state "Components can be a part of multiple segments" before describing the relationship between segments and collector pipelines, since it's a prerequisite to understanding?

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,495 @@
...
+Exported_Responded(Preceder) <= Received(Follower)
+```
+
+When an observability platform detects an out-of-range value, there is
+likely to be a defective pipeline or missing instrumentation.
+
+### Measuring Total Pipeline Loss
+
+*WIP: Figure this out. This is a bit subjective. What does an end user expect 
+when calculating total items dropped in failure?*
```
jmacd, line 348: To me, the idea of "total" signifies starting from the beginning of the pipeline at the SDKs, meaning to use a sum of all the SDK-inserted items and compare against some point later in the pipeline to measure how many are original items are lost somehow. This could mean looking at a gateway collector's exporter counts and comparing to the SDK-inserted counts, for example.

text/metrics/0238-pipeline-monitoring.md
jmacd: I wonder about adding to alternatives considered:

[The collector's logic for `fanoutconsumer` uses `multierr`](https://github.com/open-telemetry/opentelemetry-collector/blob/effe26710d251afb58ce4efc5ae5aac52d9b61a0/internal/fanoutconsumer/traces.go#L47), building on Go's `Unwrap() []error` idiom to enclose all the errors. In my opinion, it would be nice to see `fanoutconsumer` manage the logic of deciding how to transform multiple errors into a single error, so that we could specify that fanoutconsumer dictates the N-to-1 problem and the observability mechanism just follows whatever it decides.

text/metrics/0238-pipeline-monitoring.md
```diff
@@ -0,0 +1,495 @@
...
+
+*WIP: add details*
+
+### Exporterhelper and batch processor
+
+*WIP: add details*
+
+### Retries
+
+*WIP: add details*
```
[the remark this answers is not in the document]
jmacd: I think it would be nice if there was some kind of conservation property we could maintain in the presence of retries where we have N errors and need to a single success/failure status.  It seems to me, also, something similar will be applicable for the case of fanout-consumer logic.

Should we have a separate metric for the extra fanout factor which is `N-1` in both of these cases? This number will be needed somewhere to understand pipeline conservation through fanout and retries, I think.

jmacd (commented):
This looks great, thank you @kristinapathak. 

While I think I would approve it if all the "WIP" sections were flushed out, but they're minor and I think this document stands in sufficient detail for an implementation to be prototyped. Probably the next step is to prototype these metrics in the collector and an SDK.

jmacd:
Apologies to @kristinapathak, I think we should close this. 

See https://github.com/open-telemetry/opentelemetry-collector/pull/11311
https://github.com/open-telemetry/opentelemetry-collector/pull/11406

tsloughter:
@jmacd those only cover the collector, right? Is the plan that a separate PR that focuses on the SDK metrics only be opened?

PR #263, opened 2024-08-12
Introduce environment variables for propagating context to subprocesses

yangskyboxlabs:
This OTEP attempts to address the usage of TRACEPARENT and TRACESTATE environment variables as a means to propagate tracing context to subprocesses.  Such usage is already implemented by various tools in the wild, and has essentially become a de facto standard.

https://github.com/open-telemetry/opentelemetry-specification/issues/740

This aims to replace an earlier OTEP 258 which appears to have stalled. While the core idea presented is the same, this OTEP has a slightly narrow scope, and do not derive from OTEP 258 or its unpublished predecessor OTEP 241.


adrielp:
@yangskyboxlabs - #258 was geared for this and hasn't stalled as much as I haven't ave been other in progress things taking immediate priority and cognitive load on my part within the community. There's an outstanding iteration I need to do based on feedback already received. I had opened [#1392](https://github.com/open-telemetry/semantic-conventions/issues/1392) to track that work as part of the CI/CD working group given its partial origination there.  I'm absolutely happy to have the help in making in updates for sure. Based on feedback the key is the details around text map propagation, and less so the names of the environment variables, though at least defining the common set is useful. 

These OTEPs are likely going to get moved over to the Specification repository which is where we're discussing them anyways, see #266 . If you'd like to assist this, or combine forces, then happy to have the help, please feel free to reach out to me directly on CNCF slack or hit up the CICD SIG calls (or Spec calls which is where this will end up getting discussed and approved). I'd prefer to keep only one PR open that addresses the concern to minimize overhead, disparity, and gives the based chance of success.

yangskyboxlabs:
I'd happy to merge efforts.  It would appear to me that #226 is definitely happening, and I'm willing to wait until that transition happens, or add to #258 if the next iteration happens before then.

For what it's worth, this was born out confusion when I had asked an IC to actually implement tracing propagation according to OTEP 254.

carlosalberto:
#258 has been merged - I will close this unless you think there is a reason to keep it open @yangskyboxlabs

mtwo:
Closing this, as #258 was merged instead

PR #266, opened 2024-08-30
Propose that we move OTEPs into the specification repository.

jsuereth:


text/0266-move-oteps-to-spec.md
```diff
@@ -0,0 +1,66 @@
...
+To solve these, let's move OTEPs into a directory within the [specification repository](github.com/open-telemetry/opentelemetry-specification).
+We would also update all tooling and expected reviews to match existing standards for OTEPs.  Given the maintainers of OTEPs are the same as
+maintainers of the specification, this should not change the bar for acceptance.
+
+## Internal details
+
+The following changes would occur:
+
+- The following files would be moved to the specification repo:
+  - `text/` directory -> `oteps/text/`
```
lmolkova, line 37: People that are not part of otel community sometimes assume that OTEPs have power similar to the spec. If OTEPs are moved to the spec repo, it'd be even more subtle. 

I'd prefer if we put them in the `proposals/`, `designs/`, or other directory that makes the status obvious (and does not rely on the reader knowledge of what the `OTEP` is)

As an alternative or an addition, we could add a short warning/explanation at the beginning of each OTEP that it represents the original design decision and the spec is the actual source of truth.

lmolkova (approved):

text/0266-move-oteps-to-spec.md
yurishkuro: I would prefer to keep the directory name as oteps, as that is our official name for the RFC process. The README in the directory can emphasize the point that these are proposals, not the official spec. 

+1 to have the same disclaimer repeated in the OTEPs themselves, as part of the template.

austinlparker (approved):

text/0266-move-oteps-to-spec.md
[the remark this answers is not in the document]
jsuereth: I'm happy to remove it, was going for least-disruptive change initially, and I didn't know what `text` was originally there for either.

text/0266-move-oteps-to-spec.md
```diff
@@ -0,0 +1,66 @@
...
+## Prior art and alternatives
+
+OTEPs were originally based on common enhancement proposal processes in other ecosystems, where enhancements live outside core repositories and follow a more rigorous criteria and evaluation. We are finding this
+problematic for OpenTelemetry for reasons discussed above. Additionally, unlike many other ecosystems where enhancement/design is kept separate from core code, OpenTelemetry *already* keeps its design separate
+form core code via the Specification vs. implementation repositories. Unlike these other OSS projects, our Specification generally requires rigorous discussion, design and prototyping prior to acceptance.  Even
+after acceptance into the specification, work is still required for improvements to roll out to the ecosystem. Effectively: The OpenTelemetry specification has no such thing as a "small" change: There are only medium changes that appear small, but would be enhancements in other proejcts or large changes that require an OTEP.
+
+## Open questions
+
+What are the important portions of the OTEP process to bring over? Have we missed anything in this description?
```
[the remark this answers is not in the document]
jsuereth: Hoping folks comment on whether I missed anything.  From my perspective, everything *I know* was important is captured.

So what are we doing?  Review comments on this proposal.

text/0266-move-oteps-to-spec.md
jsuereth: Great points both.  I think adding disclaimers to the readme and oteps themselves is suficient, I'll add that to the text here.

text/0266-move-oteps-to-spec.md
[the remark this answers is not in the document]
trask: > OTEPs require more approvals than specification PRs

I don't think there's a way to set a different requirement per directory. We could probably roll a custom github action check.

In the meantime, we could create an OTEP PR template and include the min approval bar in the merge checklist.

trask (approved):

pellared (approved):

text/0266-move-oteps-to-spec.md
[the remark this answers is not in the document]
dyladan: I would prefer to remove `/text` part of the path. It will forever just be a directory with a single directory in it providing no value. To my knowledge the only reason `/text` exists in the oteps repo is so that the top-level of the repo doesn't have too many files (weak reason imo) and so tooling can find relevant files easily (linters etc). Both concerns covered by just having an `/oteps` dir.

dyladan: I like `oteps/` with a solid README. Other names like `designs/` or `proposals/` make it sounds like after-the-fact explainer docs or things that are not really agreed upon yet.

dyladan (approved):
I support this idea, but would remove the `text/` part of the directory path.

text/0266-move-oteps-to-spec.md
```diff
@@ -0,0 +1,68 @@
...
+## Internal details
+
+The following changes would occur:
+
+- The following files would be moved to the specification repo:
+  - `text/` directory -> `oteps/text/`
+  - `0000-template.md` -> `oteps/0000-template.md`
+- Update the specification `Makefile` to include linting, spell checking, link checking and TOC-ing the oteps directory.
+- A one-time cleanup of OTEP markdown upon import to the specification repository.
+- Close existing OTEP PRs and ask folks to reopen against the specification repository.
```
pyohannes, line 41: What about issues in the OTEP repo? Should those be closed too like PRs?

text/0266-move-oteps-to-spec.md
```diff
@@ -0,0 +1,68 @@
...
+
+Moving OTEPs back into the specification solves two main issues:
+
+- Maintaining its tooling infrastructure (currently woefully out of date)
+- Bringing it into the existing triage and voting process currently within the
+  Specification.
+
+## Explanation
+
+Originally, OTEPs were kept as a separate repository to keep disjoint/disruptive designs as a separate repository.  There are a few differences between a normal PR and an OTEP:
```
pyohannes, line 15: There are a few OTEPs that don't concern the specification, but semantic conventions. Should those go with the specification, and, if need arises, be opened in the specification repository in the future? Or should we have a separate OTEPs in the semantic conventions repo?

pyohannes (approved):
I appreciate the effort to reduce overhead, I just added a few questions about details.

text/0266-move-oteps-to-spec.md
pellared: I guess they are going to be moved to https://github.com/open-telemetry/opentelemetry-specification

jsuereth: I missed these, but yes - I think those move to the spec repo.  I think all of them should have been OTEPs, and it's likely we close and/or ask folks to open formal OTEPs for these.

They current fall below the radar of triage and we need to address that.

text/0266-move-oteps-to-spec.md
jsuereth: I believe these move with the specification, e.g. the split of semconv repository, similarly to how proto changes are proposed as OTEPs.  Any semconv proposal that impacts more than semconv should be an OTEP and done in the spec repository.

Anything that ONLY involves semantic conventions should be proposed / dealt with within semconv itself, similar to how language-specific proposals happen in language-specific repositories.  E.g. the proposal around restructuring yaml files or creating the attribute registry -> those are and *were* done only within the semconv repository.

tedsuo (approved):

carlosalberto (approved):

austinlparker:
Just a note from the spec meeting today; We should ensure that when this move happens that we clarify the use of normative language in OTEPs (and where needed, ensure that existing OTEPs are normalized to that decision)

jack-berg:
@jsuereth do you have a plan / schedule in mind for merging and executing this plan?

text/0266-move-oteps-to-spec.md
```diff
@@ -0,0 +1,68 @@
...
+aspects of the current OTEP status.
+
+## Prior art and alternatives
+
+OTEPs were originally based on common enhancement proposal processes in other ecosystems, where enhancements live outside core repositories and follow a more rigorous criteria and evaluation. We are finding this
+problematic for OpenTelemetry for reasons discussed above. Additionally, unlike many other ecosystems where enhancement/design is kept separate from core code, OpenTelemetry *already* keeps its design separate
+form core code via the Specification vs. implementation repositories. Unlike these other OSS projects, our Specification generally requires rigorous discussion, design and prototyping prior to acceptance.  Even
+after acceptance into the specification, work is still required for improvements to roll out to the ecosystem. Effectively: The OpenTelemetry specification has no such thing as a "small" change: There are only medium changes that appear small, but would be enhancements in other proejcts or large changes that require an OTEP.
+
+## Open questions
```
yurishkuro, line 60: One question about history preservation: OTEP numbers are intentionally equal to PR number to make it easier to locate the discussion that occurred. If we move the commits, this relationship will be broken. I assume blame would still be usable, but perhaps as part of the move it would be good to add a URL of the original PR into the text.

bogdandrutu (approved):

carlosalberto:
We will merge this OTEP as it has enough approvals and will start the work, tuning further details, such as:

* Retain git history
* Issue migration
* OTEP number assignation
* Use of normative language (clarify the expectation, as we currently do not use it much here)
* Other items.

XSAM (approved):

PR #267, opened 2024-10-01
Structured Logs definition

MSNev:
With the definition of Events there has been additional contentious and confusion around the difference between what IS and Event vs a Structured Log when they are both represented and transported as a `type` of OpenTelemetry Log.

This also avoids overloading of the "Events" definition to include Structured Logs, as they have different intents and purposes.

And this avoids the "Hack" approach like that used for Zero length Spans to simulate Events, before Events had been defined. With the introduction of Events and support in the OpenTelemetry Semantic Convention tooling it is an extremely trivial task to define a new `type` definition for Structured Logs whch can be used to define the structure of the log within Semantic Conventions yaml and avoiding the multiple needs of overloading the term "Event", overloading the usage of the `event.name` and to support the Semantic differences between Events and Structured Logs.

Like supporting "Unnamed" Structured Logs, which would not be possible with the current Event definition as events MUST have a `name` field and all events with the same `name` are comparable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@
 # VS Code
 .vscode
 
+# Visual Studio
+.vs
+
 # Text formatting and spelling tools.
 .tools
 
--- a/text/0267-structured-logs.md
+++ b/text/0267-structured-logs.md
@@ -1,0 +1,105 @@
+# Structured Logs
+
+## Motivation
+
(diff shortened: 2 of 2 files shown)


MSNev:
We have decided in the Logs Sig to not support the production of specific Structured Logs at this point.

